PR #4, opened 2020-09-03
Annotate ExtensionSlotReact component

denniskigen:
Adds type annotation to `ExtensionSlotReact` and deletes an unused interface definition.

### DIFF
--- a/packages/esm-extension-manager/src/extension-slot-react.component.tsx
+++ b/packages/esm-extension-manager/src/extension-slot-react.component.tsx
@@ -9,10 +9,6 @@ export interface ExtensionSlotReactProps {
   children?: ReactNode;
 }
 
-interface CancelLoading {
-  (): void;
-}
-
 interface ExtensionContextData {
   extensionSlotName: string;
   extensionName: string;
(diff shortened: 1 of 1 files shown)


packages/esm-extension-manager/src/extension-slot-react.component.tsx
```diff
@@ -23,7 +19,7 @@ const ExtensionContext = React.createContext<ExtensionContextData>({
   extensionName: "",
 });
 
-export const ExtensionSlotReact = ({
+export const ExtensionSlotReact: React.FC = ({
```
FlorianRappl, line 22: Missing the props. Can you place them in there, too? 

```ts
React.FC<ExtensionSlotProps>
```

FlorianRappl (commented):
Great :beers:!

packages/esm-extension-manager/src/extension-slot-react.component.tsx
denniskigen: Updated

FlorianRappl (approved):
LGTM - thanks :beers:!

PR #5, opened 2020-09-09
Move initial script out of module-spa

brandones:
https://github.com/openmrs/openmrs-module-spa/pull/46


FlorianRappl (commented):
I think this looks good. However, I would suggest giving it a better name. Either `esm-core` (as this is the core), `esm-app-shell`, `esm-shell`, `esm-bundle` or `esm-loader` (all of them sound correct to me).

What are your thoughts @brandones ?

brandones:
Sure, I've never been quite clear on those definitions. This repository, which is for that category of "cross-cutting concerns," I have named `esm-core`. I think `esm-app-shell` is a good name for this.

FlorianRappl (approved):
LGTM - thanks a lot!

PR #9, opened 2020-09-15
Fix tsc error; fixes build failures

denniskigen:
Fixes build failures arising from `tsc` failing with this error:

Screenshot 2020-09-15 at 12 03 57

Which is, more specifically:

Screenshot 2020-09-15 at 11 56 37

Ultimately, per @FlorianRappl, we should disable type checking / compilation errors for test files. Developers have it hard enough already 😄


FlorianRappl (approved):
Thanks. for the work @denniskigen ! 🐝  🍻

PR #10, opened 2020-09-17
MF-314: Create minimal UI editor mode

denniskigen:
https://issues.openmrs.org/browse/MF-314

Paired with @brandones. Thanks, Brandon!

When the UI editor is toggled on:

Screenshot 2020-09-21 at 10 38 40

The UI editor becomes aware of **extensions** and **extension slots** and applies a **yellow outline to the extension slot** and a **cyan background to the extension within it**.

Screenshot 2020-09-21 at 10 24 10

### DIFF
--- a/packages/esm-extension-manager/src/extension-slot-react.component.tsx
+++ b/packages/esm-extension-manager/src/extension-slot-react.component.tsx
@@ -2,6 +2,7 @@ import React, { useState, useContext, ReactNode } from "react";
 import {
   renderExtension,
   getExtensionNamesForExtensionSlot,
+  getIsUIEditorEnabled,
 } from "./extensions";
 
 export interface ExtensionSlotReactProps {
@@ -30,7 +31,7 @@ export const ExtensionSlotReact: React.FC<ExtensionSlotReactProps> = ({
   }, [extensionSlotName]);
 
   return (
-    <>
+    <div style={{ backgroundColor: getIsUIEditorEnabled() ? "cyan" : "" }}>
       {extensionNames.map((extensionName) => (
         <ExtensionContext.Provider
           key={extensionName}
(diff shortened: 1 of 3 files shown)


brandones:
So, about the weird issue with the two switches getting state mixed up: I just realized that `Switch` creates and `input` with a fixed ID: https://github.com/openmrs/esm-implementer-tools-app/blob/f65a3ef8724dbe6e938b3e341d68b6751b21cb50/src/configuration/switch.component.tsx#L9 . So if you render two switches, it creates two different `input`s that have the same id. Which is not allowed. I suspect that's what's causing the weird behavior. I guess the ID should be made dynamic in some way.

Ooh, check out this answer, "2019 hooks update" https://stackoverflow.com/a/35047487/1464495

denniskigen:
> So, about the weird issue with the two switches getting state mixed up: I just realized that `Switch` creates and `input` with a fixed ID: https://github.com/openmrs/esm-implementer-tools-app/blob/f65a3ef8724dbe6e938b3e341d68b6751b21cb50/src/configuration/switch.component.tsx#L9 . So if you render two switches, it creates two different `input`s that have the same id. Which is not allowed. I suspect that's what's causing the weird behavior. I guess the ID should be made dynamic in some way.
> 
> Ooh, check out this answer, "2019 hooks update" https://stackoverflow.com/a/35047487/1464495

This worked a treat, thanks Brandon.

denniskigen:
Why this is still a draft:

The current implementation of this does not leverage React's change detection mechanism to update the extension/extension slot state whenever the UI editor is toggled. This is WIP.

brandones:
@denniskigen I think we won't make it use React's change detection until we have a more general solution for global state. I think we should leave that as a TODO and not block this PR on it.

brandones (approved):

PR #12, opened 2020-09-30
Allow passing style and class props intended for the top-level div of an extension slot

brandones:
This makes it so there's no weird intermediate div you can't style

### DIFF
--- a/packages/esm-extension-manager/src/extension-slot-react.component.tsx
+++ b/packages/esm-extension-manager/src/extension-slot-react.component.tsx
@@ -6,12 +6,14 @@ import {
   getIsUIEditorEnabled,
 } from "./extensions";
 
-export interface ExtensionSlotReactProps {
+interface ExtensionSlotBaseProps {
   extensionSlotName: string;
   children?: ReactNode;
-  style?: {};
-  [key: string]: any; // rest of params passed to div
+  style?: React.CSSProperties;
 }
+
+// remainder of props are for the top-level <div>
+export type ExtensionSlotReactProps<T = {}> = ExtensionSlotBaseProps & T;
 
 interface ExtensionContextData {
   extensionSlotName: string;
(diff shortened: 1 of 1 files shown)


packages/esm-extension-manager/src/extension-slot-react.component.tsx
```diff
@@ -9,6 +9,8 @@ import {
 export interface ExtensionSlotReactProps {
   extensionSlotName: string;
   children?: ReactNode;
+  style?: {};
```
FlorianRappl, line 12: What's that? Is that React.CSSProperties?

packages/esm-extension-manager/src/extension-slot-react.component.tsx
```diff
@@ -9,6 +9,8 @@ import {
...
+  [key: string]: any; // rest of params passed to div
```
FlorianRappl, line 13: I would prefer to collect them in a prop, e.g., `params`. Reason: Right now we cannot distinguish; what if `style` or even `extensionSlotName` is the desired prop name (for whatever reason)? Alternatively, I'd prefix the hard-wired props to make the distinction more obvious.

For typing reasons I would not have `[key]: any`, but rather a generic `<T = {}>` to be something like

```ts
interface ExtensionSlotBaseProps {
  extensionSlotName: string;
  children?: ReactNode;
  style?: {};
}

type ExtensionSlotReactProps <T = {}> = ExtensionSlotBaseProps & T;
```

packages/esm-extension-manager/src/extension-slot-react.component.tsx
```diff
@@ -40,8 +44,12 @@ export const ExtensionSlotReact: React.FC<ExtensionSlotReactProps> = ({
     ).then((names) => setExtensionNames(names));
   }, [extensionSlotName, moduleName]);
 
+  const divStyle = getIsUIEditorEnabled()
+    ? Object.assign(style || {}, { backgroundColor: "cyan" })
```
FlorianRappl, line 48: Any reason for mutating `style`? I'd never mutate a prop.

Just `{...style, backgroundColor: "cyan" }` - no need for the check and for all the braces (plus no mutation).

brandones: Oh yeah, this is absolutely the better way. Thanks!

brandones:
Very useful feedback, thanks @FlorianRappl . Ready for re-review.

FlorianRappl (approved):
LGTM - thanks!

PR #13, opened 2020-10-01
New App Shell

FlorianRappl:
Brings the new app shell to live.

- Allows local first development, right now uses a set of ESMs defined in the CDN as a placeholder
- Comes with alias possibilities, currently aliasing `@openmrs/esm-module-config` to `@openmrs/esm-config`
- Moved `@openmrs/esm-api`, `@openmrs/esm-styleguide`, `@openmrs/esm-error-handling` into the `esm-core` repository
- Has a framework independent global state container (contained in `esm-api`), which allows store creation per esm (multiple if wanted), but already comes with a central one shared from the app
- Builds on the central Lerna tooling
- Has an HTML template that is flexible regarding where the OpenMRS API is and where the SPA is placed
- Aligned / unified the dependencies and the way of consumption

There's more to it. I'll demo everything tomorrow.


FlorianRappl:
> Starting into it. This might involve a tricky rebase, but do you think you could merge the other repos in in a way that preserves their histories, for example using lerna import?

I'd love to avoid that. Its unnecessary work imho - if you want to do it then please go ahead. Otherwise, I'd just archive the other repos (if one *really* needs to know the past the old repos are still there).

.prettierignore
```diff
@@ -1,3 +1,4 @@
 dist/
 README.md
-docs/
\ No newline at end of file
+docs/
+packages/esm-styleguide/src/buttons-and-links/buttons-and-links.stories.html
```
brandones, line 4: Should this just be `packages/esm-styleguide/src/**/*.html`?

FlorianRappl: I just brought it over as-is - I guess it would be better as you suggested, but it also has (had) no impact so far.

packages/esm-api/src/setup.ts
```diff
@@ -0,0 +1,24 @@
+import { defineConfigSchema } from "@openmrs/esm-config";
+import { refetchCurrentUser } from "./shared-api-objects/current-user";
+
+export function setupApiModule() {
+  defineConfigSchema("@openmrs/esm-api", {
```
brandones, line 5: Why wrap this stuff in a function, instead of letting it run at module import time?

packages/esm-api/src/state.ts
```diff
@@ -0,0 +1,65 @@
+import createStore, { Store } from "unistore";
```
brandones, line 1: Looking forward to seeing a usage example for this. I appreciate your emphasis on keeping it simple. Redux inspires in me some profound dread, the sense of a Kuthuluesque darkness filling the room.

My other point of reference for a way to implement global state for the app would be to export an object as a "global store" that can be hooked into using [useGlobalState](https://github.com/dai-shi/react-hooks-global-state) (or some custom wrapper thereof).

I'm happy to sign off on this state/store implementation for now, and we'll talk about it more once we see how it works.

brandones: I am skeptical about MF-specific stores (again, not going to block on it). You said that the medication orders basket is a good example use case. I look forward to seeing it. But my bias is "the best state management system for React is React."

packages/esm-app-shell/src/assets/importmap.json
```diff
@@ -0,0 +1,18 @@
+{
+  "imports": {
+    "@openmrs/esm-patient-chart-widgets": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-chart-widgets/153e295c939a49ba3dd4c33c78bffcb21b0f9e15/openmrs-esm-patient-chart-widgets.js",
+    "@openmrs/esm-patient-chart-widgets/": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-chart-widgets/153e295c939a49ba3dd4c33c78bffcb21b0f9e15/",
```
brandones, line 4: Are these `/`-suffixed deps necessary?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -0,0 +1,18 @@
...
+  "react-dom": () => require("react-dom"),
+  "react-router-dom": () => require("react-router-dom"),
+  "react-i18next": () => require("react-i18next"),
+  "single-spa": () => require("single-spa"),
+  "@openmrs/esm-api": () => require("@openmrs/esm-api"),
+  "@openmrs/esm-error-handling": () => require("@openmrs/esm-error-handling"),
+  "@openmrs/esm-config": () => require("@openmrs/esm-config"),
+  "@openmrs/esm-module-config": () => require("@openmrs/esm-config"),
+  "@openmrs/esm-extension-manager": () =>
+    require("@openmrs/esm-extension-manager"),
```
brandones, line 13: Should we rename `esm-extension-manager` to `esm-extensions`?

The aliasing mechanism is really cool.

packages/esm-app-shell/src/layout.ts
```diff
@@ -0,0 +1,2 @@
+import "carbon-components/css/carbon-components.min.css";
```
brandones, line 1: I expect something called `layout` to actually structure where things are on the screen. Maybe this file should be `style.ts` or something like that?

packages/esm-api/src/setup.ts
FlorianRappl: Because its not ready then.

When we import the module we have not set yet the variables on window. Beforehand we asynchronously imported the module, now its imported (evaluated) *before* the setup.

FlorianRappl: (Also its cleaner as the evaluation does not have the side effect)

packages/esm-api/src/state.ts
FlorianRappl: I fully agree, but the problem is obviously when you have different component trees mounted at different times, but requiring the same information source. The order basket is a good example when that happens. And in this case we solve it.

packages/esm-app-shell/src/assets/importmap.json
FlorianRappl: For us: Not. This is just because the new import map deployer uses them and I transported them here.

Personally, I would not use them.

(in theory they allow submoduling, but for that to work the package needs to follow the conventions - which barely any package does; this only works reliably for commonjs).

packages/esm-app-shell/src/dependencies.ts
FlorianRappl: We can do that!

packages/esm-app-shell/src/layout.ts
FlorianRappl: Sure -  we can do it.

packages/esm-error-handling/.travis.yml
```diff
@@ -0,0 +1,28 @@
+language: node_js
```
brandones, line 1: Is this file supposed to be here?

packages/run-openmrs/src/cli.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+}
+
+yargs.command(
+  "debug",
+  "Starts a new debugging session of the OpenMRS app shell.",
+  (argv) =>
+    argv
+      .number("port")
+      .default("port", 8080)
```
brandones, line 34: If you remove this default, won't webpack-dev-server just use 8080 or the first available port if 8080 is taken?

packages/esm-error-handling/.travis.yml
FlorianRappl: Actually no - great catch! Thanks!

packages/run-openmrs/src/cli.ts
FlorianRappl: No, as we use WDS programmatically and insert this. The auto lookup would require the import / use of a different package.

(the choice here as no impact on using WDS in the monorepo - here you use it via `--port` as usual)

packages/run-openmrs/src/utils/index.ts
```diff
@@ -0,0 +1,73 @@
...
+export function getImportmap(value: string): ImportmapDeclaration {
+  if (!/https?:\/\//.test(value)) {
+    const path = resolve(process.cwd(), value);
+
+    if (existsSync(path)) {
+      return {
+        type: "inline",
+        value: readFileSync(path, "utf8"),
+      };
+    } else if (checkImportmapJson(value)) {
```
brandones, line 61: Instead of letting this function return as if its a url if this check fails, can we throw an error with a helpful error message?

FlorianRappl: The check may fail if the provided string is like "my-import-map.json", which may be still valid at runtime. This is a valid (relative) URL.

The case where we may warn is if we know for sure that the name will be definitely an invalid URL or if the `readFileSync` fails the `checkImportmapJson`.

But for now I'd say lets leave it. If its really invalid it will be seen when running the shell anyway.

brandones:
Awesome work. Just requesting those four minor changes above.

brandones (approved):

packages/run-openmrs/src/utils/index.ts
FlorianRappl: Added another check!

FlorianRappl:
All done!

jonathandick:
Great work Florian!

PR #15, opened 2020-10-07
First NPM Release

FlorianRappl:
This brings the `assemble` subcommand for `openmrs`. Essentially, this command makes it possible to "assemble" an importmap given a configuration, such as 

```json
{
  "microfrontends": {
    "@openmrs/esm-primary-nav": "^1.0.0"
  }
}
```

or using a survey, shown below

1a8f224c-78ff-460c-8a46-580b584dd061

**Note**: The list of microfrontends in the survey is right now computed via all packages having the `openmrs` keyword and being suffixed with `-app`. We thus need to update the existing microfrontends to appear there (we should anyway, to improve their `dependencies` and `devDependencies` etc. declarations).

**Important**: This is not the end of the work on esm-core / app shell / openmrs CLI tool, but rather just a good point to have a first release to allow everyone to continue working and then improve / refine later on.

**Also**: I've updated the CI/CD pipeline. The NPM packages should (hopefully?) be published when we add a tag. So once this is in `master` we just tag the "3.0.1" release and we are good. Everything should run / be published automatically.


brandones (approved):

brandones:
Whoops, sorry, accidental approve—I've only read that last commit, not the whole thing. Let me take a proper look.

packages/openmrs/src/commands/start.ts
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable no-console */
+
+export interface StartArgs {
+  port: number;
+}
+
+export function runStart(args: StartArgs) {}
```
brandones, line 7: Is this file a stub? Should it print a message saying so?

FlorianRappl: Its a stub. I'll add something useful in a future release!

brandones (approved):
Ok, looks good! Thanks Florian!

FlorianRappl:
I'll try to create a release now. Keep fingers crossed, hehe.

PR #16, opened 2020-10-15
Add support for passing configuration to extensions

brandones:
Resolves https://issues.openmrs.org/browse/MF-319

This is a pretty bloated PR. Sorry.

- Creates a hook that extensions can use to obtain configuration that is specific to them
- Adds support for extension IDs rather than just names, which will be needed for mounting multiple of the same extension into a slot
- Adds react-root-decorator to esm-core
- Centralizes React Context in a separate module to avoid circular dependencies between esm-extensions and esm-config

### DIFF
--- a/packages/esm-config/package.json
+++ b/packages/esm-config/package.json
@@ -41,6 +41,7 @@
     "single-spa": "4.x"
   },
   "devDependencies": {
+    "@openmrs/esm-core-context": "file:../esm-core-context",
     "@types/ramda": "^0.26.44",
     "@types/react": "^16.9.46",
     "@types/systemjs": "^6.1.0",
--- a/packages/esm-config/src/react-hook/use-extension-config.test.tsx
+++ b/packages/esm-config/src/react-hook/use-extension-config.test.tsx
@@ -1,0 +1,145 @@
+import React from "react";
+import { render, screen, cleanup, waitFor } from "@testing-library/react";
+import { ModuleNameContext, ExtensionContext } from "@openmrs/esm-core-context";
+import {
+  clearAll,
+  defineConfigSchema,
+  provide,
+} from "../module-config/module-config";
+import { useExtensionConfig, clearConfig } from "./use-extension-config";
+
+xdescribe(`useExtensionConfig`, () => {
+  afterEach(clearAll);
+  afterEach(cleanup);
+  afterEach(clearConfig);
+
+  it(`can return extension config as a react hook`, async () => {
+    defineConfigSchema("ext-module", {
+      thing: {
+        default: "The first thing",
+      },
+    });
+
+    render(
+      <React.Suspense fallback={<div>Suspense!</div>}>
+        <ModuleNameContext.Provider value="slot-module">
+          <ExtensionContext.Provider
+            value={{
+              extensionModuleName: "ext-module",
+              extensionSlotName: "fooSlot",
+              extensionId: "barExt#id1",
+            }}
+          >
+            <RenderConfig configKey="thing" />
+          </ExtensionContext.Provider>
+        </ModuleNameContext.Provider>
+      </React.Suspense>
+    );
+
+    await waitFor(() => {
+      expect(screen.getByText("The first thing")).toBeTruthy();
+    });
+  });
+
+  it(`can handle multiple extensions`, async () => {
+    defineConfigSchema("foo-module", {
+      thing: {
+        default: "foo thing",
+      },
+    });
+
+    defineConfigSchema("bar-module", {
+      thing: {
+        default: "bar thing",
+      },
+    });
+
+    render(
+      <React.Suspense fallback={<div>Suspense!</div>}>
+        <ModuleNameContext.Provider value="slot-module">
+          <ExtensionContext.Provider
+            value={{
+              extensionSlotName: "fooSlot",
+              extensionModuleName: "foo-module",
+              extensionId: "fooExt#id1",
+            }}
+          >
+            <RenderConfig configKey="thing" />
+          </ExtensionContext.Provider>
+          <ExtensionContext.Provider
+            value={{
+              extensionSlotName: "fooSlot",
+              extensionModuleName: "bar-module",
+              extensionId: "barExt",
+            }}
+          >
+            <RenderConfig configKey="thing" />
+          </ExtensionContext.Provider>
+        </ModuleNameContext.Provider>
+      </React.Suspense>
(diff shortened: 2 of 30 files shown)


packages/esm-config/package.json
```diff
@@ -41,6 +41,7 @@
     "single-spa": "4.x"
   },
   "devDependencies": {
+    "@openmrs/esm-core-context": "file:../esm-core-context",
```
FlorianRappl, line 44: Why not `^3.0.3` and run `lerna bootstrap`?

packages/esm-config/src/react-hook/use-extension-config.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { render, screen, cleanup, waitFor } from "@testing-library/react";
+import { ModuleNameContext, ExtensionContext } from "@openmrs/esm-core-context";
+import {
+  clearAll,
+  defineConfigSchema,
+  provide,
+} from "../module-config/module-config";
+import { useExtensionConfig, clearConfig } from "./use-extension-config";
+
+xdescribe(`useExtensionConfig`, () => {
```
FlorianRappl, line 11: Is it `describe`?

packages/esm-extensions/package.json
```diff
@@ -41,6 +41,7 @@
   },
   "devDependencies": {
     "@openmrs/esm-config": "^3.0.3",
+    "@openmrs/esm-core-context": "file:../esm-core-context",
```
FlorianRappl, line 44: Same as above, just reference it via `^3.0.3`

packages/react-root-decorator/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@openmrs/react-root-decorator",
+  "version": "3.2.0",
```
FlorianRappl, line 3: How does this fit into the unified version scheme? I think we now need to go to independent versions?

Also the last published version is `3.3.1` - should we make this 3.3.2?

packages/react-root-decorator/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "url": "https://github.com/openmrs/openmrs-react-root-decorator/issues"
+  },
+  "homepage": "https://github.com/openmrs/openmrs-react-root-decorator#readme",
+  "peerDependencies": {
+    "i18next": "*",
+    "react": ">=16.0.0",
+    "react-i18next": ">=9"
+  },
+  "devDependencies": {
+    "@openmrs/esm-core-context": "file:../esm-core-context"
```
FlorianRappl, line 43: Do we need this? Can't it be part of the same package?

I mean I'd love to reduce packages in some way. If `core-context` and the root decorator become just one package (maybe replace the root decorator by the core-context) I'd not start crying...

FlorianRappl (commented):
LGTM in general even though I am not 100% convinced on the ID of extensions. A name should be sufficient. The combination microfrontend name + extension name + registration index should be unique enough.

packages/esm-config/src/react-hook/use-extension-config.test.tsx
brandones: Whoops! Need to re-enable that, thanks!

packages/react-root-decorator/package.json
brandones: Yeah I'm not really sure what to do about this, what would you advise?

packages/react-root-decorator/package.json
brandones: Ooh yeah, those two definitely can be combined. Let's see.

FlorianRappl: Yeah its a pain. If we merge core-context and the root decorator we could merge it into core-context and avoid "touching" root decorator at all. After all, we'd need to upgrade / change anyway in the microfrontends.

What are your thoughts?

(so essentially we rename root decorator to core context and extend it with the context functionality)

packages/react-root-decorator/package.json
FlorianRappl: Yeah its a pain. If we merge core-context and the root decorator we could merge it into core-context and avoid "touching" root decorator at all. After all, we'd need to upgrade / change anyway in the microfrontends.

What are your thoughts?

(so essentially we rename root decorator to core context and extend it with the context functionality)

brandones: Yeah, that makes sense to me. Should it just be `@openmrs/esm-context`?

FlorianRappl: I love that proposal. Yes, let's do it like that!

packages/esm-config/package.json
brandones: Nice, thanks, still getting the hang of Lerna...

brandones:
Thanks for the feedback! This is ready for re-review.

The rationale for extension IDs is in the [RFC](https://github.com/openmrs/openmrs-rfc-frontend/pull/27/files#diff-3459e36c346a78265afd01c37bf9463564c8d2eb0176d9630ce76edcf35ee350R31-R33):

> An **Extension ID** uniquely identifies an extension *within a specific extension slot*. By default, if an extension is only attached to an extension point once, the extension ID is the extension name.

The motivating example [in the slides](https://docs.google.com/presentation/d/1ParNFdehbBexycC_XzdvpPNXBCea-4GYwAuoPFtvYIY/edit#slide=id.g932ba70969_0_32) is a generic "dashboard" widget (now I guess we'd call it a "tab" or "page" of the patient chart) to which the implementer can attach widgets using the `add` config. Another example might be if we had a "line graph" widget, and an implementer wanted a tab of the patient chart to have two separate graphs of two different things. See the RefApp Patient Summary Widgets for an example of what this looks like in RefApp 2.0.

FlorianRappl:
Sure I understand the rationale - thanks for citing the RFC for this @brandones . However, as I pointed out the given combination is already unique and provides less room for error on the implementers side (which in the given case needs to specify not only the extension slot, but also the extension ID).

My suggestion on the matter would be: Leave the extension ID *optional*. If not explicitly stated we'll use the combo module name + extension slot + specified extension index as ID. Our recommendation would be to use explicit IDs, but the implicit one is legit, too.

FlorianRappl (approved):
LGTM - thanks for the nice work 👍 !

(I'll leave it free to you to either have an optional extension ID or not)

brandones:
Thanks for the review!

PR #17, opened 2020-10-16
MF-363 Application should update live in response to temporary config changes

brandones:
https://issues.openmrs.org/browse/MF-363

update

### DIFF
--- a/packages/esm-config/src/react-hook/use-extension-config.tsx
+++ b/packages/esm-config/src/react-hook/use-extension-config.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from "react";
+import { useContext, useEffect } from "react";
 import {
   ExtensionContext,
   ModuleNameContext as SlotModuleNameContext,
--- a/packages/esm-config/src/react-hook/use-extension-config.tsx
+++ b/packages/esm-config/src/react-hook/use-extension-config.tsx
@@ -1,11 +1,15 @@
-import React, { useContext } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import {
   ExtensionContext,
   ModuleNameContext as SlotModuleNameContext,
 } from "@openmrs/esm-context";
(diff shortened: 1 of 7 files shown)


jonathandick:
Very cool!

denniskigen:
Very cool indeed @brandones. Might I suggest using `screen.findByText` instead of `waitFor` for your text based matchers per this [advice](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find)?

FlorianRappl:
Unfortunately, the last release did not work so well.

image

I only see a tag created (well, that's sufficient to create the release, even though not really pleasant as GitHub still shows last released version 3.0.3).

The other thing is that the `publishConfig` was forgotten in the `esm-context`. I fear that I missed that one in the review. See https://github.com/openmrs/openmrs-esm-core/blob/master/packages/esm-app-shell/package.json#L53 for an example how to configure it.

packages/esm-config/src/react-hook/use-extension-config.tsx
```diff
@@ -21,32 +25,41 @@ export function useExtensionConfig() {
...
+    });
+    return () => sub.unsubscribe();
+  }, [slotModuleName]);
+
+  if (error) {
+    // Suspense will just keep calling useConfig if the thrown promise rejects.
+    // So we check ahead of time and avoid creating a new promise.
+    throw error;
+  }
+  const uniqueExtensionLookupId = extensionSlotName + "-" + extensionId;
```
FlorianRappl, line 58: For such constructs I'd prefer a template string.

FlorianRappl (approved):

brandones:
@FlorianRappl Oops, I did the release using `lerna publish` instead of by creating the GitHub tag. My bad. And yes, I'll fix the esm-context package.json, good catch there.

denniskigen (approved):
LGTM, thanks @brandones!

PR #19, opened 2020-10-23
Add devtools and implementer-tools to esm-core

brandones:


packages/esm-app-shell/package.json
```diff
@@ -49,7 +51,8 @@
     "systemjs": "^6.3.3"
   },
   "devDependencies": {
-    "@types/systemjs": "^6.1.0"
+    "@types/systemjs": "^6.1.0",
+    "css-modules-typescript-loader": "^4.0.1"
```
FlorianRappl, line 55: Don't we need this to compile?

Keep in mind that for the app shell package we need to be able to resolve all dependencies necessary for building. The app shell is only used as a dependency in `openmrs` and here we have to be able to build.

Alternative is that `css-modules-typescript-loader` goes into the package.json of the monorepo itself and the `openmrs` package.

packages/esm-app-shell/src/assets/importmap.json
```diff
@@ -5,9 +5,7 @@
     "@openmrs/esm-home-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-home-app/fd1737e3d6ecd8f9b8e4a684e468db73e8ec1d17/openmrs-esm-home.js",
     "@openmrs/esm-primary-navigation-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-primary-navigation-app/12a1fffb9910809dd8facb946e43eefb3eb7459a/openmrs-esm-primary-navigation.js",
     "@openmrs/esm-login-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-login-app/ba86336f68ddc4fa364534b605a4184fea5ac1eb/openmrs-esm-login.js",
-    "@openmrs/esm-devtools-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-devtools-app/75bc1045345193218fccb1058e061d24b6baa98b/openmrs-esm-devtools.js",
```
FlorianRappl, line 8: I'd keep them lazy loaded / opt-in.

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -6,14 +6,20 @@ export const sharedDependencies = {
...
+  "@openmrs/esm-error-handling": () => require("@openmrs/esm-error-handling"),
   "@openmrs/esm-extensions": () => require("@openmrs/esm-extensions"),
   "@openmrs/esm-extension-manager": () => require("@openmrs/esm-extensions"),
   "@openmrs/esm-styleguide": () => require("@openmrs/esm-styleguide"),
   "carbon-components": () => require("carbon-components"),
   "carbon-icons": () => require("carbon-icons"),
   rxjs: () => require("rxjs"),
 };
+
+export const coreApps = {
```
FlorianRappl, line 21: I'd drop this.

The reason for moving over these microfrontends is more from a dev perspective and less from a "backed-in" perspective imho.

packages/esm-app-shell/webpack.config.js
```diff
@@ -62,13 +62,27 @@ module.exports = {
         },
       },
       {
-        test: /\.css$/,
+        test: /.+\.css$/,
```
FlorianRappl, line 65: Is this change really necessary? Beforehand we find everything that ends with `.css` now its everything that ends with `.css` but has at least one character beforehand. Is it a viable case (i.e., is there a file called `.css` somewhere)?

packages/esm-app-shell/package.json
```diff
@@ -30,8 +30,10 @@
     "@openmrs/esm-api": "^3.0.6",
     "@openmrs/esm-config": "^3.0.6",
     "@openmrs/esm-context": "^3.0.6",
+    "@openmrs/esm-devtools": "^3.0.6",
```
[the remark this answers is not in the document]
brandones: Aha. Maybe I misinterpreted [this Slack thread](https://openmrs.slack.com/archives/GQJA12F4L/p1603148361041700). Is the idea that devtools and implementer-tools should just live in esm-core, but not be loaded as a "cross-cutting concern" module? Treat it like a normal microfrontend, but which is in the esm-core monorepo?

packages/esm-app-shell/package.json
brandones: Nope, cruft, sorry

packages/esm-app-shell/src/dependencies.ts
brandones: Aha, okay, makes sense.

packages/esm-app-shell/package.json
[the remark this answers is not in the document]
brandones: Oops, seeing your later comments—makes sense, thanks!

packages/esm-app-shell/webpack.config.js
brandones: I actually did a crazy amount of messing with CSS stuff to try and get it to work; this is just incidental diff noise, probably copy-pasted from a different module's webpack.

brandones:
Tested with `yarn run:shell` and `cd packages/esm-implementer-tools; ../openmrs/dist/cli.js debug`

FlorianRappl (approved):
LGTM - thanks :beers:!

Regarding a release of this - I also have some changes in the pipe (incl. version change to 3.0.7), so let's try to bring these in, too? What's your opinion? Release on Monday or Tuesday?

brandones:
Thanks @FlorianRappl ! Sure, whenever. I was only releasing to facilitate development on implementer-tools, and now that problem is out of the way :)

PR #20, opened 2020-10-26
MF-342  Extension attachments should be configurable in the Implementer Tools

brandones:
https://issues.openmrs.org/browse/MF-342

ext-conf

### DIFF
--- a/packages/esm-implementer-tools/src/configuration/extensions-config-tree.tsx
+++ b/packages/esm-implementer-tools/src/configuration/extensions-config-tree.tsx
@@ -7,7 +7,7 @@ import {
 import styles from "./configuration.styles.css";
 import EditableValue from "./editable-value.component";
 
-interface ExtensionsConfigTreeProps {
+export interface ExtensionsConfigTreeProps {
   config: { [key: string]: any };
   moduleName: string;
 }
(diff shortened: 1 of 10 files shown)


packages/esm-implementer-tools/src/configuration/extensions-config-tree.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useMemo, useState, useEffect } from "react";
+import { ExtensionSlotConfig } from "@openmrs/esm-config";
+import {
+  getExtensionSlotsForModule,
+  getExtensionIdsForExtensionSlot,
+} from "@openmrs/esm-extensions";
+import styles from "./configuration.styles.css";
+import EditableValue from "./editable-value.component";
+
+interface ExtensionsConfigTreeProps {
```
FlorianRappl, line 10: Should this be exported? Just because its being used in exports.

(BTW: If d.ts are generated then the d.ts generation will fail on such cases)

FlorianRappl:
In general looks great to me - thanks @brandones. Just had one remark.

packages/esm-implementer-tools/src/configuration/extensions-config-tree.tsx
brandones: Oh sure, yeah, it's best practice to export these? That makes sense

FlorianRappl: Hm personally I always export anything that is reached / touched by another export, e.g., `export function foo(arg1: Foo, arg2: Bar): Qxz` would need to export the types `Foo`, `Bar`, Qxz` if they are defined internally. If they are collected by some imports then there is no need (already exported somewhere else).

This way you are also save that TS will be able to properly resolve it when creating a d.ts (if you want it) + that users can actually reuse your types.

PR #21, opened 2020-10-26
MF-364: Disable MF when a backend dependency is missing or incorrect version

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-364


packages/esm-app-shell/src/dependencies.ts
```diff
@@ -1,4 +1,8 @@
-export const sharedDependencies = {
+import { openmrsFetch } from "@openmrs/esm-api";
```
FlorianRappl, line 1: This module should be exclusively used to determine shared dependencies.

Please put the code in another (new?) module. Thanks!

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+  }
+}
+
+function fetchInstalledBackendModules() {
+  return openmrsFetch(`/ws/rest/v1/module?v=custom:(uuid,version)`, {
+    method: "GET",
+  });
+}
+
+function getMissingBackendModules(requiredBackendModules) {
```
FlorianRappl, line 72: Missing type

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+      setTimeout(() => {
+        throw err;
+      });
+    }
+  } else {
+    checkIfModulesAreInstalled(module);
+  }
+}
+
+function checkIfModulesAreInstalled(module) {
```
FlorianRappl, line 45: Type?

samuelmale:
@brandones / @FlorianRappl any chance we could look into this?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
   "carbon-icons": () => require("carbon-icons"),
   rxjs: () => require("rxjs"),
 };
+
+let installedBackendModules: any[] = [];
+let modulesWithMissingBackendModules: MissingBackendModules[] = [];
+let modulesWithWrongBackendModulesVersion: any[] = [];
+
+async function checkBackendDeps(module: any) {
```
FlorianRappl, line 29: Is this really any? Can we do better (be more specific)?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+        throw err;
+      });
+    }
+  } else {
+    checkIfModulesAreInstalled(module);
+  }
+}
+
+function checkIfModulesAreInstalled(module) {
+  let missingBackendModule = getMissingBackendModules(
```
FlorianRappl, line 46: These are `const`, right? I don't see any reassignment.

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+    method: "GET",
+  });
+}
+
+function getMissingBackendModules(requiredBackendModules) {
+  const requiredBackendModulesUuids = Object.keys(requiredBackendModules);
+  const installedBackendModuleUuids = installedBackendModules.map(
+    (res) => res.uuid
+  );
+  let missingModules = difference(
```
FlorianRappl, line 77: Why not `const`?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+  return missingModules.map((key) => {
+    return { uuid: key, version: requiredBackendModules[key] };
+  });
+}
+
+function getInstalledAndRequiredBackendModules(requiredBackendModules) {
+  let requiredModules = Object.keys(requiredBackendModules).map((key) => {
+    return { uuid: key, version: requiredBackendModules[key] };
+  });
+  let installedAndRequiredBackendModules = requiredModules.filter(
```
FlorianRappl, line 90: You can directly return this.

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+  let missingModules = difference(
+    requiredBackendModulesUuids,
+    installedBackendModuleUuids
+  );
+  return missingModules.map((key) => {
+    return { uuid: key, version: requiredBackendModules[key] };
+  });
+}
+
+function getInstalledAndRequiredBackendModules(requiredBackendModules) {
```
FlorianRappl, line 86: Type?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+  let requiredModules = Object.keys(requiredBackendModules).map((key) => {
```
FlorianRappl, line 87: This is not reassigned. Make it `const`.

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+    (requiredModule) => {
+      return installedBackendModules.find((installedModule) => {
+        return requiredModule.uuid === installedModule.uuid;
+      });
+    }
+  );
+  return installedAndRequiredBackendModules;
+}
+
+function getMisMatchedBackendModules(installedAndRequiredBackendModules) {
```
FlorianRappl, line 100: Type?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+        uuid: moduleName,
+        requiredVersion: requiredVersion,
+        installedVersion: installedVersion,
+      });
+    }
+  }
+  return misMatchedBackendModules;
+}
+
+function isVersionInstalled(requiredVersion, installedVersion) {
```
FlorianRappl, line 118: Types?

packages/esm-app-shell/src/dependencies.ts
```diff
@@ -17,3 +21,125 @@ export const sharedDependencies = {
...
+  return semver.eq(
```
FlorianRappl, line 119: I am not sure if OpenMRS backend modules follow semver. Are we using true semver specifiers (^ ~ etc.) to declare backend modules?

The semver package is not giant, but not small either (https://bundlephobia.com/result?p=semver@7.3.2).

packages/esm-app-shell/src/index.ts
```diff
@@ -35,9 +41,19 @@ function loadApps() {
  * This function returns an object that is used to feed Single
  * SPA.
  */
-function setupApps(modules: Array<[string, System.Module]>) {
+async function setupApps(modules: Array<[string, System.Module]>) {
   for (const [appName, appExports] of modules) {
-    registerApp(appName, appExports);
+    if (appExports.backendDependencies) {
+      await checkBackendDeps({ ...appExports, moduleName: appName });
```
FlorianRappl, line 47: Please make this all parallel. Right now we evaluate module *after* module, but we should evaluate *all* modules in parallel.

packages/esm-app-shell/src/dependencies.ts
brandones: They do, more or less. See [OpenMRS Versioning doc](https://wiki.openmrs.org/display/docs/Versioning).

That is... much bigger than I would expect a version string comparison library to be. Maybe we want to use [compare-versions](https://www.npmjs.com/package/compare-versions)?

I'm inclined to use true semver specifiers, unless there's a strong reason to do something else.

packages/esm-app-shell/src/index.ts
samuelmale: I have tried evaluating the modules in parallel but I think for this specific case we should use a sequential flow, I agree  with you that parallel flow should make code more efficient in terms of performance, but for this specific case, **No**; why?  

With sequential, we shall make only one server request([the request that initialises the `installedBackendModules`](https://github.com/openmrs/openmrs-esm-core/blob/9288a45ececeff42d610164229f0f7230ef20173/packages/esm-app-shell/src/dependencies.ts#L30-L39)) for the `module` at index 0 then for the next proceeding modules it won't make any server request, hence more efficiency. When working in parallel, each module evaluation would make a server request hence will be slower. 

When I tested locally, running in parallel takes around 9-13secs to load the page while in sequence it takes 2-4secs. It's a big difference in performance. @FlorianRappl and @brandones tell me what you think on this, correct me if I missed the point

FlorianRappl: I think the point here is that the given flow is wrong for performance issues. Instead of

1. Run for each module
2. Get backend modules
3. Check dependencies

You should go for

1. Get backend modules
2. Run for each module
3. Check dependencies

Thus there is no need to be sequential or parallel. The backend modules are received in a single step before checking the modules.

samuelmale:
@FlorianRappl let me mention this here, 95% of these changes are a copy/paste from [openmrs-backend-dependencies](https://github.com/openmrs/esm-implementer-tools-app/blob/master/src/backend-dependencies/openmrs-backend-dependencies.ts) in `esm-implementer-tools-app` , my plan was to reuse this logic from `esm-implementer-tools-app`, if it's okay we shall have a followup PR on `esm-implementer-tools-app`. This means `esm-implementer-tools-app` will depend on the app-shell, all good on this?

brandones:
No, because then there would be a circular dependency between app-shell and esm-implementer-tools-app.

FlorianRappl:
As @brandones explained you cannot directly import / use the logic. But you could place the relevant information as a global on `window` and re-use this.

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
```diff
@@ -0,0 +1,103 @@
+import { openmrsFetch } from "@openmrs/esm-api";
+import * as semver from "semver";
+import { difference } from "lodash-es";
+import { MissingBackendModules, Module, BackendModule } from "./types";
+
+let installedBackendModules: any[] = [];
```
FlorianRappl, line 6: Is this really any? Can we use `Array<T>` instead of `T[]`? The rest uses this, too, in this codebase.

Also please use `const` if you don't reassign the variable.

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+function fetchInstalledBackendModules() {
+  return openmrsFetch(`/ws/rest/v1/module?v=custom:(uuid,version)`, {
+    method: "GET",
+  });
+}
+
+function getMissingBackendModules(requiredBackendModules: object) {
```
FlorianRappl, line 48: `object`? So its a `Record<string, ?>` what is `?`, is it `any` or something specified? It seems to be `string` from the code (its a map string - string that maps names to versions).

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
```diff
@@ -0,0 +1,103 @@
...
+  const missingModules = difference(
+    requiredBackendModulesUuids,
+    installedBackendModuleUuids
+  );
+  return missingModules.map((key) => {
+    return { uuid: key, version: requiredBackendModules[key] };
+  });
+}
+
+function getInstalledAndRequiredBackendModules(requiredBackendModules: object) {
```
FlorianRappl, line 62: Same, use `Record<string, string>`.

FlorianRappl (commented):
Looks good - the typings need a bit of work.

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
samuelmale: Fixed

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
samuelmale: Fixed

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
samuelmale: Fixed

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
```diff
@@ -3,9 +3,9 @@ import * as semver from "semver";
 import { difference } from "lodash-es";
 import { MissingBackendModules, Module, BackendModule } from "./types";
 
-let installedBackendModules: any[] = [];
-let modulesWithMissingBackendModules: MissingBackendModules[] = [];
-let modulesWithWrongBackendModulesVersion: MissingBackendModules[] = [];
+let installedBackendModules: Array<Record<string, string>> = [];
```
FlorianRappl, line 6: Is this re-assigned somewhere? If so, should it be?

samuelmale: Yes, [here](https://github.com/openmrs/openmrs-esm-core/blob/a88c5c2daafa469722a3eb2f8863d9f5af256eee/packages/esm-app-shell/src/openmrs-backend-dependencies.ts#L13)

FlorianRappl: Why don't you just `push(...newModules)`? No need to reassign.

samuelmale: Done!

packages/esm-app-shell/src/openmrs-backend-dependencies.ts
```diff
@@ -3,14 +3,14 @@ import * as semver from "semver";
...
-let installedBackendModules: Array<Record<string, string>> = [];
+const installedBackendModules: Array<Record<string, string>> = [];
 const modulesWithMissingBackendModules: MissingBackendModules[] = [];
 const modulesWithWrongBackendModulesVersion: MissingBackendModules[] = [];
 
 export async function initInstalledBackendModules() {
   try {
     const response = await fetchInstalledBackendModules();
-    installedBackendModules = response.data.results;
+    installedBackendModules.push(response.data.results);
```
FlorianRappl, line 13: I thought these are many items (`Array<T>`)? If so then please don't add many items as one item. Spread them!

```ts
installedBackendModules.push(...response.data.results);
```

samuelmale: Aaahh sorry, I don't no how I missed this :)

samuelmale: Resolved!

FlorianRappl (approved):
Thanks 🍻 !

PR #22, opened 2020-10-29
Make implementer tools aware of the config schema

brandones:
Also refactor the concept lookup box

conf

### DIFF
--- a/packages/esm-implementer-tools/src/configuration/concept-search.tsx
+++ b/packages/esm-implementer-tools/src/configuration/concept-search.tsx
@@ -1,0 +1,100 @@
+import React, { useState, useEffect, useMemo } from "react";
+import { debounce, uniqueId } from "lodash";
+import {
+  fetchConceptByUuid,
+  performConceptSearch,
+} from "./concept-search.resource";
+import styles from "./concept-search.styles.css";
+
+interface ConceptSearchBoxProps {
+  setConcept: (concept) => void;
+}
+
+export function ConceptSearchBox({ setConcept }: ConceptSearchBoxProps) {
+  const [conceptName, setConceptName] = useState("");
+  const [searchTerm, setSearchTerm] = useState("");
+  const [searchResults, setSearchResults] = useState([]);
+  const [activeConceptUuid, setActiveConceptUuid] = useState<any>("");
+  const searchTimeoutInMs = 300;
+
+  const id = useMemo(() => uniqueId(), []);
+
+  const handleUuidChange = (concept) => {
+    setActiveConceptUuid(concept.uuid);
+    resetSearch();
+    setConcept(concept);
+  };
+
+  const resetSearch = () => {
+    setSearchTerm("");
+    setSearchResults([]);
+  };
+
+  const handleSearchTermChange = debounce((searchTerm) => {
+    setSearchTerm(searchTerm);
+  }, searchTimeoutInMs);
+
+  useEffect(() => {
+    if (activeConceptUuid) {
+      const conceptString = activeConceptUuid;
+
+      fetchConceptByUuid(conceptString).then(({ data }) => {
+        setConceptName(data.name.display);
+      });
+    }
+  }, [activeConceptUuid]);
+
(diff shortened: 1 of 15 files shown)


packages/esm-implementer-tools/src/configuration/value-editor.tsx
```diff
@@ -1,18 +1,20 @@
...
 
 export default function ValueEditor({
-  valueString,
+  element,
   handleSave,
   handleClose,
 }: ValueEditorProps) {
   const ref = useRef(null);
-  const [tmpValue, setTmpValue] = useState(valueString);
+  const [tmpValue, setTmpValue] = useState(JSON.stringify(element._value));
```
FlorianRappl, line 17: We could use `() => JSON.stringify(element._value)` to only stringify when an initial state is demanded.

packages/esm-implementer-tools/src/configuration/concept-search.tsx
```diff
@@ -0,0 +1,100 @@
...
+        setConceptName(data.name.display);
+      });
+    }
+  }, [activeConceptUuid]);
+
+  useEffect(() => {
+    const ac = new AbortController();
+
+    if (searchTerm && searchTerm.length >= 2) {
+      performConceptSearch(searchTerm).then(({ data: { results } }) => {
```
FlorianRappl, line 51: Should `ac` be used here?

packages/esm-implementer-tools/src/configuration/concept-search.tsx
```diff
@@ -0,0 +1,100 @@
...
+    setSearchResults([]);
+  };
+
+  const handleSearchTermChange = debounce((searchTerm) => {
+    setSearchTerm(searchTerm);
+  }, searchTimeoutInMs);
+
+  useEffect(() => {
+    if (activeConceptUuid) {
+      const conceptString = activeConceptUuid;
```
FlorianRappl, line 39: Why the alias?

packages/esm-implementer-tools/src/configuration/concept-search.tsx
```diff
@@ -0,0 +1,100 @@
...
+export function ConceptSearchBox({ setConcept }: ConceptSearchBoxProps) {
+  const [conceptName, setConceptName] = useState("");
+  const [searchTerm, setSearchTerm] = useState("");
+  const [searchResults, setSearchResults] = useState([]);
+  const [activeConceptUuid, setActiveConceptUuid] = useState<any>("");
+  const searchTimeoutInMs = 300;
+
+  const id = useMemo(() => uniqueId(), []);
+
+  const handleUuidChange = (concept) => {
```
FlorianRappl, line 22: Should we use `useCallback` here?

Same for the other handlers.

FlorianRappl (commented):
LGTM :beers:

packages/esm-implementer-tools/src/configuration/concept-search.tsx
brandones: This whole component was copy-pasted out of EditableValue (not to exonerate myself—I reviewed the original code!). I'll fix it now, anyway.

jonathandick:
Brandon do you mind adding a few comments on the significance of this ticket? why it was important? what now becomes possible because of this.

packages/esm-implementer-tools/src/configuration/concept-search.tsx
brandones: I don't think so—creating the function is not itself expensive, and re-creating the function shouldn't cause any unnecessary re-renders (I think—feel free to correct me here).

I have consolidated the other function into this one though.

brandones:
@jonathandick With this work, we can show the default, description, and validators for each config element directly in the interface.

Once we add a `type` property to config elements, we can make implementer-tools do specific fancy things for different types of config elements. i.e. concept search for concepts (until then, for the purpose of this PR, I have implemented `isConcept` as a completely evil and fragile and temporary hack).

PR #25, opened 2020-11-06
Enable running the openmrs tooling against the local repo without errors

manuelroemer:


packages/esm-devtools/webpack.config.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
         use: ["style-loader", "css-loader"],
       },
       {
-        test: /src\/.+\.css$/i,
+        test: /\.css$/,
```
manuelroemer, line 36: I made these changes in order to be able to build/prettify on Windows. Without these, I ran into errors. Just tell me if these changes are undesired - if so, I'll revert them.

FlorianRappl: I'd say its fine!

FlorianRappl (approved):
Awesome!

PR #26, opened 2020-11-06
Rework extension contexts to work based on the extensionSlotName instead of the extensionId / Improve extension registration in the app shell

manuelroemer:
* Allows auto-attaching extensions in `setupOpenMRS` via a `slot` property (see example below).
* Adds console warnings when an extension registration is incorrect.
* Fixes/Enables the extension context: It now works without having to manually attach the extension to the actual extension slot.

`setupOpenMRS` example:
```ts
function setupOpenMRS() {
  return {
    pages: [
      {
        load: () => import('./spa-order-basket-app'),
        route: /^patient\/.+\/drugorder\/basket/,
      },
    ],
    extensions: [
      {
        id: 'drugorder-widget',
        slot: 'patient-chart-dashboard-medications',
        load: () => import('./spa-medication-summary-extension'),
      },
      {
        id: 'order-basket-workspace',
        slot: '/patient/:patientUuid/drugorder/basket',
        load: () => import('./spa-order-basket-extension'),
      },
    ],
  };
}
```

### DIFF
--- a/packages/esm-extensions/src/extensions.ts
+++ b/packages/esm-extensions/src/extensions.ts
@@ -125,6 +125,13 @@ export async function getExtensionIdsForExtensionSlot(
   let extensionIds =
     attachedExtensionsForExtensionSlot[extensionSlotName] ?? [];
 
+  const extensionsMatchingUrlPattern = Object.keys(
+    extensions
+  ).filter((extensionId) =>
+    getExtensionComponent(extensionId).component?.matcher(extensionSlotName)
+  );
+  extensionIds.push(...extensionsMatchingUrlPattern);
+
   if (config.add) {
     extensionIds = extensionIds.concat(config.add);
   }
@@ -156,7 +163,7 @@ function getExtensionComponent(extensionName: string) {
       const routeProps = extensions[name].matcher(extensionName);
 
       if (routeProps) {
-        return { component: extensionName[name], routeProps };
+        return { component: extensions[name], routeProps };
       }
     }
   }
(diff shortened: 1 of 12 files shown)


FlorianRappl (approved):
Awesome!

packages/esm-extensions/src/extensions.ts
```diff
@@ -117,29 +114,50 @@ export function getExtensionRegistration(
...
+export async function getAttachedExtensionInfoForSlotAndConfig(
+  actualExtensionSlotName: string,
   moduleName: string
-): Promise<Array<string>> {
-  const config = await getExtensionSlotConfig(extensionSlotName, moduleName);
-  let extensionIds =
-    attachedExtensionsForExtensionSlot[extensionSlotName] ?? [];
+): Promise<Array<AttachedExtensionInfo>> {
+  const config = await getExtensionSlotConfig(
+    actualExtensionSlotName,
```
brandones, line 132: Sorry I'm reviewing this late, apparently.

This should be the `attachedExtensionSlotName`. Configuration should not depend on e.g. the particular patient being viewed. e.g., we want a configuration like

```js
extensions: {
  "/mySlot/:uuidParameter": {
    remove: ["some-extension"]
  }
}
```

rather than like

```js
extensions: {
  "/mySlot/213da954-87a2-432d-91f6-a3c441851726": {
    remove: ["some-extension"]
  }
}
```

FlorianRappl: Yeah I think this is a good point. 
cc @manuelroemer

PR #27, opened 2020-11-07
MF-366 	Config elements should have `type` 

brandones:
https://issues.openmrs.org/browse/MF-366

This removes support for the `is****` validators, replacing them with `type` declarations in the schema.

All modules' configs must be updated simultaneously because this will break everything. I will change this to a normal PR once PRs against all the other modules are ready.

### DIFF
--- a/packages/esm-config/src/module-config/module-config.ts
+++ b/packages/esm-config/src/module-config/module-config.ts
@@ -119,8 +119,8 @@ export async function getImplementerToolsConfig(): Promise<object> {
 }
 
 function getSchemaWithValuesAndSources(schema) {
-  if (schema.hasOwnProperty("default")) {
-    return { ...schema, _value: schema.default, _source: "default" };
+  if (schema.hasOwnProperty("_default")) {
+    return { ...schema, _value: schema._default, _source: "default" };
   } else {
     return Object.fromEntries(
       Object.entries(schema).map(([key, subtree]) => [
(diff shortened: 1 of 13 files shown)


FlorianRappl (approved):
LGTM :beers:!

brandones:
Thanks @FlorianRappl ! I just added one more commit, to enable Typescript validation for theses "types".

packages/esm-config/src/module-config/module-config.ts
```diff
@@ -646,18 +660,17 @@ export interface ConfigObject extends Object {
...
-  "Array",
-  "Boolean",
-  "ConceptUuid",
-  "Number",
-  "Object",
-  "String",
-  "UUID",
-] as const;
-export type ConfigValueType = typeof configValueTypes[number]; // = "Array" | "Boolean" | "ConceptUuid" | ...
+export enum Type {
```
FlorianRappl, line 665: This generates a runtime object. Is this what you want? To have the correct typing above you could have also chosen `as const` for the individual constant strings.

Difference:

```ts
const foo = {
  bar: 'bar',
}; // { bar: string }


const foo = {
  bar: 'bar' as const,
}; // { bar: 'bar' }
```

brandones: Yes, because when I made the typing of the ConfigSchema more strict, Typescript was unwilling to accept a schema like

```jsx
      { bar: { default: 0, type: "Number" } }
```

because "Number" would be taken as a string, rather than the const literal `"Number"`.

So now the types can be used (in a type-safe way) as

```jsx
      { bar: { default: 0, type: Type.Number } }
```

or

```jsx
      { bar: { default: 0, type: "Number" as Type } }
```

brandones:
Made a few more small changes while updating the other modules: https://github.com/openmrs/openmrs-esm-core/pull/27/commits/623914be1469a437441195debf29647801a758c5

FlorianRappl (approved):
LGTM! 🍻

PR #29, opened 2020-11-10
[WIP] Rework the extension slot config to use the attached extension slot name

manuelroemer:
**This is WIP. The draft PR is only here to discuss the general idea. Checklist of missing things:**

- [ ] Feedback
- [ ] Adapt tests
- [ ] Code cleanup of the changes (quite rough atm)

As a follow-up to https://github.com/openmrs/openmrs-esm-core/pull/26#discussion_r520139010_, this PR updates how the extension slot config is handled. After looking at the situation, I came to the conclusion that it's not possible to pass an `attachedExtensionSlotName` to the config's `getExtensionSlotConfig` function. This is because we don't have any attached slot name at this point in time ([this function](https://github.com/openmrs/openmrs-esm-core/pull/26/files/969da03fe318d56ad1ce88f730514355b430e777#diff-5acad908a3628b9641be28a5c4767b2279c16245b6106b344c82febba61e21f1R127) is atm called in the root of the `ExtensionSlotReact`, a point where we only have the `actualExtensionSlotName` - finding out what the corresponding `attachedExtensionSlotName`s are is exactly the point of `getAttachedExtensionInfoForSlotAndConfig`).

My initial approach to handle such situations is a slight rework of how the config is handled. Instead of handling **one** `ExtensionSlotConfig` which matches the extension slot name _exactly_, I'd from now on iterate through **all** extension slot configs and check whether their name pattern match the `actualExtensionSlotName`. If so, continue with the previous logic.

I'd love some feedback on this approach as I'm not sure whether this is the ideal way to do this.
The whole scenario also creates a new situation where it is theoretically possible to have multiple extension slots matching the current `actualExtensionSlotName` (e.g. `/slot/:foo` and `/slot/:bar`). For now, I simply allow both of these to render their extensions in order, but I guess that this is up for debate...

CC @FlorianRappl @brandones

### DIFF
--- a/packages/esm-config/src/module-config/module-config.ts
+++ b/packages/esm-config/src/module-config/module-config.ts
@@ -198,16 +198,17 @@ export function clearTemporaryConfig(): void {
 /**
  * @internal
  */
-export async function getExtensionSlotConfig(
+export async function getExtensionSlotConfigs(
   slotName: string,
   moduleName: string
-): Promise<ExtensionSlotConfigObject> {
+): Promise<Record<string, ExtensionSlotConfigObject>> {
   await loadConfigs();
   const moduleConfig = mergeConfigsFor(moduleName, getProvidedConfigs());
-  const inputConfig = moduleConfig?.extensions?.[slotName] ?? {};
-  validateExtensionSlotConfig(inputConfig, moduleName, slotName);
-  const config = processExtensionSlotConfig(inputConfig);
-  return config;
+  const allExtensionSlotConfigs: Record<string, ExtensionSlotConfig> = moduleConfig?.extensions ?? {};
+  for (const config of Object.values(allExtensionSlotConfigs)) {
+    validateExtensionSlotConfig(config, moduleName, slotName);
+  }
+  return processExtensionSlotConfigs(allExtensionSlotConfigs);
 }
 
 function validateExtensionSlotConfig(
@@ -285,18 +286,21 @@ function validateExtensionSlotConfig(
   }
 }
 
-function processExtensionSlotConfig(
-  config: ExtensionSlotConfig
-): ExtensionSlotConfigObject {
-  const result: ExtensionSlotConfigObject = {};
-  if (config.remove) {
-    result.remove = config.remove;
-  }
-  if (config.order) {
-    result.order = config.order;
-  }
-  if (config.add) {
-    result.add = config.add.map((e) => e.extension);
+function processExtensionSlotConfigs(
+  allExtensionSlotConfigs: Record<string, ExtensionSlotConfig>
+): Record<string, ExtensionSlotConfigObject> {
+  const result: Record<string, ExtensionSlotConfigObject> = {};
+  for (const [key, config] of Object.entries(allExtensionSlotConfigs)) {
+    result[key] = {};
+    if (config.remove) {
+      result[key].remove = config.remove;
+    }
+    if (config.order) {
+      result[key].order = config.order;
+    }
+    if (config.add) {
+      result[key].add = config.add.map((e) => e.extension);
+    }
   }
   return result;
 }
--- a/packages/esm-extensions/src/extensions.ts
+++ b/packages/esm-extensions/src/extensions.ts
@@ -1,6 +1,6 @@
 import { mountRootParcel } from "single-spa";
 import { pathToRegexp, Key } from "path-to-regexp";
-import { getExtensionSlotConfig } from "@openmrs/esm-config";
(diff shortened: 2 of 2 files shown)


brandones:
Why don't you just put the call to `getExtensionSlotConfigs`after the call to `getAttachedExtensionInfoForSlot`?

manuelroemer:
> Why don't you just put the call to `getExtensionSlotConfigs`after the call to `getAttachedExtensionInfoForSlot`?

Because then adding extensions via the config might not work. As an example: When nothing is attached to anything (i.e. `attachedExtensionsForExtensionSlot` is `{}`), but we've got the following config:

```js
extensions: {
  "/mySlot/:uuidParameter": {
    add: ["some-extension"]
  }
}
```

the `getAttachedExtensionInfoForSlot` function will return nothing. This means that we've got no slot with which we could call `getExtensionSlotConfigs`, even though the config clearly defines a matching slot name. This means that nothing will be displayed at the end of the day, even though it's correctly configured.

In my opinion, we have to check for each config entry, whether the configured slot name matches the actual one (either as a strict match, or as a path-regex match). But if there's an easier solution/if I'm missing something, feel free to tell me!

brandones:
It sounds like you should just decompose `getAttachedExtensionInfoForSlot` into two functions: `getAttachedExtensionIdsForSlot` and `getAttachedSlotNameForSlot`.

The attached name of a slot should be available to it whether or not it has extensions. Meanwhile the `extensionInfo` returned by `getAttachedExtensionInfoForSlot` should have the same values for `actualExtensionSlotName` and `attachedExtensionSlotName` for every item in the returned array (unless I'm missing something).

FlorianRappl:
@manuelroemer and I looked into the problem in full detail. We think what needs to be done is:

1. The logic for retrieving the actual extensions to render should not be in the React component; it should be (at best) in the generic extension renderer.
2. Actually, there should not be any logic at all involved in rendering the extension. Instead, all the extension information (which extension slots are there, what extensions are available, which extensions are assigned to what slots) should be kept up-to-date.
3. We will try to implement this using a global state container, which therefore updates (and sends a signal) when it changes. This signal will result in a re-rendering of the extension slot - keeping it up to date independent of what changed when.
4. We will still distinguish between the actual extension slot (currently used; potentially resolved with parameters) and the attached extension slot (the originally defined one - having the URL blueprint, if any).
5. There is a 1-1 relationship between definitions for an extension slot in a config, and the extension slot. Thus if multiple would match (which should never occur) we'll just take the first match. All others are ignored (implementation error / undefined behavior).

As next step we will start placing the extension slots in a state container. This will be done step by step and not in one sweep. The original (internal) behavior will remain as-is as long as possible.

Finally, we identified that the config definition is potentially not well placed. We think that the configuration may / should be placed already in the top-level module. Otherwise, parts of the config can be only read (and defined) when (a part of) the application activates.

How does that sound @brandones ?

brandones:
@FlorianRappl That all sounds reasonable to me. Thank you and @manuelroemer for doing such a thorough analysis!

One question, for clarification—by "configuration should be placed in the top-level module," you mean that `extensions` should be a top-level key, on level with the module names? So a config file might look like

```js
{
  "@openmrs/esm-login-app": { logo: { alt: "Acme" } },
  "@openmrs/esm-home-app": { foo: "bar" },
  extensions: {
    "/mySlot/:uuidParameter": {
      add: ["some-extension"]
    }
  }
}
```

?

FlorianRappl:
> by "configuration should be placed in the top-level module," you mean that extensions should be a top-level key, on level with the module names? 

Yeah sorry that was not really well written there @brandones . What I meant was that we have the  `defineConfig` in the page usually, i.e., it gets loaded / run / executed *after* activation of a microfrontend. Surely, this (`defineConfig`) could be also in an extension, but then again - it only gets run when the extension is activated.

Would it make sense to `defineConfig` / apply the config already in the top-level module (i.e. the one that exports the `setupOpenMRS` function)? This way, the configuration would be loaded / used at a well-known point in time (at the application's startup) and it could influence already activation functions (e.g., not exporting certain extensions, changing paths, ...). What is your opinion?

brandones:
Assuming you mean `defineConfigSchema`, yes, I think that seems totally reasonable! I don't think that will require any core change, just a change in how `defineConfigSchema` gets used in microfrontends.

FlorianRappl:
> Assuming you mean defineConfigSchema, yes, I think that seems totally reasonable! I don't think that will require any core change, just a change in how defineConfigSchema gets used in microfrontends.

Yes indeed - just wanted to check first what is your opinion on that matter. Awesome - I'll go ahead here then!

PR #32, opened 2020-11-11
MF-320 All module-config keywords should be underscore-prefixed 

brandones:
This is the other big, breaking config change (alongside https://github.com/openmrs/openmrs-esm-core/pull/27 ).

This PR is against https://github.com/openmrs/openmrs-esm-core/pull/27 just for the clean diff. It's built on top of it. Once this PR is approved, I'll merge it into that branch. Then, once PRs are out against all the other modules to update their config schemas as will be necessary, we can merge https://github.com/openmrs/openmrs-esm-core/pull/27 to master.

I think it would also be a moment to major-version esm-core. Unless we're still informally considering esm-core 0-versioned.

### DIFF
--- a/packages/esm-implementer-tools/src/configuration/editable-value.component.tsx
+++ b/packages/esm-implementer-tools/src/configuration/editable-value.component.tsx
@@ -1,10 +1,13 @@
 import React, { useState, useEffect, useRef } from "react";
 import { isEqual } from "lodash";
-import { setTemporaryConfigValue, ConfigValue } from "@openmrs/esm-config";
+import {
+  setTemporaryConfigValue,
+  ConfigValue,
+  Validator,
+} from "@openmrs/esm-config";
 import styles from "./editable-value.styles.css";
 import ValueEditor from "./value-editor";
 import { useGlobalState } from "../global-state";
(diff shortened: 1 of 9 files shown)


packages/esm-implementer-tools/src/configuration/editable-value.component.tsx
```diff
@@ -14,9 +14,9 @@ export interface EditableValueProps {
 export interface ConfigValueDescriptor {
   _value: ConfigValue;
   _source: string;
-  default: ConfigValue;
-  description?: string;
-  validators?: Array<Function>;
+  _default: ConfigValue;
+  _description?: string;
+  _validators?: Array<Function>;
```
FlorianRappl, line 19: Could we be more specific here? Is this, e.g., `Array<ConfigValidator>` where `type ConfigValidator = (config: any) => boolean` or similar?

FlorianRappl (approved):
LGTM - thanks 🍻 !

packages/esm-implementer-tools/src/configuration/editable-value.component.tsx
brandones: Oh yeah, great idea!

PR #34, opened 2020-11-14
Remove esnext from typescript required libs

brandones:
I can't get esm-login to build against the new version of esm-core because of `fromEntries` and `flatMap` usages.


brandones:
This is failing a check because apparently [matchAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) also depends on esnext?

denniskigen (approved):
Looks good to me @brandones! Thanks.

FlorianRappl:
> This is failing a check because apparently matchAll also depends on esnext?

Yes. Just change the tsconfig and you are good.

packages/esm-config/package.json
```diff
@@ -34,7 +34,9 @@
   },
   "homepage": "https://github.com/openmrs/openmrs-esm-core#readme",
   "dependencies": {
+    "@types/string.prototype.matchall": "^4.0.0",
     "ramda": "^0.26.1",
+    "string.prototype.matchall": "^4.0.2",
```
FlorianRappl, line 39: Not needed.

https://caniuse.com/mdn-javascript_builtins_string_matchall

FlorianRappl (changes requested):
I don't get this PR.

If we start shimming this we should really think about our browser strategy and I'd see we shim a lot more APIs, too (e.g., `async` / `await` using regenerator runtime). But IIRC our goal is not old browsers but current ones. Here, this API is fully supported.

brandones:
@FlorianRappl Here's the problem I'm trying to address. I added `esnext` to the `esm-config` Typescript config so that I could use `fromEntries`. However, this seems to have "poisoned" the library, so that now any *client code* to esm-config has to also include `esnext` as a Typescript lib.

So this PR is my attempt to remove `esnext` from the esm-config Typescript config. When I removed it, Typescript compilation started to fail on `fromEntries`, `matchAll`, and `flatMap`. The shims are not there to get the library to work in the browser, but rather to get Typescript to compile without the `esnext` lib.

Please let me know what's the better way to approach this!

FlorianRappl:
> @FlorianRappl Here's the problem I'm trying to address. I added esnext to the esm-config Typescript config so that I could use fromEntries. However, this seems to have "poisoned" the library, so that now any client code to esm-config has to also include esnext as a Typescript lib.

> So this PR is my attempt to remove esnext from the esm-config Typescript config. When I removed it, Typescript compilation started to fail on fromEntries, matchAll, and flatMap. The shims are not there to get the library to work in the browser, but rather to get Typescript to compile without the esnext lib.

> Please let me know what's the better way to approach this!

I think the whole approach could be avoided as this (in the end) is a type error in one of the libs. So just ignore type errors from the libs (which is a sensible approach as we don't care if the libs themselves are correctly typed).

However, going broader I'd say that the apps (e.g., esm-login) "lie" when they don't use "esnext". After all, its what the core targets anyway, so they can also target it. Worst case they don't make use of the available APIs. Best case they don't feel limited.

So again, I'd just change the apps tsconfig (either `skipLibCheck` or adding the necessary ts config entry). Its in my opinion the most honest approach and its the one I applied to the apps in the past, too.

brandones:
Okay, sounds good. Thanks for clarifying that. I'll leave all this as-is.

PR #35, opened 2020-11-14
Fix typescript error from esm-login

brandones:
A much more conservative change than https://github.com/openmrs/openmrs-esm-core/pull/34 , which should allow esm-login to build.

Right now, `npm run typescript` in esm-login fails with

```
node_modules/@openmrs/esm-config/src/module-config/module-config.ts:597:35 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Config'.
  Index signature is missing in type '{}'.

597             setDefaults(elements, conf)
                                      ~~~~
```

I don't know why Typescript in esm-core itself works fine, but it does. This same failure causes `npm run build` in esm-login to fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "raw-loader": "^3.1.0",
     "style-loader": "^1.0.0",
     "ts-loader": "^8.0.3",
-    "typescript": "^4.0.2",
+    "typescript": "^4.0.5",
     "webpack": "^4.44.1",
     "webpack-cli": "^3.3.12",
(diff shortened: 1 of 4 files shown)


FlorianRappl:
> I don't know why Typescript in esm-core itself works fine, but it does.

Are they using the same version of TS?

brandones:
esm-core is on 4.0.3, esm-login is on 4.0.5. I'll upgrade the esm-core one.

PR #36, opened 2020-11-19
Update importmap to latest from openmrs-spa.org

brandones:


packages/esm-app-shell/src/assets/importmap.json
```diff
@@ -1,13 +1,13 @@
 {
```
jonathandick, line 1: Are we going to need to manually manage this process of keeping the import map for this repo up to date? 

@FlorianRappl ?

FlorianRappl: For now yes. This is just the default import map - not the one being used, e.g., by the ref app.

FlorianRappl (approved):
LGTM - thanks :beers:!

PR #39, opened 2020-11-20
Fix auth failure redirect

brandones:


packages/esm-api/jest.config.js
```diff
@@ -6,8 +6,7 @@ module.exports = {
     "lodash-es": "lodash",
     "@openmrs/esm-error-handling":
       "<rootDir>/__mocks__/openmrs-esm-error-handling.mock.ts",
-    "@openmrs/esm-config":
-      "<rootDir>/__mocks__/openmrs-esm-module-config.mock.ts",
+    "@openmrs/esm-config": "<rootDir>/__mocks__/openmrs-esm-config.mock.ts",
```
FlorianRappl, line 9: This was not committed. I only see that the ...module-config.mock.ts was removed, but I don't see the file added (or the original one renamed).

FlorianRappl (commented):
I think this is a good change.

packages/esm-api/jest.config.js
brandones: Whoops! Good catch, thanks. Fixed.

FlorianRappl (approved):
LGTM - thanks :beers:!

PR #42, opened 2020-11-25
[MF-372] Carbonize toast component

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-372

**Demo:**

**Code Example:**

```javascript
  showToast({description: <span>Demo error <a href="#example">See details</a></span>, kind: "error"});

```
**The above would yield something like:**
Screenshot 2020-11-25 at 23 40 38


samuelmale:
@FlorianRappl kindly checkout this PR when you get some time.

packages/esm-styleguide/package.json
```diff
@@ -34,6 +34,9 @@
   },
   "homepage": "https://github.com/openmrs/openmrs-esm-core#readme",
   "dependencies": {
+    "carbon-components": "^10.23.2",
```
FlorianRappl, line 37: `carbon-components` and `carbon-icons` are dev/peer dependencies. They come with the shell.

packages/esm-styleguide/src/toasts/active-toasts.component.tsx
```diff
@@ -1,11 +1,12 @@
 import React from "react";
 import { toastsSubject } from "./toasts";
-import Toast from "./toast.component";
+import { Toast } from "./toast.component";
 
 export default function ActiveToasts() {
-  const [toasts, setToasts] = React.useState([]);
+  const [toasts, setToasts] = React.useState(new Array());
```
FlorianRappl, line 6: Not sure why you use `new Array()`, but the literal (`[]`) is preferred.

packages/esm-styleguide/src/toasts/toast.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+};
+
+interface ToastProps {
+  toast: ToastDescriptor;
+  closeToastRef: any;
+  isClosing: any;
+}
+
+export interface ToastDescriptor {
+  id?: Number;
```
FlorianRappl, line 16: use `number`, `Number` is the constructor

packages/esm-styleguide/src/toasts/toast.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+  kind?: ToastType;
+  title?: String;
```
FlorianRappl, line 18: Use `string`

packages/esm-styleguide/src/toasts/toasts.tsx
```diff
@@ -2,13 +2,16 @@ import React from "react";
 import ReactDOM from "react-dom";
 import { Subject } from "rxjs";
 import ActiveToasts from "./active-toasts.component";
+import { ToastDescriptor } from "./toast.component";
+import { isEmpty } from "lodash";
```
FlorianRappl, line 6: `import isEmpty from 'lodash-es/isEmpty'`

packages/esm-styleguide/package.json
```diff
@@ -34,6 +34,9 @@
...
+    "carbon-components-react": "^7.22.0",
+    "carbon-icons": "^7.0.7",
     "kremling": "^2.0.1",
     "systemjs-webpack-interop": "^2.1.2"
```
FlorianRappl, line 41: I miss a dependency to `lodash-es` here.

packages/esm-styleguide/webpack.config.js
```diff
@@ -56,7 +56,7 @@ module.exports = {
   externals: Object.keys(peerDependencies),
   resolve: {
     modules: ["node_modules"],
-    extensions: [".js", ".jsx", ".json"],
+    extensions: [".js", ".jsx", ".json", ".tsx"],
```
FlorianRappl, line 59: For completeness we could also include `.ts` here

packages/esm-styleguide/webpack.config.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
         ],
       },
       {
-        test: /\.(js|jsx)$/,
+        test: /\.(js|jsx|tsx)$/,
```
FlorianRappl, line 37: `ts|tsx`

packages/esm-styleguide/src/toasts/active-toasts.component.tsx
samuelmale: Because I run into a Typescript error:

```shell

/Users/samuel/git-mirror/openmrs-esm-core/packages/esm-styleguide/src/toasts/active-toasts.component.tsx
../esm-styleguide/src/toasts/active-toasts.component.tsx
[tsl] ERROR in /Users/samuel/git-mirror/openmrs-esm-core/packages/esm-styleguide/src/toasts/active-toasts.component.tsx(14,17)
      TS2345: Argument of type 'unknown[]' is not assignable to parameter of type 'SetStateAction<never[]>'.
  Type 'unknown[]' is not assignable to type 'never[]'.
    Type 'unknown' is not assignable to type 'never'.
```

From this stalkoverflow: https://stackoverflow.com/questions/52054140/why-push-shows-argument-of-type-any-is-not-assignable-to-parameter-of-type/56394664#56394664

packages/esm-styleguide/package.json
samuelmale: Resolved

packages/esm-styleguide/src/toasts/toast.component.tsx
samuelmale: Resolved

packages/esm-styleguide/src/toasts/toast.component.tsx
samuelmale: Resolved

packages/esm-styleguide/package.json
samuelmale: Resolved

packages/esm-styleguide/webpack.config.js
samuelmale: Resolved

packages/esm-styleguide/webpack.config.js
samuelmale: Resolved

packages/esm-styleguide/src/toasts/active-toasts.component.tsx
FlorianRappl: This is not ideal. Just because there is a warning light you unmount the whole dashboard. Not good.

Instead, why don't you use `React.useState<T>([])` and specify `T` to be `Array<ToastDescriptor>`?

packages/esm-styleguide/src/toasts/toast.component.tsx
```diff
@@ -0,0 +1,80 @@
...
+  | "success"
+  | "warning"
+  | "warning-alt";
+
+export const Toast: React.FC<ToastProps> = ({
+  toast,
+  closeToast,
+  isClosing,
+}) => {
+  const { millis, title, description, kind, action } = Object.assign(
```
FlorianRappl, line 36: Unnecessary.

Look into `master` how its done there. No need for `Object.assign` etc.

samuelmale: If we don't assign the props to one object, that means we shall have two declarations ie. 

```javascript

  const { title, description, kind, action } = toast;
  const { millis } = defaultOptions;

```

Or access `millis` property directly from `defaultOptions` constant?

samuelmale: Updated

packages/esm-styleguide/src/toasts/active-toasts.component.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
-import Toast from "./toast.component";
+import { Toast, ToastDescriptor } from "./toast.component";
 
 export default function ActiveToasts({ subject }) {
-  const [toasts, setToasts] = React.useState([]);
+  const [toasts, setToasts] = React.useState<ToastDescriptor[]>([]);
```
FlorianRappl, line 5: Please use `Array<ToastDescriptor>` (as I wrote). We should always use `Array<T>` in TS. Makes things much more readable. `[]` in TS should be reserved for tuples to avoid confusion with them.

packages/esm-styleguide/src/toasts/toast.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+  | "success"
+  | "warning"
+  | "warning-alt";
+
+export const Toast: React.FC<ToastProps> = ({
+  toast,
+  closeToast,
+  isClosing,
+}) => {
+  const { title, description, kind, action } = toast;
```
FlorianRappl, line 36: I thought millis is one of the options of `toast`? Should it be? Why destructure the default options (or call them default options) if they are not settable from the outside?

```ts
const { millis = defaultOptions.millis, description, title, kind, action } = toast;
```

packages/esm-styleguide/src/toasts/active-toasts.component.tsx
samuelmale: Thanks for the clarification :)

samuelmale: Updated

packages/esm-styleguide/src/toasts/toast.component.tsx
samuelmale: Updated

packages/esm-styleguide/src/toasts/toast.component.tsx
```diff
@@ -33,12 +34,15 @@ export const Toast: React.FC<ToastProps> = ({
   closeToast,
   isClosing,
 }) => {
-  const { title, description, kind, action } = toast;
-  const { millis } = defaultOptions;
-
+  const {
+    millis = toast.millis ?? defaultOptions.millis,
```
FlorianRappl, line 38: hehe not sure why the ceremony;

```ts
millis = defaultOptions.millis
```

This is all you need.

Consider the following code:

```ts
const defaultFoo = '';
const options: { foo?: string } = {};
const { foo = defaultFoo } = options;
```

The result in ES5 is:

```js
var defaultFoo = '';
var options = {};
var _a = options.foo, foo = _a === void 0 ? defaultFoo : _a;
```

Let's compare this one to:

```ts
const defaultFoo = '';
const options: { foo?: string } = {};
const { foo = options.foo ?? defaultFoo } = options;
```

which transpiles to

```ts
var _a;
var defaultFoo = '';
var options = {};
var _b = options.foo, foo = _b === void 0 ? (_a = options.foo) !== null && _a !== void 0 ? _a : defaultFoo : _b;
```

While the second one does a lot more, it has 100% the same effect.

samuelmale: Interesting, let even get some coffee :) 

(Updated)

dkayiwa: > some coffee

Not good for your health. 👎

samuelmale: > Not good for your health. 👎

Sure @dkayiwa, I'm taking caffein free coffee 😄

samuelmale:
@FlorianRappl I updated this PR, any chance you could have a look?

FlorianRappl (approved):
LGTM - thanks @samuelmale !

PR #43, opened 2020-11-28
Some refactoring of the extension and extension slot configuration system

brandones:
The functional changes are
- That the extension slot config API has been simplified
- That configuration can now be passed through to extensions using `getExtensionConfig`/`useExtensionConfig`
- To achieve the above, there is now an `openmrsExtensionDecorator` which should be used in place of `openmrsRootDecorator` for extensions.
- Added `getExtensionLifecycle` and `getAsyncExtensionLifecycle` for extensions, to parallel `getRootLifecycle`.

I've made some minor changes in naming and function signatures just to make the code clearer.

Screenshot from 2020-11-27 19-15-34


packages/esm-config/src/module-config/module-config.ts
```diff
@@ -719,12 +705,14 @@ export enum Type {
...
 export interface ExtensionSlotConfig {
-  add?: Array<{
-    extension: string;
-    config: {};
-  }>;
+  add?: string[];
   remove?: string[];
   order?: string[];
+  configure?: ExtensionSlotConfigureValueObject;
+}
```
brandones, line 712: This is the important change to the config API.

packages/esm-extensions/src/extensions.ts
```diff
@@ -12,11 +12,11 @@ interface ExtensionRegistration extends ExtensionDefinition {
 }
 
 export interface ExtensionStore {
-  slots: Record<string, ExtensionSlotDefinition>;
+  slots: Record<string, ExtensionSlotInfo>;
   extensions: Record<string, ExtensionRegistration>;
 }
 
-export interface ExtensionSlotDefinition {
+export interface ExtensionSlotInfo {
```
brandones, line 19: Renamed this. It doesn't make sense for a "definition" to contain state. And I didn't want to call it `ExtensionSlotState`, since I'm not sure it makes sense for something called "state" to contain basic definitional information about the slot. So we climb a layer of abstraction and wind up with `ExtensionSlotInfo`.

packages/esm-extensions/src/extensions.ts
```diff
@@ -53,13 +53,12 @@ export interface ExtensionSlotDefinition {
...
   /**
-   * Returns a value indicating whether extensions of this extension slot definition can be rendered
-   * into an arbitrary extension slot component with the specified `actualExtensionSlotName`.
+   * Returns whether the given extension slot name corresponds to this ExtensionSlotInfo.
    * @param actualExtensionSlotName The actual extension slot name into which the extensions might be rendered.
    * For URL like extension slots, this should be the name where parameters have been replaced with actual values
    * (e.g. `/mySlot/213da954-87a2-432d-91f6-a3c441851726`).
    */
-  canRenderInto(actualExtensionSlotName: string): boolean;
+  matches(actualExtensionSlotName: string): boolean;
```
brandones, line 61: I think that this suggests a mental model which is easier to think about.

packages/esm-extensions/src/extensions.ts
```diff
@@ -12,11 +12,11 @@ interface ExtensionRegistration extends ExtensionDefinition {
 }
 
 export interface ExtensionStore {
-  slots: Record<string, ExtensionSlotDefinition>;
+  slots: Record<string, ExtensionSlotInfo>;
```
brandones, line 15: It should be noted that since extension slot names are not guaranteed to be unique across the application, this state element should be indexed both by module name and slot name. Otherwise there is no way to refer to a unique slot in configuration.

packages/esm-extensions/src/extensions.ts
```diff
@@ -176,28 +175,37 @@ export function renderExtension(
...
+      );
+      const extensionContextProps = {
+        _extensionContext: {
+          extensionId,
+          actualExtensionSlotName,
+          attachedExtensionSlotName,
+          extensionSlotModuleName,
+          extensionModuleName: extensionRegistration.moduleName,
+        },
+      };
```
brandones, line 201: These props are used to "forward" the ExtensionContext info across the single-spa mount point.

jonathandick (approved):
This looks good to me. 

Does this create any breaking changes? Or for existing extensions, do we need to change file structure/how it is decorated? If so, can you identify one and create PR for how to fix it so others can follow the pattern?

Thanks and great work. 

JJ

brandones:
Thanks @jonathandick !

It does not break anything. `useExtensionConfig` never worked. Now it will work if you declare your extensions using the new `getExtensionLifecycle` function.

This does not fix the issue where `add`, `remove`, and `order` are broken. That will need to be fixed separately.

PR #44, opened 2020-11-29
Make useConfig also work for extension config

brandones:


packages/esm-react-utils/src/useConfig.test.tsx
```diff
@@ -91,6 +93,176 @@ describe(`useConfig`, () => {
   });
 });
 
+describe(`useConfig in an extension`, () => {
```
brandones, line 96: Tests copy-pasted from `useExtensionConfig.test.tsx`

jonathandick (approved):
Thanks for moving all logic into one hook. I think this is will be easier to understand.

brandones:
Yeah, thanks for the good idea!

PR #45, opened 2020-12-01
Create activePage element in global app state

brandones:
This creates an app state variable `activePage` which changes according to what page is active.

By "page" I mean a page defined in `setupOpenMRS`. I have adjusted the app registration code to only update the `activePage` when the page has been declared using the `pages` array in `setupOpenMRS`. This allows that we can declare things like esm-primary-navigation which should not affect `activePage` by declaring them with the `lifecycle` and `activate` parameters.

What this allows immediately is to declare slots whose name depends on the active page. This allows the extension framework to stay completely declarative while being sensitive to routes.

After [updating the left nav slot](https://github.com/openmrs/openmrs-esm-primary-navigation/pull/44) and doing

```
attach("nav-menu-home", "active-visits-link");
attach("nav-menu-patient-chart", "reports-link");
```

we obtain

home-ext
pt-chart-ext

### DIFF
--- a/packages/esm-api/src/state.ts
+++ b/packages/esm-api/src/state.ts
@@ -54,12 +54,23 @@ export function getGlobalStore<TState = any>(
   return available.value;
 }
 
-export interface AppState {}
+export interface AppState {
+  /**
+   * The simplified route of the currently active page.
+   * Pages are declared in `setupOpenMRS`, as a `pages` array.
+   *
+   * For example, for route `/^patient\/.+\/chart/` the activePage
+   * will be `patient-chart`.
+   */
+  activePage: string | null;
+}
 
 export function createAppState(initialState: AppState) {
   return createGlobalStore("app", initialState);
 }
 
 export function getAppState() {
-  return getGlobalStore<AppState>("app", {});
+  return getGlobalStore<AppState>("app", {
+    activePage: null,
+  });
 }
--- a/packages/esm-app-shell/src/apps.ts
+++ b/packages/esm-app-shell/src/apps.ts
@@ -1,3 +1,4 @@
+import { getAppState } from "@openmrs/esm-api";
 import {
   attach,
   ExtensionDefinition,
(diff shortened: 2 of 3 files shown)


jonathandick (approved):
LGTM

packages/esm-api/src/state.ts
```diff
@@ -54,12 +54,23 @@ export function getGlobalStore<TState = any>(
   return available.value;
 }
 
-export interface AppState {}
+export interface AppState {
+  /**
+   * The simplified route of the currently active page.
+   * Pages are declared in `setupOpenMRS`, as a `pages` array.
```
FlorianRappl, line 60: Hm the `pages` will not work due to Single SPA. Unfortunately, it does not allow registering multiple activators, which is an oversight. So while technically still true, there is no advantage over using the current approach (using `activate` and `load`).

packages/esm-app-shell/src/apps.ts
```diff
@@ -104,3 +109,42 @@ function tryRegisterExtension(
     attach(slot, name);
   }
 }
+
+function createActivePageSetterHook(route: string, activator: Activator) {
+  const simplifiedRoute = simplify("" + route); // cast to string
```
FlorianRappl, line 114: Why do we need this? According to the declaration `route` is already a string.

FlorianRappl (commented):
Code looks fine to me - just again I don't understand why we need this. From the other change it seems that this is then used in `primary-navigation`, however, I do not feel the need for this (nor do I think we would benefit from it; I may be wrong here so I just want to know the reasoning).

packages/esm-api/src/state.ts
brandones: How were you imaging `pages` should be used? It seems like the following should work fine:

```jsx
  pages: [
    {
      load: () => getAsyncLifecycle(() => import('./page1.component'), options),
      route: 'page1'
    },
    {
      load: () => getAsyncLifecycle(() => import('./page2.component'), options),
      route: 'page2'
    }
  ]
```

packages/esm-app-shell/src/apps.ts
brandones: Yeah, I'm wondering why that declaration is like that? I believe you wrote it. Why are regexes not necessary?

brandones:
@FlorianRappl Up to you whether you think that this is useful apart from https://github.com/openmrs/openmrs-esm-primary-navigation/pull/44 . Feel free to close this PR if you don't think so.

brandones:
Actually—if you think this is still useful, please feel free to reopen :)

PR #46, opened 2020-12-02
Fixed Extension Config

FlorianRappl:
There is something larger that I have been working on locally, but I first want to get the config reestablished.

After all, this introduces the `useExtensionSlot` hook and tries to avoid rerenders quite a bit. It seems to work fine so far, but I don't know how dynamically we change (or expected changes) to a configuration. Presumably, it would be good to have the config also in a state container to allow listening to changes. Then config changes can reactively trigger extension slot updates. But I first wanted to hear an opinion on this one / plus get feedback if this is working at all as you expect.

### DIFF
--- a/packages/esm-extensions/src/extensions.ts
+++ b/packages/esm-extensions/src/extensions.ts
@@ -1,6 +1,7 @@
 import { mountRootParcel } from "single-spa";
 import { pathToRegexp, Key } from "path-to-regexp";
 import { createGlobalStore } from "@openmrs/esm-api";
+import { getExtensionSlotConfig } from "@openmrs/esm-config";
 
 export interface ExtensionDefinition {
   name: string;
@@ -324,3 +325,55 @@ export const reset: () => void = extensionStore.action(() => {
     extensions: {},
   };
 });
+
+/**
+ * Returns information describing all extensions which can be rendered into an extension slot with
+ * the specified name.
+ * The returned information describe the extension itself, as well as the extension slot name(s)
+ * with which it has been attached.
+ * @param actualExtensionSlotName The extension slot name for which matching extension info should be returned.
+ * For URL like extension slots, this should be the name where parameters have been replaced with actual values
+ * (e.g. `/mySlot/213da954-87a2-432d-91f6-a3c441851726`).
+ * @param moduleName The module name. Used for applying extension-specific config values to the result.
+ */
+export async function getAttachedExtensionInfoForSlotAndConfig(
+  actualExtensionSlotName: string,
+  moduleName: string
+): Promise<Array<AttachedExtensionInfo>> {
+  const config = await getExtensionSlotConfig(
+    actualExtensionSlotName,
+    moduleName
+  );
+  let extensionIds = getAttachedExtensionInfoForSlot(actualExtensionSlotName);
+
+  if (config.add) {
+    extensionIds = extensionIds.concat(
+      config.add.map((id) => ({
+        extensionId: id,
+        attachedExtensionSlotName: actualExtensionSlotName,
+        actualExtensionSlotName,
+      }))
+    );
+  }
+
+  if (config.remove) {
+    extensionIds = extensionIds.filter(
+      (n) => !config.remove?.includes(n.extensionId)
+    );
+  }
+
+  if (config.order) {
+    extensionIds = extensionIds.sort((a, b) =>
+      config.order?.includes(a.extensionId)
+        ? config.order.includes(b.extensionId)
+          ? config.order.indexOf(a.extensionId) -
+            config.order.indexOf(b.extensionId)
+          : -1
+        : config.order?.includes(b.extensionId)
+        ? 1
+        : 0
+    );
+  }
+
+  return extensionIds;
+}
--- a/packages/esm-extensions/src/extensions.ts
+++ b/packages/esm-extensions/src/extensions.ts
@@ -83,11 +83,17 @@ export const extensionStore = createGlobalStore<ExtensionStore>("extensions", {
   extensions: {},
 });
 
-function updateStore<U extends keyof ExtensionStore>(
-  updater: (state: ExtensionStore) => Pick<ExtensionStore, U>
+type MaybeAsync<T> = T | Promise<T>;
+
+async function updateStore<U extends keyof ExtensionStore>(
+  updater: (state: ExtensionStore) => MaybeAsync<Pick<ExtensionStore, U>>
 ) {
   const state = extensionStore.getState();
-  extensionStore.setState(updater(state));
+  const newState = await updater(state);
+
+  if (newState !== state) {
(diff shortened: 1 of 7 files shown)


packages/esm-react-utils/src/ExtensionSlot.tsx
```diff
@@ -24,41 +19,18 @@ export interface ExtensionSlotBaseProps {
 export type ExtensionSlotProps<T = {}> = ExtensionSlotBaseProps & T;
 
 export const ExtensionSlot: React.FC<ExtensionSlotProps> = ({
-  extensionSlotName,
+  extensionSlotName: actualExtensionSlotName,
```
brandones, line 22: I thiiiiink this is supposed to be `attachedExtensionSlotName`. `actualExtensionSlotName`s have interpolated routeProps.

(I still don't understand why we have all this functionality for route patterns as extension slot names. I think it makes dealing with extension slot names very confusing. We should maybe talk about it elsewhere.)

packages/esm-react-utils/src/ExtensionSlot.tsx
```diff
@@ -24,41 +19,18 @@ export interface ExtensionSlotBaseProps {
...
-
-  useEffect(() => {
-    registerExtensionSlot(slotModuleName, extensionSlotName);
-    return () => unregisterExtensionSlot(slotModuleName, extensionSlotName);
-  }, []);
+  const {
+    attachedExtensionSlotName,
+    extensionIdsToRender,
+    extensionSlotModuleName,
+  } = useExtensionSlot(actualExtensionSlotName);
```
brandones, line 33: Beautiful :clap: clap: clap:

packages/esm-extensions/src/extensions.ts
```diff
@@ -324,3 +337,122 @@ export const reset: () => void = extensionStore.action(() => {
...
+  if (Array.isArray(config.add)) {
+    config.add.forEach((extensionId) => {
+      if (!extensionSlot.addedIds.includes(extensionId)) {
+        extensionSlot = {
+          ...extensionSlot,
+          addedIds: [...extensionSlot.addedIds, extensionId],
+        };
+      }
+    });
+  }
```
brandones, line 405: I think we need to make a decision about whether extension slots are essentially situated within a particular module or essentially global. To be concrete about it, what you've implemented treats slot configuration as being global across all slots that share that name, and the configs are merged. So if my application is (forgive the very pseudo code)

```jsx
<Module id="primary-navigation">
  <ExtensionSlot name="foo"/>
</Module>
<Module id="home">
  <ExtensionSlot name="foo"/>
</Module>
```

and I have extensions A, B, and C all attached to slot name "foo", then if I provide the configuration

```js
{
  "primary-navigation": {
    extensions: {
      foo: { remove: ['B'] }
    }
  },
  "home": {
    extensions: {
      foo: { remove: ['A'] }
    }
  }
```

I will wind up with

```jsx
<Module id="primary-navigation">
  <C />
</Module>
<Module id="home">
  <C />
</Module>
```

I think this is surprising. What I suggest is that extension slot configurations should be per-module, so the result would be

```jsx
<Module id="primary-navigation">
  <A /> <C />
</Module>
<Module id="home">
  <B /> <C />
</Module>
```

Of course, this is not likely to come up in practice very soon, so there's no hurry to get this straightened out.

brandones (approved):
Excellent, thank you!

I fully agree that the next step is to make esm-config use the store. Then everything can be a chain of `subscribe`s, all fully responsive to changes anywhere. Full push-model. I can take that on, unless you have a strong idea about it.

The one tricky thing is that the store will need to come out of esm-api, to avoid a circular dependency. Maybe just esm-store? Unless there's something else we should bundle it in with?

packages/esm-extensions/src/extensions.ts
FlorianRappl: This is a fair point. Right now the assigned IDs are for all slots which is not what we want. Thanks for recognizing this and the detailed explanation!

I'll improve this with the larger PR.

FlorianRappl:
> The one tricky thing is that the store will need to come out of esm-api, to avoid a circular dependency. Maybe just esm-store?

Yeah this is one option. The other is to move the store to `esm-config`. It all depends on how we see `esm-config`. Personally, I regard it as "esm-base", i.e., the one core / base package that all others depend implicitly on. If we say "this is only for configuration" (hence the name) then introducing either `esm-base` (explicitly) or `esm-store` (better name, but would be constrained to only contain the store) would make sense.

brandones:
Yeah, it's a good question. I think any of these options would be totally reasonable. Though if we put config and store together, we should definitely call it esm-base and not keep calling it esm-config (I assume this is what you mean but just want to be sure).

But also the boundaries of a esm-base would be an open question—any reason not to put esm-config, esm-store, esm-extensions, esm-error-handling, and esm-api all in there (well, just the parts of esm-api that are actually "api" stuff—not sure what to do about the rest)? That's all the core "libraries." Then esm-core has this kind of nice structure, where there's just esm-base, esm-app-shell, openmrs, and a couple MFEs.

It seems like as long as we put source code in separate subdirectories, i.e. `src/config`, `src/store`, it shouldn't create any mess to put things together.

On the other hand, Lerna is great, and I haven't really found the overhead of maintaining the additional packages too burdensome.

FlorianRappl:
> But also the boundaries of a esm-base would be an open question—any reason not to put esm-config, esm-store, esm-extensions, esm-error-handling, and esm-api all in there (well, just the parts of esm-api that are actually "api" stuff—not sure what to do about the rest)? That's all the core "libraries." Then esm-core has this kind of nice structure, where there's just esm-base, esm-app-shell, openmrs, and a couple MFEs.

I like this. I mean what we can do is to 

- create `esm-store` package (contains the store)
- create `esm-base` package (re-exports its dependencies, `esm-api`, `esm-config`, `esm-extensions`, `esm-error-handling`, `esm-store`)
- declare `esm-api` a shared dependency
- declare `esm-base` a shared dependency
- all microfrontends would almost exclusively use `esm-base` as a dependency, but some may want to be more fine-grained and could also do that (e.g., you could either import the `defineConfigSchema` from `esm-config` *or* `esm-base` - the difference would be if you depend on the different packages, or "just" the base package)

Does that go in a good direction?

brandones:
Yeah, absolutely.

> some may want to be more fine-grained

For what reason? Performance? Or semantics? If the latter, should we just make esm-base export subpackages, `@openmrs/esm-base/config` etc?

FlorianRappl:
> For what reason? Performance? Or semantics?

To reduce dependencies, e.g., if some microfrontend would only depend on `esm-config` then having all the others esms as (implicit) dependencies may not be wanted. Thinking beyond this one app shell it may make sense to be more fine-grained.

brandones:
Sure, makes sense :+1:

PR #48, opened 2020-12-04
Prettier UI Editor

brandones:
screenshot--2020-12-03--222930

### DIFF
--- a/packages/esm-react-utils/webpack.config.js
+++ b/packages/esm-react-utils/webpack.config.js
@@ -5,7 +5,10 @@ const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const { peerDependencies } = require("./package.json");
 
 module.exports = {
-  entry: [resolve(__dirname, "src/index.ts")],
+  entry: [
+    resolve(__dirname, "src/set-public-path.ts"),
+    resolve(__dirname, "src/index.ts"),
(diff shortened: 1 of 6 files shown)


packages/esm-react-utils/webpack.config.js
```diff
@@ -5,10 +5,7 @@ const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const { peerDependencies } = require("./package.json");
 
 module.exports = {
-  entry: [
-    resolve(__dirname, "src/set-public-path.ts"),
-    resolve(__dirname, "src/index.ts"),
-  ],
+  entry: [resolve(__dirname, "src/index.ts")],
```
FlorianRappl, line 8: Why?

If one really wants to use the direct esm output (e.g., to reuse a microfrontend somewhere else) then the public path should be set to avoid having to deal with invalid URLs (i.e., otherwise things like side bundles cannot be found).

FlorianRappl (approved):
LGTM - I'd just recommend leaving the public path in there for the ESM output.

packages/esm-react-utils/webpack.config.js
brandones: Whoops! I don't remember why I removed this but it was certainly a mistake. Thanks!

PR #51, opened 2020-12-17
Fix freeform object defaulting and interpolation

brandones:
Addresses the bug @ivange94 reported on Slack.

### DIFF
--- a/packages/esm-config/src/module-config/module-config.test.ts
+++ b/packages/esm-config/src/module-config/module-config.test.ts
@@ -237,43 +237,6 @@ describe("getConfig", () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/esm-config/src/module-config/module-config.test.ts
```diff
@@ -237,43 +237,6 @@ describe("getConfig", () => {
     });
   });
 
-  it("supports freeform object elements, which have no structural validation", async () => {
```
brandones, line 240: This was just moved down, to be with the other freeform object tests.

ivange94 (approved):
Thanks @brandones

PR #55, opened 2021-01-03
Config store

brandones:
config-store


packages/esm-config/src/module-config/state.ts
```diff
@@ -0,0 +1,119 @@
...
+import { equals } from "ramda";
+import { Config, ConfigObject, ConfigSchema, ProvidedConfig } from "../types";
+
+/*
+ * Internal store
+ *   A store of the inputs and internal state
+ */
+
+export interface ConfigInternalStore {
+  // Configs added using the `provide` function
```
FlorianRappl, line 11: For such documentation comments I'd use the standard docgen format (using block comments /* */).

packages/esm-config/src/module-config/state.ts
```diff
@@ -0,0 +1,119 @@
...
+    lastValueOfDevDefaultsAreOn = state.devDefaultsAreOn;
+  }
+});
+
+function setAreDevDefaultsOn(value: boolean): void {
+  localStorage.setItem("openmrs:configAreDevDefaultsOn", JSON.stringify(value));
+}
+
+function getAreDevDefaultsOn(): boolean {
+  return JSON.parse(
```
FlorianRappl, line 47: I'd place this in a `try` - `catch` block. If, for whatever reason, this key `openmrs:configAreDevDefaultsOn` gets "corrupted" with some string such as `foo` then the application crashes here.

packages/esm-config/src/module-config/state.ts
```diff
@@ -0,0 +1,119 @@
...
+
+let lastValueOfDevDefaultsAreOn = getAreDevDefaultsOn();
+configInternalStore.subscribe((state) => {
+  if (state.devDefaultsAreOn != lastValueOfDevDefaultsAreOn) {
+    setAreDevDefaultsOn(state.devDefaultsAreOn);
+    lastValueOfDevDefaultsAreOn = state.devDefaultsAreOn;
+  }
+});
+
+function setAreDevDefaultsOn(value: boolean): void {
```
FlorianRappl, line 42: I would not be more explicit then we need to.

Write as many types as required, but not more. `void` is inferred here.

packages/esm-config/src/module-config/state.ts
```diff
@@ -0,0 +1,119 @@
...
+
+let lastValueOfTemporaryConfig = getTemporaryConfig();
+temporaryConfigStore.subscribe((state) => {
+  if (equals(state, lastValueOfTemporaryConfig)) {
+    setTemporaryConfig(state);
+    lastValueOfTemporaryConfig = state;
+  }
+});
+
+function setTemporaryConfig(value: Config): void {
```
FlorianRappl, line 113: void can be inferred here.

packages/esm-config/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+export type ConfigValue =
+  | string
+  | number
+  | boolean
+  | void
+  | Array<any>
+  | object;
+
+export interface ExtensionSlotConfig {
+  add?: string[];
```
FlorianRappl, line 39: I think esm-core fully uses the `Array<T>` notation that I also find much more readable. Reason is that `[]` is actually a tuple. Furthermore, especially in combined cases it gets a mess as you don't start with the real type (array), but with inner type (e.g., string). So something like `Promise<Array<Record<string, T>>>` is much more readable (and consistent) to me than `Promise<Record<string, T>[]>`.

packages/esm-react-utils/src/useConfig.ts
```diff
@@ -1,15 +1,18 @@
...
+  ConfigStore,
 } from "@openmrs/esm-config";
 import { ModuleNameContext } from "./ModuleNameContext";
 import { ExtensionContext } from "./ExtensionContext";
 import { useForceUpdate } from "./useForceUpdate";
+import { ConfigObject } from "@openmrs/esm-config/src/types";
+import isEqual from "lodash-es/isEqual";
 
 let error: Error | undefined;
+let promises: Record<string, Promise<ConfigObject>> = {};
```
FlorianRappl, line 14: Should these two be `const`? The promises and configs are objects that should not be reassigned.

FlorianRappl (commented):
I think this is great. Definitely the way to go!

brandones:
@FlorianRappl I have everything apparently working in the browser at this point. This might be a good time to do another quick review. All that's left for this PR is to fix the tests.

One thing that I'm not doing in this PR is getting the extension slots to automatically receive new extension slot configs (add, remove, order). That will require some reworking of the extension state.

packages/esm-react-utils/src/useConfig.ts
```diff
@@ -1,15 +1,17 @@
...
-  configCacheNotifier,
-  getExtensionConfig,
+  getConfigStore,
+  getExtensionConfigStore,
+  ConfigStore,
 } from "@openmrs/esm-config";
 import { ModuleNameContext } from "./ModuleNameContext";
 import { ExtensionContext } from "./ExtensionContext";
-import { useForceUpdate } from "./useForceUpdate";
+import { ConfigObject } from "@openmrs/esm-config/src/types";
```
FlorianRappl, line 9: Can this be imported from `@openmrs/esm-config` directly?

FlorianRappl (approved):
I think this all looks fine from my POV!

Great job! :beers:

packages/esm-config/src/module-config/module-config.ts
```diff
@@ -1,4 +1,4 @@
-import "systemjs/dist/system";
+// import "systemjs/dist/system";
```
FlorianRappl, line 1: Just remove it

brandones: Yep, I still have a good amount of cleanup to do once I get all the tests working. I'll ping you once everything is ready.

FlorianRappl (approved):
Let's let this baby fly and then let's see and improve. :beers:

PR #56, opened 2021-01-03
Fix overlapping act calls warning

denniskigen:
Fixes: 

Screenshot from 2021-01-03 13-27-40


brandones (approved):
Nice, thank you @denniskigen !

PR #57, opened 2021-01-03
Small UI changes to the implementer tools panel

denniskigen:
Before:
Screenshot from 2021-01-03 15-32-39

After:
Screenshot from 2021-01-03 15-29-54

- Adds padding around the `ContentSwitcher` and the Close (X) panel button - there's no longer a space between the rounded edges of the ContentSwitcher buttons and the implementer tools panel edges.
- Restores the grid layout to the `Configuration` component with rows and columns. 
- Adds styling to the `Clear Temporary Config` button that denotes it as a button with destructive effects.

### DIFF
--- a/packages/esm-implementer-tools-app/src/popup/popup.styles.css
+++ b/packages/esm-implementer-tools-app/src/popup/popup.styles.css
@@ -18,7 +18,7 @@
 }
 
 .content {
-  margin-top: 40px;
+  margin-top: 48px;
   width: 100%;
 }
 
@@ -28,12 +28,13 @@
 
 .topBar {
   position: fixed;
-  background-color: #d0d0d0;
+  align-items: center;
+  background-color: #e0e0e0;
+  padding: 0 0.5rem;
   width: 100%;
-  height: 42px;
+  height: 48px;
   display: flex;
   justify-content: space-between;
-  border-bottom: 1px solid #a0a0a0;
   z-index: 2;
 }
 
--- a/packages/esm-implementer-tools-app/src/popup/popup.styles.css
+++ b/packages/esm-implementer-tools-app/src/popup/popup.styles.css
@@ -41,8 +41,8 @@
   padding: 0 1em;
 }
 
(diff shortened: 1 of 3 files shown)


brandones:
Nice! I really appreciate the initiative. A couple things:

I kind of liked the top bar for the tabs and close button. It emulates familiar desktop OS windows. Maybe instead of making the ContentSwitcher and close button floating, we can just remove the rounding from the ContentSwitcher corners?

(We're abusing Carbon a  little bit here—we're using ContentSwitcher as if its [Tabs](https://www.carbondesignsystem.com/components/tabs/usage), just because the Carbon Components React implementation of Tabs does not work in a top bar for much more intractable reasons (hence switching it to use ContentSwitcher). But I think we should be thinking about these as Tabs.)

Looks like we've got a bit of background peeking through on the right side. Probably related to the content margin.

denniskigen:
Interesting - makes sense. Let me make those changes and see how it turns out.

denniskigen:
Screenshot from 2021-01-04 20-59-42 Thoughts @brandones?

packages/esm-implementer-tools-app/src/popup/popup.styles.css
```diff
@@ -41,3 +40,11 @@
   min-height: 3em;
   padding: 0 1em;
 }
+
+:global(.bx--content-switcher-btn:first-child),
+:global(.bx--content-switcher-btn:last-child) {
```
brandones, line 45: Will these poison the CSS in the rest of the app? If so, we should scope it with something like

```css
.topBar :global(.bx--content-switcher-btn:first-child),
.topBar :global(.bx--content-switcher-btn:last-child) {
```

brandones (approved):
Looks great, modulo a note about global CSS. Thanks @denniskigen

packages/esm-implementer-tools-app/src/popup/popup.styles.css
denniskigen: Fixed, thanks

PR #59, opened 2021-01-22
Various Enhancements

FlorianRappl:
Just a few things that accumulated over time ...

1. Improved config of the app shell regarding the `spaPath` (now has 0 artifacts *after* the build, i.e., can still be fully configured at runtime)
2. Added ability to specify configuration URLs which are then automatically loaded and provided to the config system
3. Included basic logic for breadcrumbs
4. Added "core UI components" (which are extensions); first of these is an extension to display the breadcrumbs (right now this is a draft implementation that is rather static; to be determined fully later...)

### DIFF
--- a/packages/esm-api/src/breadcrumbs/filter.ts
+++ b/packages/esm-api/src/breadcrumbs/filter.ts
@@ -1,0 +1,68 @@
+import { getBreadcrumbs } from "./db";
+import { BreadcrumbRegistration } from "./types";
+
+function getExact(
+  breadcrumbs: Array<BreadcrumbRegistration>,
+  path: string
+): BreadcrumbRegistration {
+  const [bc] = breadcrumbs.filter((m) => m.matcher.test(path));
+  return bc;
+}
+
+function getClosest(
+  breadcrumbs: Array<BreadcrumbRegistration>,
+  path: string
+): BreadcrumbRegistration | undefined {
+  const segments = path.split("/");
+
+  while (segments.length > 1) {
+    segments.pop();
+    const newPath = segments.join("/");
+    const next = getNext(breadcrumbs, newPath);
+
+    if (next) {
+      return next;
+    }
+  }
+
+  return undefined;
+}
+
+function getNext(
+  breadcrumbs: Array<BreadcrumbRegistration>,
+  path: string | undefined
+) {
+  if (path) {
+    return getExact(breadcrumbs, path) || getClosest(breadcrumbs, path);
+  }
+
+  return undefined;
+}
+
+export function filterBreadcrumbs(
+  list: Array<BreadcrumbRegistration>,
+  path: string
+) {
+  if (list.length > 0) {
+    const current = getNext(list, path);
+
+    if (current) {
+      const links = [current];
+      let previous = getNext(list, current.settings.parent);
+
+      while (previous !== undefined) {
+        links.push(previous);
+        previous = getNext(list, previous.settings.parent);
+      }
+
+      return links.reverse();
+    }
+  }
+
+  return [];
+}
+
+export function getBreadcrumbsFor(path: string) {
+  const breadcrumbs = getBreadcrumbs();
+  return filterBreadcrumbs(breadcrumbs, path);
+}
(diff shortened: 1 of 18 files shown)


packages/esm-app-shell/src/ui/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  const options = {
+    featureName: "app-shell",
+    moduleName: appName,
+  };
+
+  return [
+    {
+      id: "breadcrumbs-widget",
+      slot: "breadcrumbs",
+      load: getAsyncLifecycle(() => import("./breadcrumbs"), options),
```
brandones, line 15: Should be `getAsyncExtensionLifecycle`

packages/esm-api/src/breadcrumbs/filter.ts
```diff
@@ -0,0 +1,68 @@
...
+    segments.pop();
+    const newPath = segments.join("/");
+    const next = getNext(breadcrumbs, newPath);
+
+    if (next) {
+      return next;
+    }
+  }
+
+  return undefined;
```
brandones, line 28: Nit: I think `null` would make more sense and be more idiomatic here.

packages/esm-api/src/breadcrumbs/types.ts
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Gets the path of breadcrumb for navigation purposes.
+   */
+  path: string;
+  /**
+   * Gets a custom matching function to know if the breadcrumb should be selected.
+   *
+   * In case of a missing matcher it uses the path to regexp result of the given path.
+   *
+   * In case of a string it uses the path to regexp result of the given matcher.
```
brandones, line 11: This could be clarified a bit.

```suggestion
   * A string or RegEx that determines whether the breadcrumb should be displayed.
   * It is tested against the current location's path.
   * 
   * If `matcher` is a string, it can contain route parameters. e.g. `/foo/:bar`.
   *
   * Can be omitted; the value of `path` is used as the default value.
```

packages/esm-app-shell/src/ui/breadcrumbs.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { Breadcrumb, BreadcrumbItem } from "carbon-components-react";
+import { getBreadcrumbsFor } from "@openmrs/esm-api";
+import { ConfigurableLink } from "@openmrs/esm-react-utils";
+
+const Breadcrumbs: React.FC = () => {
+  const path = location.pathname;
+  const breadcrumbs = getBreadcrumbsFor(path);
```
brandones, line 8: This is probably fine, since this probably won't re-render often at all? But the other way you could implement it would be to create an "output store" containing just the list of breadcrumbs that should be shown for that location, which gets updated by a `popstate` event listener.

brandones (changes requested):
Generally looks great. Really excited to try out the config url mechanism.

The only blocker I think is using `getAsyncExtensionLifecycle`.

packages/esm-app-shell/src/ui/index.ts
FlorianRappl: Hm is this already fully decided? Sorry was not available in my codebase.

Personally, again, I'd prefer simplicity here and have a single mechanism. I still don't understand why we need to distinguish.

packages/esm-api/src/breadcrumbs/filter.ts
FlorianRappl: Why? You should never use `null` (exception: when an API expects this, e.g., React when not rendering something from a component). `undefined` is the way to communicate nothing is returned, `null` is reserved for native APIs.

packages/esm-app-shell/src/ui/breadcrumbs.tsx
FlorianRappl: See description of the PR:

> right now this is a draft implementation that is rather static; to be determined fully later...

FlorianRappl: Note the word `static` -> its not meant to rerender at the moment as I first want to see use cases and then see how it works together with different routing mechanism (top-level single SPA vs React's router -> usually all should be covered by the DOM's history API as you mentioned, but I am not 100% sure yet).

brandones (approved):
LGTM! Thanks!

PR #62, opened 2021-02-05
Update README with current Styleguide links

gracepotma:
Updated our Styleguide link to:
* match the latest guidance our UX Designer Ciaran added to Zeplin
* to call our Carbon as our meta-guide
* to point people to the overrides folder


FlorianRappl (approved):
🏆

PR #63, opened 2021-02-12
add carbon table css overrides

nmalyschkin:


nmalyschkin:
How tables look now:
image

packages/esm-styleguide/src/carbon-overrides.css
```diff
@@ -1,3 +1,45 @@
...
+}
+
+.bx--data-table thead,
+.bx--data-table tr,
+.bx--data-table td,
+.bx--data-table--sort th,
+.bx--data-table--sort th .bx--table-sort__flex {
+  height: 2rem;
+  min-height: unset;
+  color: #525252;
```
FlorianRappl, line 44: The `td` previously was white. Is this intentional?

nmalyschkin: Why shouldn't it be intentional, this is according to Ciaran sketches.

FlorianRappl: Ok, but @brandones explicitly overwrote them as white.

The question is now: Why the change or why the original override?

brandones: I didn't have the designs on hand; I had just heard "Brandon can you help make the tables white" and I said "sure, can do." The point was really to initialize carbon-overrides, making a skeleton for this kind of work, anyway. I trust that this is the correct change.

nmalyschkin: The background color is still white. `td` background is just transparent and now the whole table container is white.
Otherwise we get tiny grey gaps:
image

Thanks @brandones for initializing the overrides, 3 weeks ago I was looking for them not knowing that there weren't any 😅

PR #64, opened 2021-02-16
Add shared logic for fetching patient photo

samuelmale:
**Dependant PRs**
- https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/199
- https://github.com/openmrs/openmrs-esm-home/pull/56

### DIFF
--- a/packages/esm-api/src/shared-api-objects/current-patient.ts
+++ b/packages/esm-api/src/shared-api-objects/current-patient.ts
@@ -2,6 +2,8 @@ import { ReplaySubject, Observable } from "rxjs";
 import { fhir } from "../fhir";
 import { mergeAll, filter, map } from "rxjs/operators";
 import { FetchResponse } from "../types";
+import { openmrsFetch } from "../openmrs-fetch";
+import { getConfig } from "@openmrs/esm-config";
 
 let currentPatientUuid: string;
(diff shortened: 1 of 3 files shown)


packages/esm-api/src/shared-api-objects/current-patient.ts
```diff
@@ -32,6 +34,29 @@ function getPatientUuidFromUrl() {
   return match && match[1];
 }
 
+export async function fetchPatientPhotoUrl(
+  patientUuid: string,
+  abortController: AbortController
+): Promise<PhotoUrl> {
+  const { concepts } = await getConfig("@openmrs/esm-patient-registration-app");
```
FlorianRappl, line 41: No. Don't make some helper libraries dependent on any microfrontend!

FlorianRappl (changes requested):

packages/esm-api/src/shared-api-objects/current-patient.ts
FlorianRappl: If you really want the concepts to be defined elsewhere then expect this here as an argument.

samuelmale: Thanks Florian, closing this PR since extensions will solve this problem.

PR #67, opened 2021-02-26
Button color codes changes to be applied globally.

vasharma05:
These changes are In connection to issue MF-488.


brandones (approved):
This seems like the right move to me!

@donaldkibet @FlorianRappl feel free to chime in if you disagree.

PR #68, opened 2021-02-27
Implementer tools should cache all the slots it knows about in case any of them disappear

brandones:
The reason @jonathandick and Grace couldn't find the `patient-chart-nav-menu` slot is because the implementer tools were only showing it when it was mounted. This gives the implementer tools a cache of the slots.


FlorianRappl (approved):
LGTM - thanks! :beers:

PR #70, opened 2021-03-01
MF-488: Added additional global css changes to button

vasharma05:
This pull request adds required global CSS changes to Primary, Tertiary and Ghost Buttons.


brandones:
The color of buttons isn't supposed to change at all between states? That seems surprising.

Screenshots for these kinds of changes would be much appreciated.

vasharma05:
Hello @brandones ! I did the changes viewing the designs at https://zpl.io/2jz5J3q . As you can see all the properties of primary, tertiary and ghost buttons (these terms are reference from Carbon design), I made the changes accordingly. About the additional properties(hover, active, focus) it was using the default carbon design blue color, which was also to be changed to teal, and hence are these changes requested.
Thanks.

packages/esm-styleguide/src/carbon-overrides.css
```diff
@@ -44,23 +44,33 @@
   color: #525252;
 }
 
-.bx--btn--primary{
+.bx--btn--primary,
+.bx--btn--primary:hover,
+.bx--btn--primary:active,
+.bx--btn--tertiary:hover,
+.bx--btn--tertiary:active,
+.bx--btn--tertiary:focus{
```
brandones, line 52: @FlorianRappl We should make sure this file gets linted & prettier'd.

brandones (approved):
Ok, sounds good @vasharma05 !

vasharma05:
Thank You!

PR #72, opened 2021-03-03
Give implementer tools a cache of the extension IDs assigned to each slot

brandones:
Allows the extension slot config interface to work.

impl-tools-fix


jonathandick (approved):
Lgtm. Thanks for quick fix.

PR #76, opened 2021-03-05
fix typings in openmrs package

nmalyschkin:
This type errors hinder @samuelmale from building this project. I can't tell why they don't occur on my machine or in the CI pipeline (although my IDE does mark them for me).

### DIFF
--- a/packages/openmrs/src/utils/importmap.ts
+++ b/packages/openmrs/src/utils/importmap.ts
@@ -76,7 +76,7 @@ export function runProject(
 
 export async function mergeImportmap(
   decl: ImportmapDeclaration,
-  imports: Record<string, string>
+  imports: Record<string, string> | false
 ) {
   if (imports && Object.keys(imports).length > 0) {
     if (decl.type === "url") {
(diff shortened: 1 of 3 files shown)


packages/openmrs/src/utils/importmap.ts
```diff
@@ -76,7 +76,7 @@ export function runProject(
 
 export async function mergeImportmap(
   decl: ImportmapDeclaration,
-  imports: Record<string, string>
+  imports: Record<string, string> | false
```
FlorianRappl, line 79: This does not look right. I guess it should `| undefined`.

FlorianRappl:
> I can't tell why they don't occur on my machine or in the CI pipeline (although my IDE does mark them for me).

It's different versions of TypeScript / the underyling typings.

packages/openmrs/src/utils/importmap.ts
nmalyschkin: Well yes, but [this](https://github.com/openmrs/openmrs-esm-core/blob/a71bf5f7ab8f65d4bfa2dd046ba7ddd6eddb3d01/packages/openmrs/src/cli.ts#L83).
I can make it a ternary operator here, would be cleaner anyway.

FlorianRappl: Good point - thanks!

PR #78, opened 2021-03-10
adjust carbon table overrides to fit widget overview

nmalyschkin:


packages/esm-styleguide/src/carbon-overrides.css
```diff
@@ -18,15 +18,17 @@
 
 .bx--data-table-header {
   padding: 0.75rem;
+  text-align: start;
+  width: calc(100% - 220px);
```
FlorianRappl, line 22: Looks arbitrary.

packages/esm-styleguide/src/carbon-overrides.css
```diff
@@ -18,15 +18,17 @@
...
 }
 
-.bx--data-table-header::after {
+/* .bx--data-table-header::after {
```
FlorianRappl, line 25: Is this now a battle who edited that file last?

Either this is necessary or not. Why is (or was) that in? Why isn't it no longer required? Can we then remove it?

nmalyschkin: > The battles that you fight are within your own mind.

I added the above code myself and temporarily removed it to adjust the styling to the recent results overview in patient chart. I will use it again as soon as I integrate the whole overview to the patient chart.

packages/esm-styleguide/src/carbon-overrides.css
nmalyschkin: You mean the `220px`, right? Unfortunately the carbon table header is misbehaving otherwise and overlaying the header text with the action buttons if the text gets to long.

packages/esm-styleguide/src/carbon-overrides.css
FlorianRappl: Hm can we just keep it in there then? Why have several PRs / modifications if we can do it in one. Maybe I don't understand the problem yet.

packages/esm-styleguide/src/carbon-overrides.css
FlorianRappl: Still seems arbitrary. Why not 240? Why not 200? What if the table requires 260? What if a translation has a longer text and requires 280? Can we solve this differently or leave it to the microfrontends / respective components or tables (i.e., not requiring support here in the app shell)?

nmalyschkin: You are right, while I can't go without the `calc(100% - 220px)`, it should not be in the global override.

FlorianRappl: Yeah maybe you can a local one, e.g., based also on some custom class name or via a more specific selector.

FlorianRappl:
I think this can be closed, right? @nmalyschkin

PR #79, opened 2021-03-11
[MF-180] Factor out mocks to core tooling package

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-180


packages/esm-devtools-app/jest.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   moduleNameMapper: {
     "\\.(css)$": "identity-obj-proxy",
     "@openmrs/esm-framework":
-      "<rootDir>/__mocks__/openmrs-esm-framework.mock.tsx",
+      "@openmrs/esm-framework/mocks/openmrs-esm-framework.mock.tsx",
```
FlorianRappl, line 9: Would a module name like `@openmrs/esm-framework/mock` work? Assuming that we place `mocks/openmrs-esm-framework.mock.tsx` in `mock.tsx`.

packages/esm-framework/mocks/react-i18next.mock.js
```diff
@@ -0,0 +1,54 @@
+const React = require("react");
```
FlorianRappl, line 1: Hm why is this file in there? It's not part of the framework.

FlorianRappl (commented):
Goes in the right direction, but we should work on the dev experience. The module name / path should be as simple as possible.

Also only make a mock for the framework. Don't provide mocks for anything else.

packages/esm-devtools-app/jest.config.js
samuelmale: > Assuming that we place mocks/openmrs-esm-framework.mock.tsx in mock.tsx.

@FlorianRappl I don't understand what you mean here.

samuelmale: You want me to rename `mocks/openmrs-esm-framework.mock.tsx` to `mocks/mock.tsx`?

FlorianRappl: yes

FlorianRappl: See original request

> "@openmrs/esm-framework": "@openmrs/esm-framework/mock"

packages/esm-framework/mocks/react-i18next.mock.js
brandones: Yeah, but it is unfortunately necessary to mock it this way in all of the MFEs. I'd advocate including it.

FlorianRappl: Why?

We have other things that qualify, too, in that area - so for consistency reasons I'd prefer to keep the framework package to framework mocks. This should not end up as a trash repository to just bring anything, but things that are esm-framework related. `react-i18next` is not.

FlorianRappl (changes requested):
Please keep it limited to the framework mock.

packages/esm-framework/mocks/react-i18next.mock.js
samuelmale: Removed react-i18next.mock.js file.

FlorianRappl (changes requested):
Still missing:

- reportError
- interpolateString
- navigate
- getCurrentUser
- Extension
- ConfigurableLink


I got that list only by looking at some of the esm-framework.mock files in the other microfrontends.

FlorianRappl (approved):
LGTM - thanks 🍻 !

PR #81, opened 2021-03-11
MF-491  Patient registration button opens up Edit patient page

brandones:
https://issues.openmrs.org/browse/MF-491

This ensures that the `currentPatientUuid` is updated *prior* to new components mounting, and ensures that if `currentPatientUuid` is null, then `currentPatient` is treated as null as well.


packages/esm-error-handling/README.md
```diff
@@ -21,3 +21,5 @@
 
 
 
+
```
FlorianRappl, line 24: Hm there might be a bug in the document.sh - maybe you can check it.

Just noticed that 2 new empty lines are appended each time it runs (or so it seems).

FlorianRappl (commented):
LGTM! :beers:

packages/esm-error-handling/README.md
brandones: Uh-oh, you're right. Looks like it's a bug in markdown-toc. I've added a mitigation to this PR.

FlorianRappl: Awesome - you are the man! :beers:

FlorianRappl (approved):

PR #83, opened 2021-03-12
Move core types from esm-framework to esm-globals

brandones:
Previously, esm-config's `interpolateString.ts` was importing core types from the app shell using

```ts
import type {} from "@openmrs/esm-app-shell/src/types";
```

esm-app-shell is not a dep of esm-config, so this broke the TS build for MFEs that imported `interpolateString`. esm-config is a dependency of esm-app-shell, so esm-app-shell cannot be made into a proper dependency of esm-config. I've resolved the problem by moving the core types to a core lib that does not depend on any other core libs (and so can be depended on by any of them)—~esm-state~ esm-globals.


FlorianRappl:
Hm `esm-state` is not the right place for this.

I'd say we could extract this part of the app-shell in a dedicated lib called `esm-globals`, which would contain these types plus

```
function setupPaths(config: SpaConfig) {
  window.openmrsBase = config.apiUrl;
  window.spaBase = config.spaPath;
  window.spaEnv = config.env || "production";
  window.spaVersion = process.env.BUILD_VERSION;
  window.getOpenmrsSpaBase = () => `${window.spaBase}/`;
}
```

The `esm-globals` package would be contained in `esm-framework`. Packages (except framework / app shell) would, however, only have a `devDependency` to it as only types would be used from it.

brandones:
Sounds good, makes sense. Will update soon.

FlorianRappl (approved):
LGTM - thanks! :beers:

FlorianRappl:
(I'll merge despite the failing build; will correct and push directly)

PR #84, opened 2021-03-16
Correct useCurrentPatient typing

samuelmale:
Typescript infers the return type to `boolean[]`.

This breaks the Typing here:

Screenshot 2021-03-16 at 11 55 18 (2)


FlorianRappl (approved):
Thanks :beers:!

PR #86, opened 2021-03-22
MF-509: OpenMRS fetch should handle empty response data gracefully

samuelmale:
Ticket: https://issues.openmrs.org/browse/MF-509

### DIFF
--- a/packages/esm-api/src/openmrs-fetch.ts
+++ b/packages/esm-api/src/openmrs-fetch.ts
@@ -130,10 +130,21 @@ export function openmrsFetch<T = any>(
         return response;
       } else {
         // HTTP 200s - The request succeeded
-        return response.json().then((data) => {
-          response.data = data;
-          return response;
-        });
+        return response
+          .json()
+          .then((data) => {
+            response.data = data;
+            return response;
+          })
+          .catch((err) => {
+            // Server responded with empty response data or invalid json
+            if (err instanceof SyntaxError) {
+              response.data = (null as unknown) as T;
+              return response;
+            } else {
+              throw err;
+            }
+          });
       }
     } else {
       /* HTTP response status is not in 200s. Usually this will mean
--- a/packages/esm-api/src/openmrs-fetch.ts
+++ b/packages/esm-api/src/openmrs-fetch.ts
@@ -130,21 +130,15 @@ export function openmrsFetch<T = any>(
         return response;
       } else {
         // HTTP 200s - The request succeeded
-        return response
-          .json()
-          .then((data) => {
-            response.data = data;
-            return response;
-          })
-          .catch((err) => {
-            // Server responded with empty response data or invalid json
-            if (err instanceof SyntaxError) {
-              response.data = (null as unknown) as T;
-              return response;
-            } else {
-              throw err;
-            }
-          });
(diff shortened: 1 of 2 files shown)


packages/esm-api/src/openmrs-fetch.ts
```diff
@@ -130,10 +130,21 @@ export function openmrsFetch<T = any>(
...
-          return response;
-        });
+        return response
+          .json()
+          .then((data) => {
+            response.data = data;
+            return response;
+          })
+          .catch((err) => {
+            // Server responded with empty response data or invalid json
```
FlorianRappl, line 140: Not sure if this is the kind of gracefully that I imagine. Can't you just check if its JSON (e.g., via the `content-type`?). Alternatively, you can use `.text()` and do a `JSON.parse` in a try-catch. That way we know for sure its the content and not something else.

Otherwise looks fine to me.

samuelmale: Updated

packages/esm-api/src/openmrs-fetch.ts
```diff
@@ -130,8 +130,13 @@ export function openmrsFetch<T = any>(
...
       } else {
         // HTTP 200s - The request succeeded
-        return response.json().then((data) => {
-          response.data = data;
+        return response.text().then((responseText) => {
+          try {
+            response.data = JSON.parse(responseText);
+          } catch (err) {
+            // Server didn't respond with json, so just go with the response text string
+            response.data = responseText as any;
```
FlorianRappl, line 138: Hm I would just leave it empty.

packages/esm-api/src/openmrs-fetch.ts
```diff
@@ -130,8 +130,13 @@ export function openmrsFetch<T = any>(
         return response;
       } else {
         // HTTP 200s - The request succeeded
-        return response.json().then((data) => {
-          response.data = data;
+        return response.text().then((responseText) => {
+          try {
```
FlorianRappl, line 134: Before trying this can we check if we got any text at all?

packages/esm-api/src/openmrs-fetch.ts
samuelmale: What of cases were the server responds with plain text?

FlorianRappl: Which is useful for what?

So, yes, if we want to use that generically then you are right. But my impression was that we are in particular looking at the `POST .../session` problem.

Generally, this is why the *caller* should decide what to expect. Is JSON or plain text expected? Right now I am not aware of any plain text response, so I'd argue we keep this JSON only and introduce an optional argument that allows to get the text response (which would, however, also not enter the try-catch block; as we are  not interested in JSON in this case).

samuelmale: Okay for now let's assume it's json only.. I mean, it was `json` before as matter of fact that we were using `response.json()...`..

packages/esm-api/src/openmrs-fetch.ts
samuelmale: fixed

FlorianRappl (approved):

PR #89, opened 2021-03-26
Added a Service Worker, PWA-Manifest and Network-First Caching of static app shell files

manuelroemer:
In preparation for offline support, this PR adds the fundamental building blocks of a PWA:

* A manifest + associated icons
* A service worker + its integration from `index.js`
* **Preliminary and crude** caching of app shell files, in preparation for offline support

With these changes, it is possible to install the site as a PWA and to make the app shell render offline.
The latter can be be seen in the following screenshot which was made without an internet connection. While there's not much rendered, one can see that parts of the app shell (e.g. error notifications and the logo) are rendered and that the console lists no failing requests to static assets like CSS/JS files.
image

### Caching Strategy 

Caching/The Service Worker is implemented on top of Google's Workbox packages. At the moment, only the **Network First** strategy is used. I think that this brings the best out of both worlds: 

1. It does cache files for a future offline experience.
2. It is quite save in the sense that users will (with a network connection) always see the newest file versions. For the moment, this should prevent any stale caches of users/developers and allow development in this area to progress quickly without the fear of breaking things. This is subject to change, but I think that this is the best solution for the moment.

### Cached Files

For the beginning, only static app shell files are cached. The files are assembled at build time and provided to the service worker which precaches them during installation.

The cache looks like this after initial load:
image

### Open Points

* Cache invalidation is something that should be tracked. When the SW is updated in the future, old files currently remain in the cache. This is something to be considered/fixed.


packages/esm-app-shell/package.json
```diff
@@ -48,10 +48,15 @@
...
+    "systemjs": "^6.8.3",
+    "workbox-core": "^6.1.2",
+    "workbox-routing": "^6.1.2",
+    "workbox-strategies": "^6.1.2"
   },
   "devDependencies": {
     "@types/systemjs": "^6.1.0",
-    "css-modules-typescript-loader": "^4.0.1"
+    "css-modules-typescript-loader": "^4.0.1",
+    "webpack-pwa-manifest": "^4.3.0",
```
FlorianRappl, line 59: These two deps are also required in "openmrs" (as real `dependencies`) as it has to be capable of building the app shell.

FlorianRappl: (Actually, maybe we can get around that - I could use some trickery as with the `openrms develop` command... Let me see about that...)

packages/esm-app-shell/src/service-worker.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+const indexPath = prefixWithSpaBase("index.html");
+const wbManifest = self.__WB_MANIFEST;
+const absoluteWbManifestUrls = distinct(
+  wbManifest.map(({ url }) => prefixWithSpaBase(url))
+);
+const omrsCachePrefix = "omrs";
+const omrsCacheName = `${omrsCachePrefix}-spa-cache-v1`;
```
FlorianRappl, line 19: When can / should we update this name (e.g., to `v2`)?

packages/esm-app-shell/webpack.config.js
```diff
@@ -119,9 +120,22 @@ module.exports = (env, argv = {}) => {
     },
     plugins: [
       new CleanWebpackPlugin(),
+      new WebpackPwaManifest({
+        name: "OpenMRS",
+        short_name: "OpenMRS",
+        description:
+          "Open source Health IT by and for the entire planet, starting with the developing world.",
```
FlorianRappl, line 127: This is good - where is this from? We may want to be able to change that (e.g., some distributor may want to have a different slogan in here). We could use an env variable get that string (like with the other options). This string could be the fallback.

FlorianRappl (commented):
I think this is superb!

Thanks for the great job @manuelroemer :beers:!

packages/esm-app-shell/webpack.config.js
manuelroemer: This is ~stolen~ borrowed from the GitHub organization: 
image

packages/esm-app-shell/src/service-worker.ts
manuelroemer: I named it like this this because of "if the guys at Google do it, I can't be that bad of an idea": https://developers.google.com/web/tools/workbox/guides/configure-workbox#configure_cache_names

I guess that having a suffix can potentially be a good idea in case there's ever a breaking change in the things that are cached. Tbh, I'm not sure if that's ever going to happen (or if such a breaking change is even possible), but my initial thought was that having such a suffix in place won't hurt - at worst, we wouldn't ever increment it.

With that being said, I think that we could also go without it and use a simplified name like `omrs-cache` or something alike (the chance of having breaking changes seems slim at best). I'm open to both options.

packages/esm-app-shell/package.json
manuelroemer: Alright - just tell me if I should move/remove them ;)

FlorianRappl: For the moment I'd say copy them to the dependencies of `openmrs`.

=> https://github.com/openmrs/openmrs-esm-core/blob/master/packages/openmrs/package.json#L59

I'll take care of refactoring soon, but that way its always building / working.

packages/esm-app-shell/src/service-worker.ts
FlorianRappl: No its fine. So only if we make "drastic" changes (or want to flush the cache in any way) we'd change it. Thanks for the clarification!

FlorianRappl (approved):

PR #90, opened 2021-03-30
remove cloneDeep call

nmalyschkin:


packages/esm-extensions/src/render.ts
```diff
@@ -16,6 +14,18 @@ export interface CancelLoading {
   (): void;
 }
 
+const immutableSet = <T extends Record<string, any>>(
+  obj: T,
+  [parm, ...resPath]: string[],
```
FlorianRappl, line 19: Can we use `Array<string>`? What is `parm` and `resPath` (are better / more descriptive names viable here)?

brandones:
This antipattern exists a handful of times in the codebase (I think). Maybe we can add this as a helper function to `esm-state`?

And can we call it just `set`? Or `update` following the example of [React's update](https://reactjs.org/docs/update.html)?

packages/esm-extensions/src/render.ts
```diff
@@ -16,6 +14,18 @@ export interface CancelLoading {
   (): void;
 }
 
+const immutableSet = <T extends Record<string, any>>(
```
FlorianRappl, line 17: I guess this could / should be a `function`

FlorianRappl:
> 
> 
> This antipattern exists a handful of times in the codebase (I think). Maybe we can add this as a helper function to `esm-state`?
> 
> And can we call it just `set`? Or `update` following the example of [React's update](https://reactjs.org/docs/update.html)? Or (if needed) `immutablySet`? `immutableSet` sounds like an object.

Sounds good to me. `update` and placed in `esm-state` has my vote.

FlorianRappl (approved):

PR #93, opened 2021-03-31
MF-497 : implementer-tool shifted to top-nav-bar

nk183:
ticket https://issues.openmrs.org/browse/MF-497

esm-implementer-tools is registering an extension "implementer-tools-button" and attach it to "top-nav-actions-slot."
but the button is not visible I am not able to understand
@brandones @FlorianRappl can u please help

It is` not Complete` yet ,I am stuck at above attaching

image

### DIFF
--- a/packages/esm-devtools-app/src/devtools/devtools.styles.css
+++ b/packages/esm-devtools-app/src/devtools/devtools.styles.css
@@ -4,7 +4,7 @@
   height: 40px !important;
   width: 40px !important;
   bottom: 10px;
-  right: 54px;
+  right: 8px;
   position: fixed;
   border-radius: 2px;
 }
--- a/packages/esm-implementer-tools-app/src/index.ts
+++ b/packages/esm-implementer-tools-app/src/index.ts
@@ -8,11 +8,6 @@ function setupOpenMRS() {
   };
 
   return {
-    lifecycle: getAsyncLifecycle(
-      () => import("./implementer-tools.component"),
-      options
-    ),
-    activate: () => true,
     extensions: [
       {
         id: "implementer-tools-button",
--- a/packages/esm-implementer-tools-app/src/index.ts
+++ b/packages/esm-implementer-tools-app/src/index.ts
@@ -1,15 +1,28 @@
 import { getAsyncLifecycle } from "@openmrs/esm-framework";
 
 function setupOpenMRS() {
+  const moduleName = "@openmrs/esm-implementer-tools-app";
+  const options = {
+    featureName: "Implementer Tools",
+    moduleName,
+  };
+
   return {
     lifecycle: getAsyncLifecycle(
       () => import("./implementer-tools.component"),
-      {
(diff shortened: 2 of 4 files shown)


brandones:
The button is absolutely positioned right now.

You also need to make the button into a separate component, so that extension is only the button. The main popup component should remain as it is.

nk183:
hi @brandones I have tried implementing this is what it looks like :
image

can you please guide which CSS file should I work on I am a little confused as [ExtensionSlot](https://github.com/openmrs/openmrs-esm-primary-navigation/blob/797e828d4c63213919e8f7dcf8d4413df1b83ee2/src/components/navbar/navbar.component.tsx#L90) is showing boot the button

packages/esm-implementer-tools-app/src/index.ts
```diff
@@ -1,15 +1,28 @@
...
 
 function setupOpenMRS() {
+  const moduleName = "@openmrs/esm-implementer-tools-app";
+  const options = {
+    featureName: "Implementer Tools",
+    moduleName,
+  };
+
   return {
     lifecycle: getAsyncLifecycle(
```
FlorianRappl, line 11: The whole lifecycle thing is no longer needed if you provide the button into the top nav.

Remove `lifecycle` and `activate`.

FlorianRappl (changes requested):

packages/esm-implementer-tools-app/src/index.ts
nk183: hi @FlorianRappl I have removed it Can you please check

packages/esm-devtools-app/src/devtools/devtools.styles.css
```diff
@@ -4,7 +4,7 @@
   height: 40px !important;
   width: 40px !important;
   bottom: 10px;
-  right: 54px;
+  right: 8px;
```
FlorianRappl, line 7: Why do we need this trigger button? I thought it should go into the top nav bar?

I'd say let's remove this class.

nk183: @FlorianRappl  I think you have mistaken this class is for `devtool button` .this change shifts the button to right.
The `implementer tool` is supposed to shift to `nav bar`.

FlorianRappl: Yeah you are right. This is in `devtools`!

FlorianRappl (approved):

nk183:
hi @FlorianRappl  there was a bug in UI as I have mentioned above see this https://issues.openmrs.org/secure/attachment/60110/MF-497.mp4

FlorianRappl:
@nk183 and?

1. Fix it?
2. I don't think the bug was in this PR, or? I rather think it is in primary navigation / the slot usage.

nk183:
yes you are right sorry
and I am woking on it

PR #96, opened 2021-04-08
Service Worker: Bootstrapping Support for Offline MFs

manuelroemer:
This PR improves the previous service worker implementation from #89 with the following features:

* The Service Worker is now able to precache files from MFs defined in the import map.
* The Service Worker invalidates obsolete files when caching MFs from an import map.
* The app shell now supports an exported `setupOpenMRSOffline` function.

The following sections should provide more details about the respective implementation changes.

## MF Caching Based on the Import Map

Once the app shell loaded its import map, it sends it to the Service Worker via an event.
The Service Worker iterates through the import map's entries and, based on the address value,
generates the address for a so called **build manifest file** which informs the Service Worker
about the cacheable files provided by the MF.
This file can be generated by MFs at build time and should conform to the following conventions:

* Its URL is the import map address + the `.buildmanifest.json` suffix.  
  The corresponding file to the address `http://example.com/my-microfrontend.js` would be
  `http://example.com/my-microfrontend.js.buildmanifest.json`.
* It must list the MFs cacheable files in the following structure:

```ts
interface BuildManifest {
  chunks?: Array<{
    files?: Array<string>;
  }>;
}
```

> ℹ **Note:**  
> This structure is the default output structure of the [webpack-stats-plugin](https://www.npmjs.com/package/webpack-stats-plugin)
> package. We can/should debate whether we want to utilize a custom structure here.

Once the Service Worker has retrieved all of these file URLs, it can automatically add them to its
cache and display them in offline mode.

## Cache Invalidation

Whenever the Service Worker caches a new set of files resolved from an import map, it checks against
the cache's current value and removes any entries which:

1. Don't appear in the new import map's file list.
2. Don't originate from the app shell itself.

If, for whatever reason, an import map or a build manifest file's content changes, the previously
cached entries will be removed from the cache.

## The `setupOpenMRSOffline` function

When the app shell loads without an internet connection, it will not invoke the default
`setupOpenMRS` function (mainly because no existing MF would work in its current state and also
because not every MF will support offline features in the future).
Instead, it checks if a module exports the `setupOpenMRSOffline` function and uses this instead
for setting up the SPA.

Apart from the name, the two functions have the exact same body/behavior.
MFs desiring offline support have to export both functions though.

## Open Questions

* Should we have a custom build manifest format?
* Is the current build manifest name/suffix (i.e. `.buildmanifest.json`) okay or is there a better version?
* General thoughts on the caching strategy as it is right now?


manuelroemer:
Creating this as a Draft PR for the moment to give others time to review the changes and think about whether the general direction of the changes here are alright (or if they need to be changed from the ground up).

FlorianRappl:
> This structure is the default output structure of the webpack-stats-plugin package. We can/should debate whether we want to utilize a custom structure here.

I would strongly recommend to take the default. Right now I don't see any reason for a custom one, so taking *any* format is acceptable. As the default just works I wouldn't change it.

> Is the current build manifest name/suffix (i.e. .buildmanifest.json) okay or is there a better version?

I think we can stay with that. Personally, a name like `<mf-name>.offline.json` would have been sound, too, but then we would need more work to derive it. Is the asset contains general information and is not just relevant for offline use I think sticking to "....buildmanifest.json" is totally fine.

> General thoughts on the caching strategy as it is right now?

I think its a good strategy. We should always go network first and only use the cached items if we are offline. This strategy, however, does only apply to resources like the importmap and the index.html. All hashed files (such as all MFs and assets of the app shell) could be used cache-first. Problem with cache-first is where to draw the line. Some (unknown) resources may still require network first (especially during development) or otherwise will become an issue.

packages/esm-app-shell/src/helpers.ts
```diff
@@ -12,3 +12,7 @@ export function routeRegex(regex: RegExp, location: Location) {
 export function distinct<T>(values: Array<T>) {
```
FlorianRappl, line 12: These functions could be also placed in `esm-utils`. There, we could generally use them. (Sorry the `distinct` is on me - but I just saw it when you introduced `flatten`.

packages/esm-app-shell/src/index.ts
```diff
@@ -12,11 +12,20 @@ function wireSpaPaths() {
 }
 
 function registerServiceWorker() {
-  if ("serviceWorker" in navigator) {
+  if ("serviceWorker" in navigator && navigator.onLine) {
```
FlorianRappl, line 15: Should the registration really be avoided when offline (honest question)?

I guess only the importmap update part we would like to avoid in offline mode, or maybe I am wrong here.

packages/esm-app-shell/webpack.config.js
```diff
@@ -161,7 +161,7 @@ module.exports = (env, argv = {}) => {
         analyzerMode: env && env.analyze ? "static" : "disabled",
       }),
       new InjectManifest({
-        swSrc: resolve(__dirname, "./src/service-worker.ts"),
+        swSrc: resolve(__dirname, "./src/service-worker/service-worker.ts"),
```
FlorianRappl, line 164: I guess `service-worker/index.ts` would be a bit better.

FlorianRappl: (In this case I guess you'll need to set `swDest` to `service-worker.js` or so)

packages/esm-app-shell/src/helpers.ts
manuelroemer: Now that you mention it, do we have access to lodash in the app shell? It's not explicitly referenced in its package.json, but I think that I've seen it in some others packages in this repo.
Then we could replace these custom functions entirely.

packages/esm-app-shell/src/index.ts
manuelroemer: > only the importmap update part we would like to avoid in offline mode
This is actually the reason why I introduced this check, so it's absolutely a valid question. I figured that registering the service worker in offline mode wouldn't make much sense since it would not be able to fetch anything anyway and thus would not be able to provide any meaningful offline experience.

With that being said, maybe it makes sense to still register it. We could have some kind of offline.html fallback in the future, maybe? Then having it registered would be a good thing.

I'll give you the last choice on that.

packages/esm-app-shell/webpack.config.js
manuelroemer: Sounds good, will change it!

packages/esm-app-shell/src/helpers.ts
FlorianRappl: Its not one of the dependencies. Its also not a shared dependency. We usually bundle in selected functions.

We can do as you want. You can use `lodash-es` - but only with direct imports from the modules.

packages/esm-app-shell/src/index.ts
FlorianRappl: Yeah for consistency (and flexibility) I'd prefer to always register it.

yarn.lock
```diff
@@ -10525,7 +10525,7 @@ locate-path@^6.0.0:
   dependencies:
     p-locate "^5.0.0"
 
-lodash-es@^4.17.15, lodash-es@^4.17.20:
+lodash-es@4.17.21, lodash-es@^4.17.15, lodash-es@^4.17.20:
```
FlorianRappl, line 10528: We may have a mix of dependencies specs in there (4.17.21, ^4.17.15, ...) -> I would align them all to `^4.17.21`.

FlorianRappl (approved):
LGTM!

PR #97, opened 2021-04-14
Enhanced SW to Window communication and dynamic route caching registrations

manuelroemer:
Adds support for caching the latest GET request to the user session endpoint which in turn enables a lot of other functions in offline mode (e.g. getting details about the user, his/her location, etc.).

The request/response is always cached using the NetworkFirst strategy so that the website always has the latest data when online.

In addition, reworks the message handling of the service worker with respect to future changes.


packages/esm-app-shell/webpack.config.js
```diff
@@ -156,6 +156,7 @@ module.exports = (env, argv = {}) => {
       }),
       new DefinePlugin({
         "process.env.BUILD_VERSION": JSON.stringify(`${version}-${timestamp}`),
+        "process.env.OMRS_API_URL": JSON.stringify(openmrsApiUrl),
```
FlorianRappl, line 159: We need to be careful here.

Yes, the API Url is already fixed at build-time, however, it can still be changed afterwards. We are not using this constant in the JS - it only gets transported into the HTML file. I'd suggest to give the service-worker also a variable. Potentially, this is done via communication with the app shell, which was configured using the variables.

Look at index.html template file to see what I'm referring to.

FlorianRappl (commented):
In general great - just one thing we need to be careful about.

packages/esm-app-shell/webpack.config.js
manuelroemer: Alright, I wasn't aware that it could actually change beyond page initialization. I'll make the changes tomorrow.

Sending the value from the app shell seems to be the way to go then. If I'm going to introduce this new kind of message, I'll also include [workbox-window](https://developers.google.com/web/tools/workbox/modules/workbox-window#window_to_service_worker_communication) tomorrow as part of this PR. This will make future message development (and the one for this case) easier.

packages/esm-app-shell/package.json
```diff
@@ -1,63 +1,64 @@
...
-  },
-  "devDependencies": {
-    "@types/systemjs": "^6.1.0",
-    "css-modules-typescript-loader": "^4.0.1",
-    "webpack-pwa-manifest": "^4.3.0",
-    "workbox-webpack-plugin": "^6.1.2"
-  }
-}
+{
+  "name": "@openmrs/esm-app-shell",
```
FlorianRappl, line 2: Hm there is presumably something wrong with the gitattributes or so. I don't see any changes so I assume the LF has been replaced with CRLF. Can you have a look?

packages/esm-app-shell/src/index.ts
```diff
@@ -13,22 +14,29 @@ function wireSpaPaths() {
...
 
+    window.addEventListener("load", () => {
       if (navigator.onLine) {
-        navigator.serviceWorker.ready.then((sw) => {
-          window.importMapOverrides.getCurrentPageMap().then((importMap) => {
-            sw.active?.postMessage({
-              type: "importMap",
-              importMap,
-            });
+        window.importMapOverrides.getCurrentPageMap().then((importMap) => {
```
FlorianRappl, line 22: Are we sure we want to do that?

Reason: The override is a dev-only feature and I see no reason to use this in production code. In dev, however, the URL will not change and I fear that it may be problematic w.r.t. caching. Maybe we shouldn't cache then?

packages/esm-app-shell/package.json
manuelroemer: Sure, I'll have a look. I'll delay it until this is ready for review again, so that it doesn't change again by accident in between.

packages/esm-app-shell/src/index.ts
manuelroemer: I used `getCurrentPageMap` mainly because I figured that it was the best way to access the final import map when it's eventually loaded. What would be the best alternative to this one?

FlorianRappl: I see - yes that makes sense. Well, we also get it like that (https://github.com/openmrs/openmrs-esm-core/blob/master/packages/esm-app-shell/src/run.ts#L39). It's fine. Thanks for the explanation!

packages/esm-app-shell/package.json
```diff
@@ -37,6 +37,7 @@
     "carbon-components": "^10.19.0",
     "carbon-icons": "^7.0.7",
     "dayjs": "^1.10.4",
+    "dexie": "^3.0.3",
```
FlorianRappl, line 40: Sure you want to add that here? Would it be good to have a dedicated package (and part of `esm-framework`) instead? Maybe make it part of `esm-state` or have `esm-db` or something similar?

FlorianRappl (approved):
LGTM! :beers:

packages/esm-app-shell/package.json
manuelroemer: Resolved in chat: We delay the creation of such a package until we provide actually shared functionality for MFs (Dexie is currently only used in the SW itself).

PR #98, opened 2021-04-19
minor module config improvements

nmalyschkin:
minor changes while I was looking into configs

### DIFF
--- a/packages/esm-config/src/module-config/module-config.ts
+++ b/packages/esm-config/src/module-config/module-config.ts
@@ -1,4 +1,4 @@
-import { clone, map, reduce, mergeDeepRight, prop } from "ramda";
+import { clone, reduce, mergeDeepRight } from "ramda";
 import {
   Config,
   ConfigObject,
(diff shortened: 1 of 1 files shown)


packages/esm-config/src/module-config/module-config.ts
```diff
@@ -546,16 +533,16 @@ function mergeConfigsFor(
...
+    .filter((c) => !!c);
 
   return mergeConfigs(allConfigsForModule);
 }
 
+const mergeDeepAll = reduce(mergeDeepRight);
 function mergeConfigs(configs: Array<Config>) {
-  const mergeDeepAll = reduce(mergeDeepRight, {});
-  return mergeDeepAll(configs);
+  return mergeDeepAll({}, configs) as Config;
```
brandones, line 545: Not sure why you'd prefer a binary `mergeDeepAll` that requires a constant "initial value" parameter. Not a big deal, anyway.

nmalyschkin: This way we don't create a new `mergeDeepAll` on every function call. But as you said – not a big deal.

PR #99, opened 2021-04-21
add useExtensionSlotMeta

nmalyschkin:


packages/esm-react-utils/src/useExtensionSlotMeta.ts
```diff
@@ -0,0 +1,32 @@
...
+  extensionStore,
+  ExtensionStore,
+  ExtensionMeta,
+} from "@openmrs/esm-extensions";
+
+/**
+ * Extract meta data from all extension for a given extension slot
+ * @param actualExtensionSlotName
+ */
+export const useExtensionSlotMeta = (actualExtensionSlotName: string) => {
```
FlorianRappl, line 12: Use `extensionSlotName`.

packages/esm-react-utils/src/useExtensionSlotMeta.ts
```diff
@@ -0,0 +1,32 @@
...
+  const [meta, setMeta] = useState<Record<string, ExtensionMeta>>({});
+
+  useEffect(() => {
+    const update = (state: ExtensionStore) => {
+      if (state.slots[actualExtensionSlotName])
+        setMeta(
+          Object.fromEntries(
+            state.slots[actualExtensionSlotName].attachedIds.map((id) => [
```
FlorianRappl, line 20: I don't think you want `attachedIds`. I think you want the `assignedIds` from the instance. The attached extensions may not be assigned and thus do not render.

packages/esm-react-utils/tsconfig.json
```diff
@@ -16,8 +16,10 @@
       "es2015",
       "es2015.promise",
       "es2016.array.include",
-      "es2018"
-    ]
+      "es2018",
+      "esnext"
+    ],
+    "target": "ESNext"
```
FlorianRappl, line 22: Hm why?

packages/esm-devtools-app/tsconfig.json
```diff
@@ -12,8 +12,10 @@
       "es2015",
       "es2015.promise",
       "es2016.array.include",
-      "es2018"
+      "es2018",
+      "esnext"
     ],
+    "target": "ESNext",
```
FlorianRappl, line 18: Why?

packages/esm-react-utils/src/useExtensionSlotMeta.ts
nmalyschkin: I'm not sure, but I believe that `attachedIds` is what I want here. `useExtensionSlotMeta` should return meta data on registered extensions, not only rendered extensions.

packages/esm-devtools-app/tsconfig.json
nmalyschkin: otherwise typescript was complaining about `Object.fromEntries`.

packages/esm-react-utils/src/useExtensionSlotMeta.ts
FlorianRappl: Nope you only want the assigned IDs. Difference is that attached is all, but assigned are those that also respect configuration (e.g., removed or added extensions via config).

packages/esm-devtools-app/tsconfig.json
FlorianRappl: Nope. The target is what TS compiles to. There is no reason to change this.

packages/esm-react-utils/src/useExtensionSlotMeta.ts
nmalyschkin: Can I access `assignedIds` before the extensions are mounted? Otherwise I need to check how the get computed on mount.

FlorianRappl: Yes they are independent. Once a slot is opened / registered its assigned extensions are computed.

FlorianRappl (approved):
Yeah I think these are all quite some nice hooks!

PR #100, opened 2021-04-26
MF-566: Fix Trend-line and charts turning black

donaldkibet:
### Description
[ticket MF-566](https://issues.openmrs.org/projects/MF/issues/MF-566?filter=allissues)
At the moment [patient-chart - feature/monorepo](https://github.com/openmrs/openmrs-esm-patient-chart/tree/feature/monorepo) doesn't provide chart styles this would provide the chart styles
### Screenshoot
Screenshot 2021-04-26 at 16 02 59

### DIFF
--- a/packages/esm-styleguide/package.json
+++ b/packages/esm-styleguide/package.json
@@ -39,6 +39,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@carbon/charts": "^0.41.57",
     "carbon-components-react": "7.31.0",
     "lodash-es": "^4.17.21",
     "systemjs-webpack-interop": "^2.1.2"
--- a/packages/esm-styleguide/src/_all.scss
+++ b/packages/esm-styleguide/src/_all.scss
@@ -1,5 +1,6 @@
 @import "./vars";
 @import "~carbon-components/scss/globals/scss/styles";
+@import "~@carbon/charts/styles/styles.scss";
 
 @import "./components/general";
 @import "./components/canvas";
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -25,6 +25,8 @@
     "@babel/preset-env": "^7.11.0",
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
+    "@carbon/charts": "^0.41.57",
```
FlorianRappl, line 28: Wrong place. These are now rather dependencies of `esm-styleguide`.

package.json
```diff
@@ -41,6 +43,7 @@
     "css-loader": "^3.2.0",
     "cssnano": "^4.1.10",
     "css-minimizer-webpack-plugin": "^1.2.0",
+    "d3": "5.x",
```
FlorianRappl, line 46: Same applies here.

Also I don't see any reference to it - so do we really need this?

package.json
donaldkibet: Moved this to `esm-styleguide`

package.json
donaldkibet: `carbon/charts` uses `d3` as a dependent

FlorianRappl: Yes, but we only use it for stylesheets right now. So are the sheets depending on D3?

packages/esm-styleguide/src/_all.scss
```diff
@@ -1,5 +1,6 @@
 @import "./vars";
 @import "~carbon-components/scss/globals/scss/styles";
+@import "~@carbon/charts/styles/styles.scss";
```
FlorianRappl, line 3: Are these sheets depending on D3?

packages/esm-styleguide/package.json
```diff
@@ -39,7 +39,10 @@
     "access": "public"
   },
   "dependencies": {
+    "@carbon/charts": "^0.41.57",
+    "@carbon/charts-react": "^0.41.57",
     "carbon-components-react": "7.31.0",
+    "d3": "5.x",
```
FlorianRappl, line 45: Again, not clear why we have / need this.

packages/esm-styleguide/package.json
```diff
@@ -39,7 +39,10 @@
     "access": "public"
   },
   "dependencies": {
+    "@carbon/charts": "^0.41.57",
+    "@carbon/charts-react": "^0.41.57",
```
FlorianRappl, line 43: Why charts-react?

I only see you using `@carbon/charts`.

FlorianRappl (approved):
LGTM!

PR #101, opened 2021-04-29
Added esm-offline Package / Show Notifications when Static Files are Precached

manuelroemer:
Adds the new `esm-offline` package which contains shared functionality targeting offline development.

Updated the app shell setup to leverage the new functions provided by `esm-offline`:
* Show a notification when the static app shell/MF files are precached.
* Elevate offline data synchronization invocation to the app shell.


FlorianRappl (approved):
Alright, looks great to me! :beers:

PR #102, opened 2021-05-10
Breadcrumbs style mods

denniskigen:
- Set background color to `$openmrs-background-grey` so it matches the [designs](https://app.zeplin.io/project/605a0def8c1a5c07401482c1/screen/605a1ca5c27e8c0704aba8e3).
I've moved the styles to an `.scss` file to be able to take advantage of sass variables.

### DIFF
--- a/packages/esm-app-shell/webpack.config.js
+++ b/packages/esm-app-shell/webpack.config.js
@@ -39,7 +39,7 @@ const cssLoader = {
   options: {
     modules: {
       localIdentName:
-        "esm-patient-allergies__[name]__[local]___[hash:base64:5]",
+        "esm-patient-app-shell__[name]__[local]___[hash:base64:5]",
     },
   },
 };
@@ -97,7 +97,11 @@ module.exports = (env, argv = {}) => {
         },
         {
           test: /\.s[ac]ss$/i,
-          use: ["style-loader", cssLoader, "sass-loader"],
+          use: [
+            mode === "production"
+              ? { loader: require.resolve(MiniCssExtractPlugin.loader) }
+              : { loader: require.resolve("sass-loader") },
+          ],
         },
         {
           test: /\.(woff|woff2|png)?$/,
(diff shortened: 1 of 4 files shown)


packages/esm-app-shell/webpack.config.js
```diff
@@ -34,6 +34,16 @@ const openmrsConfigUrls = (process.env.OMRS_CONFIG_URLS || "")
   .map((url) => JSON.stringify(url))
   .join(", ");
 
+const cssLoader = {
+  loader: "css-loader",
+  options: {
+    modules: {
+      localIdentName:
+        "esm-patient-allergies__[name]__[local]___[hash:base64:5]",
```
FlorianRappl, line 42: The local ident name should maybe be something like `esm-app-shell_...`

packages/esm-app-shell/webpack.config.js
```diff
@@ -85,6 +95,10 @@ module.exports = (env, argv = {}) => {
             { loader: require.resolve("css-loader") },
           ],
         },
+        {
+          test: /\.s[ac]ss$/i,
+          use: ["style-loader", cssLoader, "sass-loader"],
```
FlorianRappl, line 100: We should use the MiniCssExtractPlugin.loader for production workloads.

Otherwise we end up with a large JS file instead of having a dedicated CSS file

packages/esm-app-shell/webpack.config.js
denniskigen: Whoops! 👀

packages/esm-app-shell/webpack.config.js
```diff
@@ -97,7 +97,11 @@ module.exports = (env, argv = {}) => {
         },
         {
           test: /\.s[ac]ss$/i,
-          use: ["style-loader", cssLoader, "sass-loader"],
+          use: [
```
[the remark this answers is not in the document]
denniskigen: Fixed. Thanks a lot for the tip @FlorianRappl!

FlorianRappl (approved):
LGTM - thanks :beers:!

PR #103, opened 2021-05-11
Expose an event from the SW which notifies about failed network requests

manuelroemer:
(Also removes Workbox log entries).


FlorianRappl (approved):
LGTM - thanks! :beers:

PR #105, opened 2021-05-15
feat(mf-463) : Implement search experience for nav bar

donaldkibet:
### What does this PR do

Implements the search experience for the nav bar

### Screenshoot

MF-463

### To do

1. Bring in test
- [x] Remove search from [esm-home](https://github.com/openmrs/openmrs-esm-home)


FlorianRappl:
In my opinion has nothing to do with primary navigation. This is rather related to the domain of patients.

It's now up for debate if this should be in patient-chart microfrontend itself, in a new microfrontend hosted in the patient chart monorepo, or in a new microfrontend.

We should use this chance to put it in the right domain (and not just put it where it technically "fits").

donaldkibet:
Closing this PR, this functionality will now be moved to [openmrs-esm-patient-management](https://github.com/openmrs/openmrs-esm-patient-management)

PR #107, opened 2021-05-19
Fix: Offline syncing API did not sync at the correct time and did not properly invoke the sync callbacks

manuelroemer:
Basically fixes the order in which the sync callbacks are invoked. We must give the app the change to setup the cbs before actually invoking them.


FlorianRappl (approved):
Awesome - thanks!

PR #109, opened 2021-05-21
fix login and primary navigation translation

donaldkibet:
### Description

At the moment the `initReactI18next` instance isn't passed to both login and primary navigation as a result of missing peer dependency of `react-i18next`


FlorianRappl (approved):
Great improvement!

PR #110, opened 2021-05-21
Add ability to show inline notifications and toasts

denniskigen:
Presently, the way we're handling notifications doesn't quite align with the Carbon [notifications pattern](https://www.carbondesignsystem.com/patterns/notification-pattern/).

Carbon offers [two main variants](https://www.carbondesignsystem.com/components/notification/usage/) of notifications: inline notifications and toasts. What we're referring to as toast notifications are inline notifications in the Carbon sense. We don't currently have a way to represent toasts.

~To remedy this, I've factored out the current toast notification implementation into a `NotificationComponent` that provides a type `prop` with which one can specify whether they want to show either an inline notification or a toast.~ 

To remedy this, I've refactored the existing `ToastComponent` to render toasts. I've also added a new `NotificationComponent` that renders inline notifications.

~Though their implementations differ, there are enough functional similarities between the two notification types to merit grouping them in a single component.~

Thanks, @nmalyschkin, for helping with this!

Screenshot 2021-05-22 at 00 53 33

_Toasts shown in the top right hand corner in the screenshot above while an inline notification is shown at the bottom._


packages/framework/esm-styleguide/src/notifications/active-notifications.component.tsx
```diff
@@ -0,0 +1,53 @@
+import React from "react";
```
FlorianRappl, line 1: `esm-styleguide` should be framework independent.

Can we somewhat rewrite this w/o using React? I think it would improve the solution, too - as right now need to create a dedicated React render tree (and face issues with the de-sync).

FlorianRappl (commented):
Yeah so in general I like this, but its breaking and we should try to minimize that.

Right now most changes deal with renaming `toast -> notification`.

Why don't we just have "notification" = inline notification and "toast" = "toast notification"? So toast could stay and would just change implementation etc.

FlorianRappl (approved):
LGTM - thanks!

PR #111, opened 2021-05-21
Add extraction of login and primary nav translation

donaldkibet:
### Description

Add extraction of translation on esm-login and esm-primary-navigation

### DIFF
--- a/packages/apps/esm-primary-navigation-app/package.json
+++ b/packages/apps/esm-primary-navigation-app/package.json
@@ -14,7 +14,8 @@
     "build": "webpack --mode=production",
     "analyze": "webpack --mode=production --env.analyze=true",
     "typescript": "tsc",
-    "lint": "eslint src --ext ts,tsx"
+    "lint": "eslint src --ext ts,tsx",
+    "extract-translations": "i18next 'src/**/*.component.tsx'"
   },
   "keywords": [
     "openmrs",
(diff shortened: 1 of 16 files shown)


packages/apps/esm-primary-navigation-app/package.json
```diff
@@ -14,7 +14,8 @@
     "build": "webpack --mode=production",
     "analyze": "webpack --mode=production --env.analyze=true",
     "typescript": "tsc",
-    "lint": "eslint src --ext ts,tsx"
+    "lint": "eslint src --ext ts,tsx",
+    "extract-translations": "i18next 'src/**/*.component.tsx'"
```
FlorianRappl, line 18: How about implementer tools and devtools?

Should we add this there, too?

donaldkibet: I think it would be great if all MFs could have translation, am adding translation for implementer tools and the devtools

FlorianRappl: Awesome! Thanks :rocket:

donaldkibet: At the moment translation for `dev-tool` will be difficult to achieve as the component rendering the overrides [here](https://github.com/joeldenning/import-map-overrides/blob/master/src/ui/list/list.component.js#L159) is a dependency.

trans

FlorianRappl: All good - thanks!

FlorianRappl (approved):
LGTM - thanks :beers:!

brandones:
@jonathandick Really? Seems crazy that this could produce a 500 on the server...

denniskigen:
@donaldkibet  Can this PR be restored? It appears like it was reverted to fix an unrelated issue.

PR #116, opened 2021-05-26
Refinements to notifications

denniskigen:
Following a conversation with Ciaran earlier today, I've made a few adjustments to notifications:

- Notifications can have either low or high contrast, with critical  system notifications or patient-related notifications being high contrast.
- Inline notifications should not dismiss automatically.
- Toasts should dismiss automatically after 5 seconds.

### DIFF
--- a/packages/framework/esm-styleguide/src/notifications/notification.component.tsx
+++ b/packages/framework/esm-styleguide/src/notifications/notification.component.tsx
@@ -8,8 +8,8 @@ export interface NotificationProps {
 export interface NotificationDescriptor {
   description: React.ReactNode;
   action?: React.ReactNode;
-  kind?: any;
-  lowContrast?: boolean;
+  kind?: InlineNotificationType;
+  critical?: boolean;
   millis?: number;
   title?: string;
 }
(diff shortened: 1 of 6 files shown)


packages/framework/esm-styleguide/src/notifications/notification.component.tsx
```diff
@@ -1,19 +1,15 @@
...
 export interface NotificationProps {
   notification: InlineNotificationMeta;
-  closeNotification(): void;
 }
 
 export interface NotificationDescriptor {
   description: React.ReactNode;
   action?: React.ReactNode;
   kind?: any;
+  lowContrast?: boolean;
```
FlorianRappl, line 12: Can we translate this into domain specific language?

I'd call this "critical" or "important" or so. `lowContrast` does not give any indicator what it should be about.

packages/framework/esm-styleguide/src/notifications/notification.component.tsx
```diff
@@ -30,38 +26,16 @@ export type InlineNotificationType =
...
-        actions={action}
-        kind={kind || "info"}
-        subtitle={description}
-        title={title || ""}
-      />
-    </div>
+    <InlineNotification
+      actions={action}
+      kind={kind || "info"}
+      lowContrast={lowContrast ?? true}
```
FlorianRappl, line 36: Boolean React attributes are usually chosen to be `false` by default. That way `<Foo bar />` is sufficient and no `<Foo bar={false} />` is usually not needed.

FlorianRappl:
Is this still in development @denniskigen ?

packages/framework/esm-styleguide/src/notifications/notification.component.tsx
denniskigen: Does this apply to optional parameters as well? Because if a value is not provided for `critical` it gets set to `undefined`. I want it to default to `false` while remaining an optional parameter.

FlorianRappl (approved):
LGTM - thanks!

PR #117, opened 2021-05-29
MF-607 : Inconsistent z-order in forms component and patient search

donaldkibet:
### Description

1. Updated `importmap.json` file
2. Fix inconsistent z-index for forms
3. Fix z-index on patient-search results to display correctly on home page

### Screenshoot

before 

Screenshot 2021-05-29 at 21 27 28

Screenshot 2021-05-29 at 21 25 31

after

Screenshot 2021-05-29 at 21 27 59


HerbertYiga:
cc @FlorianRappl

packages/shell/esm-app-shell/src/assets/importmap.json
```diff
@@ -1,25 +1,28 @@
...
-    "@openmrs/esm-home-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-home-app/2021-05-05_04bf41050f5e5e52d312de43c5317dc8f2216b65/openmrs-esm-home-app.js",
-    "@openmrs/esm-primary-navigation-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-primary-navigation-app/2021-05-05_dbb417c7346ba274618d5550c7d5b920b6c05b0e/openmrs-esm-primary-navigation-app.js",
+    "@openmrs/esm-patient-chart-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-chart-app/2021-05-27_1bd9478782616d67d0915f8cfdfa4b50903e3dae/openmrs-esm-patient-chart-app.js",
+    "@openmrs/esm-home-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-home-app/2021-05-16_009bf396e360d394a67d22755aa518935fefe16c/openmrs-esm-home-app.js",
+    "@openmrs/esm-primary-navigation-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-primary-navigation-app/2021-05-12_42dc53fd711a731d54c13c5beda48d6e13174deb/openmrs-esm-primary-navigation-app.js",
```
FlorianRappl, line 5: The core apps should be removed. They are anyway part of the bundle.

It's a good reminder; I should also remove them from the `import-map.json` of the ref app.

packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -101,3 +101,11 @@
     }
   }
 }
+
+.bx--side-nav {
+  z-index: 0;
```
FlorianRappl, line 106: This always feels arbitrary. Any way we get a better layout anyway? Not depending on any `z-index` would be great. If not, then its fine (but I'd like to know how you tested that this does not conflict with anything else...).

packages/shell/esm-app-shell/src/assets/importmap.json
donaldkibet: removed core-app from the import-map

packages/shell/esm-app-shell/src/assets/importmap.json
```diff
@@ -1,25 +1,25 @@
...
-    "@openmrs/esm-patient-programs-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-programs-app/2021-05-06_13b665933b4c3a7adf65d83af1b2a6e9654a2f01/openmrs-esm-patient-programs-app.js",
-    "@openmrs/esm-patient-allergies-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-allergies-app/2021-05-06_13b665933b4c3a7adf65d83af1b2a6e9654a2f01/openmrs-esm-patient-allergies-app.js",
-    "@openmrs/esm-patient-test-results-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-test-results-app/2021-05-06_13b665933b4c3a7adf65d83af1b2a6e9654a2f01/openmrs-esm-patient-test-results-app.js",
-    "@openmrs/esm-patient-clinical-view-app": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/esm-patient-clinical-view-app/2021-05-06_13b665933b4c3a7adf65d83af1b2a6e9654a2f01/openmrs-esm-patient-clinical-view-app.js"
+    "config-file": "https://spa-modules.nyc3.digitaloceanspaces.com/@openmrs/config/config.json",
```
FlorianRappl, line 3: If we want to use the config file we should configure the build to use this by default for the config file path.

We already have a way of including (an unlimited amount of) config files and we should use this. Import maps are / should be exclusively used for MFs.

packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -101,3 +101,7 @@
     }
   }
 }
+
+.bx--content-switcher-btn.bx--content-switcher--selected {
```
FlorianRappl, line 105: I only see trouble awaiting with this declaration.

I'd rather see this being set explicitly in a nested selected pointing to the specific area where there is trouble. In general, if we need to rely on z-index we did do something wrong in layouting - and should rather put corrections there.

With techniques such as flexbox or grid it should be easy to avoid the common layouting pitfalls that lead to excessive use of z-index.

PR #118, opened 2021-06-01
Set up docsify

brandones:
I'm proposing that we use Docsify to generate a documentation site. The [Docsify website](https://docsify.js.org/#/) is made using Docsify, and is a good example of its use. It's all just Markdown. In fact, our existing plain Markdown docs constitute a valid Docsify site.

Screenshot from 2021-06-01 16-56-09

So it seems like basically zero risk and high potential. If we hate it eventually, then we are left with perfectly good Markdown files.

And people (like non-developers) can keep editing these markdown files without worrying about Docsify.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Document</title>
+  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+  <meta name="description" content="Description">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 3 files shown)


FlorianRappl (commented):
And where should this run?

As long as its not published I don't see any value add. Also since GitHub does a good job rendering the markdown I also don't see much value add.

It also does not solve the biggest challenge with documentation in our setup: Fragmentation.

I'd first tackle the fragmentation issue before adding more things that need maintenance. I see that this could shine well, but right now we lack documentation, coherence, and centralization in that area. Docsify (or literally any other tool that does the same) don't help much for that.

brandones:
It would go on GitHub Pages. There is basically no maintenance overhead and the UI is nice. I'm going to be overhauling the docs this week.

brandones:
Also @FlorianRappl your input would be very welcome about how to structure the docs large-scale.

PR #119, opened 2021-06-08
MF-603: Fix Location selector shows the wrong list of locations

donaldkibet:
### Description

1. Add `login locations` tag to fetch the correct locations
2. [ticket description](https://issues.openmrs.org/projects/MF/issues/MF-603?filter=allissues)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,6 +2,7 @@ dist/
 README.md
 docs/
 *.svg
+*.png
 *.ico
 *.ejs
 packages/esm-styleguide/src/**/*.html
--- a/packages/apps/esm-login-app/src/choose-location/choose-location.component.tsx
+++ b/packages/apps/esm-login-app/src/choose-location/choose-location.component.tsx
@@ -52,7 +52,9 @@ export const ChooseLocation: React.FC<ChooseLocationProps> = ({
   React.useEffect(() => {
     if (isLoginEnabled) {
       const ac = new AbortController();
-      queryLocations("", ac).then((locations) => setLoginLocations(locations));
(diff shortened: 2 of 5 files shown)


jonathandick (approved):
LGTM

jonathandick (commented):
Sorry, one change, can we make this a configuration because not all implementations may use tags. Maybe a boolean which defaults to true called "useLoginLocationTag"

brandones:
@jonathandick Wait, what's the real-world alternative? Are there implementations for which all locations are login locations (including Unknown Location?)? Or is it that a different tag might be used?

jonathandick:
> @jonathandick Wait, what's the real-world alternative? Are there implementations for which all locations are login locations (including Unknown Location?)? Or is it that a different tag might be used?

Currently at ampath, we don't use tags at all. We haven't set up a location hierarchy for our sites. Perhaps we should but I would like to avoid having to set that up in order to get started with O3.0.

mseaton:
@brandones / @jonathandick - just happened to see this, but definitely agree that choosing to have a specific login location attached to the session should be optional, our login screen should not present this question if the server is not configured to support it, and none of the rest of the application is dependent upon the existence of a login location having been selected.

jonathandick:
> @brandones / @jonathandick - just happened to see this, but definitely agree that choosing to have a specific login location attached to the session should be optional, our login screen should not present this question if the server is not configured to support it, and none of the rest of the application is dependent upon the existence of a login location having been selected.

I think this is more about deciding what list of locations to show on login, rather than associating a session with a location.

FlorianRappl (approved):
LGTM - just run prettier and we are fine. :beers:

PR #120, opened 2021-06-11
Add support for JSON parameterization of build. And microfrontends.json -> frontend.json

brandones:
This is to support OpenMRS SDK integration. An implementer will be able to specify these properties in their distribution.

### DIFF
--- a/packages/tooling/openmrs/src/cli.ts
+++ b/packages/tooling/openmrs/src/cli.ts
@@ -179,7 +179,7 @@ yargs.command(
       )
       .default("fresh", false)
       .string("build-config")
-      .describe("build-config", "Path to a frontend build/assemble config JSON")
+      .describe("build-config", "Path to a SPA build config JSON")
       .string("spa-path")
       .default("spa-path", "/openmrs/spa/")
       .describe("spa-path", "The path of the application on the target server.")
@@ -228,8 +228,8 @@ yargs.command(
       .default("registry", "https://registry.npmjs.org/")
       .describe("registry", "The NPM registry used for getting the packages.")
       .string("config")
-      .default("config", "frontend.json")
-      .describe("config", "Path to a frontend build/assemble config JSON.")
+      .default("config", "spa-build-config.json")
+      .describe("config", "Path to a SPA build config JSON.")
       .boolean("fresh")
       .describe(
         "fresh",
@@ -279,7 +279,7 @@ yargs.command(
 
 yargs
   .epilog(
-    "The frontend build/assemble config JSON is a JSON file, typically `frontend.json`, which defines parameters for the `build` and `assemble` " +
+    "The SPA build config JSON is a JSON file, typically `frontend.json`, which defines parameters for the `build` and `assemble` " +
       "commands. The keys used by `build` are `apiUrl`, `spaPath`, `configUrls`, and `importmap`, each of " +
       "which is equivalent to the corresponding command line argument. The keys used by `assemble` are:\n" +
       "  microfrontends  \tAn object which specifies which microfrontends to include. It should have package names " +
--- a/packages/tooling/openmrs/src/cli.ts
+++ b/packages/tooling/openmrs/src/cli.ts
@@ -163,7 +163,7 @@ yargs.command(
 
 yargs.command(
   "build",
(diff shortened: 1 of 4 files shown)


packages/tooling/openmrs/src/cli.ts
```diff
@@ -178,6 +178,8 @@ yargs.command(
         "Determines if the output directory should be cleaned before the run."
       )
       .default("fresh", false)
+      .string("build-config")
+      .describe("build-config", "Path to a frontend build/assemble config JSON")
```
FlorianRappl, line 182: What's the default here? This should be `optional`.

packages/tooling/openmrs/src/cli.ts
```diff
@@ -225,11 +228,8 @@ yargs.command(
       .default("registry", "https://registry.npmjs.org/")
       .describe("registry", "The NPM registry used for getting the packages.")
       .string("config")
-      .default("config", "microfrontends.json")
-      .describe(
-        "config",
-        "The configuration for gathering the list of microfrontends to include."
-      )
+      .default("config", "frontend.json")
```
FlorianRappl, line 231: I think this name is not really right. Beforehand the name was more fitting. If you want to fit the general use case why don't you default to `openmrs.json`?

packages/tooling/openmrs/src/commands/build.ts
```diff
@@ -1,26 +1,48 @@
...
+  configUrls: Array<string>;
+  importmap: string;
+  spaPath: string;
+}
+
+function loadBuildConfig(buildConfigPath?: string): BuildConfig {
+  if (buildConfigPath) {
+    return JSON.parse(readFileSync(buildConfigPath, "utf8"));
+  } else {
+    return {} as BuildConfig;
```
FlorianRappl, line 33: If the file was invalid or not existing we should error out.

packages/tooling/openmrs/src/cli.ts
```diff
@@ -163,7 +163,7 @@ yargs.command(
 
 yargs.command(
   "build",
-  "Builds a new app shell using the provided configuration.",
+  "Builds a new app shell.",
```
FlorianRappl, line 166: Right now this does a bit more than that. Also, it is now inconsistent with `debug`, so some of these changes (using a config file) should be brought in there, too.

packages/tooling/openmrs/src/cli.ts
brandones: The default is undefined; yargs treats all arguments as optional unless [`demandOption`](https://github.com/yargs/yargs/blob/master/docs/api.md#demandOption) is used.

packages/tooling/openmrs/src/cli.ts
brandones: The application we're working on is being marketed as "OpenMRS Frontend 3.0," no longer as "OpenMRS Microfrontends." And this change makes this JSON file no longer only a list of microfrontends, but rather describe the configuration for the frontend application generally. `openmrs.json` would introduce namespace clash with server-side configuration files.

packages/tooling/openmrs/src/commands/build.ts
brandones: Yes, that is how it behaves. `readFileSync` throws an error if the file doesn't exist, and `JSON.parse` throws an error if the file is invalid. `{}` is returned only in the case that no build config is provided.

packages/tooling/openmrs/src/cli.ts
brandones: Yeah, definitely. For now, any suggestion about how to describe what `build` does? It builds the app shell, and what else?

packages/tooling/openmrs/src/cli.ts
FlorianRappl: Not sure I agree.

1. This is only for the default name; the user can give the file any name
2. Why would this tooling care about any other tooling (that is not directly used by it) or how something is marketed?

I'd say let's make the name `spa-build-config.json`.

brandones: Sure, Florian.

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #121, opened 2021-06-11
gracepotma-carbon-doc-suggestions

gracepotma:
Draft content for the carbon section of the Dev Guide. Totally open to changes and I know this doesn't include the more technical dev guidance we still need - just thought it would be good to have some starter content (and this is something I could actually help with a little).


docs/main/carbon
```diff
@@ -0,0 +1,39 @@
+# Using Carbon and the Styleguide
```
FlorianRappl, line 1: Looks good to me - thanks!

Can we name the file "carbon.md"?

brandones (approved):
Awesome, thanks so much Grace! Eventually I suspect that some of this should be part of the project-plan-type docs on the Wiki, and referenced from here? But glad to have this written anyway.

gracepotma:
> Awesome, thanks so much Grace! Eventually I suspect that some of this should be part of the project-plan-type docs on the Wiki, and referenced from here? But glad to have this written anyway.

Thanks @brandones! Yeah, FWIW the logic/background of Carbon, Design Systems vs Style Guides, etc etc is all documented in the Wiki; thought we may want some brief background in this guide since changing the UI is rather controversial. However definitely feel free to cut out any text you don't find adds value (for devs). I'll rearrange it so that the dev-actionable pieces are at the start rather than the end.

brandones:
Sounds good. What do you mean by "changing the UI is rather controversial," though? Not disagreeing, just clarifying.

PR #122, opened 2021-06-11
Upgrade to husky 6

brandones:
Following https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v6


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx pretty-quick --staged && lerna run lint && lerna run test && lerna run typescript && lerna run document
```
FlorianRappl, line 4: Doesn't run in windows.

We should have a platform agnostic solution here.

Personally, I'd drop husky anyway. At least drop it as pre-commit and make it pre-push. Commits should not be dictated (or blocked) by some tooling. Pushes should be checked.

brandones: I don't think it matters so much since you can always `--no-verify`, but sure, I've switched it to running `pre-push`.

I have no idea what kind of shell syntax Windows does and doesn't support, and I don't have a Windows machine to test on. If you can point me to some helpful documentation on the subject I'd be happy to make it cross-platform; otherwise would you mind fixing it up yourself, assuming you have a Windows machine?

FlorianRappl: Sorry I don't use Windows. The shell situation on Windows is also not straight forward as there are multiple shells (most popular PowerShell and Command) that go with different syntax (at least for anything more than just "run this command").

I think my assumption here was wrong though. I think husky runs this script and tries to use bash if available somewhere (which should be the case for most Windows systems that installed git).

Still we could improve the situation slightly by changing the script to:

```
npx pretty-quick --staged
lerna run lint
lerna run test
lerna run typescript
lerna run document
```

I'm also unsure about the `. "$(dirname "$0")/_/husky.sh"` part. Is that necessary? Haven't seen it before.

brandones: Right, that makes sense, based on what I'm reading now. About the `husky.sh`, see https://github.com/typicode/husky/issues/929 .

FlorianRappl (approved):
:beers:

.husky/pre-commit
FlorianRappl: Yeah thanks for these pointers. Some of my doubts with Husky v6 have been confirmed in this issue https://github.com/typicode/husky/issues/921, but I also see its advantages. Not sure if we gain something by the update (besides having that infamous .husky folder).

In any case we should try to roll this out to other repos to have it (at least for our stuff) consistent. Would be appreciated. Awesome work - thanks!

PR #123, opened 2021-06-15
Allow specifying offline response data via HTTP headers

manuelroemer:
Used in form-entry. Essentially allows a MF to control certain aspects of an offline fallback response produced by the SW (before, we simply returned e.g. status code 204 No Content for any failure - now the MF can actively decide what happens to a certain degree).


FlorianRappl (approved):
LGTM!

PR #125, opened 2021-06-16
MF-607: Fix Inconsistent z-order in forms component

donaldkibet:
### Description

 Fixed the inconsistent content z-index of the forms, this happened because the entire primary-navigation app was `position fixed` and in the document flow resulting in the errant display.

### Screenshoot

MF-607

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/root.scss
+++ b/packages/apps/esm-primary-navigation-app/src/root.scss
@@ -23,11 +23,9 @@ $labeldropdown: #c6c6c6;
 }
 
 .headerPanel {
-  min-height: fit-content;
+  height: max-content;
 }
 
-.primaryNavContainer {
-  position: fixed;
-  top: 0;
-  z-index: 3;
+.primaryNavContainer header {
+  z-index: 3 !important
 }


packages/apps/esm-primary-navigation-app/src/root.scss
```diff
@@ -23,11 +23,9 @@ $labeldropdown: #c6c6c6;
...
 .headerPanel {
-  min-height: fit-content;
+  height: max-content;
 }
 
-.primaryNavContainer {
-  position: fixed;
-  top: 0;
+.primaryNavContainer header {
   z-index: 3;
```
FlorianRappl, line 30: I think it makes sense to actually remove the fixed position for the primary nav bar. It should not have any z index at all (neither the bar, nor its children).

I've seen you removed the fixed position already from the container - why does the header need a z-index?

donaldkibet: carbon-components provide a header z-index of `8000` which is higher than the workspace z-index of 10 in tablet mode workspace [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/dd79d473beb52a686333933460441a2b3bd362af/packages/esm-patient-chart-app/src/workspace/context-workspace.scss#L6)

FlorianRappl: Hm but if we want to override the one from carbon I'd suggest making it an `!important`. Otherwise, you'll never know if they remain on the same specifity. In general, it not use any value though. I'd write `unset`.

FlorianRappl (approved):
LGTM - thanks!

PR #126, opened 2021-06-16
Add Markdown link check to GitHub Actions

brandones:
Adds a GitHub Action that uses https://github.com/gaurav-nelson/github-action-markdown-link-check to check that all the links in our documentation are working. cc @gracepotma


brandones:
I think this check made an amazing case for itself: https://github.com/openmrs/openmrs-esm-core/runs/2842350437?check_suite_focus=true

.github/workflows/docs.yml
```diff
@@ -0,0 +1,19 @@
+name: Check documentation
```
brandones, line 1: This file is the actual change this PR is about; all the rest is just fixing errors that this check identified.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+  schedule:
+    - cron: "0 9 * * *"
```
brandones, line 9: Runs this check daily at 9am, in case any of the external links get stale.

jonathandick (approved):
LGTM

.github/workflows/docs.yml
FlorianRappl: Good thinking!

FlorianRappl (approved):
:beers:

PR #127, opened 2021-06-17
MF-623: show search icon on far left

seankwon:
# Description

Using row -reverse to show the search icon first. if someone could let me know how to do this better, please tell me.

the major culprit is here: https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-search-app/src/index.ts#L22 where we are loading the extension

Screen Shot 2021-06-16 at 8 35 05 PM

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
@@ -1,3 +1,4 @@
 .topNavActionSlot {
   display: flex;
+  flex-direction: row-reverse;
 }


packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
```diff
@@ -1,3 +1,4 @@
 .topNavActionSlot {
   display: flex;
+  flex-direction: row-reverse;
```
FlorianRappl, line 3: Hehe. This is not how we handle this.

It should be ordered correctly via configuration. Since the search icon is an extension with the given commit we would never be sure that it remains on the left.

seankwon: yea, i dont really know how to it via loading the extension since i dont know how to load extensions in certain orders.

brandones: Yeah, this is a conversation that we've tried to have before, [with limited success](https://openmrs.slack.com/archives/GQJA12F4L/p1615998719009600). I maintain that it would be sensible to add an order index parameter to extension definitions. Extensions would be ordered within slots by their order index.

```typescript
extensions: [
  {
    id: "foo",
    slot: "bar-slot",
    // ... everything as usual
    order: 4
]
```

You're correct that we do need to solve this problem.

FlorianRappl: Well, this problem is already solved if you ask me.

1. We have the generic meta that solves it (just put the `order` there and leave it up to the slot to use it).
2. We have configurations (which I refer to in this issue / and in the conversation on Slack when Sean mentioned the issue).

seankwon: sorry, not asking if it exists - just need documentation to refer to or code examples

PR #128, opened 2021-06-17
Use typedoc-plugin-no-inherit to simplify docs

brandones:


FlorianRappl (approved):
LGTM - thanks!
Just the build needs to be working.

PR #132, opened 2021-06-21
Fixed login location picker showing wrong no of locations

donaldkibet:
### Description

1. Fixed location picker showing the wrong number of locations when a user searches for a location and clears the search, it displays the wrong number of locations as shown in the gif below.
2. Add height for location results

before
Screenshot 2021-06-21 at 09 59 24

after
Screenshot 2021-06-21 at 09 59 39

### Screenshoot
wrongNoOfLocations

### DIFF
--- a/packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
+++ b/packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
@@ -47,6 +47,11 @@
 .locationResultsContainer {
   display: flex;
   overflow-y: auto;
+  margin-top: 0.5rem;
+}
+
+:global(.omrs-breakpoint-gt-tablet) .locationResultsContainer {
+  height: 23rem;
 }
 
 .locationRadioButton {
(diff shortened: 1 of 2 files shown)


packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
```diff
@@ -47,6 +47,8 @@
 .locationResultsContainer {
   display: flex;
   overflow-y: auto;
+  margin-top: 0.5rem;
+  height: 23rem;
```
FlorianRappl, line 51: Sure about that? What about mobile here?

donaldkibet: Since mobile and tablet have keyboard appear as one searches for location not having a height suits better. I have applied the height on desktop devices only

FlorianRappl (approved):
LGTM ! Thanks 🍻

PR #134, opened 2021-06-21
Tooling: Fix build config resolution

brandones:


packages/tooling/openmrs/package.json
```diff
@@ -7,7 +7,7 @@
     "openmrs": "./dist/cli.js"
   },
   "scripts": {
-    "start": "npm run watch",
+    "start": "./dist/cli.js",
```
FlorianRappl, line 10: You mean "node ./dist/cli.js"?

FlorianRappl: (Yes, the file has a shebang, but still if its not +x or if its on other systems it might not work)

FlorianRappl (approved):
Great!

PR #136, opened 2021-06-22
MF-632: Follow-up fix

denniskigen:
Constrain the height of the search results div to the dimensions specified in the design. The height should not change when the number of results displayed changes.

Screenshot 2021-06-22 at 12 58 12

### DIFF
--- a/packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
+++ b/packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
@@ -37,6 +37,7 @@
 .searchResults {
   @extend .bodyShort02;
   margin: 1rem 0rem;
+  height: 26.125rem;
 }
 
 .resultsCount {


packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
```diff
@@ -37,6 +37,7 @@
 .searchResults {
   @extend .bodyShort02;
   margin: 1rem 0rem;
+  height: 26.125rem;
```
FlorianRappl, line 40: Should that be max-height? How is this on mobile?

FlorianRappl (commented):
LGTM!

packages/apps/esm-login-app/src/location-picker/location-picker.component.scss
denniskigen: Not great. I've asked Ciaran about it and it appears the smallest viewport we're designing for right now is a 7-inch tablet. That said, I can reduce the height to something like `21rem` on the mobile viewport. Should I do so in this instance?

FlorianRappl: Well, why not have a `max-height: 100%` or `max-height: 100vh` rule in there? You can then leave the original height, but its always cut if its exceeding the full height.

FlorianRappl (approved):
LGTM! :beers:

PR #137, opened 2021-06-22
Restored extraction of login and primary nav translation

donaldkibet:
### Description

Restored as it has been removed as mentioned [here](https://github.com/openmrs/openmrs-esm-core/pull/111#issuecomment-865796360)


FlorianRappl (approved):
LGTM :beers:!

PR #140, opened 2021-06-24
Service Worker Resilience: Retries of caching attempts

manuelroemer:
Still WIP, but I want to gather some initial feedback of whether this approach is right. Feel free to review!

# Open Tasks / Questions

- [x] Testing of the behavior.
- [x] Q: When should we retry? Like it is currently (during _every_ caching attempt) or only in selected cases? 
- [x] Q: Are the retry "values", i.e. num of attempts, duration etc. good enough? (I think they might have to be increased).

### DIFF
--- a/packages/framework/esm-api/src/types/visit-resource.ts
+++ b/packages/framework/esm-api/src/types/visit-resource.ts
@@ -1,6 +1,7 @@
 import { OpenmrsResource } from "./openmrs-resource";
 
 export interface NewVisitPayload {
+  uuid?: string;
   location: string;
   patient?: string;
   startDatetime: Date;
--- a/packages/shell/esm-app-shell/src/service-worker/caching.ts
+++ b/packages/shell/esm-app-shell/src/service-worker/caching.ts
@@ -1,4 +1,4 @@
-import { ImportMap } from "@openmrs/esm-framework";
+import { ImportMap } from "@openmrs/esm-globals";
 import { retry } from "@openmrs/esm-utils";
 import { absoluteWbManifestUrls, omrsCacheName } from "./constants";
 import { fetchUrlsToCacheFromImportMap } from "./importMapUtils";
--- a/packages/shell/esm-app-shell/src/service-worker/caching.ts
+++ b/packages/shell/esm-app-shell/src/service-worker/caching.ts
@@ -1,4 +1,5 @@
-import { ImportMap, retry } from "@openmrs/esm-framework";
+import { ImportMap } from "@openmrs/esm-framework";
+import { retry } from "@openmrs/esm-utils";
 import { absoluteWbManifestUrls, omrsCacheName } from "./constants";
 import { fetchUrlsToCacheFromImportMap } from "./importMapUtils";
 import { ServiceWorkerDb } from "./storage";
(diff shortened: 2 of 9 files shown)


FlorianRappl:
> When should we retry? Like it is currently (during every caching attempt) or only in selected cases?

Sounds fine to me. Everything we want to cache is something we should really try to get loaded. So yes.

> Are the retry "values", i.e. num of attempts, duration etc. good enough? (I think they might have to be increased).

Personally, I think they are fine. It should be between 3 and 5 entries and the total max. length may not exceed 40 seconds if you ask me. But the chosen values are fine.

Regarding the testing of the behavior; good question. I need to think about it. Honestly, we need to have an E2E testing strategy here (automated). I think we need to ask for support here as presumably there is already the QA team active.

packages/shell/esm-app-shell/src/service-worker/caching.ts
```diff
@@ -1,5 +1,6 @@
-import { ImportMap } from "@openmrs/esm-globals";
-import { absoluteWbManifestUrls, indexUrl, omrsCacheName } from "./constants";
+import { ImportMap } from "@openmrs/esm-framework";
+import { retry } from "@openmrs/esm-utils";
```
manuelroemer, line 2: I faced some weird issue during debugging just now where the SW could no longer be loaded because apparently the JS file was invalid. This happens specifically when importing `retry` from `esm-framework`. It works just fine with the `esm-utils` import here. 

Any idea what might be causing this?

FlorianRappl: Yes, don't import framework. Its very heavy and the service worker should - if any - just rely on `esm-offline`. `esm-framework` also contains `esm-globals`, which works against the DOM. In the service worker you don't have access to the DOM, so this will not work.

FlorianRappl: Note that the import of `ImportMap` does not have any impact as this is a type, so TS will erase the actual import (and you will not face the DOM access issue there).

manuelroemer: That makes a lot of sense suddenly, thanks!

manuelroemer: I went over the SW again and replaced one lingering esm-framework import with esm-global.

packages/framework/esm-api/src/types/visit-resource.ts
```diff
@@ -1,6 +1,7 @@
 import { OpenmrsResource } from "./openmrs-resource";
 
 export interface NewVisitPayload {
+  uuid?: string;
```
manuelroemer, line 4: Smuggling this one in so that I don't have to do another PR for this small change.

FlorianRappl: ;) Works!

packages/shell/esm-app-shell/src/service-worker/caching.ts
FlorianRappl: Great!

FlorianRappl (approved):
Great! :beers:

PR #141, opened 2021-06-24
Standardise UI shell right panel button behaviour

denniskigen:
Ensures that the icon buttons for items in the UI shell right panel header change to a dismiss button (using the `Close20` icon) when the panel is expanded. 

This applies to the App menu and the User menu. In future, this will also apply to the Notifications menu.


FlorianRappl (approved):
Great, thanks!

PR #142, opened 2021-06-24
Remove offline setup success notification

denniskigen:
The success notification being shown upon successful setup of the offline mode has been superseded by the new offline mode designs.

Screenshot 2021-06-24 at 17 40 32

We can now remove this notification per @manuelroemer and Ciaran.

### DIFF
--- a/packages/shell/esm-app-shell/src/run.ts
+++ b/packages/shell/esm-app-shell/src/run.ts
@@ -213,13 +213,6 @@ async function setupServiceWorker() {
   if (navigator.onLine) {
     try {
       await Promise.all([precacheImportMap(), precacheSharedApiEndpoints()]);
-
-      showNotification({
-        title: "You can now go offline",
-        description:
-          "The application is done preparing the offline mode. You can now use the website without an internet connection.",
-        kind: "info",
-      });
     } catch (e) {
       showNotification({
         critical: true,


packages/shell/esm-app-shell/src/run.ts
```diff
@@ -213,13 +213,6 @@ async function setupServiceWorker() {
   if (navigator.onLine) {
     try {
       await Promise.all([precacheImportMap(), precacheSharedApiEndpoints()]);
-
-      showNotification({
```
FlorianRappl, line 217: Yes this is now removed. But I miss the implementation of the new designs.

FlorianRappl:
I know the PR is titled

> Remove offline setup success notification

but it does not feel right just to remove something without having the proper replacement included.

denniskigen:
I hear you, Florian, and agree fully. I think the primary motivation for getting rid of it is that it's a bit of a nag having to dismiss it manually every time.

FlorianRappl:
> The primary motivation for getting rid of it is that it's a bit of a nag having to dismiss it manually every time.

But by that logic we should also remove, e.g., the "found missing backend dependencies". And regarding the manual dismissal; this - I think - was done separately and as told my Cairan IIRC. The dialogs should all dismiss after a couple of seconds.

denniskigen:
Closing in favour of waiting for work on the new offline mode designs before removing this notification.

PR #143, opened 2021-06-29
MF-637 : Add ability to close side-nav when user clicks on a link

donaldkibet:
### Description

Add ability to close the side-nav bar on tablet-mode when a user clicks on a link item

### Screenshoot
close-side-nav

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
@@ -5,7 +5,7 @@ import { SideNav } from "carbon-components-react/es/components/UIShell";
 import styles from "./side-menu-panel.component.scss";
 
 interface SideMenuPanelProps extends SideNavProps {
-  hidePanel: Function;
+  hidePanel: () => void;
 }
 
 const SideMenuPanel: React.FC<SideMenuPanelProps> = ({
@@ -13,18 +13,22 @@ const SideMenuPanel: React.FC<SideMenuPanelProps> = ({
   hidePanel,
 }) => {
   const menuRef = useRef(null);
-  const current = menuRef?.current;
 
   useEffect(() => {
     const handleClickOutside = (event: MouseEvent) => {
-      if (current && !current.contains(event.target)) {
+      if (menuRef?.current && !menuRef.current.contains(event.target)) {
         hidePanel();
       }
     };
 
     document.addEventListener("click", handleClickOutside);
     return () => document.removeEventListener("click", handleClickOutside);
-  }, [current, hidePanel]);
+  }, [menuRef, hidePanel]);
+
+  React.useEffect(() => {
+    window.addEventListener("popstate", hidePanel);
+    return window.addEventListener("popstate", hidePanel);
+  }, [hidePanel]);
 
   return (
     expanded && (
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -78,7 +78,10 @@ const Navbar: React.FC<NavbarProps> = ({
             <HeaderMenuButton
               aria-label="Open menu"
               isCollapsible
-              onClick={() => togglePanel("sideMenu")}
+              onClick={(event) => {
+                togglePanel("sideMenu");
+                event.stopPropagation();
+              }}
               isActive={isActivePanel("sideMenu")}
             />
           )}


packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -35,7 +35,10 @@ const SideMenuPanel: React.FC<SideMenuPanelProps> = ({
         isChildOfHeader={expanded}
         className={styles.link}
       >
-        <ExtensionSlot extensionSlotName="nav-menu-slot" />
+        <ExtensionSlot
+          onClick={() => hidePanel()}
```
FlorianRappl, line 39: This also closes when *anything* was clicked; not only a link.

What we could do instead, is that we listen for navigation events in the nav menu. If a navigation happened, then we just set the state to hide.

This way it would also close, e.g., when you press on the home link above. Which is - I think - what should happen.

Could be similar to:

```ts
React.useEffect(() => {
  history.addEventListener('pop', hidePanel);
  return () => history.removeEventListener('pop', hidePanel);
}, []);
```

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -3,6 +3,7 @@ import { ExtensionSlot } from "@openmrs/esm-framework";
 import { SideNav } from "carbon-components-react/es/components/UIShell";
 import { SideNavProps } from "carbon-components-react";
 import styles from "./side-menu-panel.component.scss";
+import { useHistory } from "react-router-dom";
```
FlorianRappl, line 6: Why?

I think we should stay generic here. `useHistory` only works if we are in a router context.

Why not use the generic browser history API as I outlined?

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -26,6 +28,13 @@ const SideMenuPanel: React.FC<SideMenuPanelProps> = ({
     return () => document.removeEventListener("click", handleClickOutside);
   }, [current, hidePanel]);
 
+  React.useEffect(() => {
+    const unlisten = history.listen(
```
FlorianRappl, line 32: Unnecessary ceremony. Just `return history.listen(...)`. But then again, I'd not use `react-router-dom` to avoid coupling.

FlorianRappl:
A review was requested, but I don't see the last comments being "outdated" / resolved.

FlorianRappl:
Hi @donaldkibet since I saw you pushing to the branch... are you planning to address the comments?

donaldkibet:
> Hi @donaldkibet since I saw you pushing to the branch... are you planning to address the comments?

Hi, am addressing the comments

FlorianRappl:
Hi @donaldkibet are you still on that? Just wanted to know if you need help / when you plan to work on it. Thanks!

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -12,24 +12,14 @@ const SideMenuPanel: React.FC<SideMenuPanelProps> = ({
...
-        hidePanel();
-      }
-    };
-
-    document.addEventListener("click", handleClickOutside);
-    return () => document.removeEventListener("click", handleClickOutside);
-  }, [current, hidePanel]);
+  React.useEffect(() => {
+    window.addEventListener("popstate", () => hidePanel());
+    return () => window.removeEventListener("popstate", () => hidePanel());
```
FlorianRappl, line 17: That will not work as you create a new function and thus the reference does not equal.

Why don't you just pass in the `hidePanel` function?

```js
useEffect(() => {
    window.addEventListener("popstate", hidePanel);
    return () => window.removeEventListener("popstate", hidePanel);
}, [hidePanel]);
```

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -1,35 +1,25 @@
-import React, { useEffect, useRef } from "react";
+import React from "react";
```
FlorianRappl, line 1: Any reason for this change?

Why do I ask? I just aligned the whole `esm-core` repository to use direct imports, i.e., `import React, { useEffect, useRef } ...` in this case. Would be great to follow one style.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -1,35 +1,25 @@
...
 
 const SideMenuPanel: React.FC<SideMenuPanelProps> = ({
   expanded,
   hidePanel,
 }) => {
-  const menuRef = useRef(null);
-  const current = menuRef?.current;
-
-  useEffect(() => {
-    const handleClickOutside = (event: MouseEvent) => {
```
FlorianRappl, line 19: Shouldn't that be kept, too?

I mean we still want to close when the user clicks somewhere else, right?

FlorianRappl (approved):
LGTM - thanks!

PR #145, opened 2021-06-30
Removed the background color from breadcrumbs

vasharma05:
### What does this PR include?

I removed the fixed grey background color from the breadcrumbs.

CC. @FlorianRappl


FlorianRappl (approved):
LGTM!

PR #146, opened 2021-07-05
Add modal functionality to esm-code

nmalyschkin:
@FlorianRappl This is after our conversation on Friday. Is this going in the direction you imagined?


packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -30,6 +30,8 @@
   <body>
     <%= htmlWebpackPlugin.tags.bodyTags %>
     <div class="omrs-toasts-container"></div>
+    <div class="omrs-inline-notifications-container"></div>
+    <div class="omrs-modals-container" style="visibility: hidden;"></div>
```
FlorianRappl, line 34: This is good. The `visibility: hidden` may not be needed. Maybe we should set the initial style to hidden, or have a dedicated class for it, which may be toggled and is initially given.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -301,6 +302,7 @@ export function run(configUrls: Array<string>) {
   const provideConfigs = createConfigLoader(configUrls);
 
   integrateBreakpoints();
+  setupModalsContainer();
```
FlorianRappl, line 305: Hm I expected that the app shell provides the container to the library. The library itself should not (need to) know the container.

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  );
+
+  const instance: Partial<ModalInstance> = {
+    container,
+    mounted: false,
+  };
+
+  const popFromStack = () => {
+    const index = modalStack.findIndex((x) => x === instance);
+    if (~index) {
```
FlorianRappl, line 76: Yes this is clever. I am just not sure if `index !== -1` is not more readable. We can keep it as is for now.

packages/shell/esm-app-shell/src/run.ts
nmalyschkin: So it would look like this here: `setupModalsContainer(document.querySelector<HTMLElement>(".omrs-modals-container"))`?

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface ModalInstance {
+  container: HTMLElement;
+  close: () => void;
+  mounted: boolean;
+}
+
+let modalContainer: HTMLElement | null;
+const modalStack: Array<ModalInstance> = [];
+
+export function setupModalsContainer() {
```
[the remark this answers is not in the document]
nmalyschkin: I was considering using a subscribable state container but it seemed over the top.
Will place the `modalStack` and `modalContainer`!

packages/shell/esm-app-shell/src/run.ts
FlorianRappl: Yes, even though I would make it symmetrical to the toasts etc. (i.e., calling it `renderModals`).

packages/framework/esm-react-utils/src/showReactModal.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useContext } from "react";
```
FlorianRappl, line 1: I think we can remove this file.

packages/framework/esm-styleguide/src/modals/_modals.scss
```diff
@@ -0,0 +1,18 @@
...
+  position: fixed;
+  width: 100vw;
+  height: 100vh;
+  background-color: rgba(0, 0, 0, 0.5);
+  z-index: 9000;
+  visibility: hidden;
+
+  div {
+    position: fixed;
+    width: 100vw;
```
FlorianRappl, line 11: Should the modals content be really 100%?

Shouldn't it be rather fixed (e.g., 600x800) with max-width and max-height being 100%?

Also the `position: fixed` is not needed - its container is already fixed.

packages/framework/esm-styleguide/src/modals/_modals.scss
```diff
@@ -0,0 +1,18 @@
+.omrs-modals-container {
+  position: fixed;
+  width: 100vw;
+  height: 100vh;
+  background-color: rgba(0, 0, 0, 0.5);
+  z-index: 9000;
+  visibility: hidden;
+
+  div {
```
FlorianRappl, line 9: You mean `> div`? Right now every descendant div in this container would get this.

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+  modalStack: Array<ModalInstance>;
+}
+
+const modalStore = createGlobalStore<ModalState>("globalModalState", {
+  modalContainer: null,
+  modalStack: [],
+});
+
+// handling any modal state change here, updating the container
+modalStore.subscribe(({ modalStack, modalContainer }) => {
```
FlorianRappl, line 21: I think the subscribe would only start within the `renderModals`. That way, you also don't need to have any instance global here.

FlorianRappl: (Just to be clear: Let's move the subscribe *inside* the `renderModals` function).

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+ * @param onClose
+ * @returns
+ */
+export function showModal(
+  extensionId: string,
+  props: Record<string, any> = {},
+  onClose: () => void = () => {}
+) {
+  const container = document.createElement("div");
+  const cleanup = renderExtension(
```
FlorianRappl, line 69: I think this would be implicit by the `renderModals`. The `showModal` would only place the desired modal in the state container.

packages/framework/esm-styleguide/src/modals/index.tsx
nmalyschkin: I had it this way initially but what if someone (for some stupid reason) decides to call the `renderModals` more than once (e.g. execute `run` a second time). If this is not a concern I'll move it into `renderModals`.

packages/framework/esm-styleguide/src/modals/_modals.scss
FlorianRappl: Hm - I was under the (potentially false) impression that the container should be responsible for rendering the chrome of the modal. It should render the close button and provide the canvas where the content would be placed in. Right now it also does this functionally (see escape key handling + close handling).

packages/framework/esm-styleguide/src/modals/index.tsx
FlorianRappl: I think this is no real concern (otherwise, it would be the same with toasts etc.).

packages/framework/esm-styleguide/src/modals/index.tsx
nmalyschkin: This is done here because `showModals` is returning the dispose function, which needs to call this cleanup function. We place the modal instance in the state container, therefore we need to construct the close function in `showModals`.

FlorianRappl: Hm - I am not sure actually.

What should be returned is a way to remove the dialog from the state container. Once its removed, internally, the dispose should be called. We shouldn't return the dispose to the caller, but rather a way to trigger this process.

Therefore, the whole dispose action would be inside the renderModals, which subscribes to the store and handles the rendering mechanism.

packages/framework/esm-styleguide/src/modals/_modals.scss
nmalyschkin: I considered providing a framework for modals (close button, fixed size etc) but I couldn't find a smallest common denominator (the yes/no dialog probably wouldn't want to have a close button). If you want we can elevate this question in the slack channel.

packages/framework/esm-styleguide/src/modals/index.tsx
nmalyschkin: However we turn it, we need to pass back a function that triggers the destruction process. I'll restructure it so the `renderExtension` call is made inside the store update handler.

packages/framework/esm-styleguide/src/modals/_modals.scss
FlorianRappl: Well, since you implemented the escape key its already known that you can always close.

We now have two options:

- Either we set some defaults that need to be used
- Or we rely on the user to pass in options (second or third argument) for showModals

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #148, opened 2021-07-07
Updated useVisit to also use getStartedVisit subject | Added global event for precaching static data when loading the app

manuelroemer:


FlorianRappl (approved):
LGTM! Thanks :beers:

PR #149, opened 2021-07-11
Provide notifications panel extension slots

denniskigen:
- Expose an extension slot for the notifications menu button in the navbar. The SVG button icon rendered in that slot would vary based on the following conditions:

  - `Notification20` as the default.
  - `NewNotification20` when there are new unread notifications.
  - `Close20` when the notifications panel is open.

- Expose an extension slot for the notifications menu panel. The patient alerts logic would ostensibly live in a patient management-type microfrontend which would handle how to fetch and render (and potentially store) those reminders. This extension slot would provide information about whether the panel is open or not (via the `expanded`) boolean prop.

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/notifications-menu-panel.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/notifications-menu-panel.component.tsx
@@ -1,9 +1,12 @@
 import React from "react";
+import { useTranslation } from "react-i18next";
+import { ExtensionSlot } from "@openmrs/esm-framework";
 import { HeaderPanel, HeaderPanelProps } from "carbon-components-react";
 import styles from "./notifications-menu.component.panel.scss";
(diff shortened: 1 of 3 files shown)


packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/notifications-menu-panel.component.tsx
```diff
@@ -17,11 +20,10 @@ const NotificationsMenuPanel: React.FC<NotificationsMenuPanelProps> = ({
...
     >
       <h1 className={styles.heading}>{t("notifications", "Notifications")}</h1>
-      <div>
-        <p className={styles.emptyNotifications}>
-          {t("noNotifications", "You currently have no notifications")}
-        </p>
-      </div>
+      <ExtensionSlot
+        extensionSlotName="notifications-nav-menu-slot"
+        state={{ expanded: expanded }}
```
FlorianRappl, line 25: `state` should be memoized.

PR #150, opened 2021-07-14
Found modules with unresolved backend dependencies.

alaboso:
- The current implementation comparing required backend dependencies with actual installed dependencies compares with an incorrect version. e.g. openmrs-spa -> implementer tools -> backend modules (webservices.rest installed _**2.2**_ whereas actual is _**2.31.0-SNAPSHOT.7e24fb**_
 - semver.satisfies(installedVersion, requiredVersion) ("2.31.0", "^2.2") return false. @samuelmale  @FlorianRappl - Is this the expected outcome?
 - Kindly

### DIFF
--- a/packages/apps/esm-implementer-tools-app/src/backend-dependencies/openmrs-backend-dependencies.ts
+++ b/packages/apps/esm-implementer-tools-app/src/backend-dependencies/openmrs-backend-dependencies.ts
@@ -102,7 +102,9 @@ function getMisMatchedBackendModules(
   for (let uuid in installedAndRequiredBackendModules) {
     const requiredVersion = installedAndRequiredBackendModules[uuid].version;
     const moduleName = installedAndRequiredBackendModules[uuid].uuid;
-    const installedVersion = installedBackendModules.find((mod) => mod.uuid == moduleName)?.version ?? "";
+    const installedVersion =
+      installedBackendModules.find((mod) => mod.uuid == moduleName)?.version ??
+      "";
 
     if (!isVersionSatisfied(requiredVersion, installedVersion)) {
       misMatchedBackendModules.push({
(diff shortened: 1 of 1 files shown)


alaboso:
The installed modules in the backend: (App UI Module --> 1.9.0  && Rest Web Services -- > 2.31.0-SNAPSHOT.7e24fb)

spa-installed-backend-modules

The implementer tools display of "installed modules:" ((App UI Module --> 1.2.6-SNAPSHOT  && Rest Web Services -- > 2.2))

spa-impl-tools-backend-modules


Also - Would it be better to change the versions as `greater than` instead of strict `equals to`? case: webservices rest module v2.31.* is **required** by login esm (currentProvider attribute in session json) but the same feature may not be required by another esm (which may require same module but not necessarily same feature of currentProvider and therefore could work with a lower version)??

FlorianRappl:
> Also - Would it be better to change the versions as greater than instead of strict equals to?

Not sure what you mean. Usually, the versions are specified using semver constraints such as `^`.

> case: webservices rest module v2.31.* is required by login esm

Not sure I understand your case. If a MF requires the BE module then a MF requires it. It does not matter if other MFs require this or not. The fact that you load a MF that requires a certain BE is reason enough to alert if that constraint cannot be resolved.

Also note that your change

- does not come with any unit test
- makes a statement I cannot follow ("emver.satisfies(installedVersion, requiredVersion) ("2.31.0", "^2.2") return false.", this is unit tested and working)
- seems to deal with a change in the data model

Some clarification on these points would be desired.

packages/apps/esm-implementer-tools-app/src/backend-dependencies/openmrs-backend-dependencies.ts
```diff
@@ -100,9 +100,11 @@ function getMisMatchedBackendModules(
 ) {
   let misMatchedBackendModules: BackendModule[] = [];
   for (let uuid in installedAndRequiredBackendModules) {
-    const installedVersion = installedBackendModules[uuid].version;
     const requiredVersion = installedAndRequiredBackendModules[uuid].version;
     const moduleName = installedAndRequiredBackendModules[uuid].uuid;
+    const installedVersion =
```
alaboso, line 105: I did some cleanup and removed the getInstalledVersion function in favor of find()

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #151, opened 2021-07-21
Fixed the css of the modal rendered in the styleguide

vasharma05:
### What does this PR include?
This PR includes the CSS changes to the modal rendered, which was first taking full width and height and wasn't according to what it should be. The difference is highlighted in the screenshots attached.

Before: 
image

After:
image


FlorianRappl (approved):
LGTM :beers:!

PR #152, opened 2021-07-22
Adjust sidenav width

denniskigen:
This makes it so that on desktop, the side nav is flush with the chart container in the patient chart summary view.

- Before (note how the left margin of the side nav encroaches into the patient chart container):

Screenshot 2021-07-22 at 12 12 14

- After:

Screenshot 2021-07-22 at 12 09 29


FlorianRappl (approved):
LGTM!

PR #154, opened 2021-07-22
Offline Banner

manuelroemer:
Integrates an offline banner into the app shell.


packages/apps/esm-primary-navigation-app/src/components/offline-banner/offline-banner.component.tsx
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import styles from "./offline-banner.component.scss";
+import { ConnectionSignalOff16 } from "@carbon/icons-react";
+import { subscribeConnectivity } from "../../../../../framework/esm-globals/src";
```
FlorianRappl, line 4: `@openmrs/esm-framework`

packages/apps/esm-primary-navigation-app/src/components/offline-banner/offline-banner.component.tsx
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import styles from "./offline-banner.component.scss";
+import { ConnectionSignalOff16 } from "@carbon/icons-react";
```
FlorianRappl, line 3: Import from submodule, e.g., `@carbon/icons-react/es/.../`

packages/apps/esm-primary-navigation-app/src/components/offline-banner/offline-banner.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const [lastUpdated, setLastUpdated] = React.useState(new Date());
+
+  const readAndSetLastUpdated = () => {
+    const value = localStorage.getItem("offline-last-updated");
+    const date = value ? new Date(value) : new Date();
+    setLastUpdated(date);
+  };
+
+  const writeLastUpdated = (date = new Date()) => {
+    localStorage.setItem("offline-last-updated", date.toISOString());
```
FlorianRappl, line 48: We could also put it in the indexDB (and let it be managed by `esm-offline`)? That way it could store when it was online / offline / synced ... such metrics.

FlorianRappl (commented):
LGTM so far - thanks! :beers:

packages/apps/esm-primary-navigation-app/src/components/offline-banner/offline-banner.component.tsx
manuelroemer: Now that you mention it - depending on how you interpret the designs, this current approach may be temporary. What's happening right now is that the date _when you go offline_ is saved. The design kind of suggests the time of the last sync (which would, I guess, have to come from the offline patient list).
If the latter is what we need to go with, the entire local storage approach will probably have to be replaced anyway.

FlorianRappl: Let's leave it as-is for now, but I'd say we should improve it in the future.

FlorianRappl (approved):
:beers:

PR #155, opened 2021-07-23
Include Extension Slot, required for Name, Programme switcher

larslemos:
Screen Shot 2021-07-23 at 03 59 13
Screen Shot 2021-07-23 at 04 04 33
While creating an extension slot to allow customization for a drop-down. The below over-position is found: 

Extension_Slot_v5

Also additional changes to Z-index oft the style guides. 

The hack needed here to override:

Screen Shot 2021-07-22 at 08 46 44

The solution requires the  z-Index on the root.scss to be removed:

Screen Shot 2021-07-22 at 08 46 53

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
@@ -1,5 +1,6 @@
 .topNavActionSlot {
   display: flex;
+  z-index: 9999 !important;
 }
 
 .headerGlobalBar button {
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
@@ -1,6 +1,5 @@
 .topNavActionSlot {
-  display: flex;
-  z-index: 9999 !important;
+  display: flex; 
 }
 
 .headerGlobalBar button {
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -93,7 +93,7 @@ const Navbar: React.FC<NavbarProps> = ({
           </HeaderLink>
           <ExtensionSlot
             className={styles.dividerOverride}
-            extensionSlotName="top-navigation-slot"
+            extensionSlotName="top-nav-info-slot"
           />
           <HeaderGlobalBar className={styles.headerGlobalBar}>
             <ExtensionSlot
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -91,6 +91,10 @@ const Navbar: React.FC<NavbarProps> = ({
           >
             <Logo />
           </HeaderLink>
+          <ExtensionSlot
+            className={styles.dividerOverride}
+            extensionSlotName="top-navigation-slot"
+          />
           <HeaderGlobalBar className={styles.headerGlobalBar}>
             <ExtensionSlot
               extensionSlotName="top-nav-actions-slot"
--- a/packages/apps/esm-primary-navigation-app/src/root.scss
+++ b/packages/apps/esm-primary-navigation-app/src/root.scss
@@ -21,6 +21,3 @@
   height: var(--omrs-topnav-height);
 }
 
-.primaryNavContainer header {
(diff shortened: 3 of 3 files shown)


packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -25,8 +25,8 @@ import Logo from "../logo/logo.component";
 import styles from "./navbar.component.scss";
 import { isDesktop } from "../../utils";
 import { UserSession } from "../../types";
+import { HeaderNavigation } from "carbon-components-react";
```
FlorianRappl, line 28: Can you make it a direct submodule import? It looks something like `carbon-components-react/es/components/...` I guess it may com from `UIShell`, but I don't know.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -71,6 +71,35 @@ const Navbar: React.FC<NavbarProps> = ({
 
   const render = useCallback(() => {
     return (
+      <Header aria-label="OpenMRS">
```
FlorianRappl, line 74: Something in here looks like a merge gone wrong. Can you check again?

.gitignore
```diff
@@ -70,4 +70,8 @@ dist
 .idea/
 
 storybook-static
-package-lock.json
\ No newline at end of file
+package-lock.json
+
+#Monorepos folders
+*.package.json
+*.yarn.lock
```
FlorianRappl, line 77: What are those? If you operate the monorepo correctly there shouldn't be any additional rules necessary.

Make sure to use `npx lerna bootstrap` to resolve dependencies etc.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
larslemos: @FlorianRappl  Working on that, will add an additional commit.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
larslemos: Yes, reviewing the history for the merge.

.gitignore
larslemos: Yes, I run the command. Forget to remove these, wont be necessary.

larslemos:
@FlorianRappl  I've updated the commits to reflect the suggested changes.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -91,6 +91,10 @@ const Navbar: React.FC<NavbarProps> = ({
           >
             <Logo />
           </HeaderLink>
+          <ExtensionSlot
+            className={styles.dividerOverride}
+            extensionSlotName="top-navigation-slot"
```
FlorianRappl, line 96: How about `top-nav-info-slot`?

FlorianRappl (commented):
LGTM - thanks! :beers:

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.scss
```diff
@@ -1,5 +1,6 @@
 .topNavActionSlot {
   display: flex;
+  z-index: 9999 !important;
```
FlorianRappl, line 3: Not sure why that's needed but I think it should be removed.

jonathandick:
This doesn't feel to me like you should be using the same slot for changing the user dashboard as for creating a place to load the patient name when in a patient's chart. @FlorianRappl ?

FlorianRappl:
> This doesn't feel to me like you should be using the same slot for changing the user dashboard as for creating a place to load the patient name when in a patient's chart. 

Not sure what you refer to. This PR only adds an additional slot for flexibility to the top bar. Their purpose is to bring another selection menu in here, but independent of that purpose having another slot there makes sense for flexibility.

larslemos:
> This doesn't feel to me like you should be using the same slot for changing the user dashboard as for creating a place to load the patient name when in a patient's chart. @FlorianRappl ?

@jonathandick  there was no slot previously there. The Slot, either for Programme Switcher or Patient Name, will be override in the esm overriding the slot. It will be mounted.

FlorianRappl (approved):
LGTM!

larslemos:
Hi @FlorianRappl  and @jonathandick  following up on this. 
There was serveral suggestion on how to override the sidenav, and have the Progamme-switcher with an higher Z-index.
None worked within out team. How should we proceed ?

PR #157, opened 2021-07-26
Add note to include the protocol when using `openmrs <cmd> --backend`

brandones:


packages/tooling/openmrs/src/cli.ts
```diff
@@ -40,7 +40,7 @@ yargs.command(
       .describe("host", "The host name or IP for the server to use.")
       .string("backend")
       .default("backend", "https://openmrs-spa.org/")
-      .describe("backend", "The backend to proxy API requests to.")
+      .describe("backend", "The backend to proxy API requests to. Include the protocol.")
```
FlorianRappl, line 43: Rephrase to `The URL for the backend to proxy the API requests to.` - as it is right now I do not see any value. What is protocol here?

brandones: The protocol is the part of the URL which specifies the network protocol, e.g. `http://` or `https://`.

FlorianRappl: I am aware of that. I wrote that question to illustrate that somebody who does not know "how to talk to the backend" (i.e., via URLs) does not know what a protocol is either. By specifying that we refer to the backend via an URL we imply that a protocol has to be given. Making the whole remark obsolete.

FlorianRappl: (btw the default is already a great example of a legit URL; I don't see how somebody could miss specifying the protocol here...)

brandones: Well both Jayasanka and I did so independently, so 🤷. I certainly know what a protocol is, and Jayasanka probably does too.

FlorianRappl: > I certainly know what a protocol is, and Jayasanka probably does too.

That I know. My point was: The remark regarding a protocol is in my opinion obsolete if we explicitly tell the user to use a valid URL. Last time I checked "foo.com" was not a valid URL, but "https://foo.com" was. The scheme is the crucial part here.

PR #158, opened 2021-07-30
Add support for using 'openmrs' with the --sources argument multiple times

brandones:


packages/tooling/openmrs/package.json
```diff
@@ -31,7 +31,7 @@
     "@babel/preset-env": "^7.11.0",
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
-    "@openmrs/esm-app-shell": "^3.1.10",
+    "@openmrs/esm-app-shell": "next",
```
FlorianRappl, line 34: We are int he monorepo here! You need to reference the package with the correct version.

Otherwise `npx lerna bootstrap` won't work...

yarn.lock
```diff
@@ -2655,6 +2655,34 @@
   dependencies:
     "@octokit/openapi-types" "^9.1.0"
 
+"@openmrs/esm-app-shell@next":
```
FlorianRappl, line 2658: You import the app shell form the registry!

Don't do that!

Use it from the monorepo.

FlorianRappl (changes requested):

packages/tooling/openmrs/package.json
brandones: Yeah, for some reason the app shell I was getting contained an old version of `esm-framework` (or for whatever reason, certain newer functions in it weren't available), so I was just messing with this. It didn't solve the issue.

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #160, opened 2021-08-05
Make Caching URLs More Resilient in the SW

manuelroemer:
I ran into this problem right now while working with patient-registration. The SW tried to precache a lot of files (as it should), but some of them failed (for example there was a file of esm-form-entry that always returned 404 - Not Found).

The issue here is that the SW uses `cache.addAll`. As soon as one of these files failed, everything else was also discarded. This means that if a single (potentially unimportant) MF fails, everything else fails too.

I migrated the SW to manually caching each URL using `cache.add` and further improved the debug logs which should make it much easier to troubleshoot *what* cannot be cached (before we'd only get a list of everything - quite useless when there are about 200 URLs in there).


FlorianRappl (approved):
Great!

PR #165, opened 2021-08-12
MF-650 Back button from Login Password page should lead to Login Username Page

maureennduta:
### What does this PR do

This fixes browser navigation in the event a user inputs the wrong username and continues to the password page. They can go back and change the username 

Browser Button

### DIFF
--- a/packages/apps/esm-login-app/src/login/login.component.tsx
+++ b/packages/apps/esm-login-app/src/login/login.component.tsx
@@ -39,10 +39,17 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
   const [t] = useTranslation();
 
   useEffect(() => {
+
     if (user) {
       history.push("/login/location", location ? location.state : undefined);
     }
+    if (!location.hash) {
+      setShowPassword(false);
+      setUsername("");
+    }
+
   }, [user, history, location]);
+
 
   useEffect(() => {
     const field = showPassword
@@ -58,6 +65,7 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
     const field = usernameInputRef.current;
 
     if (field.value.length > 0) {
+      history.push("/login/#user");
       setShowPassword(true);
     } else {
       field.focus();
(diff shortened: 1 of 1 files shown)


packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -58,6 +65,7 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
     const field = usernameInputRef.current;
 
     if (field.value.length > 0) {
+      history.push("/login/#user");
```
FlorianRappl, line 68: I disagree. We use browser (history) routing.

Either make it a dedicated page and properly push or don't have this functionality. I also don't see the benefit of having the back button here working like that. It's either two pages (i.e., URLs) or not.

FlorianRappl (changes requested):
Should be sound. Right now its not.

packages/apps/esm-login-app/src/login/login.component.tsx
manuelroemer, line 68: If we do go with the hash here I'd prefer `/login#password` - it reflects the UI better (`#password` is present when the password field is shown).

But tbh, it feels wrong to me to (ab)use a hash here and I think that this could be solved differently. @FlorianRappl What do you think about this?

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -39,11 +39,18 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
   const [t] = useTranslation();
 
   useEffect(() => {
+
     if (user) {
       history.push("/login/location", location ? location.state : undefined);
     }
+    if (!location.hash) {
+      setShowPassword(false);
+      setUsername("");
```
manuelroemer, line 48: It may make sense to _not_ reset the username here. In my opinion this would improve some flows, e.g. for long usernames where only one character is wrongly typed. -> The user wouldn't have to reenter the entire name, but could just fix the one character which is wrong.

packages/apps/esm-login-app/src/login/login.component.tsx
manuelroemer: Oh nevermind, Florian was faster 😆
Please ignore my comments then.

PR #166, opened 2021-08-12
Switch from babel-eslint to @babel/eslint-parser

denniskigen:
The ESLint Babel parser we're currently using has been [deprecated](https://www.npmjs.com/package/babel-eslint) in favour of `@babel/eslint-parser`. This PR installs `@babel/eslint-parser` and configures ESLint to use it as the default parser.


brandones (approved):
Thanks @denniskigen !

PR #168, opened 2021-08-13
MF-633: Align navbar menu item button interactions

denniskigen:
This PR aligns the interaction behaviour for menu items in the navbar as suggested in Ciaran's [QA design review doc](https://www.notion.so/Top-Nav-6d97f575ec0d46e0a1fd199b7df6b079). More specifically, we want that when menu item buttons are first clicked (to open a menu panel or similar), the menu item button icon gets replaced by an 'X'. Later, when the menu item button gets clicked (to close a menu panel or similar), the 'X' gets replaced with the original button icon.

Open state:
Screenshot 2021-08-13 at 11 49 31

Closed state:
Screenshot 2021-08-13 at 11 49 20

Using `isOpen` from the implementer tools store offers a nice advantage - being able to change the navbar icon state by clicking the close button directly from the implementer tools panel.

Screenshot 2021-08-13 at 11 49 51

### DIFF
--- a/packages/apps/esm-implementer-tools-app/src/implementer-tools.button.tsx
+++ b/packages/apps/esm-implementer-tools-app/src/implementer-tools.button.tsx
@@ -1,24 +1,30 @@
 import React from "react";
-import Tools20 from "@carbon/icons-react/es/tools/20";
-import styles from "./implementer-tools.styles.css";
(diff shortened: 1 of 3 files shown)


manuelroemer (approved):
LGTM, thanks!

PR #169, opened 2021-08-19
add ability to close app menu and use name when user clicks outside

KorirC:
### Description 
added ability to close app menu and user menu when user clicks outside.


donaldkibet:
I will close this PR since a better solution to solve the issue has been done by @FlorianRappl as seen [here](https://github.com/openmrs/openmrs-esm-core/commit/e84e2dd96b6cf4158845f0dce54e1b1d888e414e) we need to use it to achieve the correct actions

PR #172, opened 2021-08-24
Format in pre-push

brandones:


donaldkibet (approved):
LGTM 🚀

manuelroemer (approved):
LGTM!

.husky/pre-push
```diff
@@ -3,7 +3,7 @@
 
 set -e  # die on error
 
-npx pretty-quick --staged
+yarn prettier
```
FlorianRappl, line 6: Sorry this PR does not make much sense (@donaldkibet @manuelroemer). If the push occurs and a formatting issue would be found, then we cannot detect it as prettier is run before lint. However, the changes from prettier are not committed, so even though linting says OK the branch / PR will be invalid and not pass lint...

So overall not a good change. I'd remove this line (and `npx lerna run document`) from this script.
@brandones

brandones: Good catch. Fixed in https://github.com/openmrs/openmrs-esm-core/pull/174

PR #173, opened 2021-08-24
Updated SW API to Function Without SW (MF-538) | Centralized Offline Patient Sync API | Scaffolded esm-offline-tools-app

manuelroemer:
## Requirements

## Summary

**⚠ Important:** This PR only contains parts of the implementation. Another part is provided via extensions in esm-patient-list-app. There will be another PR in that repository too.

This PR does three things:

### Updated SW API to Function Without SW

The functions interacting with the SW should now work without the SW being present and report appropriate errors when it's missing.

### Centralized Offline Patient Sync API
Since syncing patient data is central to the entire app, it makes sense to move it from esm-patient-list-app into esm-offline. esm-patient-list-app still manages *how* and *when* patients will be available offline (i.e. when they are on at least one offline list), but the *actual synchronization* is now handled by esm-offline.

Here we also provide a global store with the current synchronization state. This enables MFs (mainly offline-tools-app) to display the current synchronization state of a given patient.

### esm-offline-tools

**Disclaimer: Offline tools is obviously not nearly finished yet and requires follow-up work in the future. The current state is a good one to integrate though as it already contains important functionality for managing offline patients.** 

The PR adds the esm-offline-tools-app MF which already contains several components that allow management of offline patients. Currently included:

* Home page with offline patients overview table
* Patient sync info (downloading, errors, last synced at)
* Filtering of offline patient table
* Info about which offline patient data handlers failed to synchronize

## Screenshots

Webaufnahme_24-8-2021_183317_localhost
Webaufnahme_24-8-2021_183331_localhost
Webaufnahme_24-8-2021_183424_localhost
Webaufnahme_24-8-2021_183433_localhost
Webaufnahme_24-8-2021_18351_localhost

## Related Issue
https://issues.openmrs.org/browse/MF-538

## Other

@FlorianRappl I know that this is a large one and that reviewing is going to be hard/annoying with that many files. It would be great if you could still focus on the following things since it can very well be that I messed something up there:

* The esm-offline-tools-app project and its integration into esm-app-shell
* The used extensions in offline tools (there are two `ExtensionSlot`s used in the MF)
* The public API of the patient data synchronization module in esm-offline


FlorianRappl (approved):
LGTM - awesome work :beers:!

PR #174, opened 2021-08-25
Improve pre-commit and pre-push hooks

brandones:
This adds a pre-commit hook which formats and documents. It formats only staged files, and documents only packages which have been changed.

It also makes the pre-push checks only run for packages that have had changes since `master`.

Fixes https://issues.openmrs.org/browse/MF-742


FlorianRappl (approved):
That's great - thanks! :beers:

PR #176, opened 2021-08-31
bug fix failing webpack-dev-server on running `yarn run:shell` command

donaldkibet:
### What does this PR do? 

At the moment when you try running 
```sh 
  yarn run:shell or npm run run:shell
```

you get the error below as a result of using the beta version of 
```sh
"webpack-dev-server": "4.0.0-beta.3"
```

Screenshot 2021-08-31 at 08 53 48


FlorianRappl (approved):
Cool!

PR #177, opened 2021-08-31
Add ability to pass props to extensions on using `attach`

donaldkibet:
### What does this PR do?

1. Add ability to pass props to attached extensions using the `attach` function

#### Intended usage

- Adding a prop

```sh

  attach(extensionSlotName, extensionID, props)

```

- Retrieving the props

```sh

const { slots } = useExtensionStore();
const { props } = slots[extensionSloName]

```

This is to enable this [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/322#discussion_r691305749) to move ahead

### DIFF
--- a/packages/framework/esm-extensions/src/extensions.ts
+++ b/packages/framework/esm-extensions/src/extensions.ts
@@ -68,7 +68,11 @@ export const registerExtension: (
   }
 );
 
-export function attach(extensionSlotName: string, extensionId: string) {
+export function attach(
+  extensionSlotName: string,
+  extensionId: string,
+  props?: Record<string, any>
+) {
   updateExtensionStore((state) => {
     const existingSlot = state.slots[extensionSlotName];
 
@@ -80,6 +84,7 @@ export function attach(extensionSlotName: string, extensionId: string) {
           [extensionSlotName]: {
             ...createNewExtensionSlotInfo(extensionSlotName),
             attachedIds: [extensionId],
+            props,
           },
         },
       };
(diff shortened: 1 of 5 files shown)


packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -68,7 +68,11 @@ export const registerExtension: (
   }
 );
 
-export function attach(extensionSlotName: string, extensionId: string) {
+export function attach(
+  extensionSlotName: string,
+  extensionId: string,
+  props?: object
```
FlorianRappl, line 74: Not `object`. Rather `Record<string, any>`.

FlorianRappl (changes requested):
Hm I am confused here.

Props are entering through the *slot* which wants to display extension components. The *initial* props are defined with the slot - not when `attach` is called.

Right now I also do not see these props used which makes me more confused.

FlorianRappl:
I think the referenced PR comment was misunderstood.l The props need to be supplied by the workspace. Not by the one calling `attach`. There, you could always capture variables - you don't need this mechanism.

donaldkibet:
> I think the referenced PR comment was misunderstood.l The props need to be supplied by the workspace. Not by the one calling `attach`. There, you could always capture variables - you don't need this mechanism.

That is my intention to extract the props on [context-workspace](https://github.com/openmrs/openmrs-esm-patient-chart/blob/702f27b85d/packages/esm-patient-chart-app/src/workspace/context-workspace.component.tsx) , The props are then supplied by the workspace.

PR #179, opened 2021-09-01
MF-762: Introduce a way to cache a single specific HTTP request without using URL patterns

manuelroemer:
## Requirements

## Summary

Adds the new `x-omrs-offline-caching-strategy` header which allows users to tell the SW that the network-first caching strategy should be used for specific request (leading to that request being cached).
In that context, refactored the existing header handling and added support for specifying headers with the `fhir.read` method typings.

## Screenshots

*None.*

## Related Issue
https://issues.openmrs.org/browse/MF-762

## Other

*None.*


FlorianRappl (approved):
LGTM!

PR #184, opened 2021-09-14
MF-779: Harmonise RefApp colour palette

denniskigen:
## Requirements

## Summary

This PR harmonises the esm-styleguide colour variable references with those specified in the [Official style guide colour palette](https://app.zeplin.io/styleguide/60d5ecb9efdcd81256117e7d/colors) following a review process with Ciaran.

Also included in this PR are app-wide refactors to accommodate the update colour variable names.


packages/framework/esm-styleguide/src/_vars.scss
```diff
@@ -1,31 +1,52 @@
...
-$field-01: #f4f4f4;
-$grey-2: #e0e0e0;
-$labeldropdown: #c6c6c6;
+$open-mrs-background-gray: #ededed;
+$state-hover-selected-ui: #cacaca;
+$support-02: #24a148;
+$teal-bg-10: #f2f5f5;
+$teal-bg-20: #dae0e0;
+$ui-01: #f4f4f4;
+$ui-04: #8d8d8d;
```
FlorianRappl, line 47: Where is `$ui-02` etc.? Seems odd to miss 2 here. Also why are the usage-named colors not referring to a color named variable? Makes not much sense to have that one unique and not colorized.

Ideally, all color variable names would be context / usage named. But its fine for me if we give them color-like names and never change them (or only change them in slight variation, still adhering to their name).

But then the context named variables should refer to the colors to indicate the binding. After all, the color named variables make the palette, and the context named variables make the theme.

denniskigen: Sure, I'll field that with Ciaran and revert.

denniskigen:
@FlorianRappl I've updated the variables after discussions with Ciaran.

packages/framework/esm-styleguide/src/_vars.scss
```diff
@@ -1,31 +1,53 @@
...
+$color-teal-60: #007d79;
+$color-teal-70: #005d5d;
+$color-teal-80: #004144;
+$color-teal-90: #022b30;
+$color-teal-bg-10: #f2f5f5;
+$color-teal-bg-20: #dae0e0;
+$color-white: #ffffff;
+$color-yellow-10: #fff8e1;
+$color-yellow-30: #feecae;
+$color-yellow-100: #f1c21b;
```
FlorianRappl, line 48: In general this is fine.

Just two things:

1. Updating the framework and rebuilding will be super difficult. Right now there is no mapping / replacement for the used variable names. How is this going to be communicated?
2. Are we sure that dropping context aware names is the right thing to do? The color names are good - and should stay, but having the theme outlined here is in my opinion necessary. Otherwise, we will have some issues when changing "colors" or replacing a color later. See (1) for the current issue.

brandones:
@denniskigen Do you have the tools and information you need to move this forward?

denniskigen:
@brandones not quite with this one. I think further consultations are warranted here.

brandones:
Cool. I don't fully understand what the project here is, so if you can get the advice you need from @FlorianRappl that would be ideal. But if he's off or otherwise unavailable, we could have a quick call or something for you to get me up to speed.

brandones:
@denniskigen What should we do about this one?

denniskigen:
I'm not sure how to advance this. I'll close for now.

PR #185, opened 2021-09-16
Fix eslint error in implementer-tools introduced by 83753e15b1bfd8c7fd2d5ed9df6b8c9894f5ec98

brandones:
@FlorianRappl left ESLint errors instead of cleaning up his work in implementer-tools. The build has been broken on this account. He might be encouraged to use PRs in the future, which make trivial issues like this apparent.

EDIT: Also fixed some typescript errors in esm-primary-navigation-app.


FlorianRappl (approved):
Great! 🍻

PR #186, opened 2021-09-16
MF-784 Fix '`document` is not defined' at `svg-utils.ts:9`

brandones:


FlorianRappl (commented):
@brandones this is not a fix and it won't work (have you ever reproduced the issue and tried it with the proposed fix?).

If `document` is really not defined in line 9 it would also not be defined in line 16. Why don't you fix the root cause instead?

The root cause is most likely that the file is used in a different context that has no access to `document`, such as a web worker. So finding out how that can be and improving this is the way to fix MF-784.

brandones:
Oh yeah, fair. Closing for now. Might take a more serious crack at it if I have time later.

PR #187, opened 2021-09-17
Initial Version of Offline Actions

manuelroemer:
## Requirements

## Summary

The initial version of the offline actions. When comparing with the designs, there are still several differences to be done, but this is a good intermediary state.

## Related Issue

*None.*

## Other

*None.*


FlorianRappl (approved):
LGTM - thanks! :beers:

brandones:
Hey @manuelroemer , what's the status of this ticket?

manuelroemer:
Oh right, this is still open, I forgot about it. I'll have to fix the build and then we can merge from my side.

packages/apps/esm-implementer-tools-app/src/configuration/editable-value.component.tsx
```diff
@@ -136,7 +136,7 @@ export default function EditableValue({
                 hasIconOnly
                 onClick={() => {
                   temporaryConfigStore.setState(
-                    unset(temporaryConfigStore.getState(), ["config", ...path])
+                    unset(temporaryConfigStore.getState(), ["config", ...path]) as any
```
manuelroemer: @brandones I hacked in the `as any` here to move the PR forward. By added the `lodash-es` types, TS started to complain here that `unset` returns a `boolean` which is not assignable to `Config`. Not sure if that indicates a bug or if that's the way it should be. Maybe this needs some extra attention though and I think that you know most about the config API, hence the tag.

brandones: Oh shoot, thanks for the heads up!

PR #188, opened 2021-09-23
Smarter config mocks

brandones:
## Summary

This makes it so that by default, test authors don't have to override the `useConfig` implementation. It also ensures that the tests stay up to date with changes in the config schema.


denniskigen (approved):
LGTM, thanks @brandones!

PR #189, opened 2021-09-24
Change everything to refer to 'frontend modules' instead of 'microfrontends'

brandones:
@gracepotma and @jonathandick have made the decision to change our naming from "microfrontends" to "frontend modules." This is a pretty big shift.

The only meaningful code change here is that the `microfrontends` parameter of the frontend build config is now called `frontendModules`. Heads up @rbuisson , the 3.x distro code will have to change.

We'll also need to update https://wiki.openmrs.org/display/projects/Frontend+3.0+Documentation+for+Implementers

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ All libraries are aggregated in the `@openmrs/esm-framework` package:
 
 - [@openmrs/esm-framework](packages/framework/esm-framework)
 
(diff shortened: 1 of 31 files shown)


manuelroemer (approved):
Looks good from my side!

PR #191, opened 2021-09-29
Make 'uuid' a dependency, not devDependency, of esm-offline

brandones:
Compiling a microfrontend with typescript I got

```
> @pih/esm-referrals-queue-app@1.0.5 typescript
> tsc

node_modules/@openmrs/esm-offline/src/uuid.ts:1:20 - error TS2307: Cannot find module 'uuid' or its corresponding type declarations.

1 import { v4 } from "uuid";
                     ~~~~~~
```

I'm hoping this solves it?


manuelroemer (approved):
Whoops, I should have made that a dependency in the first place. Thanks for updating this!

PR #194, opened 2021-10-04
Add support for automatically downloading the core apps

ibacher:
## Requirements

## Summary

Much more detail [on Slack](https://openmrs.slack.com/archives/GQJA12F4L/p1633101215112700).

This adds a new flag to the `build` command of the `openmrs` tool to enable automatically downloading and extracting the "core" apps of the appshell (those listed as `devDependencies` of the appshell).

Rationale: to maintain a dev server with the current [importmap overrides](https://spa-modules.nyc3.digitaloceanspaces.com/import-map.json), we need to bundle the core applications into the appshell, since they aren't included in that importmap. When building from the monorepo, this is done automatically, by using a relative path. However, when building an OpenMRS distribution using `npx openmrs build`, we don't have the entire monorepo checked out. Thus we need some way to identify the apps from the monorepo and ensure they are in a location that Webpack can find when performing the build.

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/tooling/openmrs/src/cli.ts
+++ b/packages/tooling/openmrs/src/cli.ts
@@ -191,6 +191,9 @@ yargs.command(
         "target",
         "The target directory where the build artifacts will be stored."
       )
+      .string("registry")
+      .default("registry", "https://registry.npmjs.org/")
+      .describe("registry", "The NPM registry used for getting the packages.")
       .boolean("fresh")
       .describe(
         "fresh",
@@ -230,6 +233,12 @@ yargs.command(
       .describe(
         "importmap",
         "The import map to use. Can be a path to an import map to be taken literally, an URL, or a fixed JSON object."
+      )
+      .boolean("download-coreapps")
+      .default("download-coreapps", false)
+      .describe(
+        "download-coreapps",
+        "Allows automatically downloading the core app shell apps if they are not currently avaialble."
       ),
   async (args) =>
     runCommand("runBuild", {
@@ -238,6 +247,7 @@ yargs.command(
       configUrls: args["config-url"],
       pageTitle: args["page-title"],
       supportOffline: args["support-offline"],
+      downloadCoreapps: args["download-coreapps"],
       ...args,
       importmap: args.importmap,
       buildConfig:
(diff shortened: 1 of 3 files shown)


packages/tooling/openmrs/src/cli.ts
```diff
@@ -230,6 +233,12 @@ yargs.command(
       .describe(
         "importmap",
         "The import map to use. Can be a path to an import map to be taken literally, an URL, or a fixed JSON object."
+      )
+      .boolean("download-coreapps")
+      .default("download-coreapps", false)
+      .describe(
+        "download-coreapps",
+        "Allows automatically downloading the core app shell apps if they are not currently avaialble."
```
brandones, line 241:
```suggestion
        "Downloads and bundles the core apps. For cases where the core apps are not in the import map."
```

Do I understand this correctly?

ibacher: Yeah, that reads much better.

packages/tooling/openmrs/src/commands/build.ts
```diff
@@ -33,9 +46,100 @@ function loadBuildConfig(buildConfigPath?: string): BuildConfig {
...
+    const devDependencies: Record<string, string> =
+      require("@openmrs/esm-app-shell/package.json").devDependencies;
+
+    const apps = Object.keys(devDependencies)
+      .filter((dep) => dep.endsWith("-app"))
+      .map((dep) => [dep, devDependencies[dep]]);
+
+    if (apps.length > 0) {
+      const cacheDir = resolve(process.cwd(), ".cache");
+      !existsSync(cacheDir) && mkdirSync(cacheDir, { recursive: true });
```
brandones, line 109:
```suggestion
      if (!existsSync(cacheDir)) {
        mkdirSync(cacheDir, { recursive: true });
      }
```

Preferable for readability. Short-circuiting is a side-effect, not the purpose of the `&&` operator.

brandones (approved):
Really excellent. Thanks @ibacher .

I'm still not sure why these don't just go in the import map, but if we have to have bundled "core apps," this is a good thing to have for it.

PR #197, opened 2021-10-06
App shell: Webpack shouldn't call InjectManifest in dev mode

brandones:
## Summary

Changes the esm-app-shell webpack config to only call InjectManifest in prod mode. @manuelroemer should confirm that this is the right thing to do.

## Screenshots

When running `yarn run:shell`:

Before:
Screenshot from 2021-10-06 13-22-42

After, there is no such error.


denniskigen (approved):
Having to kill and restart the app shell after rebuilding your app in order to see your changes propagated in dev mode is such a frustrating experience. Thanks for fixing this!

PR #198, opened 2021-10-07
Disable link check for _sidebar.md, which only works in docsify

brandones:


denniskigen (approved):
LGTM @brandones. Does this fix the failing check?

PR #199, opened 2021-10-09
MF-855: Versioning core should create a release version

brandones:
We need to be able to create proper release versions (`X.Y.Z`). The current release behavior is to create an `alpha.0` version. This is a problem because all our packages refer to each other with peerDependency version `3.x`, and prerelease versions do not satisfy `3.x`.

Use 'lerna version --conventional-graduate' to release versions with CI.

https://issues.openmrs.org/browse/MF-855


prapakaransp:
Sounds like a good plan for maintenace.

PR #205, opened 2021-10-14
MF-865  Page "order" attribute not working on Chrome

brandones:
## Summary

Fix invalid comparison function used with `sort`. See https://stackoverflow.com/a/1969200/1464495

Fixes error introduced in https://github.com/openmrs/openmrs-esm-core/pull/196

## Screenshots

Screenshot from 2021-10-14 14-04-06

## Related Issue

https://issues.openmrs.org/browse/MF-865


ibacher:
LGTM. Are we still on the two approval system?

brandones:
I wasn't aware we ever were!

PR #206, opened 2021-10-15
Tests for the configurability of the extension system

brandones:
Create tests for the configurability of the extension system. These live in esm-framework, because they depend on esm-config, esm-extensions, and esm-react-utils. This also includes some minor refactoring.

The potentially contentious thing introduced here is `debug`. Have a look at the [docs](https://github.com/visionmedia/debug#readme). You enable it in node (for tests) with an env variable `DEBUG=openmrs:xyz`, or in the browser with `localStorage.debug = 'openmrs:xyz'`.

I imagine integrating this into esm-framework to make it easier to debug state changes, config loading, extension attachments, etc. Probably we'd want to codemod it out of production builds. We'd also need to be quite careful that it isn't misused or overused, or it could become really annoying to anyone who wants to use it (remember how LoggingAdvice in OpenMRS Core logged every function entrance and exit at level INFO for like 5 years and everyone hated it?). Your thoughts and suggestions are welcome.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "css-loader": "^5.2.4",
     "css-minimizer-webpack-plugin": "^1.2.0",
     "cssnano": "^4.1.10",
-    "debug": "^4.3.2",
     "docsify": "^4.12.1",
     "ejs": "^2.6.2",
     "eslint": "^7.10.0",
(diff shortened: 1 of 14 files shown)


brandones:
Ping @manuelroemer @denniskigen @jonathandick

PR #207, opened 2021-10-19
MF-863: Make the Offline Tools MF dynamically composable via extensions

manuelroemer:
## Requirements

## Summary

Modifies the offline tools module such that other modules can dynamically extend it with pages using the extension meta API. The naming/concept is ~stolen~ taken from the patient chart.

## Screenshots

*None.*

## Related Issue

https://issues.openmrs.org/browse/MF-863

## Other

The linked issue will still require some work (at the moment the migration of the offline tool's dashboard where the different tiles are rendered - these should also be dynamically composable via extensions).

### DIFF
--- a/packages/apps/esm-offline-tools-app/src/root.component.tsx
+++ b/packages/apps/esm-offline-tools-app/src/root.component.tsx
@@ -17,25 +17,6 @@ const Root: React.FC = () => {
       <div className={`omrs-main-content ${styles.mainContentContainer}`}>
         <Route path="/offline-tools" exact component={Home} />
         <Route path="/offline-tools/:page" component={OfflineToolsPage} />
-        {/* <Route
-          exact
-          path="/offline-tools/patients"
-          component={OfflinePatients}
-        />
-        <Route
-          exact
-          path="/offline-tools/patients/:patientUuid/offline-data"
-          component={OfflinePatientSyncDetails}
-        />
-        <Route
-          exact
-          path="/offline-tools/actions"
-          render={() => (
-            <OfflineActions
-              canSynchronizeOfflineActions={canSynchronizeOfflineActions}
-            />
-          )}
-        /> */}
       </div>
     </BrowserRouter>
   );


packages/apps/esm-offline-tools-app/src/offline-tools-page/offline-tools-page.component.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <>
+        <ExtensionSlot extensionSlotName="breadcrumbs-slot" />
+        <ExtensionSlot
+          key={pageConfig.slot}
+          extensionSlotName={pageConfig.slot}
+          state={{ basePath }}
+        />
+      </>
+    );
+  };
```
brandones, line 36: Any reason not to just do

```suggestion
    const page = match.params.page;

    return (
      <>
        <ExtensionSlot extensionSlotName="breadcrumbs-slot" />
        <ExtensionSlot
          key={page}
          extensionSlotName={`offline-tools-page-${page}-slot`}
          state={{ basePath }}
        />
      </>
    );
  };
```

? And then get rid of all the `meta` objects? This would simplify the extension API of this module substantially (in addition to the code being simpler and easier to understand).

manuelroemer: Good observation actually, that would be possible as long as we only require a `page/name` meta from the extensions.

My arguments for using the meta approach are the following:
* I could see more info being provided by the extensions at some point (this will depend on the designs), e.g. a meta value to control the presence/absence of the breadcrumbs, a fixed page title (like in the patient chart, ...). While it's not here right now, it would be very simple to add when the meta approach is already used.
* Consistency with the patient chart. And both "areas" have the same problem to be solved: Providing pages and page links in the sidebar via extensions from foreign modules. I think that being consistent here is good idea for the code base's health ("If you know one, you know the other.").
* Regarding "the code being simpler and easier to understand":
  I think that either approach of how to register/setup the extensions will have to be documented somewhere. I believe that both are a little opaque in their own way. The meta one because it's "heavy", i.e. because you have to setup multiple extensions the right way and your suggestion because it's "magic", i.e. you need to know that the URL segment maps to the extension slot name. In either case a new user would have to scan and understand the code in order to extend it (without docs, that is).

Though I do have to say, maybe this is indeed prematurely optimized. I think in the end I do lean towards using the meta, but only slightly and am fine with using either suggestion. Thoughts are welcome!

PR #209, opened 2021-10-22
Fixup for  Make 'pre' versions come before the corresponding release version #208 

brandones:


brandones:
Merging because @ibacher already more or less reviewed, and this needs fixing.

PR #210, opened 2021-10-26
MF-894: Changed hex to variables in styleguide.

vasharma05:
## Requirements

## Summary

Changed hex of colors to variables in `_overrides.scss `in style guide.

## Screenshots

*None.*

## Related Issue

https://issues.openmrs.org/browse/MF-894

## Other

*None.*


denniskigen (approved):
I like the change from colour hex codes to variables. However, I'm not too keen on the use of the carbon spacing scale tokens. I find the extra overhead of having to look up the spacing scale to find the corresponding rem or px value annoying. Just my 2 cents.

brandones:
Thanks @denniskigen ! Yeah I wonder if we should try to come to consensus about the use of spacing variables. I agree that they're of limited value. I wonder if their value will become more evident when we upgrade Carbon?

PR #217, opened 2021-11-06
Allow building a config file into the distro

brandones:
This adds an optional argument to `openmrs` which allows providing a config file which will be built into the distro. This will simplify the build pipeline for a PIH distro, and probably many future distros.


manuelroemer (approved):
Looks good to me!

PR #219, opened 2021-11-17
O3-909

FlorianRappl:
## Requirements

## Summary

This is based on the [sample repository I created on GitHub](https://github.com/FlorianRappl/module-federation-with-systemjs-importmap). It implements using Webpack Module Federation for sharing dependencies.

## Related Issue

[O3-909](https://issues.openmrs.org/browse/O3-909)

## Other

We also should adopt this change to the existing modules. Right now the solution is backwards-compatible, but ideally I'd like to throw out that layer.


manuelroemer (approved):
Looks awesome! 🚀 Also, +1 on eventually removing the backwards compatibility layer if this is integrated.

brandones (approved):
Looks great. Thank you @FlorianRappl !

brandones:
What will we need to change in order to remove the backwards-compatibility layer?

FlorianRappl:
> What will we need to change in order to remove the backwards compatibility layer?

Well, not much - the MFs just need to be rebuild / republished using the updated tooling.

Old way:

- using `external`
- rely on System.import for resolving shared dependencies
- root module exports the same things as the src/index.ts (e.g., setupOpenMRS function)

New way:

- using `shared` of the Webpack Module Federation (WMF) plugin
- rely on the container / shared chunks from WMF
- root module is a container having `init` and `get` function - `get` is required to obtain the shared "app" module (exports src/index.ts)

Right now the loading takes care of these differences, but ideally only the new way is considered.

PR #220, opened 2021-11-17
Enhanced the SW route registration table with a strategy column and reworked the routing to be strategy driven.

manuelroemer:
## Requirements

## Summary

As a temporary step towards fixing unwanted cache invalidation, the SW's internal route table was updated with a new `strategy` column. This allows the SW to determine if a matching request should be actively (re)cached or simply be kept during the invalidation.

## Screenshots

*None.*

## Related Issue

Addresses/Fixes the underlying issue described by https://issues.openmrs.org/browse/O3-915, but does not implement the proposed API.

## Other

While this should fix the cache invalidation problem for now, I still believe that a better approach would be to generally move the cache handling out of the SW.

### DIFF
--- a/packages/framework/esm-offline/src/service-worker-http-headers.ts
+++ b/packages/framework/esm-offline/src/service-worker-http-headers.ts
@@ -5,6 +5,8 @@ export const omrsOfflineResponseStatusHttpHeaderName =
 export const omrsOfflineCachingStrategyHttpHeaderName =
   "x-omrs-offline-caching-strategy";
 
+export type OmrsOfflineCachingStrategy = "default" | "network-first";
+
 /**
  * Defines the keys of the custom headers which can be appended to an HTTP request.
  * HTTP requests with these headers are handled in a special way by the SPA's service worker.
@@ -30,7 +32,7 @@ export type OmrsOfflineHttpHeaders = {
    *   The service worker decides the strategy to be used.
    * * `network-first`: The service worker will make the request and cache its response.
    */
-  [omrsOfflineCachingStrategyHttpHeaderName]?: "default" | "network-first";
+  [omrsOfflineCachingStrategyHttpHeaderName]?: OmrsOfflineCachingStrategy;
 };
 
 export type OmrsOfflineHttpHeaderNames = keyof OmrsOfflineHttpHeaders;
--- a/packages/framework/esm-offline/src/service-worker-http-headers.ts
+++ b/packages/framework/esm-offline/src/service-worker-http-headers.ts
@@ -5,7 +5,17 @@ export const omrsOfflineResponseStatusHttpHeaderName =
 export const omrsOfflineCachingStrategyHttpHeaderName =
   "x-omrs-offline-caching-strategy";
 
-export type OmrsOfflineCachingStrategy = "default" | "network-first";
+/**
+ *
+ *
+ * * `cache-or-network`: The default strategy, equal to the absence of this header.
+ *   The SW attempts to resolve the request via the network, but falls back to the cache if required.
+ *   The service worker decides the strategy to be used.
+ * * `network-first`: See https://developers.google.com/web/tools/workbox/modules/workbox-strategies#network_first_network_falling_back_to_cache.
+ */
+export type OmrsOfflineCachingStrategy =
+  | "network-only-or-cache-only"
+  | "network-first";
 
 /**
  * Defines the keys of the custom headers which can be appended to an HTTP request.
@@ -26,11 +36,6 @@ export type OmrsOfflineHttpHeaders = {
   [omrsOfflineResponseStatusHttpHeaderName]?: `${number}`;
   /**
    * Instructs the service worker to use a specific caching strategy for this request.
-   * The supported values are:
-   *
-   * * `default`: Equal to the absence of this header/an invalid value.
-   *   The service worker decides the strategy to be used.
-   * * `network-first`: The service worker will make the request and cache its response.
    */
   [omrsOfflineCachingStrategyHttpHeaderName]?: OmrsOfflineCachingStrategy;
 };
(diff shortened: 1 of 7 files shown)


packages/framework/esm-offline/src/service-worker-http-headers.ts
```diff
@@ -5,6 +5,8 @@ export const omrsOfflineResponseStatusHttpHeaderName =
 export const omrsOfflineCachingStrategyHttpHeaderName =
   "x-omrs-offline-caching-strategy";
 
+export type OmrsOfflineCachingStrategy = "default" | "network-first";
```
FlorianRappl, line 8: Is default the same as cache-first? Can we maybe have "default" | "cache-first" | "network-first"? Right now its unclear what default is (and it may actually changed - depending on how default is configured / interpreted).

packages/shell/esm-app-shell/src/service-worker/routing.ts
```diff
@@ -115,3 +85,57 @@ async function handleWithNetworkOnlyAndCacheFallback(
...
 }
+
+const knownHandlers: Record<
+  OmrsOfflineCachingStrategy,
+  (options: RouteHandlerCallbackOptions) => Promise<Response>
+> = {
+  ["network-first"]: (options) => networkFirst.handle(options),
+  ["default"]: async (options: RouteHandlerCallbackOptions) => {
+    try {
+      return await networkOnly.handle(options);
```
FlorianRappl, line 96: Ah so its only fallback to cache if network unavailable!

FlorianRappl (commented):
Yeah looks great - just maybe include a cache-first or so strategy / possibility imho.

packages/shell/esm-app-shell/src/service-worker/routing.ts
manuelroemer: Exactly - I am still thinking about a better/shorter name than "network with cache fallback" (I am considering `network-or-cache ` atm).

The important point is that it's _not_ adding requested files to the cache (as opposed to Workbox's `CacheFirst` strategy which gradually adds all files to the cache which aren't found there - that's something we don't want).

manuelroemer: I am open to rename default though. I was also considering `none` as a replacement to indicate that no real caging strategy should be applied. But again, not sure on the naming.

brandones: What's the difference between the two strategies? It looks like `default` first attempts to request from the network, and then falls back to the cache—which is what I would expect `network-first` to do.

manuelroemer: The difference is that any non-"default" strategy adds items to the cage when they are requested through the network. The custom one doesn't do that.

Network First, for example, makes a NW request and, when successful, _caches the response_. It alternatively falls back to the cache when the NW request fails.

brandones: Oh I see, so the difference between `default` and `network-first` is that `default` only caches the things that we explicitly tell the SW to cache, right?

manuelroemer: > that `default` only caches the things that we explicitly tell the SW to cache

Mhh, not quite. `default` should be seen as the fallback strategy that is used, well, by default. It does two things only:
1. It tries to resolve any request via the network first (essentially the normal behavior that any request goes through).
2. If (1) fails, it looks into the cache and tries to resolve the request there. (This is what happens when the app is in offline mode. While offline, any NW request will fail, but the app continues working because this default strategy uses the cache to resolve requests).

The important point is that the default strategy doesn't actively cache anything. It simply looks into _all_ cached resources (no matter _how_ they ended up in the cache) and checks whether one of these can be used to resolve a request.

Network first, on the other hand, is a strategy that is implemented by workbox. Most of these (with the exception of `CacheOnly` and `NetworkOnly`) _**do** add_ responses resolved via the network to the cache.
It makes sense to use `NetworkFirst` for some requests (e.g. for the `/session` endpoint - we always want to have the latest value in the cache), but not by default (that would very quickly pollute the cache).

brandones: Awesome, thanks for the explanation. I might suggest `network-and-cache-all` for `NetworkFirst` and `network-or-cache` (as you suggested) for this default/fallback mechanism (or `get-without-caching`, or `network-and-cache-none`).

brandones: (though also understandable if you want to keep `network-first` aligned with the Workbox strategy name)

packages/shell/esm-app-shell/src/service-worker/storage.ts
```diff
@@ -1,17 +1,27 @@
...
+   * If it matches, the URL should be cached using the `strategy`.
    */
   pattern: string;
+  /**
+   * The caching strategy to be used for caching matching URLs.
+   * Due to historical reasons, this value might be missing.
+   * In such cases, `network-first` should be assumed (the historical default).
+   *
+   * * `default`: No explicit strategy should be applied. Matching requests should remain in the cache though.
+   * * `network-first`: Use the network-first strategy. Matching requests keep being added to the cache.
```
brandones, line 22:
```suggestion
   * * `default`: Get from the network and fall back to the cache. Do not add anything to the cache.
   * * `network-first`: Use the Workbox NetworkFirst strategy. Add matching requests to the cache.
```

packages/shell/esm-app-shell/src/service-worker/routing.ts
manuelroemer: > (though also understandable if you want to keep network-first aligned with the Workbox strategy name)

Yes, that's something I'd like to do. This makes it easy to link to workbox links and docs.

I updated the code and settled on `network-only-or-cache-only` instead of `default` for now. It's long, but describes the actual strategy being employed the best, imo.

FlorianRappl (approved):
You rock!

PR #221, opened 2021-11-18
Removed obsolete "default" route handlers | Don't clear dynamic routes when app starts

manuelroemer:
## Requirements

## Summary

Removed `default` route handlers still in the code. Shame on me for missing these in the last PR.
Also used that chance to improve logging done by the SW.

Clearing all dynamic route registrations on app start was also removed. This doesn't make any sense anymore when these routes are dynamically registered **_once_** because of certain user actions (e.g. when marking a form as offline). 

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*


FlorianRappl (approved):
LGTM! Thanks

PR #223, opened 2021-11-20
Update to Typescript 4.5

brandones:


manuelroemer:
Looks good generally! Should we also update the TS version in `package.json`?

brandones:
Good call @manuelroemer , updated.

manuelroemer (approved):
LGTM!

PR #224, opened 2021-11-22
Fix typescript to 4.5.x to avoid breakages over minor versions

brandones:
We'll have to be proactive about keeping up to date, but this will make it so there's not a fire every time Typescript releases a new minor version.


denniskigen (approved):
LGTM! Thanks, @brandones.

PR #226, opened 2021-11-23
O3-939: Added extension slot for adding external links in the app-menu

vasharma05:
## Requirements

## Summary

I have added an extension slot in the navigation bar's app menu, to accommodate links that direct to the external links.

## Screenshots

image

## Related Issue

https://issues.openmrs.org/browse/O3-939

## Other

*None.*


vasharma05:
Hey @FlorianRappl ! Can you please have a look in here.
Thanks!

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/app-menu-panel.component.tsx
```diff
@@ -22,6 +22,10 @@ const AppMenuPanel: React.FC<AppMenuProps> = ({ expanded, hidePanel }) => {
         className={styles.menuLink}
         extensionSlotName="app-menu-slot"
       />
+      <ExtensionSlot
```
FlorianRappl, line 25: See comment in the other PR; shouldn't we use these new links instead of extensions?

FlorianRappl (commented):
Should check if this is the way to go.

See https://github.com/openmrs/openmrs-rfc-frontend/pull/30 for details on the link system.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/app-menu-panel.component.tsx
vasharma05: Yes, I'll shift the component here itself.

vasharma05:
Hi @FlorianRappl ! The RFC's Link system you shared, I don't think that is of any use here, since it has to be defined in the `setupOpenMRS function`, but the links are configured in the configuration file and the file is also registered in the same `setupOpenMRS function`. The `defineConfigSchema` sets the `useConfig` context, and I'm using the `useConfig` to get the links.

Correct me if I'm wrong.
Thanks!

FlorianRappl:
Hi @vasharma05 that is not fully correct.

1. You can access the configuration *anywhere*. Does not need to be a React component (=> `useConfig`). So you can access it right after you set up the config schema.
2. Not sure why these links are "configurable". In the RFC the link configuration is implicit. So that could then be just used (and is much better, see (3)).
3. As my previous argument - right now the config is not really ideal to say the least. If I just want to disable a single link I need to configure everything. Not good. It should be possible to add or remove a single link without touching everything.

.husky/pre-push
```diff
@@ -4,5 +4,4 @@
 set -e  # die on error
 
 npx lerna run lint --since master
-npx lerna run test --since master
```
FlorianRappl, line 7: Why did you remove testing?!

FlorianRappl (commented):
Let's bring back the testing then we could merge if this is really needed / required.

Otherwise, I'd recommend improving the config and actually using the link system for this.

brandones:
Hi @vasharma05 and @FlorianRappl , I wouldn't agonize too much over the implementation of this, since it will be totally replaced once we have a Link system (which is not yet implemented, only planned). And definitely don't worry about making it so that a single link can be added or removed without affecting other configured links—the config system is built assuming that config is controlled completely by the implementer. The implementer is the only person who would add or remove links. They don't need to worry about links coming from anywhere else. If that's going to not be the case, we'll need to have much higher-level conversations about config system architecture. Not here in this PR.

vasharma05:
Thanks @brandones ! 
Hey @FlorianRappl , we should move ahead with the PR. Any others suggestions are welcome.
Thanks!

FlorianRappl:
Hi @vasharma05  I still see the testing being removed. Please update the PR. Thanks!

vasharma05:
So sorry @FlorianRappl ! Reverted back the test.
Thanks for your patience!

FlorianRappl (approved):

PR #228, opened 2021-11-25
Fix for O3-941

FlorianRappl:
## Requirements

## Summary

Fixes the issue that the class constructor requires new to be invoked.

The issue was caused by TS transpiling to ES5, which, of course, is not what we want. I've changed the target to ES2015 (ES6). This adds support for classes, such that TS will not shim this and properly use `new` / `super` on it (natively).

## Screenshots

*None.*

## Related Issue

See https://issues.openmrs.org/browse/O3-941 for more details.

## Other

The change would only be required in the `app-shell`, but I find it consistent when the same target is used in all libs (and it should be used in the MFs, too).

Since Node 10 (and therefore Node 12) supports pretty much all the features I've also changed the `openmrs` target to use ES2015. Node 10 was deprecated end of April (EOL April 30th 2021) - hence I already went to suggest Node 12 (next LTS after 10) is the minimum for the tooling.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,7 +3,6 @@
 - [ ] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
 - [ ] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).
 
-
 ## Summary
 
 <!--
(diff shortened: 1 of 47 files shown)


manuelroemer (approved):
LGTM, thanks!

jonathandick:
For those interested, here's the docs on target for ts: https://www.typescriptlang.org/tsconfig#target.

denniskigen:
Sharp work, @FlorianRappl!

PR #229, opened 2021-11-30
Release 3.1.14

brandones:


FlorianRappl (commented):
The last release was 3.1.12 (see, e.g., https://www.npmjs.com/package/openmrs)

This does not make any sense. The workflow (as described) is:

- Make a GitHub release (in this case it would be 3.1.13)
- Bump the version
- The prerelease will be added automatically

Now it seems that someone (maybe you) changed this workflow and that lerna prerelease is used. In my opinion this is a bad idea. It takes away the possibility of *determining* the version of the prerelease. Instead, every prerelease is suddenly a patch, even if it contains a breaking change. Now this PR of yours would cause a 3.1.15-pre... version to be released. Not what you want, right?

I would advocate to revisit the change in the first place. The whole model is broken since https://github.com/openmrs/openmrs-esm-core/pull/208 and https://github.com/openmrs/openmrs-esm-core/pull/209.

FlorianRappl:
(To be clear, there was a misunderstand how NPM works in #208. NPM does not remove tags. So if you put a tag "next" on day 1 and then make changes that are elevated on day 2 to "latest" then "next" still points to the release of day 1. This is the reason for the workflow as described above and why the next person to merge into `master` should handle the version update / change, but until then `next` points to the `latest`, but with a pre-suffix).

brandones:
Right, the system that existed prior to #208  was based on a misunderstanding of SemVer. I'd recommend reviewing [the spec](https://semver.org/#spec-item-11). According to the spec, `1.2.3-pre.0 < 1.2.3`. According to that prior system, `1.2.3-pre.0 > 1.2.3`. This kind of disagreement has real consequences, because NPM and other tools follow the SemVer spec. The main problems we were having with the old system were:
- The `openmrs` tool could only build the last release version of the shell and framework.
- The frontend dependency checking system would flag things incorrectly. If an MF said it required at least core version `3.1.12` and core version `3.1.12-pre.100` is present, the frontend dependency checker would complain. This is because for us prior to #208 , `3.1.12` might already exist and be *before* `3.1.12-pre.100`, and contain the necessary changes, but according to SemVer, `3.1.12` is supposed to *follow* `3.1.12-pre.100`.

Yes, pre-release versions are always a patch version bump; this is a consequence of SemVer, and is familiar to Java developers, for whom `SNAPSHOT` versions have the same semantics.

As discussed on Slack, we need to ensure that pre-release versions of core get deployed to servers that are supposed to run the latest core. This is totally reasonable requirement, and consistent with a correct use of SemVer. Releasing this now is a quick fix, to get core to deploy to openmrs-spa. Given that the conference is happening now, it would be great to get this done.

ibacher:
I'm not sure it's an issue to have 3.1.15-pre out there even if there never is a 3.1.15 release and instead we release 3.2.0. Even if we do things manually, we cannot a priori know "this will be a breaking release" or "this will be a non-breaking release" (at least until we're actually doing the release). With the previous workflow, we were ending up in the situation where 3.1.12-pre was more up-to-date than 3.1.12 (because the version bump wasn't happening), which just doesn't make any sense to me (and makes it harder for MFs to declare which version of core they depend on because, again, by SemVer 3.1.12-pre is _before_ 3.1.12, but a MF might have a dependency on a fix that's been added after 3.1.12). The desireable property here is that the prerelease version is always "more recent" than the last release.

Maybe I'm looking at this from the wrong angle, but that's how it seems to me.

We (theoretically) had a 3.1.13 release, though the GitHub release itself wasn't created, so NPM wasn't updated, but the underlying issue is that we have code depending on 3.1.14-pre..., so if we do a new release, it should be 3.1.14. Releeasing 3.1.13 is trivially fixable, but this should be 3.1.14.

brandones:
Right, the system that existed prior to #208  was based on a misunderstanding of SemVer. I'd recommend reviewing [the spec](https://semver.org/#spec-item-11). According to the spec, `1.2.3-pre.0 < 1.2.3`. According to that prior system, `1.2.3-pre.0 > 1.2.3`. This kind of disagreement has real consequences, because NPM and other tools follow the SemVer spec. The main problems we were having with the old system were:
- The `openmrs` tool could only build the last release version of the shell and framework.
- The frontend dependency checking system would flag things incorrectly. If an MF said it required at least core version `3.1.12` and core version `3.1.12-pre.100` is present, the frontend dependency checker would complain. This is because for us prior to #208 , `3.1.12` might already exist and be *before* `3.1.12-pre.100`, and contain the necessary changes, but according to SemVer, `3.1.12` is supposed to *follow* `3.1.12-pre.100`.

Yes, pre-release versions are always a patch version bump; this is a consequence of SemVer, and is familiar to Java developers, for whom `SNAPSHOT` versions have the same semantics.

As discussed on Slack, we need to ensure that pre-release versions of core get deployed to servers that are supposed to run the latest core. This is totally reasonable requirement, and consistent with a correct use of SemVer. Releasing this now is a quick fix, to get core to deploy to openmrs-spa. Given that the conference is happening now, it would be great to get this done.

ibacher (commented):
Approving to ge things moving.

ibacher (approved):

denniskigen (approved):

FlorianRappl:
> Right, the system that existed prior to #208 was based on a misunderstanding of SemVer.

Hm not sure why you think this is the case. Of course, `1.2.3-pre.0 < 1.2.3`. Why do you think the previous version did not think that way? Being compatible to a preview version *has* to exclude being compatible with the then released version. There is no guarantee that changes in a preview version are the same when released. I am buffled that you think the system beforehand thought so.

And no - a preview should not "just be one patch". Otherwise, there would, e.g., never be a preview of React 17 / 18 / whatsoever, but just a final release. That's clearly not the case. You should already indicate the compatibility level in a preview; otherwise you can just make `0.0.0-buildid` releases...

> The frontend dependency checking system would flag things incorrectly. If an MF said it required at least core version 3.1.12 and core version 3.1.12-pre.100 is present, the frontend dependency checker would complain.

And it should! Obviously, this was created for a pre-release and therefore there is no guarantee that the changes in the pre-release made it into the full release... This is just logic. It's a pre-release for a reason.

ibacher:
@FlorianRappl I think I see what you're getting at now. So, the problem we had before (that #208 was trying to solve) was that the version number wasn't getting bumped correctly, so, looking at the versions in NPM, we have [3.1.12-pre.627](https://www.npmjs.com/package/openmrs/v/3.1.12-pre.627) which was create 9 days after 3.1.12 was released. This is one scenario we want to avoid.

The other scenario we want to avoid is the current state that we're in now where we've releases 3.1.14 (to unblock the demo site), but we've also released [3.1.15-pre.680](https://www.npmjs.com/package/openmrs/v/3.1.15-pre.680). The problem here is that Lerna has bumped the version in the release, but in the _repo_ the version number in lerna.json is 3.1.14, so the next PR that's merged after this will create version 3.1.14-pre.681.

I think the spirit of what #208 was trying to do (automatically increment the patch number as a fail-safe) is a good idea, but the current implementation doesn't do things quite correctly.

The flow we use for backend modules is this:

1. A release is done. This creates a tag, publishes the binaries, etc.
2. A commit is automatically made to the repo to increase the version number. This "new" version number can be user-selected (it's a variable in the CI build), but if it isn't, we default to bumping the minor version (so releasing 3.1.0 also publishes 3.2.0-SNAPSHOT).
3. When it comes time to make the next release, we review the changes and either release the SNAPSHOT version or a different version (maybe very little changed and the release gets called 3.1.1 or it becomes 4.0.0, etc.)
4. Repeat.

I think the state we want is to have some degree of automation around version numbers _for pre-releases only_ to avoid the same issue we had with 3.1.12-pre.627 and for this bumping the patch version is sufficient, but that bump in the patch version needs to actually be committed to Git so we don't end up in the situation we are currently in.

Does that summarise things fairly? Is there an easy way to fill in the gap in the current workflow or is there a better workflow we could adopt that avoids both tof these issues?

FlorianRappl:
Thanks @ibacher . Yeah that summarizes it pretty well. Personally, I'd like to distinguish between the different issues.

To avoid the scenario you described (which is certainly something to avoid) I'd have suggested a simple check that would (e.g., as part of the standard linting / pre-push) check if the current version is the same as the currently released version. If that's true then it would fail and you'd need to select a new version (which could be as simple as bump the patch version, no questions asked).

Alternatively, that thing could be done automatically, either in the release job (already pushing a preview with the patch version, but then we might have an ugly situation again imho as there might be some pre-release that is the same as a release with just an additional patch number (can be avoided, too, by, e.g., pushing that change with a `skip-ci` commit).

The flow you describe is essentially the last paragraph. In this scenario we'd avoid the sudden surprise that we face right now where - no matter what you do and what you specified - you get a patch version pre-release. This is fine in my opinion. My personal recommendation would be to check for it as described above though. In the end, a version number (change) can only (reliably) be decided by the respective developers of the changes.

brandones:
Yeah, thanks for putting that so clearly, @ibacher .

> The problem here is that Lerna has bumped the version in the release, but in the repo the version number in lerna.json is 3.1.14, so the next PR that's merged after this will create version 3.1.14-pre.681.

Fortunately, it looks like that's not what's happening:

Screenshot from 2021-12-01 10-43-12

on [npmjs](https://www.npmjs.com/package/@openmrs/esm-framework). It seems like versions are progressing in the order expected. This is because our CI job intentionally does a patch version bump + a prerelease version bump when releasing the `next` version.

I think it would be reasonable to do it the other way, also: commit the patch version bump to code right after each release, and then have CI just do a prerelease bump. Then there are sort of two approaches to versioning from there. If `lerna.json` now says version `3.1.15`, these alternatives are:
1. Developers leave the version at `3.1.15`. Each commit to master produces a `3.1.15-pre` version. At release time, the release manager looks at the changelog and chooses the next release version, whether `3.1.15`, `3.2.0`, or `4.0.0`.
2. Developers are responsible for updating the version according to their changes. A commit to master which introduces a breaking change should also change the version in `lerna.json` to `4.0.0`. CI would release the `next` version for that commit as `4.0.0-pre`. At release time, the release manager releases whatever version is in `lerna.json`.

I think any of these options are pretty reasonable. I think we should continue this conversation [in the relevant RFC](https://github.com/openmrs/openmrs-rfc-frontend/pull/31).

PR #230, opened 2021-12-01
Update Webpack Configuration

FlorianRappl:
## Requirements

## Summary

Our Webpack configuration is unfortunately a bit outdated, which may also result in some of the issues we see. For instance, there is no need for the `file-loader` (or `raw-loader`) to include resources. Since `file-loader` (in particular) has a peer dep to Webpack v4 we should be very careful (especially when using NPM v7).

This PR makes the usage consistent and improves this.

## Screenshots

_None._

## Related Issue

_None._

## Other

_None._


brandones (approved):
Thanks @FlorianRappl !

PR #232, opened 2021-12-03
Docs: node 17.2.0 required for 'yarn run:shell'

brandones:


README.md
```diff
@@ -84,6 +84,8 @@ yarn verify
 yarn run:shell
 ```
 
+This requires node 17.2.0 or higher.
```
FlorianRappl, line 87: Why? First, Node v17 is not an LTS and should never be a requirement. Its then v16 I guess. But even then I wouldn't write that. I run that command with Node v12 and Node v14 without any troubles.

So what's the reasoning (and why not use / change the `engines` field to make npm aware of that)?

brandones: I take it you're not able to reproduce [the error I posted on Slack](https://openmrs.slack.com/archives/CHP5QAE5R/p1638498354370900)?

brandones:
@manuelroemer confirmed on Slack that this should be working; he's using node 14. Closing.

PR #233, opened 2021-12-03
O3-952: Login page translations

denniskigen:
`##` Requirements

## Summary
    
This commit adds some missing translation keys and strings to the login page. These include:
    
- A translation string for the username input.
- A translation string for the password input.
- A translation key for the password input when the provided value is invalid.

Also, to make sure new translation strings and keys are kept up to date, I've added the `extract-translations` task to the pre-commit hook. The corollary of this change is that translation strings and keys get extracted from all of the existing apps inside of the packages directory. 

## Screenshots

Before:

Screenshot 2021-12-03 at 21 16 01

Screenshot 2021-12-03 at 21 16 11

After:

Screenshot 2021-12-03 at 21 14 53

Screenshot 2021-12-03 at 21 15 25


packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -139,7 +139,7 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
                   id="username"
                   type="text"
                   name="username"
-                  labelText={t("username", "UserName")}
+                  labelText={t("username", "Username")}
```
ibacher, line 142: I think we should actually leave this in. It's an easier way to smoke-test "are translations working" than trying to mess with browser language settings :grin:

tools/i18next-parser.config.js
```diff
@@ -42,18 +42,7 @@ module.exports = {
...
-    "es",
-    "fr",
-    "hi",
-    "sw_KE",
-    "sw",
-    "tr_TR",
-    "uk",
-    "zh_CN",
-  ],
+  locales: ["en"],
```
ibacher, line 45: Is this the result of the automated tooling? We at least have all of these languages for the login app, so it would be good if we could add the placeholders for the various other translations

denniskigen: Correct me if I'm wrong @brandones but I believe that we're depending on Transifex for translating from the default `en` translations to other locales. Transifex will handle generating locale-specific translation files for those other locales. So in that sense, we don't need to specify these other locales for `i18next-parser`.

brandones: Yes, that sounds right to me @denniskigen , except that we actually need to wire things up with Transifex for all our repos. Right now we are using Transifex with none of them.

ibacher: Well, in that case, I'm happy for this to be merged in. We can do the Transifex setup somewhere else. Thanks @brandones!

ibacher (approved):

PR #234, opened 2021-12-03
O3-951: Major lag caused by updateConfigExtensionStore

brandones:
## Summary

This improves the performance of loading extensions. The actual fix is in the first commit; I recommend reviewing this PR per-commit. The commits have been organized to facilitate that.

The fix prevents unnecessary calls to `configExtensionStore.setState`, which was causing lots of churn in the React reconciler. This improves the performance of clicking the user menu by about 70%; reducing the response time from about 500ms to about 140ms (when using `run:shell`, which uses a development build of React).

## Screenshots

These are profiles of what happens when you click the user menu.

Before
Screenshot from 2021-12-02 20-26-55

After
Screenshot from 2021-12-02 21-47-41

## Related Issue

https://issues.openmrs.org/browse/O3-951

### DIFF
--- a/packages/framework/esm-extensions/src/store.ts
+++ b/packages/framework/esm-extensions/src/store.ts
@@ -1,3 +1,4 @@
+import isEqual from "lodash-es/isEqual";
 import {
   configExtensionStore,
   ConfigExtensionStoreElement,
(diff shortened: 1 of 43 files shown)


jonathandick:
Awesome work @brandones !

packages/framework/esm-extensions/src/store.ts
```diff
@@ -124,5 +125,14 @@ function updateConfigExtensionStore(extensionState: ExtensionStore) {
...
+  if (
+    !isEqual(
+      configExtensionStore.getState().mountedExtensions,
+      configExtensionRecords
+    )
+  ) {
+    configExtensionStore.setState({
+      mountedExtensions: configExtensionRecords,
+    });
+  }
```
brandones, line 137: This is the important change for this ticket. Everything else is more or less unrelated.

denniskigen (approved):
Nicely done, @brandones!

ibacher (approved):

ibacher:
LGTM. Thanks @brandones!

PR #236, opened 2021-12-08
O3-918 Two+ instances of the same extension should be able to added the same slot

brandones:
## Summary

This brings our use of the term "name" and "id" into compliance with [RFC-27](https://github.com/openmrs/openmrs-rfc-frontend/blob/master/text/0027-extensions.md). A name corresponds to an extension, and an ID corresponds to an extension within a particular slot. There had been confusion on this subject; as such, providing different configurations to the same extension attached multiple times was not working.

## Related Issue

https://issues.openmrs.org/browse/O3-918


ibacher:
@brandones I'm good with the concept, but the test added to cover the feature seem to be failing in CI. Could you have a look at that?

ibacher (approved):

brandones:
Thanks @ibacher !

PR #238, opened 2021-12-09
Docs: API docs for esm-framework only

brandones:
This makes it so that we only generate API docs for `@openmrs/esm-framework`. That's really the only library people are using. This will allow us to focus more on making these docs usable. In accordance with that goal, they're now linked at the top of the main README.


packages/framework/esm-react-utils/docs/interfaces/ComponentConfig.md
```diff
@@ -17,7 +17,7 @@
 
 #### Defined in
 
-[packages/framework/esm-react-utils/src/ComponentContext.ts:11](https://github.com/openmrs/openmrs-esm-core/blob/master/packages/framework/esm-react-utils/src/ComponentContext.ts#L11)
+[packages/framework/esm-react-utils/src/ComponentContext.ts:11](https://github.com/openmrs/openmrs-esm-core/blob/39d3e79/packages/framework/esm-react-utils/src/ComponentContext.ts#L11)
```
brandones, line 20: Oh, something went wrong here. I need to fix this.

ibacher:
I think the overall change is probably fine, but it looks like we're leaving some dead links around right now?

brandones:
Fixed the broken links, @ibacher

ibacher (approved):

PR #239, opened 2021-12-09
Fix incorrect startDatetime when starting a new visit

denniskigen:
## Requirements

## Summary

Change the type annotation for the `startDatetime` from `Date` to `string` as we're now converting it into an ISO 8601 string before POSTing. 

Currently, starting a new visit doesn't always record the correct visit `startDateTime`. Converting the provided date value to an ISO string prior to saving it fixes this.

The actual fix will happen in the [VisitForm](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx) component.

### DIFF
--- a/packages/framework/esm-api/docs/interfaces/NewVisitPayload.md
+++ b/packages/framework/esm-api/docs/interfaces/NewVisitPayload.md
@@ -37,7 +37,7 @@ ___
 
 ### startDatetime
 
-• **startDatetime**: `Date`
+• **startDatetime**: `string`
 
 #### Defined in
 
@@ -47,7 +47,7 @@ ___
 
 ### stopDatetime
 
-• `Optional` **stopDatetime**: `Date`
+• `Optional` **stopDatetime**: `string`
 
 #### Defined in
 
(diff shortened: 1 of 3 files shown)


brandones:
@denniskigen Why don't we keep the parameter as a Date and do the conversion here?

denniskigen:
Just because it's less of a hassle - I'd have much preferred it if all of this visits-related stuff was colocated.

brandones:
Really, it's less of a hassle to make all users of these functions format the date in a specific way, than to just do it once within these functions?

Also in a more general sense, we should be trying to keep dates as Dates until as close as possible to the moment they are transported or displayed.

brandones:
Ping @denniskigen ? I'm open to doing it the way you've suggested, but I'm confused about your reasoning.

denniskigen:
Recent fixes to the [Visit Form component](https://github.com/openmrs/openmrs-esm-patient-chart/blob/0f2ef7cff77459488cb3828088e72931f5760ab0/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx) provided by https://github.com/openmrs/openmrs-esm-patient-chart/pull/484 obviate the need for this pull request, @brandones. 

Feel free to close.

PR #243, opened 2021-12-14
Create functions for locale-sensitive date formatting

brandones:
## Requirements

## Summary

This creates date and time formatting functions which are locale sensitive and which limit the available formats to those recommended by the design team.

### DIFF
--- a/packages/framework/esm-utils/src/omrs-dates.ts
+++ b/packages/framework/esm-utils/src/omrs-dates.ts
@@ -1,3 +1,8 @@
+/**
+ * @module
+ * @category Date and time
+ */
+import { i18n } from "i18next";
 import dayjs from "dayjs";
 import utc from "dayjs/plugin/utc";
 import isToday from "dayjs/plugin/isToday";
@@ -7,9 +12,7 @@ dayjs.extend(isToday);
 
 declare global {
   interface Window {
-    i18next: {
-      language: string;
-    };
+    i18next: i18n;
   }
 }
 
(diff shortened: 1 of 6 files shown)


donaldkibet (approved):
LGTM Thanks @brandones 🚀

brandones:
Thanks @donaldkibet !

CC @denniskigen @vasharma05 @samuelmale — I want to make sure this has support from the people who will be using it, developing UI features.

packages/framework/esm-utils/src/omrs-dates.ts
```diff
@@ -79,36 +90,144 @@ export function toOmrsIsoString(date: DateInput, toUTC = false): string {
...
+ *
+ * This is created by concatenating the results of `formatDate`
+ * and `formatTime` with a comma and space. This agrees with the
+ * output of `Date.prototype.toLocaleString` for *most* locales.
+ */
+export function formatDatetime(date: Date, mode: FormatDateMode = "standard") {
+  const dateString = formatDate(date, mode);
+  const timeString = formatTime(date);
+  return `${dateString}, ${timeString}`;
+}
```
ibacher, line 223: From the designs, I notice that date and times are usually separated by `@`, e.g., `Today @ 11:21`, `30-Sep-2021 @ 18:42`. Shouldn't this then be the format we use here?

brandones: It's true. But see Ciarán's [comment](https://openmrs.slack.com/archives/CHP5QAE5R/p1639129372442800?thread_ts=1639110796.438300&cid=CHP5QAE5R) on Slack:

> I think we could reduce the date styles down to these two:
> dd-mmm-yyyy, hh:mm
> Today, hh:mm

PR #246, opened 2021-12-16
Fix O3-979: Translations not working

brandones:
The translation lookup function had not been adapted to accommodate module federation.

## Related Issue

https://issues.openmrs.org/browse/O3-979


denniskigen (approved):
LGTM. Thanks for this, @brandones.

PR #248, opened 2021-12-17
Add parseDate function

brandones:
I'm suggesting that we have this separate `parseDate` function rather than having `formatDate` accept strings and implicitly parse them. I think it's helpful to have to be explicit about what form a date is in. It should reduce the probability of certain kinds of errors, like running `formatDate` on the same date twice (we can't guarantee it's idempotent).


ibacher (approved):
@brandones I'm happy with this as a change. Should we add some docs on using dates in the 3.0 frontend? Seems like an important issue to ensure that everyone is following an established pattern for...

PR #253, opened 2021-12-22
Do the prerelease versioning of esm-app-shell before deploying to CI

brandones:
At present, `window.spaVersion` on openmrs-spa is "3.1.14-20211221". However, this is actually the same `@openmrs/esm-app-shell` code as is pre-released as version [3.1.15-pre.735](https://www.npmjs.com/package/@openmrs/esm-app-shell/v/3.1.15-pre.735). The problem is that the deployed build is happening before the pre-release versioning. This should correct that.


jonathandick (approved):
lgtm

PR #255, opened 2021-12-22
Add global nav slot to left nav bar

jonathandick:
Presently, we have a single slot which is essentially overridden by a route (e.g. /chart) to allow route specific navigation items to be added to the nav bar. This adds a new slot (above the existing page specific slot) which can be used across mf's to add global links to that will always be present independent of the route. Designs are forthcoming for how this navigation will look like but this sets the foundation to allow this to happen.

## Requirements

- [ ] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Screenshots

_None._

## Related Issue

_None._

## Other

_None._


brandones (approved):
This seems like a reasonable approach to me. Thanks @jonathandick

PR #257, opened 2021-12-22
Add time options to formatDate 

brandones:
## Summary

This adds an `options` parameter to `formatDate`, with a `time` parameter that can be used to control when the time should be appended to the date. By default, the time is appended only when the date is "Today."


donaldkibet (approved):
LGTM Thanks @brandones

brandones:
Thanks @donaldkibet !

PR #258, opened 2022-01-05
Update context switcher hover styling

donaldkibet:
## Requirements

## Summary

- Update the style for the hover state of content-switcher, this is as a result of not matching with the designs which @Ciarán Duffy pointed out

## Screenshots

enhancement

## Related Issue

_None._

## Other

_None._

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -132,8 +132,8 @@
 
   &:focus,
   &:hover {
-    background-color: $interactive-01;
-    color: $ui-background;
+    background-color: #edf5ff;
+    color: $interactive-01;
   }
 
   &.bx--content-switcher--selected {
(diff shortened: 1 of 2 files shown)


packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -132,8 +132,8 @@
 
   &:focus,
   &:hover {
-    background-color: $interactive-01;
-    color: $ui-background;
+    background-color: #edf5ff;
```
brandones, line 135: We should add this color to [the vars](https://github.com/openmrs/openmrs-esm-core/blob/master/packages/framework/esm-styleguide/src/_vars.scss). It's in the [styleguide](https://scene.zeplin.io/project/60d59321e8100b0324762e05) as blue/10.

brandones (approved):

packages/framework/esm-styleguide/src/_overrides.scss
donaldkibet: @brandones added the color to vars

brandones: Thank you!

PR #259, opened 2022-01-07
Don't load the polyfill for url

ibacher:
This fixes things so that the SDK won't break when executing `npx openmrs@next build`.

Error this is fixing can be [found here on Slack](https://openmrs.slack.com/archives/C02Q7G0A5PE/p1641485421004000?thread_ts=1641485325.003900&cid=C02Q7G0A5PE).

### DIFF
--- a/packages/shell/esm-app-shell/webpack.config.js
+++ b/packages/shell/esm-app-shell/webpack.config.js
@@ -159,7 +159,13 @@ module.exports = (env, argv = {}) => {
     resolve: {
       mainFields: ["module", "main"],
       extensions: [".ts", ".tsx", ".js", ".jsx"],
-      fallback: { http: false, stream: false, https: false, zlib: false },
+      fallback: {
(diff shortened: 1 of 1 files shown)


jonathandick (approved):

ibacher:
@jonathandick Thanks for the review!

brandones:
@ibacher Any idea why this works?

ibacher:
@brandones Not exactly. My best guess is that webpack is trying to load a dependency we don’t actually use (node-fetch from cross-fetch from i18next’s http backend), sees that it’s trying to load Node’s URL library in a web context and bails because no polyfill is defined to emulate Node’s URL library in the web context (this latter causes an error bc webpack 4 used to automatically load an appropriate polyfill). But that’s mostly a guess.

brandones:
Well, nice work, anyway. Seems like a tricky one.

PR #260, opened 2022-01-08
Mock fixup

brandones:
Exports the esm-utils functions as their own mocks. Organizes the mocks file some. Adds an easy way to compare the real API to the mock API of `@openmrs/esm-framework`. Removes some mocks for things that don't exist any more.


denniskigen (approved):
Nice work, @brandones!

PR #262, opened 2022-01-11
Create mechanism for overriding webpack configuration

brandones:
This allows a frontend module developer to do things like

```js
// webpack.config.js

const config = require("openmrs/default-webpack-config");

// This is required for react-pdf
config.additionalConfig.resolve = { fallback: { "stream": false } };

// This is a stupid example of an "override"
config.overrides.devServer = null;

module.exports = config;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,8 @@ yarn run:shell
 
 ```sh
 cd packages/apps/esm-[xyz]-app
-yarn start
+yarn serve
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -91,7 +91,8 @@ yarn run:shell
 
 ```sh
 cd packages/apps/esm-[xyz]-app
-yarn start
+yarn serve
```
brandones, line 94: Unrelated. I think `yarn start` has not been working since the switch to module federation.

ibacher (approved):
`mergeWith` actually seems to be pretty smart!

PR #263, opened 2022-01-11
interpolateUrl should be public

brandones:


donaldkibet (approved):
LGTM Thanks @brandones

PR #267, opened 2022-01-13
O3-780: Add bundle size monitoring to PR actions for all MF repos

FlorianRappl:
## Requirements

## Summary

We should be tracking bundle size changes. We want a tool that comments on every PR with the bundle size change caused by the PR.

## Screenshots

_None._

## Related Issue

See [O3-780](https://issues.openmrs.org/browse/O3-780).

## Other

_None._


FlorianRappl:
The error above should be gone once the branch is checked in. `master` does not have the file currently. So I can only verify it when we merged in the change.

tools/size-generator.mjs
```diff
@@ -0,0 +1,28 @@
+import { generateFileSizeReport } from "@jsenv/file-size-impact";
+
+new URL("./", import.meta.url);
```
manuelroemer, line 3: Not sure, is this intended (maybe to throw an error for some weird URLs) or is this a leftover?

manuelroemer (approved):
Looks good, let's see this in action! 🚀

tools/size-generator.mjs
FlorianRappl: Ha. You found my easter egg! Seriously, great catch : rocket:!

manuelroemer: The easter bunny came early this year 😆

PR #272, opened 2022-01-19
O3-1015  Implementer should be able to configure primary color

brandones:
https://issues.openmrs.org/browse/O3-1015

Peek 2022-01-18 20-02

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -42,20 +42,24 @@
 .bx--btn--tertiary:active,
 .bx--btn--tertiary:focus {
   background-color: $brand-teal-01;
+  background-color: var(--brand-teal-01);
 }
 
 .bx--btn--primary:hover {
   background-color: $brand-01;
+  background-color: var(--brand-01);
 }
 
 .bx--btn--tertiary {
   color: $brand-teal-01;
+  color: var(--brand-teal-01);
 }
 
 .bx--btn--tertiary,
 .bx--btn--primary:focus,
 .bx--btn--tertiary:focus {
   border-color: $brand-teal-01;
+  border-color: var(--brand-teal-01);
 }
 
 .bx--btn--tertiary:hover,
@@ -67,21 +71,27 @@
 /* UI Shell header */
 
 .bx--header {
-  background-color: $brand-01;
-  border-bottom: $brand-02;
+  // background-color: $brand-01;
+  background-color: var(--brand-01);
+  // border-bottom: $brand-02;
+  border-bottom: var(--brand-02);
 }
 
 .bx--header__action:hover {
   background-color: $brand-02;
+  background-color: var(--brand-02);
 }
 
 .bx--header-panel {
   background-color: $brand-02;
+  background-color: var(--$brand-02);
 }
 
 .bx--header__action--active {
   background-color: $brand-02;
+  background-color: var(--brand-02);
   border-color: $brand-02;
+  border-color: var(--brand-02);
 }
 
 .bx--overflow-menu--flip {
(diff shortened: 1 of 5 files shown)


packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -67,21 +71,27 @@
 /* UI Shell header */
 
 .bx--header {
-  background-color: $brand-01;
-  border-bottom: $brand-02;
+  // background-color: $brand-01;
+  background-color: var(--brand-01);
+  // border-bottom: $brand-02;
+  border-bottom: var(--brand-02);
```
ibacher, line 77: Clean-up: we should probably keep the commented-out lines.

packages/framework/esm-styleguide/src/_vars.scss
```diff
@@ -27,6 +27,12 @@ $field-01: #f4f4f4;
 $grey-2: #e0e0e0;
 $labeldropdown: #c6c6c6;
 
+:root {
+  --brand-01: #005d5d;
+  --brand-02: #004144;
+  --brand-teal-01: #007d79;
+}
```
ibacher, line 34: Since this is in an scss file, wouldn't it make sense to do:

```suggestion
:root {
  --brand-01: $brand-01;
  --brand-02: $brand-02;
  --brand-teal-01: $brand-teal-01;
}
```

ibacher: Secondary point to all of this, but `brand-teal-01` could use a rename...

ibacher (commented):
I think that this is basically the right approach. The main downside is that we're requiring this slightly verbose pattern wherever this colour is used.

This might be more invasive, but maybe we could leverage a Sass mixin, e.g.,

```scss
@mixin brand-background-color {
  background-color: $brand-01;
  background-color: var(--brand-01);
}
```

Then again, maybe that's just making more trouble?

packages/framework/esm-styleguide/src/_vars.scss
brandones: Strangely, that doesn't work—if you do this then the variable `--brand-01` gets set to the string literal `"$brand-01"`.

brandones:
If we're going to use a mixin I'd say go all the way and use this one:

```scss
@mixin var($property, $varName) {
  #{$property}: map-get($vars, $varName);
  #{$property}: var(--#{$varName});
}
```

which is used like

```scss
@include var(background-color, brand-01);
```

However, I'm inclined to think that people outside the squad would rarely use this strange syntax.

Unless—and this is something I think we should do—we get rid of the `$brand-01` (et al.) SASS variables. That would force people to use either that or the CSS variable, either of which would be preferable.

packages/framework/esm-styleguide/src/_vars.scss
ibacher: Hmmm... I think `brand-light-01` at least gets the goal across better? I'm thinking that we'll definitely need some sort of documentation with annotated screenshots so people can see where the branding takes place.

ibacher:
> However, I'm inclined to think that people outside the squad would rarely use this strange syntax.

Yeah... Sass mixins don't make for easy-reading syntax. I was just trying to come up with a way to avoid copy-and-paste where we end up with just the `background-color: $brand-01` everywhere.

> we get rid of the $brand-01 (et al.) SASS variables

I don't necessarily disagree with this, but it needs some careful working through since at least some of the variables we define seem to be intended as overrides of Carbon-defined variables (though at a quick pass, it doesn't look like we're actually using different values).

> That would force people to use either that or the CSS variable, either of which would be preferable.

Or the worst-case scenario `background-color: #005d5d` 😬

ibacher (approved):
I actually like this!

PR #274, opened 2022-01-19
Added flair to point folks to the esm-template-app

gracepotma:


brandones:
Do we think that people are finding the "Tour" page but not the template app?

jonathandick (approved):
is this still active? can we either close or merge? approved if we're keeping this.

PR #278, opened 2022-01-20
Change esm-framework peer dependency versions from 3.x to >=3.1.14-pre

brandones:
This fixes these warnings. The problem is caused by the fact that, according to SemVer, pre-release versions (e.g. `3.1.14-pre.800`) do not satisfy major version caret specifiers (e.g. `3.x` or `^3`).

Screenshot from 2022-01-19 22-26-01


ibacher (approved):

ibacher:
This is one of the most annoying interpretations of SemVer out there...

PR #279, opened 2022-01-20
Remove frontendDependency system; use peerDependencies for this

brandones:
This removes the frontend dependency checking system. Now that we use module federation, frontend dependency versions are already checked. See the warnings below, which come from the module federation version checking system.

image


ibacher (approved):
Yeah... I don't understand why this is a desirable feature at all.

PR #280, opened 2022-01-20
Added emphasis that esm-template can be used to create a new MF

gracepotma:


brandones:
As with the other one, I suspect that people are just not finding the "Creating a Frontend Module" page in the first place.

gracepotma:
> As with the other one, I suspect that people are just not finding the "Creating a Frontend Module" page in the first place.

Honestly @brandones I think we need to assume that half the time, folks might just arrive on a single page to try to answer 1 specific question; i.e. probably best to assume that they might not have read all the previous pages. OK that sounds more dramatic than I mean it to, but for key resources we really really want them to take away (e.g. the esm-template-app) I thought we might as well call attention to those wherever. Does that make sense? Happy to defer to you and see how it goes; I realize this could be a bit of a large reaction to all the comments we're hearing about not having enough information to go on.

docs/main/creating_a_microfrontend.md
```diff
@@ -1,9 +1,15 @@
 # Creating a Frontend Module
 
-New frontend modules can be created from the
-[openmrs-esm-template-app](https://github.com/openmrs/openmrs-esm-template-app).
-You can fork that repository, or clone-and-copy it. Follow the instructions
-in the README to turn it into your own frontend module.
+:bulb:**New frontend modules can be created from the
+[openmrs-esm-template-app](https://github.com/openmrs/openmrs-esm-template-app)**.:bulb:
```
brandones, line 4:
```suggestion
**New frontend modules can be created from the
[openmrs-esm-template-app](https://github.com/openmrs/openmrs-esm-template-app)**. :bulb:	
```

docs/main/creating_a_microfrontend.md
```diff
@@ -1,9 +1,15 @@
...
+:bulb:**New frontend modules can be created from the
+[openmrs-esm-template-app](https://github.com/openmrs/openmrs-esm-template-app)**.:bulb:	
+
+You can fork that repository, or clone-and-copy it. 
+Follow the instructions in the README to turn it into your own frontend module.
+
+|Developer Quote|
+|------|
+|:thinking:	*"I was new to using the 3.x framework, but I was able to use the esm-template-app to set up a new microfrontend module within a few hours." :partying_face:*|
+</br>
```
brandones, line 12:
```suggestion
```

brandones:
Yeah but I can't imagine them coming to this page trying to learn something other than how to make a new microfrontend, and there it is on the first line.

I agree with you in general, and think you're right about how people generally approach docs. But I'm trying to balance with a variation on [this insight](https://twitter.com/dan_abramov/status/1470623694821208066?s=20): eventually, appealing to the lowest common denominator worsens the experience for those with a median level of competence. We have to accept that people will forever ask questions to which there are obvious answers in the docs. We will continue linking them to the appropriate places. If we're going to make decisions that add what I would consider cruft and make the docs what I would consider less readable, I would like those decisions to at least be evidence-based.

brandones (approved):

PR #286, opened 2022-01-21
Allow Proxying with Additional Cookies

FlorianRappl:
## Requirements

## Summary

For some environments we need to supply additional parts when proxying the request. For instance, due to some protection layer we'll need a cookie that is then read by a load balancer in front of the actual OpenMRS backend.

## Screenshots

_None._

## Related Issue

_None._

## Other

_None._

### DIFF
--- a/packages/shell/esm-app-shell/src/run.ts
+++ b/packages/shell/esm-app-shell/src/run.ts
@@ -210,10 +210,15 @@ async function setupServiceWorker() {
     `${window.getOpenmrsSpaBase()}service-worker.js`
   );
   registerSwEvents(sw);
+  await triggerOfflineMode();
+}
 
+async function triggerOfflineMode() {
   if (navigator.onLine) {
     try {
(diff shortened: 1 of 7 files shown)


packages/shell/esm-app-shell/src/run.ts
```diff
@@ -210,10 +210,15 @@ async function setupServiceWorker() {
...
   );
   registerSwEvents(sw);
+  await triggerOfflineMode();
+}
 
+async function triggerOfflineMode() {
   if (navigator.onLine) {
     try {
-      await Promise.all([precacheImportMap(), precacheSharedApiEndpoints()]);
+      if (localStorage.getItem("CAN_GO_OFFLINE") === "true") {
```
manuelroemer, line 219: This should probably be removed to not break caching (at least the local storage check), but otherwise this looks good to me! 🚀

FlorianRappl: Ouch good catch! Thanks!

manuelroemer (approved):

PR #292, opened 2022-01-25
(bug) O3-1003: fix login not displaying invalid login message

donaldkibet:
## Requirements

## Summary

- When a user logs in with bad password, no error message is displayed. 

## Screenshots

bug-login

## Related Issue

https://issues.openmrs.org/browse/O3-1003

## Other

_None._


jonathandick (approved):
lgtm

packages/apps/esm-login-app/translations/en.json
```diff
@@ -3,7 +3,8 @@
   "confirm": "Confirm",
   "contactAdmin": "Contact the site administrator",
   "continue": "Continue",
-  "errorMessage": "",
```
brandones, line 6: Ha! This would explain why the text never showed up in English.

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -209,11 +220,6 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
                 </Button>
               </div>
             )}
-            <div className={styles["center"]}>
-              <p className={styles["error-msg"]}>
-                {t("errorMessage", errorMessage)}
-              </p>
-            </div>
```
brandones, line 216: And the fact that nothing wraps this would explain why it *always* shows up for languages with a non-blank translation.

brandones (changes requested):
@donaldkibet Going to disagree with JJ here. There are now two competing mechanisms for handling errors—the existing one, and the one you created. The original is quite good and I think we should use it, though it needs a little repair.

What it does is throws an Error for the invalid credentials (line 108), then handles that Error—_or any other error, such as a 5XX_—in a catch block. Whatever the error, its message goes to `errorMessage` (line 111) and the inputs get reset (line 112).

Where things go wrong is on line 214, where the translation key is used. Right now, the line is

```tsx
t("errorMessage", errorMessage)
```

This will look up a translation key "errorMessage" and print it if it finds it, and print `errorMessage` only if no translation key is found—but it is always found, so it is always printed. In English it is `""` so nothing is printed. In Spanish the error text is always shown.

I think the intention of the author was probably

```
t(errorMessage, errorMessage)
```

This should print nothing if `errorMessage` is empty, a translation if `errorMessage` happens to be equal to one of the translation keys, and otherwise will print `errorMessage` itself.

We should change line 108 so that it throws something that looks more like a translation key so that the intent of the code is clearer:

```tsx
throw new Error("invalidCredentials")
```

And then make sure that `invalidCredentials` is defined as a translation key in all the translation JSON files. If we are doing automatic extraction, we may need to find a way to tell it about this key.

brandones:
Note that in the code you've written, `errorMessage` is never used. And no matter the error that comes up (even e.g. a 503), "Invalid username or password" is what gets printed, which is not good.

brandones:
See the [i18next-parser README](https://github.com/i18next/i18next-parser#caveats) about the dynamic keys. Looks like we will want something like

```js
// t('invalidCredentials')
```

donaldkibet:
@brandones I have made the requested changes

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -130,6 +130,15 @@ const Login: React.FC<LoginProps> = ({ history, location, isLoginEnabled }) => {
   if (config.provider.type === "basic") {
     return (
       <div className={`canvas ${styles["container"]}`}>
+        {errorMessage && (
+          <InlineNotification
+            kind="error"
+            style={{ width: "23rem" }}
+            subtitle={t(`${errorMessage}`)}
```
brandones, line 137:
```suggestion
            subtitle={t(errorMessage)}
```

brandones (approved):
One suggestion, but otherwise looks good! Thanks @donaldkibet .

donaldkibet:
Thanks @brandones for the review, going ahead and merging this in

PR #296, opened 2022-01-31
Docs: fix typo in Map

brandones:


ZacButko (approved):
Lint is complaining about a lot of useEffect missing dependencies which can get cleaned up later. Looks good to me

PR #297, opened 2022-02-01
Minor performance improvements to ExtensionSlot

brandones:
A couple minor performance improvements to the `ExtensionSlot` implementation.

One I had messed up in https://github.com/openmrs/openmrs-esm-core/pull/236 . I hadn't realized that defining `defaultSelect` inline would cause render churning.

The other is a result of the observations that 1. refs should [generally not](https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/) be used in dependency arrays and 2. props changing should not cause the Extension component to re-render.

### DIFF
--- a/docs/main/config.md
+++ b/docs/main/config.md
@@ -63,7 +63,7 @@ defineConfigSchema("@openmrs/esm-hologram-doctor", {
     color: {
       _type: Type.Boolean,
       _default: false,
-      _description: "Whether the cologram supports color display."
+      _description: "Whether the hologram supports color display."
     }
   }
 }
--- a/packages/framework/esm-react-utils/src/ExtensionSlot.tsx
+++ b/packages/framework/esm-react-utils/src/ExtensionSlot.tsx
@@ -48,7 +48,9 @@ export interface ExtensionSlotBaseProps {
 export type ExtensionSlotProps = ExtensionSlotBaseProps &
   React.HTMLAttributes<HTMLDivElement>;
 
-const defaultSelect = (e) => e;
+function defaultSelect(extensions: Array<ConnectedExtension>) {
+  return extensions;
+}
 
 export const ExtensionSlot: React.FC<ExtensionSlotProps> = ({
   extensionSlotName,
--- a/packages/framework/esm-react-utils/src/ExtensionSlot.tsx
+++ b/packages/framework/esm-react-utils/src/ExtensionSlot.tsx
@@ -48,9 +48,11 @@ export interface ExtensionSlotBaseProps {
 export type ExtensionSlotProps = ExtensionSlotBaseProps &
   React.HTMLAttributes<HTMLDivElement>;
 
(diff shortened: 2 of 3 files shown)


docs/main/config.md
```diff
@@ -63,7 +63,7 @@ defineConfigSchema("@openmrs/esm-hologram-doctor", {
     color: {
       _type: Type.Boolean,
       _default: false,
-      _description: "Whether the cologram supports color display."
+      _description: "Whether the hologram supports color display."
```
brandones, line 66: Unrelated

jonathandick (approved):
lgtm

brandones:
@jonathandick You understand this change, or just helping this through? I want to make sure someone familiar with dependency array comparison rules agrees with me on this.

jonathandick:
More the latter. @ZacButko what do you think?

packages/framework/esm-react-utils/src/ExtensionSlot.tsx
```diff
@@ -48,9 +48,13 @@ export interface ExtensionSlotBaseProps {
...
   React.HTMLAttributes<HTMLDivElement>;
 
+function defaultSelect(extensions: Array<ConnectedExtension>) {
+  return extensions;
+}
+
 export const ExtensionSlot: React.FC<ExtensionSlotProps> = ({
   extensionSlotName,
-  select = (e) => e,
+  select = defaultSelect,
```
ibacher, line 57: I see so this is similar to [this rule](https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent), i.e., the anonymous function always changes identity between two renders, so the component always gets re-rendered?

packages/framework/esm-react-utils/src/ExtensionSlot.tsx
```diff
@@ -83,7 +87,7 @@ export const ExtensionSlot: React.FC<ExtensionSlotProps> = ({
           {children ?? <Extension state={stateRef.current} />}
         </ComponentContext.Provider>
       )),
-    [select, extensions, extensionSlotName, stateRef.current]
+    [select, extensions, extensionSlotName]
```
ibacher, line 90: From the linked article:

> Luckily, as we just learned, including it in the dependency array doesn't make any difference anyway, so just include it!

So, similarly, excluding it wouldn't seem to have any ill-effects anyway (after all, `useRef` is ensuring that `stateRef === stateRef` on each render, even if `stateRef.current !== state`).

ibacher (approved):
@brandones IIUC, this seems correct and at least the move from anonymous function -> named function is potentially saving us a fair bit of wasted render work.

packages/framework/esm-react-utils/src/ExtensionSlot.tsx
brandones: Yep!

PR #298, opened 2022-02-01
(fix) Fix logo alignment in the navbar

denniskigen:
## Requirements

## Summary

The app logo should be centre-aligned in a flex container to match the look and feel of the rest of the elements in the navigation bar. Note that this change does not affect SVG logos.

## Screenshots

> Before
Screenshot 2022-02-01 at 15 58 25

> After
Screenshot 2022-02-01 at 15 55 30

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/logo/logo.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/logo/logo.component.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { useConfig } from "@openmrs/esm-framework";
+import styles from "./logo.component.scss";
 
 const Logo: React.FC = () => {
   const { logo } = useConfig();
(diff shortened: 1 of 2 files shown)


brandones (approved):

packages/apps/esm-primary-navigation-app/src/components/logo/logo.component.tsx
```diff
@@ -1,13 +1,20 @@
...
 
   return (
     <>
       {logo?.src ? (
-        <img src={logo.src} alt={logo.alt} width={110} height={40} />
+        <img
+          className={styles.logo}
+          src={logo.src}
+          alt={logo.alt}
+          width={110}
```
brandones, line 15: It seems funny that we hard-code the width. Seems like we'd want to support logos of different widths at some point.

denniskigen: Yeah, what do you reckon would be a reasonable default? Or maybe that's one for Ciaran.

brandones: No default; just that whatever image you drop in gets sized correctly. Only the height and the alignment actually matter. Does that make sense?

PR #299, opened 2022-02-01
(fix) Improve Lifecycle typing

brandones:
Improvement to typing, and fixing a bug which the improved type exposes.

### DIFF
--- a/packages/shell/esm-app-shell/src/apps.ts
+++ b/packages/shell/esm-app-shell/src/apps.ts
@@ -6,6 +6,7 @@ import {
   PageDefinition,
   registerExtension,
   ResourceLoader,
+  Lifecycle,
 } from "@openmrs/esm-framework";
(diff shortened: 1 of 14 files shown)


ibacher (approved):

packages/shell/esm-app-shell/src/apps.ts
```diff
@@ -65,21 +66,29 @@ function getLoader(
...
-              ...dataProps,
-              ...props,
-            });
-          },
         };
+        if (lifecycle.update) {
+          Object.assign(newLifecycle, {
+            update(props) {
+              return (
+                lifecycle.update &&
```
ibacher, line 82: Isn't this redundant?

brandones: No, because it's executed at the time that the `update` function is called, by which point `lifecycle` may have changed.

PR #301, opened 2022-02-03
(feat) Return updated `patientUuid` from usePatient

denniskigen:
## Requirements

## Summary

Presently, `usePatient` does not return the updated `patientUuid` whenever the route changes. This commit updates `usePatient` to return the latest updated `patientUuid` value.

### DIFF
--- a/packages/framework/esm-react-utils/src/usePatient.ts
+++ b/packages/framework/esm-react-utils/src/usePatient.ts
@@ -146,7 +146,7 @@ export function usePatient(patientUuid?: string) {
   return {
     isLoading: state.isLoadingPatient,
     patient: state.patient,
-    patientUuid: patientUuid ?? null,
+    patientUuid: patientUuid ?? state.patientUuid,
     error: state.err,
   };
 }


brandones (approved):
You're a hero @denniskigen

jonathandick:
Hey, i just checked (and I'm pretty sure this change was integrated) but same problem still occurring on openmrs-spa.

brandones:
Oh yeah, isn't the core deployment on openmrs-spa manual or something, such that it wouldn't have been integrated yet? I always forget how that works (we should document it [here](http://o3-dev.docs.openmrs.org/#/under_the_hood/squad_devops)).

ibacher:
> isn't the core deployment on openmrs-spa manual or something

No that was my mistake. module-spa has the ability to [load assets from a remote server](https://github.com/openmrs/openmrs-module-spa/blob/ee2b10b2c7746629c62f33059eb41c1b3e938185/omod/src/main/java/org/openmrs/module/spa/servlet/SpaServlet.java#L91) which is used as a kind of poor man's reverse proxy in front of the most recent `@openmrs/esm-app-shell` served from Digital Ocean. So, in brief, it's always up-to-date.

PR #304, opened 2022-02-03
Improve PR template

brandones:
Brevity means people read it.

We should get rid of "I checked for feature overlap with [**existing widgets**](https://om.rs/directory)." for all the other repos. That has never been meaningful to anyone.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,6 +15,7 @@
 
 ## Related Issue
 <!-- Paste the link to the Jira ticket here if one exists. -->
+<!-- https://issues.openmrs.org/browse/O3- -->
 
 ## Other
 <!-- Anything not covered above -->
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,41 +1,20 @@
...
 ## Related Issue
-
-_None._
-
-<!--
-Optional.
-If present, please link any related issue here, e.g. "https://issues.openmrs.org/browse/123").
-Don't forget to remove the *None.* above if you do fill this section.
--->
+<!-- Paste the link to the Jira ticket here if one exists. -->
```
ibacher, line 17:
```suggestion
<!-- Paste the link to the Jira ticket here if one exists. -->
<!-- https://issues.openmrs.org/browse/O3- -->
```

ibacher (approved):
One small change, but only because I actually use the URL when I have a Jira ticket

PR #307, opened 2022-02-04
(fix) O3-1061: Patient Chart is re-rendering on each page change

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Ensures that the patient object is not reloaded unnecessarily. Reloading the patient object was causing the patient header to reload in the patient chart every time the dashboard was changed.

## Related Issue
https://issues.openmrs.org/browse/O3-1061

### DIFF
--- a/packages/framework/esm-react-utils/src/usePatient.ts
+++ b/packages/framework/esm-react-utils/src/usePatient.ts
@@ -1,4 +1,4 @@
-import { useEffect, useReducer } from "react";
+import { useEffect, useMemo, useReducer } from "react";
 import { fetchCurrentPatient, PatientUuid } from "@openmrs/esm-api";
 
 type NullablePatient = fhir.Patient | null;
@@ -133,10 +133,13 @@ export function usePatient(patientUuid?: string) {
 
   useEffect(() => {
     const handleRouteUpdate = (evt) => {
-      dispatch({
-        type: ActionTypes.loadPatient,
-        patientUuid: getPatientUuidFromUrl(),
-      });
+      const newPatientUuid = getPatientUuidFromUrl();
+      if (newPatientUuid != state.patientUuid) {
+        dispatch({
+          type: ActionTypes.loadPatient,
+          patientUuid: newPatientUuid,
+        });
+      }
     };
     window.addEventListener("single-spa:routing-event", handleRouteUpdate);
     return () =>


ZacButko (commented):
For the "Unchanged files with check annotations", these are mostly useEffect missing dependencies. Can we silence these with `// eslint-disable-next-line react-hooks/exhaustive-deps` ?

jonathandick (approved):
Thanks @brandones !. I've tested this out locally and the problem is solved. Well done.

denniskigen:
Sharp work as ever, @brandones!

I wonder, should we be listening for all `single-spa:routing-events` or should we be asserting only that the chart summary landing page URL has changed. Am I wrong in assuming that this event fires every time a user navigates around the same patient's patient chart?

denniskigen (approved):

packages/framework/esm-react-utils/src/usePatient.ts
```diff
@@ -133,10 +133,13 @@ export function usePatient(patientUuid?: string) {
 
   useEffect(() => {
     const handleRouteUpdate = (evt) => {
-      dispatch({
-        type: ActionTypes.loadPatient,
-        patientUuid: getPatientUuidFromUrl(),
-      });
+      const newPatientUuid = getPatientUuidFromUrl();
```
dkayiwa, line 136: Does the `usePatient()` hook apply for `getPatientUuidFromUrl()` as here? https://github.com/openmrs/openmrs-esm-patient-chart/pull/536/files

ibacher: @dkayiwa Not sure I quite understand what you're asking... The commit you reference was done because the nav component wasn't receiving the updated patient uuid at the same time as other widgets, but that's due to [this line](https://github.com/openmrs/openmrs-esm-patient-chart/pull/536/files#diff-cf0aaeac4d8c50f016a1ae3971f18a98b8afa184e4777d3f8f797520c727882aL22) which listened to the `popstate` event instead of the `single-spa:routing-event`. So by switching from using the custom parsing to using the `usePatient` hook we ensure there's at least consistency across the UI.

ibacher:
@ZacButko Certainly better to explicitly disable the warnings than the current situation where they show up in every PR 😁

@denniskigen 
> Just a thought - is it worth comparing the URL changes for each single-spa-routing-event so we can run the patientUuid update only when a new patient chart gets loaded.

Isn't that effectively what this PR does by adding a guard to check to see if the patientUuid changed?

PR #308, opened 2022-02-04
Update to PostCSS 8

ibacher:
## Requirements

## Summary

This is a simple make-work PR to update to PostCSS 8. There is no real reason this needs to happen, but it's also pretty harmless.


brandones (approved):

brandones:
Thanks Ian! Great to keep up with stuff like this.

brandones:
This has broken the build; esm-styleguide is not building. Reverting.

PR #318, opened 2022-02-11
Fix bug where extensions weren't updating with props

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Remember https://github.com/openmrs/openmrs-esm-core/pull/297 ? I was wrong. Here's the thing:

It is not advisable to put refs in dependency arrays because ref value changes won't trigger re-renders (using "render" in the React sense of a component function run). Change state or props and the component re-renders. But ref values can change arbitrarily and the component won't notice.

However! Once the component *does* re-render, it *is* able to compare the values in the dependency array—to ask, "has this ref value changed since the last time I saw it?" We were relying on that to cause the extensions to re-mount when new props come in. Extensions won't re-render or re-mount *in response to* prop changes. But they will re-mount if they happen to re-render for another reason, and the props have changed.

## Screenshots

props-update


jonathandick (approved):
Thanks Brandon!

PR #325, opened 2022-02-11
O3-1001: Added support for searching person attribute and returning the respective UUID

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
In [O3-1001](https://issues.openmrs.org/browse/O3-1001), we need fields rendered for storing the person-attributes. To make it convenient for the implementor, I have added a search box to search for person-attribute-types, and when clicking a result, it stores the UUID of the selected person-attribute-type.

Works:
- [x] Added a new type in the configuration: PersonAttributeTypeUuid
- [x] For searching concept or person-attribute-type, I have changed the HTML input to carbon component' input field.
- [x] The selected concept/attribute type's UUID is shown above the input box.

## Screenshots
image
image

## Related Issue
https://issues.openmrs.org/browse/O3-1001

## Other
**None**


packages/framework/esm-config/src/types.ts
```diff
@@ -6,6 +6,7 @@ export enum Type {
   Object = "Object",
   String = "String",
   UUID = "UUID",
+  PersonAttributeTypeUuid = "PersonAttributeTypeUuid",
```
brandones, line 9: Please maintain the alphebetization

brandones (approved):
One comment. Generally the code looks great. I'm going to trust you on the UI (it's not like the implementer tools are a carefully crafted surprise-and-delight-the-user experience quite yet).

Really impressed with how well you seem to have gotten set up with this and found your way around in this codebase. Great work.

I imagine we should make a generic UUID search component that is used for concepts and patient attributes. I'm sure there will be plenty more types of UUIDs we will want to support.

packages/apps/esm-implementer-tools-app/src/implementer-tools.styles.scss
```diff
@@ -0,0 +1,14 @@
+@import "~carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/type/styles";
+@import "~carbon-components/src/globals/scss/vendor/@carbon/layout/scss/generated/spacing";
+@import "~@openmrs/esm-styleguide/src/vars";
+
+.popupTriggerButton {
+  height: 25px !important;
+  width: 25px !important;
```
ZacButko, line 7: Avoid `!important` if possible. Instead increase specificity of the element being styled. 
https://uxengineer.com/css-specificity-avoid-important-css/

ZacButko (commented):
Code looks sharp. Nice work @vasharma05

brandones:
Hi @vasharma05 , can we get this in? It's very close. Would be nice to get it in before conflicts come up.

packages/framework/esm-config/src/types.ts
vasharma05: Hi @brandones! In line 4, the `ConceptUuid` is also written in the same CamelCase manner as with other types, so I think it is in accordance with others.
What do you say?

ibacher: @brandones  point is that `PersonAttributeTypeUuid` should be between `Object` and `String`, so the list is in alphabetical order.

PR #326, opened 2022-02-14
(docs) Fill out contributing docs from RFC-20 text

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Rephrased and distilled the text of RFC-20 directly into the "contributing" docs.

### DIFF
--- a/docs/getting_started/contributing.md
+++ b/docs/getting_started/contributing.md
@@ -57,6 +57,24 @@ Remember—small PRs are easier to review anyway!
 
 Get a feel for what the reviewer expects before submitting a PR with a large amount of code in it.
 
+#### Your PR title should indicate the type of change it is.
+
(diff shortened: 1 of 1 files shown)


ZacButko (commented):
Could we please add links to following resources?
- Code style guide
- Filename convention style guide (do we have one?)
- PR format style guide (I think you just added a Github Action to check for this)

brandones:
We don't have a code style guide, other than what is enforced by prettier. We mostly have filename conventions but they're not written down anywhere; I guess I could take a crack at writing them here.

I just filed https://issues.openmrs.org/browse/O3-1109 because someone does need to do that, with the GitHub Action. But yeah I'll add a note about it to these docs.

ZacButko (approved):
Thanks for those changes @brandones . I think this is really helpful

PR #327, opened 2022-02-15
(fix) Bug where usePatient was causing superfluous refetches

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

There was yet another bug where `usePatient` was causing superfluous refetches. Part of the problem was a stale `useEffect` closure: the `useEffect` that creates the `routing-event` listener was missing the `patientUuid` dependency, causing `dispatch` to be called unnecessarily (because the `state.patientUuid` in the closure was out of date and therefore always `!= newPatientUuid`).

I also cleaned up the `useReducer` logic to more adequately represent the state where the patient UUID needs to be loaded from the URL.

## Screenshots

Peek 2022-02-14 21-37


denniskigen (approved):
I like the look of this, @brandones!

PR #332, opened 2022-02-21
03-1125: (bug) fix current visit retaining previous patient visit-data

donaldkibet:
## Requirements

#### For changes to apps

#### If applicable
- [x] My work includes tests, or is validated by existing tests.

## Summary
This PR addresses a number of things

- remove dependency on `rxjs`
- update `useVisit` hook to fix bug where, on searching a different patient, the previous patient `visit` data would still be displayed as the patient's visit. This is seen on the `active-visit-tag.component` on patient chart
- provide mechanism to pass in current visit. useful when entering retrospective data.

## Screenshots
current-visit

## Related Issue
[https://issues.openmrs.org/projects/MF/issues/O3-1125](https://issues.openmrs.org/projects/MF/issues/O3-1125)

### DIFF
--- a/packages/framework/esm-api/src/shared-api-objects/visit-utils.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/visit-utils.ts
@@ -13,19 +13,20 @@ import {
   Visit,
 } from "../types";
 
+export const defaultVisitCustomRepresentation =
+  "custom:(uuid,encounters:(uuid,encounterDatetime," +
+  "form:(uuid,name),location:ref," +
+  "encounterType:ref,encounterProviders:(uuid,display," +
+  "provider:(uuid,display))),patient:(uuid,uuid)," +
+  "visitType:(uuid,name,display),attributes:(uuid,display,value),location:(uuid,name,display),startDatetime," +
+  "stopDatetime)";
+
 export function getVisitsForPatient(
   patientUuid: string,
   abortController: AbortController,
   v?: string
 ): Observable<FetchResponse<{ results: Array<Visit> }>> {
-  const custom =
-    v ||
-    "custom:(uuid,encounters:(uuid,encounterDatetime," +
-      "form:(uuid,name),location:ref," +
-      "encounterType:ref,encounterProviders:(uuid,display," +
-      "provider:(uuid,display))),patient:(uuid,uuid)," +
-      "visitType:(uuid,name,display),attributes:(uuid,display,value),location:(uuid,name,display),startDatetime," +
-      "stopDatetime)";
+  const custom = v ?? defaultVisitCustomRepresentation;
 
   return openmrsObservableFetch(
     `/ws/rest/v1/visit?patient=${patientUuid}&v=${custom}`,
--- a/packages/framework/esm-react-utils/src/useVisit.ts
+++ b/packages/framework/esm-react-utils/src/useVisit.ts
@@ -1,43 +1,34 @@
-import { openmrsFetch, Visit } from "@openmrs/esm-api";
+import {
+  defaultVisitCustomRepresentation,
+  openmrsFetch,
+  Visit,
+} from "@openmrs/esm-api";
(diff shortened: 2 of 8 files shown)


packages/framework/esm-api/src/shared-api-objects/visit-utils.ts
```diff
@@ -27,22 +27,9 @@ export function getVisitsForPatient(
...
-      },
-    }
-  )
-    .pipe(take(1))
-    .pipe(
-      map((response: FetchResponse<{ results: Array<Visit> }>) => {
-        return response;
-      })
-    );
+  return openmrsFetch(`/ws/rest/v1/visit?patient=${patientUuid}&v=${custom}`, {
```
brandones, line 30: I think we should switch this to use SWR. CC @denniskigen

denniskigen: Agreed

packages/framework/esm-react-utils/src/useVisit.ts
```diff
@@ -1,49 +1,121 @@
...
+    const handleVisitUpdate = (evt: CustomEvent) => {
+      const { type, state = {} } = evt.detail;
+      dispatch({ type: ActionTypes.loadVisit, visit: state.visit });
+    };
+    window.addEventListener("visit-change", handleVisitUpdate);
+    return () => window.removeEventListener("visit-change", handleVisitUpdate);
   }, [patientUuid]);
 
-  return { currentVisit, error };
+  const updateCurrentVisit = (visit: Visit | null) => {
```
brandones, line 99: As noted [in the other PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/570/files#r812276481), this should probably go outside the hook. Also, please add a docstring to clarify its intended usage.

packages/framework/esm-react-utils/src/useVisit.ts
```diff
@@ -1,49 +1,43 @@
...
-        const currentVisit = data.results.find(
-          (visit) => visit.stopDatetime === null
-        );
+const defaultVisitCustomRepresentation =
+  "custom:(uuid,encounters:(uuid,encounterDatetime," +
+  "form:(uuid,name),location:ref," +
+  "encounterType:ref,encounterProviders:(uuid,display," +
+  "provider:(uuid,display))),patient:(uuid,uuid)," +
+  "visitType:(uuid,name,display),attributes:(uuid,display,value),location:(uuid,name,display),startDatetime," +
+  "stopDatetime)";
```
brandones, line 18: Can you import this from `esm-api`, so that `useVisit` stays in sync with `getVisitsForPatient`?

packages/framework/esm-react-utils/src/useVisit.ts
```diff
@@ -1,49 +1,43 @@
...
+/**
+ * This React hook returns a visit object. If the `patientUuid` is provided
+ * as a parameter, then the visits, currentVisit, error and mutate function
+ * for that patient visit is returned.
+ * @param patientUuid Unique patient identifier `string`
+ * @param includeInactive Boolean that specifies whether to include Inactive visits `boolean`, defaults to false
+ * @returns Object {`visits`, `error` `isValidating`, `currentVisit`, `mutate`}
+ */
+export function useVisit(
```
brandones, line 24: Now that this returns `visits`, the singular is confusing. I'd advocate two functions—`useVisits` which has an `includeInactive` parameter that defaults `true`, and `useVisit` which has no `includeInactive` parameter (it always uses `false`). If we have a guarantee from the server that there is always only one active visit, then `useVisit` should not return the `visits` object at all (if we have no such guarantee then it is probably best to include it so that it can be handled at the application level).

Sorry this is such a process, but I think it pays to get core APIs right.

brandones (approved):
Looks good, thanks @donaldkibet !

PR #335, opened 2022-02-23
Simplify extension system again

brandones:
https://github.com/openmrs/openmrs-esm-core/pull/252 lives again.

## Requirements

## Summary
Copy-pasing from before:

> This simplifies the extension system. Previously, the extension system accommodated multiple slots having the same name. This was not part of the [implementation plan](https://github.com/openmrs/openmrs-rfc-frontend/blob/master/text/0027-extensions.md) and has no use case that isn't more simply implemented by using attach. This change removes the possibility of multiple slots having the same name, establishing a 1:1 correspondence between slots and names.
> 
> This also splits the extension store into an "internal" store and an "output" store, much the way the config stores are handled. This will enable module authors to get information about the extension system state while still permitting the extension system implementation to evolve.
>
> This will make the implementation of the Link System much easier.

It is unchanged (with the exception of a few lines that I will comment) since before.

This time, workspace problems will not occur, since the workspace is now totally independent of the extension system.

Everything using the patient header is working. However, I think the patient header needs to be refactored anyway, [as discussed here](https://openmrs.slack.com/archives/CHP5QAE5R/p1641931093043900?thread_ts=1641914260.035600&cid=CHP5QAE5R).


packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -43,7 +43,7 @@ const openmrsConfigUrls = (process.env.OMRS_CONFIG_URLS || "")
   .map((url) => JSON.stringify(url))
   .join(", ");
 
-function checkDirectory(dirName) {
+function checkDirectoryExists(dirName) {
```
brandones, line 46: This is the only file that has changed since #252 . Previously it was only checking if the `dist` dir exists. Now it checks whether it has contents.

ZacButko (approved):
OK

PR #346, opened 2022-03-07
O3-1145: Person attribute labels should be loaded from the backend

vasharma05:
This PR is an extension to #325.
## Requirements

#### For changes to apps

#### If applicable

## Summary
In the work done in #325, the user can fetch the UUID of the person-attribute-type from the backend but filled the name of the attribute-type manually. In this PR, I have made changes such that when a user fetches the attribute type by UUID from the backend, the name field will also be filled from the backend's response.

## Screenshots
https://www.loom.com/share/2e2ecd6d5c1e4f19b61c84829971d867

## Related Issue
https://issues.openmrs.org/browse/O3-1143

### DIFF
--- a/packages/apps/esm-implementer-tools-app/src/configuration/value-editors/person-attribute-search.tsx
+++ b/packages/apps/esm-implementer-tools-app/src/configuration/value-editors/person-attribute-search.tsx
@@ -6,13 +6,24 @@ import styles from "./uuid-search.scss";
 import { useTranslation } from "react-i18next";
 import {
   Search,
+  StructuredListBody,
   StructuredListCell,
   StructuredListRow,
   StructuredListWrapper,
 } from "carbon-components-react";
+import { ValueEditorField } from "./value-editor-field";
+import { DisplayValue } from "../display-value";
+import { Type } from "@openmrs/esm-framework";
+import { ConceptSearchBox } from "./concept-search";
 
+interface PersonAttributeType {
+  name: string;
+  uuid: string;
+  type: string;
+  concept: any;
+}
 interface PersonAttributeTypeSearchBoxProps {
-  value: string;
+  value: PersonAttributeType;
   setPersonAttributeUuid: (personAttributeType) => void;
 }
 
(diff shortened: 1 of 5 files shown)


vasharma05:
Hi @ibacher!
I was a bit concerned about the new change. I had to change 2 fields(name and UUID) from a single input, hence I thought of adding all the fields in the person-attribute's search component in the implementor tools, as is done in packages/apps/esm-implementer-tools-app/src/configuration/value-editors/person-attribute-search.tsx. I just wanted to know if there'll be any fallout of this method in the future?
Thanks!

ibacher:
@vasharma05 I think that in light of [O3-1143](https://issues.openmrs.org/projects/O3/issues/O3-1143), I think it would be better if we refactored things to just not capture the name for the attribute type field at all or at least not as a separate element. Instead, we can change things so that this works very similar to how the `ConceptSearchBox` component already works, i.e., we're searching based on name, but storing a UUID value.

The rationale being that way we can load the attribute type name in the current user's locale rather than the implementors locale.

packages/apps/esm-implementer-tools-app/src/configuration/value-editors/value-editor-field.tsx
```diff
@@ -53,8 +53,8 @@ export function ValueEditorField({
   ) : valueType === Type.PersonAttributeTypeUuid ? (
     <PersonAttributeTypeSearchBox
       value={value}
-      setPersonAttributeUuid={(personAttributeTypeUuid) =>
-        onChange(personAttributeTypeUuid)
+      setPersonAttributeUuid={(personAttributeType) =>
+        onChange(personAttributeType)
```
ZacButko, line 57: This can be reduced to `setPersonAttributeUuid={onChange}`

packages/apps/esm-implementer-tools-app/src/configuration/value-editors/person-attribute-search.tsx
```diff
@@ -103,7 +118,8 @@ export function PersonAttributeTypeSearchBox({
                 key={personAttributeType.uuid}
                 role="option"
                 onClick={() => {
-                  handleUuidChange(personAttributeType);
+                  handleChange("uuid", personAttributeType.uuid);
+                  handleChange("name", personAttributeType.display);
```
ZacButko, line 122: This does two synchronous updates, which also calls resetSearch twice. In this case it doesn't look like it will cause many re-renders, but in more complex cases it could be noticeable. Your `handleChange` function could be restructured as below in order to handle updating multiple keys at the same time

`handleChange() = newData -> {
   setAttributes((prevAttributes) => ({
     ...prevAttributes,
     ...newData
   })
)`

ZacButko (changes requested):
Code looks OK, but I agree with @ibacher we should store UUID instead of name. We should also fix the missing useEffect dependency before mergin.

vasharma05:
After a talk with @ibacher, the requirements for ticket O3-1143 were different than the work done in this PR, hence I am closing this PR. 
I thank the developers to devote their time reviewing this PR.

PR #352, opened 2022-03-10
(chore) Fix CI build for pre-release

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Moves the update to the esm-framework version to the postversion script instead of the version script when $npm_package_version should be the version that the package is being updated to instead of the version it is currently set as.


donaldkibet (approved):
Thanks! @ibacher

PR #355, opened 2022-03-11
Fixes incorrect link to manifest in index.html

mseaton:
## Requirements

#### For changes to apps

## Summary

Loading any MFE results in multiple errors in the server log and in the browser console related to the inability to load the manifest.json file.  Researching this, indeed the generated index.html is referencing "auto/manifest.xxx.json", whereas the file itself is not located in an "auto" directory, but in the same directory as "index.html".

Based [on reading here](https://stackoverflow.com/questions/65245185/new-to-webpackautoprefix-problem-with-webpack-manifest-plugin), the intended fix is to add output.publicPath: "" to the webpack configuration.

## Screenshots
In the server log, this appears like the following:

SpaServlet.getFile(166) |2022-03-11 14:32:21,589| File with path '/home/mseaton/openmrs/mirebalais/frontend/auto/manifest.d7f50154f66448287167e3a19324d47a.json' doesn't exist

This can easily be seen by loading an existing instance and viewing the JS console.

image

### DIFF
--- a/packages/shell/esm-app-shell/webpack.config.js
+++ b/packages/shell/esm-app-shell/webpack.config.js
@@ -98,6 +98,7 @@ module.exports = (env, argv = {}) => {
       filename: "openmrs.js",
       chunkFilename: "[chunkhash].js",
       path: resolve(__dirname, outDir),
+      publicPath: "",
     },
     target: "web",
     devServer: {


ibacher (approved):
Thanks Mike!

PR #357, opened 2022-03-14
(fix) App not hot updating with slot config changes

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Fixes a bug where the app does not hot update in response to extension slot config changes.

Includes a breaking change to `useStore` and `useStoreState`, but this will not affect any existing users.

## Screenshots

remove-ext


packages/framework/esm-config/src/module-config/module-config.ts
```diff
@@ -125,9 +125,17 @@ function computeExtensionSlotConfigs(
...
-    slotStore.setState({ loaded: true, config });
+  const newSlotStoreEntries = Object.fromEntries(
+    Object.entries(slotConfigs).map(([slotName, config]) => [
+      slotName,
+      { loaded: true, config },
+    ])
+  );
+  const slotStore = getExtensionSlotsConfigStore();
+  const oldState = slotStore.getState();
+  const newState = { slots: { ...oldState.slots, ...newSlotStoreEntries } };
```
ZacButko, line 136: This will not remove oldState slots if they are meant to be removed. Is it desired to remove old unused slots from state?

packages/framework/esm-react-utils/src/useStore.ts
```diff
@@ -1,11 +1,58 @@
...
+  }
+
+  return bound;
+}
+
+const defaultSelectFunction = (x) => x;
+
+function useStore<T, U>(store: Store<T>): T;
+function useStore<T, U>(store: Store<T>, select: (state: T) => U): U;
+function useStore<T, U>(
```
ZacButko, line 24: What are all of these declarations for? You're just declaring a function then immediately overwriting it?

packages/framework/esm-config/src/module-config/module-config.ts
ibacher: I asked @brandones a similar question once and he indicated the idea is to keep the slots around so that the implementer tools can track them even if the slot itself is no longer rendered.

brandones: @ibacher is correct. There will not be so many that it will be a memory problem.

packages/framework/esm-react-utils/src/useStore.ts
brandones: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

ZacButko (approved):

PR #358, opened 2022-03-15
(feat) Add support for extension config schemas

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Currently, extensions inherit the configuration schemas of the modules in which they are defined. Some extensions (especially generic ones) have highly specific configuration requirements, which pollute the module config schema: see [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/d8cd49eb90747d05fbc1dbdd38282f36ad52d036/packages/esm-patient-chart-app/src/config-schema.ts#L35) in patient-chart.

This makes it so that extensions can have their own schema, apart from the module schema.

Unrelatedly, this removes some dead code. It is not used anywhere.


ibacher (approved):
LGTM! I really like this feature!

PR #359, opened 2022-03-15
(fix) O3-1173: Patient Chart fails to load on tablet

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This is fixed by allowing metas to be fetched for extensions that are not mounted yet. I can't think of any real reason not to allow that.

## Related Issue

https://issues.openmrs.org/browse/O3-1173

### DIFF
--- a/packages/framework/esm-extensions/src/extensions.ts
+++ b/packages/framework/esm-extensions/src/extensions.ts
@@ -40,19 +40,16 @@ function updateExtensionOutputStore(
 ) {
   const slots: Record<string, ExtensionSlotState> = {};
   for (let [slotName, slot] of Object.entries(internalState.slots)) {
-    // Only include registered slots
-    if (slot.moduleName) {
-      const { config } = getExtensionSlotConfigFromStore(
-        extensionSlotConfigs,
-        slot.name
-      );
-      const assignedExtensions = getAssignedExtensionsFromData(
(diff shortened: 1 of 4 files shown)


packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -40,19 +40,16 @@ function updateExtensionOutputStore(
 ) {
   const slots: Record<string, ExtensionSlotState> = {};
   for (let [slotName, slot] of Object.entries(internalState.slots)) {
-    // Only include registered slots
-    if (slot.moduleName) {
```
ibacher, line 44: Out of curiosity, do you have any idea why `slot.moduleName` apparently isn't set in tablet mode?

ibacher (approved):

packages/framework/esm-extensions/src/extensions.ts
brandones: This is actually an important point—it's that `slot.moduleName` gets set when the slot is registered. The slot gets registered when the `ExtensionSlot` component gets mounted. It does not get mounted in tablet mode because there is no left nav in tablet mode.

ibacher: Ah! Yes, that makes sense.

PR #360, opened 2022-03-15
(feat) Config should be able to override meta

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This allows meta to be overridden by configuration. This will make it possible to define dashboards in configuration by configuring a generic `dashboard` extension.

## Other

This raises the question—should we replace `meta` with `config` as a concept? Currently, the structure of things is that "meta" is an extension's way of telling slots how it should be used, while "config" tells the extension how to behave. But I don't think it would be unnatural for "an extension's config" to both tell the extension how to behave and to tell the slot how it should be used.

The disadvantage is that the meta would have to be included in the extension schema, rather than specified simply as an object in the extension declaration. So altogether we could expect things to look kind of like this:

In `index.ts` files:

```ts
defineExtensionConfigSchema(
  'patient-summary-dashboard', 
  getDashboardConfigSchema({
    name: "summary" ,
    title:  "Patient Summary",
    slot: "patient-chart-summary-dashboard-slot",
    columns: 4
  })
);

setupOpenMRS() {
  return {
    extensions: [
      {
        name: 'patient-summary-dashboard',
        slot: 'patient-chart-dashboard-slot',
        order: 5,
        load: getSyncLifecycle(createDashboardLink(), { featureName: 'patient summary', moduleName }),
        online: true,
        offline: true,
      },
    ]
  }
}

function getDashboardConfigSchema({ name, title, slot, columns }) {
  return  {
    name: { _type: Type.String, _default: name },
    title: { _type: Type.String, _default: title },
    slot: { _type: Type.String, _default: slot },
    columns: { _type: Type.Number, _default: columns }
  }
}
```

At [chart-review.component.tsx:32](https://github.com/openmrs/openmrs-esm-patient-chart/blob/9bd7a8df9b8ae5757ec90c454571d4da1d4fda5e/packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx#L32):

```ts
  const ungroupedDashboards = extensionStore.slots['patient-chart-dashboard-slot'].assignedExtensions.map(
    (e) => e.config,
  );
  const groupedDashboards = navGroups
    .map((slotName) => extensionStore.slots[slotName].assignedExtensions.map((e) => e.config))
    .flat();
```

One reason this is nice is that then there would be more straightforward to allow configuration through the implementer tools. As things are now, the implementer tools would have to look at the extension meta and try to infer the types of the elements there (which determines what kind of input widget is offered to the user).


ibacher:
In principle, I like the idea of having stuff typed and giving implementers more control over things without the need to write actual code. That said, there are some points about this proposal I'm not keen on. First, I think we should be restricting the configuration system to things that implementers can meaningfully configure and I'm not sure everything in `meta` meets that case, e.g., what does it mean to overwrite `slot`?

Rather than replacing or overwriting `meta` with configuration, it might make sense to either somewhat limit the overwrite (e.g., from the example, maybe we just don't allow overrides to `slot` but allow the other properties to be overwritten) or just keep `meta` around for that limited set of values that are definitely only metadata and push everything else into config.

Second, and this is a more minor quibble, this effectively makes certain configuration names unusable. E.g., I can't have a property called `title` in my configuration unless I'm intending it to exactly match the extension title.

brandones:
Those are good points @ibacher . I wonder if the right move is simply to make config available on the slot side in the same way that `meta` is. So places that use meta could also check to see if there's an overriding value in the config and use that if available. What do you think?

```ts
  const ungroupedDashboards = extensionStore.slots['patient-chart-dashboard-slot'].assignedExtensions.map(
    (e) => getDashboardDefinition(e.meta, e.config),
  );
  const groupedDashboards = navGroups
    .map((slotName) => extensionStore.slots[slotName].assignedExtensions.map((e) => getDashboardDefinition(e.meta, e.config)))
    .flat();

function getDashboardDefinition(meta, config) {
  return {  ...meta, ...config };
  // Vut actually doing something more complicated so that only the expected properties from config get included, etc.
  // The point is, it's up to the extension consumer to implement this logic.
}
```

ibacher:
> What do you think?

I actually think that's exactly the type of API I was hoping for here!

brandones:
Closing in favor of https://github.com/openmrs/openmrs-esm-core/pull/366

PR #368, opened 2022-03-23
(bugfix) Client Redirect from '/' to '/home'

ZacButko:
## Requirements

#### For changes to apps
(N/A)

#### If applicable
(N/A)

## Summary
Solves ticket [OS-782]( https://issues.openmrs.org/browse/O3-782) from Client side
This is an additional safeguard for serverless setups. Primary fix done by @ibacher [here](https://github.com/openmrs/openmrs-contrib-ansible-docker-compose/commit/fb39323ab29b70d2c575fa0786165b067a13dc12)

## Screenshots
No UI Changes

## Related Issue
[OS-782]( https://issues.openmrs.org/browse/O3-782)


packages/apps/esm-primary-navigation-app/src/index.ts
```diff
@@ -42,6 +44,14 @@ function setupOpenMRS() {
         offline: true,
         order: 0,
       },
+      {
+        load: getSyncLifecycle(() => HomeRedirect(), options),
```
brandones, line 48: I think you could actually just write the single-spa lifecycle object, something like

```ts
load: {
  bootstrap: () => { navigateToUrl(window.getOpenmrsSpaBase() + "home"); },
  mount: () => {},
  unmount: () => {}
}
```

packages/apps/esm-primary-navigation-app/src/index.ts
```diff
@@ -42,6 +44,14 @@ function setupOpenMRS() {
...
+        route: (location: Location) =>
+          location.pathname === window.getOpenmrsSpaBase(),
+        online: true,
+        offline: true,
+        order: 0,
```
brandones, line 53: This doesn't have meaning here, and is also the default value

packages/apps/esm-primary-navigation-app/src/index.ts
ZacButko: pretty close. These need to be promises

packages/apps/esm-primary-navigation-app/src/index.ts
ZacButko: Order apparently is necessary, otherwise it crashes 🙄

brandones: Oh really??? That's a bug

PR #369, opened 2022-03-23
(fix) Service Worker Housekeeping: Remove Unused SW Events | Fix SW Startup Cycle During Initial Load

manuelroemer:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Housekeeping for the service worker. Removes a now unused event (`networkRequestFailed` - was only introduced for `esm-form-entry` and is now not required anymore). Also improves the startup and SW registration (which should issues with initial caching of resources).


FlorianRappl (approved):

brandones:
Hi @manuelroemer , do you think this will fix https://issues.openmrs.org/browse/O3-1067 , or is it unrelated?

PR #371, opened 2022-03-26
Search locations in the location picker not only on the basis of prefixes

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Currently, when we look for the "Pharmacy" location, we might search with "P, Pha, ...", but searching with "macy" won't yield the required results. Hence I added `:contains` to the name parameter.

## Screenshots
image

## Related Issue
https://issues.openmrs.org/browse/O3-1193


ZacButko (approved):
Thanks Vineet!

rbuisson (approved):

PR #372, opened 2022-03-28
(fix) Loosen UUID validation

ibacher:
## Requirements
- [ ] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.

#### For changes to apps

#### If applicable

## Summary

I'm not sure why these rules were in place but they seem incorrect

### DIFF
--- a/packages/framework/esm-config/src/validators/type-validators.ts
+++ b/packages/framework/esm-config/src/validators/type-validators.ts
@@ -27,7 +27,7 @@ export const isObject = validator(
 
 export const isUuid = validator(
   (val) =>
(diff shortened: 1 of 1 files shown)


brandones:
@ibacher See https://stackoverflow.com/a/13653180/1464495

ibacher:
@brandones Yep. I opened this because not all UUIDs in existing OMRS installations are correctly generated RFC 4122 UUIDs; some are just string that look a lot like them. If it helps, this regex does match the ABNF grammar given for string representations of UUIDs in RFC 4122.

brandones:
I sort of suspected that might be the case. Thanks @ibacher .

PR #373, opened 2022-03-29
BREAKING: (feat) Allow inversion of useOnClickOutside

jnsereko:
## Requirements

#### For changes to apps

#### If applicable

## Summary

 There was need to create an inversion of [useOnClickOutside](https://github.com/openmrs/openmrs-esm-core/blob/master/packages/framework/esm-react-utils/src/useOnClickOutside.tsx) to handle other components on Dom beyond the referenced component.  Please fill free to checkout https://openmrs.slack.com/archives/CHP5QAE5R/p1648509416725489
 
 I needed this functionality when i was working with https://github.com/openmrs/openmrs-esm-patient-management/pull/136 
cc @ibacher @brandones @ZacButko

### DIFF
--- a/packages/framework/esm-react-utils/src/useOnClickOutside.tsx
+++ b/packages/framework/esm-react-utils/src/useOnClickOutside.tsx
@@ -2,7 +2,8 @@ import { useRef, useEffect } from "react";
 
 export function useOnClickOutside<T extends HTMLElement = HTMLElement>(
   handler: (event: Event) => void,
-  active = true
+  active = true,
+  invert = false
 ) {
   const ref = useRef<T>(null);
 
@@ -11,7 +12,7 @@ export function useOnClickOutside<T extends HTMLElement = HTMLElement>(
       const listener = (event: Event) => {
         const el = ref?.current;
 
-        if (el?.contains(event.target as Node)) {
+        if (el?.contains(event.target as Node) === !invert) {
           handler(event);
         }
       };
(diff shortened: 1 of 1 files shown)


ibacher (approved):

ibacher:
So in e84e2dd, which was when `useOnClickOutside()` was extracted from the primary-nav-app, a transcription error occurred so that:

```ts
if (!el || el.contains(event.target as Node)) {
  return;
}
handler(event);
```

Became:

```ts
const el = ref?.current;

if (el?.contains(event.target as Node)) {
  handler(event);
}
```

Obviously, this is a bit of an error as it changes the logic from "run this handler every time any element outside of this element is clicked on" to "run this handler any time this element or one of it's children are clicked on". We really don't need any complicated setup for that (normal DOM bubbling takes care of that), so I've updated this commit to re-implement the original logic.

I've marked this as a BREAKING change because it does completely invert the functionality of this component; however, I suspect that this breaking change will actually fix existing buggy behaviour.

brandones:
Agreed, thank you @ibacher and @jnsereko .

@jnsereko , can you please write a test that verifies this change?

jnsereko:
> Agreed, thank you @ibacher and @jnsereko .
> 
> @jnsereko , can you please write a test that verifies the corrected behavior? It should fail for the old implementation and succeed for the new implementation.

hello @brandones 
I have created a test PR for this component https://github.com/openmrs/openmrs-esm-core/pull/379 and i kindly request for a review

PR #378, opened 2022-04-01
CurrentUserContext should always render children

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Prior to this commit, the child nodes of the `CurrentUserContext` only render once loading is finished. However, in some circumstances, the `session` endpoint will return a 401 error instead of the response (e.g., when the user presents with a session cookie for an expired backend session). In these circumstances, the login screen will not render.

Since the `CurrentUserContext` is only used in the login app 🤞, it should be harmless to ensure it's children are always rendered.

Not 💯 convinced I haven't missed something, though.


ibacher:
@denniskigen FYI, I think this is the fix for the issue you mentioned

jonathandick (approved):

brandones (approved):

PR #379, opened 2022-04-03
(test): useOnClickOutside test

jnsereko:
## Requirements

#### For changes to apps

#### If applicable

### DIFF
--- a/packages/framework/esm-react-utils/src/useOnClickOutside.test.tsx
+++ b/packages/framework/esm-react-utils/src/useOnClickOutside.test.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import { render, fireEvent, screen } from "@testing-library/react";
+import { useOnClickOutside } from "./useOnClickOutside";
+
+describe("useOnClickOutside", () => {
+  const handler = jest.fn();
+  const ref = useOnClickOutside<HTMLDivElement>(handler);
+  const activeRef = useOnClickOutside<HTMLDivElement>(handler, false);
+  const buttonText = "Button Text";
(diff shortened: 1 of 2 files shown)


jnsereko:
@brandones @ibacher 
This is failing.  It is complaining `Invalid hook call. Hooks can only be called inside of the body of a function component` 
at https://github.com/openmrs/openmrs-esm-core/blob/master/packages/framework/esm-react-utils/src/useOnClickOutside.tsx#L7

image

How can i solve this?  Do i need to  pass the `ref` as a parameter instead of declaring it using the hook(which is the cause of the error)

ibacher:
The simplest solution is probably just to leverage the react-hooks testing library: https://react-hooks-testing-library.com/, which is made to solve this exact problem.

nanfuka:
you could pass that hook in the component where you used that function **useOnClickOutside**. then in the function just pass that hook as a property...

packages/framework/esm-react-utils/src/useOnClickOutside.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  test("should call handler() when clicking outside", async () => {
+    expect(handler).not.toHaveBeenCalled();
+
+    await act(async () => {
+      render(<div ref={ref as React.MutableRefObject<HTMLDivElement>}></div>);
+      fireEvent.mouseDown(document.body);
+    });
+
+    expect(handler).toHaveBeenCalledTimes(1);
```
jnsereko, line 20: hey @ibacher 
This line is failing.  its like `useOnClickOutside` is not called

PR #383, opened 2022-04-05
(fix) Precache importmap as part of the static dependency lifecycle | Remove obsolete XMLHttpRequest patches

manuelroemer:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Precaches the app's importmap references as part of the static dependency lifecycle. This has multiple advantages: In the future when we introduce a "Refresh cache" feature, these refs will be updated. It further prevents some issues where currently the system tried to cache deps while offline (which obviously fails).
In addition, I remove some old patches which have been added to make the network request interception of the service worker work with the esm-form-entry-app module. These are no longer required since the approach changed.


FlorianRappl (approved):
LGTM - thanks!

PR #386, opened 2022-04-06
(fix) Consider relative URLs when scanning the importmap for assets.

manuelroemer:
## Requirements

#### For changes to apps

## Summary
Add support for import map URLs relative to the service worker. This should fix an issue with files not being correctly precached on app load  in environments which don't use absolute URLs in the import map.


ibacher:
Shouldn't this use `window.spaBase` as the base URL?

manuelroemer:
@ibacher Fair question! The reason for using `self.location` here is that the script is running in the service worker context where `window.spaBase` is `undefined`. The service worker is always registered at `{spaBase}/service-worker.js`, meaning that `self.location` will implicitly be based upon the spa base value (and can thus be used as a substitute).

ibacher (approved):

PR #389, opened 2022-04-08
(feat) move from babel -> swc

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This moves us from using Babel to using SWC for transpiling. Using SWC gives us a minor speed boost in compile times ~5 minutes with Babel vs ~4 minutes with SWC to build esm-core on my machine. The real target here, however, is speeding up the app shell build step (i.e. `npx openmrs build`). On my machine, this gives us a pretty good speed bump (from ~70 seconds to ~47 seconds; this isn't a rigorous metric the reported values are the medians of three runs).

Some other notes:

I did a first pass of this using esbuild. Esbuild actually performs somewhat better on my machine, shaving a few extra seconds off the full build. However, I couldn't get esbuild working with Jest (I didn't spend too much time trying; I'm sure its possible). And a couple of seconds didn't seem worth it.

This also replaces babel with the TypeScript eslint parser, mostly so I could get rid of the Babel configuration files.


brandones (approved):
Cool, thanks @ibacher ! Following @denniskigen 's lead, I'm also inclined to have this repo use turborepo as well.

PR #393, opened 2022-04-11
(chore) Re-add enough Babel that webpack-config works

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Originally with the SWC migration, I tried to avoid migrating the webpack-config tooling, thinking this would enable us to avoid having to update downstream MFs. Unfortunately, this turns out not to be the case, so this PR migrates webpack-config. This does mean that other repos will need to be updated to use SWC when they are updated to the latest version, but at least this update will be consistent.

### DIFF
--- a/packages/tooling/openmrs/package.json
+++ b/packages/tooling/openmrs/package.json
@@ -32,6 +32,7 @@
   ],
   "homepage": "https://github.com/openmrs/openmrs-esm-core#readme",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


brandones:
Thanks @ibacher ! Will leave merging to you so that you can time it with the downstream repos.

ZacButko (approved):

PR #398, opened 2022-04-14
(fix) Don't crash when a slot is registered before its extensions

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

If a slot has an attachment and is registered before an extension to which it is attached, it should not crash. Rather, the slot should simply not receive the unregistered extension.

## Other

This comes up when [creating dashboards via configuration](https://github.com/openmrs/openmrs-esm-patient-chart/pull/653).


ibacher (approved):
Thank you for also cleaning up those files I missed!

PR #402, opened 2022-04-19
(feat) Offline Registered Patients in Offline Tools Card | Sync Queue Functions for Retrieving Full Sync Item | Sync Queue Manipulation Tests 

manuelroemer:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Considers offline registered patients in the offline patients card of the offline tools. As follow-up work, the offline patient table will be updated to also display such patients. This requires the addition of two new API members in `esm-offline`: Functions to retrieve the full sync item (and not only the content). I used this chance to also start out with some tests for `esm-offline`. They currently cover the functions manipulating the sync queue.


ZacButko (approved):
OK

PR #415, opened 2022-04-27
Add `age` function to esm-framework/mock

donaldkibet:
## Requirements

#### For changes to apps

#### If applicable

## Summary
- Add `age` mock function to `esm-framework/mock` file

### DIFF
--- a/packages/framework/esm-framework/mock.tsx
+++ b/packages/framework/esm-framework/mock.tsx
@@ -8,6 +8,7 @@ export {
   formatDate,
   formatDatetime,
   formatTime,
+  age,
 } from "@openmrs/esm-utils";
 export { interpolateString, interpolateUrl } from "@openmrs/esm-config";
 


packages/framework/esm-framework/mock.tsx
```diff
@@ -292,6 +292,8 @@ export const useVisit = jest.fn().mockReturnValue({
 
 export const useVisitTypes = jest.fn(() => []);
 
+export const age = jest.fn();
```
brandones, line 295: I think this one we can just pass through from the actual implementation. See the exports from `esm-util` up at the top, lines 6-11.

Also if you were adding this it should go under the `/* esm-utils */` comment anyway, not here with `esm-react-utils`.

brandones (changes requested):

brandones (approved):

PR #423, opened 2022-05-06
O3-1242  App crashes on logout

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This fixes the broken logout functionality. It also fixes a problem where clicking "Change location" from the user menu in the primary navigation bar would lead to an "An error has occurred. Please try reloading the page." page.

The general idea of the fix is to make it so that `useSession` uses Suspense, so that it always either throws or returns a valid session object—it can no longer return `null`.

There is a substantial bit of refactoring worked in. I changed `getCurrentUser` to use Unistore rather than RxJS for state management, mostly because I have a much better understanding of how Unistore works, and am confident that it will behave synchronously when I expect it to (the behavior of Observables is ambiguous in this regard).

I did my best to document the hook's implementation, since its complexity is increased substantially.

## Screenshots

Peek 2022-05-05 16-45

## Related Issue

https://issues.openmrs.org/browse/O3-1242

## Other

Ideally, this should be tested by E2E tests. CC @jayasanka-sack .


packages/tooling/webpack-config/src/index.ts
```diff
@@ -163,6 +163,9 @@ export default (
       headers: {
         "Access-Control-Allow-Origin": "*",
       },
+      devMiddleware: {
+        writeToDisk: true,
+      },
```
brandones, line 168: Unrelated: this makes it so that `yarn run:shell` plays nicer with `yarn serve`ing the apps.

jayasanka-sack:
> Ideally, this should be tested by E2E tests. CC @jayasanka-sack .

Noted with thanks! 😊

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
```diff
@@ -1,26 +1,25 @@
...
-import { useCurrentUser } from "../CurrentUserContext";
 
 export interface RedirectLogoutProps extends RouteComponentProps<{}> {
   isLoginEnabled: boolean;
 }
 
 const RedirectLogout: React.FC<RedirectLogoutProps> = ({ isLoginEnabled }) => {
   const config = useConfig();
-  const user = useCurrentUser();
+  const user = useSession();
```
ZacButko, line 12: This is confusing naming since I think (as seen in `choose-location.component.tsx`..) `{ user }` is also available in `useSession`.. did you mean to set this to `{ user }` instead? or should this be called `const session = useSession()` ?

packages/apps/esm-login-app/src/root.component.tsx
```diff
@@ -3,39 +3,34 @@ import Login from "./login/login.component";
...
-          render={(props) => (
-            <RedirectLogout {...props} isLoginEnabled={isLoginEnabled} />
-          )}
-        />
-      </BrowserRouter>
-    </CurrentUserContext>
+    <BrowserRouter basename={window.spaBase}>
+      <Route
+        exact
+        path="/login(/confirm)?"
```
ZacButko, line 16: Just checking, I haven't seen this syntax from react-router-dom before... doesn't look like plain text and doesn't look like regex?

ZacButko (approved):
Thanks for cleaning this Brandon

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
brandones: Oops, that's an error. Good catch, thanks!

packages/apps/esm-login-app/src/root.component.tsx
brandones: I didn't write this, but it looks like regex to me. Parentheses for grouping and a question mark for "optional;" i.e. this will match `/login` or `/login/confirm`.

PR #424, opened 2022-05-10
BREAKING: O3-1267  Simplify module registry and import system

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This simplifies the module loading system. It has been using a blend of Webpack Module Federation and SystemJS which produces some bugs. This switches us onto using only WMF for module loading. SystemJS is maintained only for parsing the import map.

This also removes the requirement that frontend modules in the import map be suffixed with `-app` or `-widgets` to be loaded. It is unnecessary, since the only things in the import map are frontend modules.

I don't expect anyone to really understand this deeply. I struggled with this for two weeks and had two separate consultations with the inventor of Module Federation. I have done my best to convey my limited understanding of it in the comments.

## Screenshots

No UI changes.

## Related Issue

https://issues.openmrs.org/browse/O3-1267

## Other

This is the most complete kind of breaking change: modules compiled with the webpack configuration before and after this change will be utterly incompatible. This probably warrants a major-versioning of esm-core.

When this is merged, we will need to update all the other repos immediately.


packages/shell/esm-app-shell/src/load-modules.ts
```diff
@@ -0,0 +1,76 @@
...
+ * the application is `@openmrs-esm-app-shell`. The list of remotes
+ * to be loaded is supplied in import maps, which are loaded by
+ * SystemJS. We are not using the SystemJS registry at all, only WMF.
+ *
+ * Each module ("remote") is provided as a name and URL. For each URL,
+ * a `<script>` element is appended to the DOM. Because those scripts have webpack
+ * library type == 'var', when they are loaded, they create a new
+ * global variable which has the Webpack Container Interface for that
+ * module: the `init` and `get` methods. We use those to obtain the
+ * actual module exports.
```
ZacButko, line 16: 👍

ZacButko (approved):

ibacher:
Thank you so much for this @brandones!

PR #427, opened 2022-05-11
(fix) Prevent app crashes in certain browser environments where IndexedDB is not working properly

manuelroemer:
## Requirements

#### For changes to apps

#### If applicable

## Summary
The following addresses errors which were very annoying to debug as they were thrown in arbitrary locations without any kind of stack trace attached:

In the current state, the app can crash and/or some modules and UI components don't render properly when loading the app in the InPrivate mode of Firefox. The core of the issue is that the IndexedDB API is not properly working in InPrivate mode. While the API itself is technically available (i.e. accessing anything `window.indexedDB` won't crash), any call to such APIs will throw. In Dexie, this surfaces as a `DatabaseClosedError`. These are currently thrown by several offline APIs, specifically by most sync queue related members. Invocations of these functions then lead to the aforementioned issues.

Regarding our supported browsers, Firefox seems to be the only one having these issues. In fact, these are declared as bugs in Firefox itself and will (hopefully) be fixed eventually:
https://bugzilla.mozilla.org/show_bug.cgi?id=781982
https://bugzilla.mozilla.org/show_bug.cgi?id=1639542
While I, as a developer, would love to say "let's just wait until Firefox fixes the issue", that's of course not good take for OpenMRS users. Hence, this PR works around the issue at the core API by catching the `DatabaseClosedError`s. This essentially allows InPrivate mode users to continue calling the APIs - it's just more or less a noop.

So far, I've been conservative with applying this bandaid fix. For me, the current changes in the PR allow the application to function without any observable crashes. Yes, features like the offline tools (obviously) don't work in InPrivate mode, but that's about it - the rest of the app can be used as before. My suggestion would be to enhance the error catching only when necessary, in the hope that Firefox eventually fixes the issue and allows us to remove the workaround.

## Related Issue

There is this issue which technically also addresses issues with the InPrivate mode in Firefox. Having said that, it isn't related to IndexedDB. Nontheless, in my local tests, patient registration could be accessed without any problems in the setup.
https://issues.openmrs.org/browse/O3-1073


ZacButko (approved):
Yes unfortunately I don't think this will be supported (ever) by Firefox. The issue linked is 10 years old and closed with top comment being "It is by design that this doesn't work"

Thanks for patching the crash

PR #434, opened 2022-05-20
(fix) esm-config bug with arrays of strings

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Fixes a bug where the config module would emit warnings for valid array-of-string configurations.

### DIFF
--- a/packages/framework/esm-config/src/module-config/module-config.ts
+++ b/packages/framework/esm-config/src/module-config/module-config.ts
@@ -596,18 +596,18 @@ function mergeConfigs(configs: Array<Config>) {
 // Recursively check the provided config tree to make sure that all
 // of the provided properties exist in the schema. Run validators
 // where present in the schema.
-const validateConfig = (
+function validateConfig(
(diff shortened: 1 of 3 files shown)


packages/framework/esm-config/src/module-config/module-config.ts
```diff
@@ -596,18 +596,18 @@ function mergeConfigs(configs: Array<Config>) {
...
-) => {
+) {
   for (const key of Object.keys(config)) {
     const value = config[key];
     const thisKeyPath = keyPath + "." + key;
     const schemaPart = schema[key] as ConfigSchema;
 
     if (!schema.hasOwnProperty(key)) {
-      if (key !== "extensionSlots") {
+      if (!(key === "extensionSlots" && keyPath !== "")) {
```
brandones, line 610: Unrelated—"extensionSlots" is only a valid key at the top level (i.e. when `keyPath` is `""`).

denniskigen (approved):

PR #471, opened 2022-06-30
(bug) Fix failing nav-link on patient-outpatient leftNav

donaldkibet:
## Requirements

#### For changes to apps

#### If applicable

## Summary

The `useLocation` hook from `react-router-dom` seem not to be working always returning null. This is causing the `leftNav` component on the outpatient to fail. This PR attempts to fix this bug.

### DIFF
--- a/packages/framework/esm-styleguide/src/left-nav/index.tsx
+++ b/packages/framework/esm-styleguide/src/left-nav/index.tsx
@@ -1,6 +1,6 @@
 /** @module @category UI */
 import React from "react";
-import { BrowserRouter, useLocation } from "react-router-dom";
+import { BrowserRouter } from "react-router-dom";
 
 import { ExtensionSlot, useStore } from "@openmrs/esm-react-utils";
 import { createGlobalStore } from "@openmrs/esm-state";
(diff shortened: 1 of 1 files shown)


packages/framework/esm-styleguide/src/left-nav/index.tsx
```diff
@@ -32,8 +32,7 @@ type LeftNavMenuProps = SideNavProps;
 export const LeftNavMenu = React.forwardRef<HTMLElement, LeftNavMenuProps>(
   (props, ref) => {
     const { slotName, basePath } = useStore(leftNavStore);
-    const location = useLocation();
-    const currentPath = location.pathname;
+    const currentPath = window.location ?? { pathname: "" };
```
hadijahkyampeire, line 35: Thanks for fixing this @donaldkibet just wondering can't we still keep using the `router` location? Also, where is the `pathname` coming from 🤔 , `window. location` has `href`.

donaldkibet: The [Location](https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname) has the `pathname` if am not wrong. The router location seems to be returning null, check your browser while this code is running without this fix, `location.pathname` is undefined. How do you propose we fix this issue?

Screenshot 2022-06-30 at 17 47 22

corneliouzbett (approved):

PR #472, opened 2022-07-01
Enhancing UserHasAccess component

Arjun-Go:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Existing UserHasAccess component returns null when the user is unauthorised. In this PR, we have enhanced the component such that the user sees an error message / notification or can redirect to a given URL.  

## Screenshots
image

### DIFF
--- a/packages/framework/esm-react-utils/src/UserHasAccess.tsx
+++ b/packages/framework/esm-react-utils/src/UserHasAccess.tsx
@@ -2,14 +2,14 @@
 import { getCurrentUser, LoggedInUser, userHasAccess } from "@openmrs/esm-api";
 import React, { useEffect, useState } from "react";
 
-interface UserHasAccessProps {
+export interface UserHasAccessProps {
   privilege: string;
-  unauthorisedResponse?: React.ReactNode | undefined;
+  fallback?: React.ReactNode;
 }
 
 export const UserHasAccess: React.FC<UserHasAccessProps> = ({
   privilege,
-  unauthorisedResponse,
+  fallback,
   children,
 }) => {
   const [user, setUser] = useState<LoggedInUser | null>(null);
@@ -24,6 +24,6 @@ export const UserHasAccess: React.FC<UserHasAccessProps> = ({
   if (user && userHasAccess(privilege, user)) {
     return <>{children}</>;
   } else {
-    return unauthorisedResponse ? <>{unauthorisedResponse}</> : null;
+    return fallback ? <>{fallback}</> : null;
   }
 };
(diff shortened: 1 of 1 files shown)


manuelroemer:
Hey @Arjun-Go, thanks for the PR! I think that this concrete approach can potentially cause unwanted side-effects because it is making the component do too much/make too many assumptions about what the unauthorized fallback should look like. Consider that `UserHasAccess` can be used anywhere - both for wrapping entire pages, but also only small sections, e.g. little input fields. If one page used that component 10 times (e.g. for various little form fields), you'd get 10 error messages when the page loads (given that you don't have the required privileges).

I think that, to achieve your goal of rendering some kind of fallback when a user doesn't have access, it would be much better to leave that fallback as a placeholder, e.g. like this (treat the following code as a suggestion, not the ultimate way it should look like):

```tsx
const fallbackWhenUnauthorized = (
    <div className="omrs-inline-notifications-container">
        <InlineNotification
            title="Unauthorised"
            subtitle="You cannot access this page."
            kind="error"
        />
    </div>
);

// Then somewhere else:
<UserHasAccess unauthorizedChildren={fallbackWhenUnauthorized} ...otherProps />
```

Moving the responsiblity to the user of the component keeps `UserHasAccess` clean and lean while also allowing each user to determine the best fallback content to render in that specific situation.

Arjun-Go:
Hi @manuelroemer, the suggestion really helped. I've refactored the code to incorporate your comments. Thanks!!

packages/framework/esm-react-utils/src/UserHasAccess.tsx
```diff
@@ -1,13 +1,15 @@
 /** @module @category API */
+import { getCurrentUser, LoggedInUser, userHasAccess } from "@openmrs/esm-api";
 import React, { useEffect, useState } from "react";
-import { getCurrentUser, userHasAccess, LoggedInUser } from "@openmrs/esm-api";
 
-export interface UserHasAccessProps {
+interface UserHasAccessProps {
```
brandones, line 5: This should still be exported

packages/framework/esm-react-utils/src/UserHasAccess.tsx
```diff
@@ -1,13 +1,15 @@
...
   privilege: string;
+  unauthorisedResponse?: React.ReactNode | undefined;
```
brandones, line 7: `response` doesn't make sense as a name here. It's not a response, it's a thing to render. I suggest `fallbackChildren`. Or just `fallback`. Note that `| undefined` is in most cases redundant with `?:`. Also, as an aside, we normatively use US spelling.

brandones:
```suggestion
  fallback?: React.ReactNode;
```

brandones (changes requested):

packages/framework/esm-react-utils/src/UserHasAccess.tsx
Arjun-Go: sure @brandones , changes have been made accordingly.

brandones (approved):

PR #474, opened 2022-07-08
(feat) Centralize ErrorState component

ZacButko:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This takes the ErrorState component, which is shown when data loading fails, out of esm-patient-common-lib and makes it more broadly accessible in esm-framework by way of esm-styleguide. The component itself is copy pasted. Everything else is support to get the component to render and test in the core package

## Screenshots
Same UI as was in patient chart
Screen Shot 2022-07-08 at 8 00 04 AM

## Other
This took quite a bit more work than expected. Had to install Jest and testing library for jest, this version bump caused some changes in default values, notably for testEnvironment from jsdom to node, hence the addition of jest.config.json and explicitly stating jsdom. Next test environment needed to be added as dev dependency, but this is broken with v13 for React < 18, so had to pin this to ^12.x. Next translations weren't implemented yet, so this was solved by adding __mocks__/react-i18next.js with standard mock info here.


denniskigen (approved):
Thanks, @ZacButko. Sorry for having to go through all the pain.

packages/framework/esm-styleguide/src/error-state/error-state.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { t } = useTranslation();
+  const isTablet = useLayoutType() === "tablet";
+
+  return (
+    <Tile light className={styles.tile}>
+      <div className={isTablet ? styles.tabletHeading : styles.desktopHeading}>
+        <h4>{headerTitle}</h4>
+      </div>
+      <p className={styles.errorMessage}>
+        {t("error", "Error")} {`${error?.response?.status}: `}
```
ibacher, line 25: Not a problem introduced with this PR, but I'd be curious to know what the strategy is for translating strings like this... AFAICT, there are no real translations for this anywhere...

ibacher (approved):

packages/framework/esm-styleguide/src/error-state/error-state.component.tsx
ZacButko: The translation function only translates the word "Error". The response status I'm assuming is the status code which should be a number, unless we want to translate way from Arabic numerals? As for the `response.statusText` is it reasonable to say that this is the backend's responsibility to translate?

ibacher: I think you missed my point: Say I want to say that "Error" in French should be "Erreur". Where do I put that string?

ZacButko: Do you just mean that the translation file is missing?

PR #475, opened 2022-07-09
(feat): Add implicit conditions to configuration schema

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This adds an implicit configuration section to every configuration schema called `"Display conditions"`[^1]. Currently, `"Display conditions"` has only one possible setting: `privileges` which is an array of one or more privileges that the user must have to be able to see any extension configured with that configuration schema. It defaults to not restricting. The `"Display conditions"` section is settable from anywhere in the configuration system (e.g., slot-specific configuration, extension-in-slot configuration). Note that this also means that no configuration scheme can have a top-level key named `"Display conditions"`. The validator does check for this and will log an error if an app tries to create a config schema with a `"Display conditions"` keys.

Usage:

```json
{
	"@openmrs/esm-patient-biometrics-app": {
		"Display conditions": {
			"privileges": ["View Obs"]
		}
	}
}
```

There are two important limitations here:
1. The configuration option does *not* override the `privilege` setting that can be returned as part of a page or extension definition. I'm open to discussing changing this.
2. Although this is added to all configuration schemas, the `"Display conditions"` section only has any effect on extensions.

Will ensure that extensions declared in `@openmrs/esm-patient-biometrics-app` only appear if the user has the "View Obs" privilege.

## Other

This PR also contains one technically breaking change; namely `userHasAccess()` now always returns a boolean. I checked the various repos including OHRI and did not find any uses of `userHasAccess()` that depended on it's previous return value (which was either the string of the permission checked or `undefined`).

There are a couple of additional, backwards compatible API enhancements:
* `userHasAccess()` can now take either a string for a single permission or an array of multiple permissions. If multiple permissions are specified, the user must have all of those permissions.[^2]
* The `privilege` that can be returned for pages and extensions in `setupOpenMRS()` can now likewise be an array as can the `privilege` property on the `UserHasAccess` component.

[^1]: O3-1207 suggested calling this section just `"conditions"`. However, `"conditions"` seemed ambiguous—especially since a "condition" has a very different meaning in the context of clinical practice.
[^2]: This decision seems like it might need some justification since it's may not be unreasonable to want to specify multiple privileges and show something if the user has *any* of the specified privileges. My thinking here is that the most common use-case for privileges in the frontend is to enable hiding widgets from users where the backend will reject the request anyways and permissions on the backend are always additive in this way.

### DIFF
--- a/packages/framework/esm-api/src/shared-api-objects/current-user.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/current-user.ts
@@ -1,6 +1,7 @@
 /** @module @category API */
 import { reportError } from "@openmrs/esm-error-handling";
 import { createGlobalStore } from "@openmrs/esm-state";
+import isUndefined from "lodash-es/isUndefined";
 import { Observable } from "rxjs";
 import { openmrsFetch, sessionEndpoint } from "../openmrs-fetch";
 import type {
@@ -122,15 +123,31 @@ function setUserLanguage(data: Session) {
 }
 
 function userHasPrivilege(
-  requiredPrivilege: string,
+  requiredPrivilege: string | string[],
   user: { privileges: Array<Privilege> }
 ) {
-  return user.privileges.find((p) => requiredPrivilege === p.display);
+  if (typeof requiredPrivilege === "string") {
+    return !isUndefined(
+      user.privileges.find((p) => requiredPrivilege === p.display)
+    );
+  } else if (Array.isArray(requiredPrivilege)) {
+    return (
+      requiredPrivilege.length &&
+      requiredPrivilege.every(
+        (rp) => !isUndefined(user.privileges.find((p) => rp === p.display))
+      )
+    );
+  } else {
+    console.error(`Could not understand privileges "${requiredPrivilege}"`);
+  }
+
+  return true;
 }
 
 function isSuperUser(user: { roles: Array<Role> }) {
-  const superUserRole = "System Developer";
-  return user.roles.find((role) => role.display === superUserRole);
+  return !isUndefined(
+    user.roles.find((role) => role.display === "System Developer")
+  );
 }
 
 /**
@@ -179,7 +196,7 @@ export function clearCurrentUser() {
 }
 
 export function userHasAccess(
-  requiredPrivilege: string,
+  requiredPrivilege: string | string[],
   user: { privileges: Array<Privilege>; roles: Array<Role> }
 ) {
   return userHasPrivilege(requiredPrivilege, user) || isSuperUser(user);
--- a/packages/framework/esm-api/src/shared-api-objects/current-user.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/current-user.ts
@@ -132,7 +132,6 @@ function userHasPrivilege(
     );
   } else if (Array.isArray(requiredPrivilege)) {
     return (
-      requiredPrivilege.length &&
       requiredPrivilege.every(
         (rp) => !isUndefined(user.privileges.find((p) => rp === p.display))
       )
--- a/packages/framework/esm-api/src/shared-api-objects/current-user.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/current-user.ts
@@ -131,10 +131,8 @@ function userHasPrivilege(
       user.privileges.find((p) => requiredPrivilege === p.display)
     );
   } else if (Array.isArray(requiredPrivilege)) {
-    return (
-      requiredPrivilege.every(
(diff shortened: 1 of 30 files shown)


packages/framework/esm-api/src/shared-api-objects/current-user.ts
```diff
@@ -122,15 +123,31 @@ function setUserLanguage(data: Session) {
...
+  if (typeof requiredPrivilege === "string") {
+    return !isUndefined(
+      user.privileges.find((p) => requiredPrivilege === p.display)
+    );
+  } else if (Array.isArray(requiredPrivilege)) {
+    return (
+      requiredPrivilege.length &&
+      requiredPrivilege.every(
+        (rp) => !isUndefined(user.privileges.find((p) => rp === p.display))
+      )
```
brandones, line 138: This will return `0` if `requiredPrivilege == []`. I think that intuitively, checking `userHasPrivilege([])` should always return `true`.

ibacher: Good point.

packages/framework/esm-config/src/module-config/module-config.test.ts
```diff
@@ -301,7 +315,12 @@ describe("getConfig", () => {
     Config.provide(goodConfig);
     const result = await Config.getConfig("foo-module");
     expect(result).toStrictEqual({
-      bar: { a: { b: 0 }, c: 2, diff: 2 },
+      bar: {
+        a: { b: 0 },
+        c: 2,
+        diff: 2,
+      },
+      "Display conditions": { privileges: [] },
```
brandones, line 323: Why should `Display conditions` be available in the config object? Shouldn't it just be handled behind the scenes?

ibacher: Implementation-wise this seemed simpler than adding it implicitly and then also adding it implicitly to the implementer tools.

packages/framework/esm-config/src/module-config/module-config.ts
```diff
@@ -863,3 +879,16 @@ function getExtensionNameFromId(extensionId: string) {
   const [extensionName] = extensionId.split("#");
   return extensionName;
 }
+
+/**
+ * The displayConditionsSchema is implicitly included in every configuration schema
+ */
+const displayConditionsSchema: ConfigSchema = {
```
brandones, line 886: This is quite different from how we handle the extension config. But it is a cool idea, and might actually be a better approach than what we currently have implemented. The only tricky part, at first glance, is that IMO these special keys (`extensionSlots`, `Display conditions`) should not be included in the resolved config object returned by `useConfig`/`getConfig`.

brandones:
This is awesome, thanks @ibacher .

I'm inclined to call this the implementation of https://issues.openmrs.org/browse/O3-1207 . I think you have improved on the naming, which is great. And your rationale about permissions being additive seems reasonable to me.

I think it is very important that the configured value of "permissions" override the one provided in code. It would be great if the implementer tools config object also included the existing extension permissions (the ones from code) as the "default" value.

I think we should look into bringing the implementation of `extensionSlots` and `Display conditions` into alignment. The way you've figured out might be better than what I wrote for `extensionSlots`.

packages/framework/esm-config/src/module-config/module-config.test.ts
brandones: I really like your implementation. Maybe we can just `delete` that key from the object returned by `getConfig`?

ibacher:
@brandones I've implemented the part to hide the display conditions from consuming apps. Getting this to override extension-level privileges settings turns out to require some major refactoring. The underlying problem is that we only track extension configs only for mounted extensions and this means that from the perspective of the code that hides the extensions, the configuration never exists.

packages/framework/esm-config/src/module-config/module-config.test.ts
```diff
@@ -1009,7 +1055,11 @@ describe("extension config", () => {
     updateConfigExtensionStore();
     Config.provide(moduleLevelConfig);
     const result = getExtensionConfig("barSlot", "fooExt").config;
-    expect(result).toStrictEqual({ bar: "qux", baz: "bazzy" });
+    expect(result).toStrictEqual({
+      bar: "qux",
+      baz: "bazzy",
+      "Display conditions": { privileges: [] },
```
brandones, line 1061: I expect it not to appear here, either

brandones (approved):
Awesome, thank you @ibacher !

We will eventually have to do the refactoring necessary to make configured permissions override in-code ones.

PR #481, opened 2022-07-18
(feat) Change ExtensionSlot's 'extensionSlotName' prop to just 'name'

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Does what it says on the tin. Having `extensionSlot` prefixing one of the props for `ExtensionSlot` is redundant.


ibacher (approved):
As you said, does what it says on the tin.

PR #482, opened 2022-07-18
Global button for opening Implementer Tools

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
I have made a global button for opening and closing the implementer tools in the application. Currently, we are not able to open the Implementer tools in the patient chart, due to the visit header. Seeing the designs in [here](https://app.zeplin.io/project/620a1aba48e34da85564ca12/screen/62b1f0fe10a74c1af11ed596), I have made a global button for the same.

## Screenshots
Screenshot 2022-07-19 at 3 13 47 PM
Screenshot 2022-07-19 at 3 13 55 PM
Screenshot 2022-07-19 at 3 14 05 PM
Screenshot 2022-07-19 at 3 14 17 PM

### DIFF
--- a/packages/apps/esm-implementer-tools-app/src/global-implementer-tools-button.test.tsx
+++ b/packages/apps/esm-implementer-tools-app/src/global-implementer-tools-button.test.tsx
@@ -1,0 +1,13 @@
+import { cleanup, render, screen } from "@testing-library/react";
+import React from "react";
+import "@testing-library/jest-dom";
+import GlobalImplementerToolsButton from "./global-implementer-tools.component";
+
+describe("Testing the global implementer tools button", () => {
+  afterEach(cleanup);
+  it("should render global Implementer tools", () => {
+    render(<GlobalImplementerToolsButton />);
+    const button = screen.getByTestId("globalImplementerToolsButton");
+    expect(button).toBeInTheDocument();
+  });
+});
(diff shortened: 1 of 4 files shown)


ZacButko (approved):
Awesome!

brandones:
Awesome, thank you @vasharma05 !

Since we haven't yet switched to having the implementer tools wrap around the screen, should we put the button at the bottom of the screen instead?

vasharma05:
Hi @brandones!
Actually we need to put the button for the visit header on the patient chart, since the user is not able to access the implementer tools in the patient chart. As discussed with @ZacButko, he suggested that since we'll be implementing this button for only visit header, and we have to keep this button everywhere as per design, why not to keep it for the whole application.

ZacButko:
I'm actually on board with @brandones 's suggestion. Basically what the button does now is to "pop up" the implementer tools from the bottom on the screen. Then it makes sense that the button to open them is on the bottom. I think everything else is agreed on, this would just be moving the action button from the top of the page to the bottom. @vasharma05 what do you think?

vasharma05:
Well @brandones, my suggestion was to just put this button on the visit header, since it's meant to be used at that place only, till the time we work on the new implementer's tools. Having 2 buttons doing the same action (nav action and this chevron button) at the same time only. 

I can place this button on the bottom of the screen, but I feel that it might also conflict with the UI. 
What do you say @brandones @ZacButko ?
I'll push the changes for the updated position.
Thanks!

ZacButko (approved):
🤩

brandones (approved):

PR #487, opened 2022-08-01
Fix versions

ibacher:
## Requirements
- [X] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.

#### For changes to apps

#### If applicable

## Summary
Follow-on to the previous commit to get exact version numbers. This updates the CI build to use a single step to produce the pre-release version. I think the failures in CI are caused by referencing the "patch" version without having any artefact associated with the "patch" version. Since ultimately, we want versions like 3.4.1-pre.12345, this gets us there in a single step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,8 +72,7 @@ jobs:
         if: steps.cache.outputs.cache-hit != 'true'
         run: yarn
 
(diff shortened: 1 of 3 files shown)


brandones (approved):
Nice, thanks

PR #493, opened 2022-08-09
(fix) Redirection issue in UserHasAccess component

abinaya-u:
## Requirements

#### For changes to apps

## Summary
There is a bug in the current code, when the logged in user is unauthorized. When the fallback is a redirect, user always gets redirected since an asynchronous event is present in the useEffect which sets the user state variable.

## Related Issue
https://bahmni.atlassian.net/browse/BAH-1645?atlOrigin=eyJpIjoiMzIwNzlkZGYxYTdkNGVjOWFiOGRiNTI2MjI1YzMxNjIiLCJwIjoiaiJ9

### DIFF
--- a/packages/framework/esm-react-utils/src/UserHasAccess.tsx
+++ b/packages/framework/esm-react-utils/src/UserHasAccess.tsx
@@ -24,6 +24,6 @@ export const UserHasAccess: React.FC<UserHasAccessProps> = ({
   if (user && userHasAccess(privilege, user)) {
     return <>{children}</>;
   } else {
-    return fallback ? <>{fallback}</> : null;
+    return fallback && user ? <>{fallback}</> : null;
   }
 };


ibacher:
A couple of small things: the convention for this is `(fix)` rather than `(bug)` (see the previous commits). The commit title could also be more descriptive. Finally, I'm not quite sure what this is trying to accomplish? It seems like instead of pushing this behaviour into the `UserHasAccess` element, it more properly belongs in the component you're supplying as the fallback.

abinaya-u:
> A couple of small things: the convention for this is `(fix)` rather than `(bug)` (see the previous commits). The commit title could also be more descriptive. Finally, I'm not quite sure what this is trying to accomplish? It seems like instead of pushing this behaviour into the `UserHasAccess` element, it more properly belongs in the component you're supplying as the fallback.

@ibacher Thanks for the comment. So we are not directly injecting redirect functionality to UserHasAccess component. We are addressing an issue when fallback is a Redirect component. Currently, since user is null initially it will be redirected before setting the user state variable. To avoid that we are checking whether user variable is present.

ibacher:
@abinaya-u The logic you describe is perfectly fine and makes sense. My point is just that it doesn't belong in the UserHasAcccess component. However, in your _fallback_ component you could do something like this:

```jsx
const { user } = useSession();
return !user ? null : /* redirect logic */
```

Which should solve the same problem.

abinaya-u:
> @abinaya-u The logic you describe is perfectly fine and makes sense. My point is just that it doesn't belong in the UserHasAcccess component. However, in your _fallback_ component you could do something like this:
> 
> ```js
> const { user } = useSession();
> return !user ? null : /* redirect logic */
> ```
> 
> Which should solve the same problem.

Thanks for the suggestion @ibacher. Even if we add this logic in fallback component, we need to check the user variable in UserHasAccess component before rendering the fallback. So having the check in UserHasAccess component might solve the issue.

ZacButko:
I agree its an issue of three states -> no user, user with permissions, user without permissions. - and I agree with Ian that it is outside the scope of UserHasAccess to provide the fallback functionality for user isn't loaded yet / no user. Semantically there should only be one fallback per component <sup>[citation needed]</sup>. Seeing as the user is a higher dependency I might structure the code like so:

```js
<UserLoaded fallback={<Loading />}>
	<UserHasAccess fallback={<NoAccessMessage />}>
		<p>The user exists and has access!</p>
	</UserHasAccess>
</UseLoaded>
```

UserLoaded isn't an actual component, but it would look very similar to what Ian has written above.

ZacButko:
Closing this as the fallback functionality of "no user present" is not the responsibility of `UserHasAccess`

PR #494, opened 2022-08-09
(fix) requiredPrivilege should not display an error if privilege is undefined

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Basically, this removed an error message I accidentally introduced. If `userHasPrivilege()` is called and the `requiredPrivilege` is `undefined`, then `userHasPrivilege()` should simply be a no-op returning `true`.

### DIFF
--- a/packages/framework/esm-api/src/shared-api-objects/current-user.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/current-user.ts
@@ -123,7 +123,7 @@ function setUserLanguage(data: Session) {
 }
 
 function userHasPrivilege(
-  requiredPrivilege: string | string[],
+  requiredPrivilege: string | string[] | undefined,
   user: { privileges: Array<Privilege> }
 ) {
   if (typeof requiredPrivilege === "string") {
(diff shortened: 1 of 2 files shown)


packages/framework/esm-api/src/shared-api-objects/current-user.ts
```diff
@@ -134,6 +134,7 @@ function userHasPrivilege(
     return requiredPrivilege.every(
       (rp) => !isUndefined(user.privileges.find((p) => rp === p.display))
     );
+  } else if (isUndefined(requiredPrivilege)) {
   } else {
```
vasharma05, line 138:
```suggestion
  } else if (!isUndefined(requiredPrivilege)) {
  }
```

vasharma05 (approved):

packages/framework/esm-api/src/shared-api-objects/current-user.ts
vasharma05: An empty if block seems abrupt.

ibacher: It's basically to stop being processed by the next block, which logs an error...

ibacher, line 138:
```suggestion
  } else if (!isUndefined(requiredPrivilege)) {
```

PR #505, opened 2022-08-15
(feat) allow extensions to take function as children instead of using wrap()

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

As part of the 4.0 upgrade the `wrap()` function on Extensions (which allows the consumer to wrap the React tree in a custom container) was deprecated with the note to use a child function. This implements support for the child-function mechanism. See the test for what this looks like.


packages/framework/esm-react-utils/src/ConfigurableLink.tsx
```diff
@@ -40,7 +44,7 @@ export function ConfigurableLink({
   templateParams,
   children,
   ...otherProps
-}: ConfigurableLinkProps) {
+}: PropsWithChildren<ConfigurableLinkProps>) {
```
ZacButko, line 47: ooo interesting

ZacButko (approved):

packages/framework/esm-react-utils/src/ConfigurableLink.tsx
ibacher: Yeah... types-wise at any rate, React functional components now need to opt-in to supporting children... Oddly this worked, but (at least in my VSCode instance) it was because `children` was being typed by react-i18next.

brandones:
@ibacher I think there was some confusion here... I don't know why the `<Extension>` component would have children at all. The `<Extension>` component really just marks where in the slot child tree the extension should be rendered. There are two examples in the [documentation for ExtensionSlot](https://github.com/openmrs/openmrs-esm-core/blob/cea9a99e740dea190ce38d2929cde5b9a6c485a7/packages/framework/esm-styleguide/src/_vars.scss#L63-L70). Is there a use case I'm not thinking of?

ibacher:
@brandones In retrospect, I agree. At the time, I wasn't massively comfortable with how extension slots and render functions worked, so I really just wanted a pain-free way of [migrating what had been our one use of the `wrap` property](https://github.com/openmrs/openmrs-esm-patient-chart/blob/eea68c63846af39892bdf2e6bc1d08dc9fcc4824/packages/esm-patient-chart-app/src/patient-chart/chart-review/dashboard-view.component.tsx#L91).

I think we can safely retire this and rewrite that one use-case.

PR #509, opened 2022-08-16
(fix) pin semver version so backend modules aren't reported as missing

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Pins the version of semver to a version that works. The more recent version causes some sort of infinite recursive call when trying to access the LRU cache from the "lru-cache" library. I don't know exactly why that happens, but both the semver and lru-cache libraries are targetted at Node as a runtime rather than browsers, so I suspect that the libraries don't play well with ESM semantics.

### DIFF
--- a/packages/framework/esm-utils/package.json
+++ b/packages/framework/esm-utils/package.json
@@ -47,6 +47,6 @@
     "rxjs": "6.x"
   },
   "dependencies": {
-    "semver": "^7.3.2"
+    "semver": "7.3.2"
   }
 }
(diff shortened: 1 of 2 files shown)


ibacher:
@ZacButko I think this fixes the last annoying error message, so once this is merged, we can probably do a real release of this module at least...

packages/framework/esm-utils/package.json
```diff
@@ -47,6 +47,6 @@
     "rxjs": "6.x"
   },
   "dependencies": {
-    "semver": "^7.3.2"
+    "semver": "7.3.2"
```
ZacButko, line 50: haha, you have to out-semver the semver

ZacButko (approved):

Josephkagimu1:
That's great the issue has been sorted .

PR #513, opened 2022-08-25
Added mock encryption (stringify) for syncQueue

hamza-kaizar:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This is a test PR to demonstrate the challenges with encrypting data in syncQueue.

A mock encryption (where data is stringified) is done in offline mode. The data is again parsed to it's original (JSON) format before being uploaded to the server. While testing with patient registration form, it was found that after registration, patient visit would not be automatically created. There seems to be some dependency between the forms that happens right after encryption. Request to investigate this.

### DIFF
--- a/packages/framework/esm-offline/src/sync.ts
+++ b/packages/framework/esm-offline/src/sync.ts
@@ -16,6 +16,7 @@ export interface SyncItem<T = any> {
   content: T;
   createdOn: Date;
   descriptor: QueueItemDescriptor;
+  encrypted?: boolean;
   lastError?: {
     name?: string;
     message?: string;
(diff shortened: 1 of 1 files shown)


manuelroemer:
The problem with the visit happens, because you are not running the decryption logic in, to name one example, the `getFullSynchronizationItemsFor` function (there are also other functions in `sync.ts` which must support decryption).

This causes issues [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/2fd113394638c2d19ba333499bad5ba838dda515/packages/esm-patient-common-lib/src/offline/visit.ts#L96). Instead of receiving a full visit object, the function receives the JSON string (and then throws, because that string doesn't have the expected attributes).

hamza-kaizar:
Thanks Manuel. This seems to have been what was causing the problem. It has been resolved. We will keep this in mind when working on this further.

ZacButko:
@hamza-kaizar is Manu's suggestion sufficient or is this PR still needed?

PR #515, opened 2022-08-25
feat(react-utils): extension should not re-render with state changes

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Alright, this appears to be a bit beyond me, but this is (most of) the fix to stop re-rendering the form widget constantly. The root of the problem is in the [Extension](https://github.com/openmrs/openmrs-esm-core/blob/dc64954dd108e4fb6b125086d79b69b5304156eb/packages/framework/esm-react-utils/src/Extension.tsx#L67) component. In particular, things were updated to track the current parcel in the Extension itself rather than in the `renderExtension()` function as used to be the case. This is fine because, in theory, that seems the right place to manage the parcel.

The underlying issue is that the parcel rendering is fundamentally asynchronous (since almost all of our parcels implement a `load()` function that looks like:

```ts
() => import("./root.component")
```

This means that `renderExtension()` now always returns `null` (because the parcel is basically never actually done rendering when `renderExtension()` returns) and consequently [this guard](https://github.com/openmrs/openmrs-esm-core/blob/dc64954dd108e4fb6b125086d79b69b5304156eb/packages/framework/esm-react-utils/src/Extension.tsx#L68) doesn't stop the parcel from re-rendering.

My fix is basically just returning a promise from `renderExtension()` and then using the promise to set the `parcel` ref correctly. This appears to work in practice, i.e., I haven't personally noticed any misbehaviour when running this locally but there are a large number of features supported by the extension system with extensive tests and I can't get those tests to pass, which suggests that I've broken something... but I haven't really been able to figure out where...

If anyone has a better suggestion for how to fix this, I'm definitely open to it.

It may be possible to just revert the commit that made these changes (d9d1fd), but the reversion can't be done cleanly without clobbering features added both in that commit and in subsequent commits to the extension system and I haven't had a great deal of luck straightening that out.

This also incidentally upgrades us to single-spa-react@5 which has proper support for React 18, which also seemed to fix some issues with the call to `parcel.update()`.


ZacButko (approved):

packages/framework/esm-react-utils/src/Extension.tsx
```diff
@@ -66,29 +65,48 @@ export const Extension: React.FC<ExtensionProps> = ({
...
         state
-      );
+      ).then((newParcel) => {
+        parcel.current = newParcel;
+      });
+
       return () => {
-        parcel.current && parcel.current.unmount();
+        parcel.current &&
+          typeof parcel.current.unmount === "function" &&
```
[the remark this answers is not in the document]
ibacher: Yeah... this is just me being paranoid :grin:. I'll get rid of it.

packages/framework/esm-react-utils/src/Extension.tsx
```diff
@@ -66,29 +65,48 @@ export const Extension: React.FC<ExtensionProps> = ({
...
+          parcel.current.unmount();
       };
     }
+
+    // we intentionally do not re-run this hook if state gets updated
+    // state updates are
```
brandones, line 87: state updates are... handled by the next `useEffect`?

packages/framework/esm-react-utils/src/Extension.tsx
```diff
@@ -66,29 +65,48 @@ export const Extension: React.FC<ExtensionProps> = ({
...
-      );
+      ).then((newParcel) => {
+        parcel.current = newParcel;
+      });
+
       return () => {
-        parcel.current && parcel.current.unmount();
+        parcel.current &&
+          typeof parcel.current.unmount === "function" &&
+          parcel.current.unmount();
```
brandones, line 82: The error I'm seeing in the tests is https://single-spa.js.org/error/?code=6&arg=&arg=UPDATING, so maybe you should add "and the parcel is not in status UPDATING" to the conditions here

brandones:
This seems good and right! Thanks for fixing this thing; the way I wrote it with the return statement is obviously wrong. I hope you get the tests passing without much trouble.

ibacher:
Alright, I've fixed one part of the test failures I was seeing by moving the update to the extension store to before we actually mount the parcel. This clears up the issue where the assignedExtensions were only properly updated after the extension slot was rendered. However, it seems to leave some bugs still in the extensions tests. Hopefully, I can work those out tomorrow.

PR #516, opened 2022-08-26
(fix) Display offline patient updates in offline tools

manuelroemer:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Fixes the issue that offline tools were not display offline updates made to patients. The following flow did, for example, not display the name change update in the offline tool's patient table:

1. Add a patient to the offline list.
2. Go offline.
3. Make an update to that patient's name (e.g. Lisa -> Lisa Two).
4. Go to the offline tools. They still displayed the old name pre-update.

In addition, this PR also contains some other niceties, somewhat related to the above issue:
* SWRs are now used more efficiently for fetching resources required by the offline tools. Should lead to better performance in the offline tools.
* The overview card for offline patients displayed wrong numbers when a patient was updated. This is now fixed.

### DIFF
--- a/packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
+++ b/packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
@@ -132,5 +132,9 @@ function useMergedSwr<T>(
       isValidating,
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [merge, ...swrResponses]);
+  }, [
+    merge,
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    ...swrResponses.flatMap((res) => [res.data, res.error, res.isValidating]),
+  ]);
 }
--- a/packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
+++ b/packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
@@ -77,8 +77,6 @@ export function useOfflinePatientsWithEntries() {
         ...offlineUpdates
       ) as fhir.Patient;
 
-      console.log("final", fhirPatientsSwr.data);
-
       return {
         patient: finalPatient,
         entry: offlinePatientEntry,
--- a/packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
+++ b/packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
@@ -3,39 +3,101 @@ import {
   getSynchronizationItems,
   getDynamicOfflineDataEntries,
 } from "@openmrs/esm-framework";
-import useSWR from "swr";
+import merge from "lodash-es/merge";
+import { useMemo } from "react";
+import useSWR, { SWRResponse } from "swr";
+
+function useDynamicOfflineDataEntries(type: string) {
+  return useSWR(`dynamicOfflineData/entries/${type}`, () =>
+    getDynamicOfflineDataEntries(type)
+  );
+}
+
+function useSynchronizationItems<T>(type: string) {
+  return useSWR(`syncQueue/items/${type}`, () =>
+    getSynchronizationItems<T>(type)
+  );
+}
+
+function useFhirPatients(ids: Array<string>) {
+  const stableIds = useMemo(() => [...ids].sort(), [ids]);
+  return useSWR(["fhirPatients", stableIds], () =>
+    Promise.all(
+      stableIds.map((patientId) =>
+        fetchCurrentPatient(patientId).then((res) => res.data)
+      )
+    )
+  );
+}
 
 export function useOfflineRegisteredPatients() {
-  return useSWR("offlineTools/offlineRegisteredPatients", async () => {
-    const offlinePatients = await getDynamicOfflineDataEntries("patient");
-    const syncItems = await getSynchronizationItems<{
-      fhirPatient?: fhir.Patient;
-    }>("patient-registration");
-    return syncItems
-      .filter(
-        (item) =>
-          item.fhirPatient &&
-          !offlinePatients.find(
-            (entry) => entry.identifier === item.fhirPatient.id
-          )
-      )
+  const offlinePatientsSwr = useDynamicOfflineDataEntries("patient");
+  const patientSyncItemsSwr = useSynchronizationItems<{
+    fhirPatient?: fhir.Patient;
+  }>("patient-registration");
+
+  return useMergedSwr(() => {
+    return patientSyncItemsSwr.data
+      .filter((patientRegistrationItem) => {
+        const isNewlyRegistered =
+          patientRegistrationItem.fhirPatient &&
+          !offlinePatientsSwr.data.find(
+            (offlinePatientEntry) =>
+              offlinePatientEntry.identifier ===
+              patientRegistrationItem.fhirPatient.id
+          );
+        return isNewlyRegistered;
+      })
       .map((item) => item.fhirPatient);
-  });
+  }, [offlinePatientsSwr, patientSyncItemsSwr]);
 }
 
 export function useOfflinePatientsWithEntries() {
-  return useSWR("offlineTools/offlinePatients", async () => {
-    const offlinePatientEntries = await getDynamicOfflineDataEntries("patient");
+  const offlinePatientsSwr = useDynamicOfflineDataEntries("patient");
+  const patientSyncItemsSwr = useSynchronizationItems<{
+    fhirPatient?: fhir.Patient;
+  }>("patient-registration");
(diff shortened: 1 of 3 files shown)


packages/apps/esm-offline-tools-app/src/hooks/offline-patient-data-hooks.ts
```diff
@@ -3,39 +3,99 @@ import {
...
-      fhirPatient?: fhir.Patient;
-    }>("patient-registration");
-    return syncItems
-      .filter(
-        (item) =>
-          item.fhirPatient &&
-          !offlinePatients.find(
-            (entry) => entry.identifier === item.fhirPatient.id
-          )
+function useDynamicOfflineDataEntries(type: string) {
```
ibacher, line 10: Can you help me understand why we're using SWR here? I noticed that a similar pattern seemed to be causing unnecessary re-renders on some of the apps (since SWR, by default, retriggers on events like the screen re-gaining focus)...

manuelroemer: Before I dive off into a wrong explanation I'll go with a counter question: What would be a reason not to use SWR here? If we don't want want the app to rerender on such events, it should (just) be disabled in the SWR settings, imo...

ibacher: So, I think the default SWR behaviour is something we generally want. We're this can prove problematic is when these offline hooks are used in the component tree above extensions and extension slots. Really what I'm wondering is if we're just using SWR as a sort of cache here we don't just use a unistore?

manuelroemer: > Really what I'm wondering is if we're just using SWR as a sort of cache here [...]

Yes that's part of the reason. And in my opinion, that is actually a good reason *for* using SWR here!

So, generally, the important concept about the swr library is that it is (in my own words) a library for asynchronously fetching resources. Not more, not less. A resource can be anything here: Something from the IndexedDB, something from a cache or, like in 99% of use-cases, something from a server. Any web API that we call using swr really is just that: an asynchronously fetched resource. But the same is also the case for data from an IndexedDB that is managed by the local browser (like the offline data used in this PR).

With that introduction, the point of using this library is that the flow for interacting with anything that's used asynchronously is, more or less, always the same:

1. You start the request (to the website, the DB, ...).
2. The request is pending. (No `data`, no `error`, `isValidating` is `true`).
3. The request eventually completes, either:
   1. Successfully (`data` is set).
   2. With an error (`error` is set).

SWR also provides a few other benefits, the most important one being request deduplication/caching (like you mentioned). For example, considering this component setup:

```ts
const Component1 = () => {
    const { data } = useSWR('key', async () => fetchAsyncResource());
    return <>{data}</>;
};

const Component2 = () => {
    const { data } = useSWR('key', async () => fetchAsyncResource());
    return <>{data}</>;
};

const App = () => (
    <>
        <Component1 />
        <Component2 />
    </>
)
```

Here, both `Component1` and `Component2` request the same resource. SWR deduplicates this and only calls `fetchAsyncResource` once. But the key here is that this is not only exlusively usable for calling web APIs. This deduplication also makes a lot of sense for, at the example of this PR, reducing the queries done to the IndexedDB. Therefore yes, it absolutely makes sense to also use SWR here. Accessing the offline data in the IndexedDB is just an async operation that can very well be handled by the SWR lifecycle.


> We're this can prove problematic is when these offline hooks are used in the component tree above extensions and extension slots.

I think I see why we want to minimize the number of rerenders. The thing is, the underlying problem of extensions rerendering is not unique to SWR calls accessing the IndexedDB. For example, just assume that the SWR hooks in this PR just made a normal fetch request. We'd have the same problem with the "proper" usage of SWR. If we want to reduce the number of rerenders, we should probably think about granularly disabling automatic mutations when certain events (like focusing) happen.

I hope that this answers the questions - not sure if this was really the explanation you were looking for. Please just tell me, if not. 😄

ibacher: > The thing is, the underlying problem of extensions rerendering is not unique to SWR calls accessing the IndexedDB.

Absolutely. However, while doing some deep debugging on the form re-renders, I was noticing that these `useSWR` hooks were causing the component to re-render even when the underlying data hadn't changed (for example, the `useSWR` hook was being called while online and with no results in the sync queue, but was still causing a state change in the component due to state tracked in the `useSWR` hook itself—this is probably a bug in `useSWR` and not your code, but it is something we need to work around).

This is especially problematic in components that render and `<ExtensionSlot>` and hence, on re-render, can end up re-rendering the attached extensions. I don't think there's a per-se issue with this pattern, but I think we need some additional memoization or something to ensure that `useSWR()` isn't invoking the underlying method unnecessarily. (This was basically the motivation behind [this commit](https://github.com/openmrs/openmrs-esm-patient-chart/commit/91b04476ee3c3e3e06390d501d3a1de4daec5456) in the patient chart).

manuelroemer: Hm, do you maybe have a way for me to reproduce the issues with SWR somehow? I'm really curious about this. SWR should really not be at fault for those rerenders, from my perspective. It's supposed to only reexecute the `fetcher` function when the associated key changes. The `data` returned by the `fetcher` is cached and should therefore also not cause rerenders by itself.
I also recall that SWR guarantees the following behavior:

```ts
const result = useSomeSwr();
const { data, error, isValidating } = result;
// Here, data, error and isValidating are guaranteed to be stable.
// result is, *IIRC*, *not* guaranteed to be stable though. This could (if I remember it right) cause rerenders if used wrongly.
```

---

Looking at the commit that you linked (sorry, I must have missed the PR), I do have some comments. I think this here is kind of the wrong location, but it fits into the discussion.
Taking for example the `useOfflineVisit` hook (in its old SWR version): Here are some assumptions (and issues!) that I see with it in its current form.

```ts
export function useOfflineVisit(patientUuid: string): Omit<ReturnType<typeof useVisit>, 'isLoading'> {
  const swrKey = patientUuid ? `offlineVisit/${patientUuid}` : null;
  // ^ swrKey is stable (unless, of course, the patient UUID changes).

  const offlineVisitSwr = useSWR<Visit>(swrKey, async () => {
    const offlineVisit = await getOfflineVisitForPatient(patientUuid);
    return offlineVisit ? offlineVisitToVisit(offlineVisit) : undefined;
  });
  // ^ offlineVisitSwr is *not* stable, but its `data`, `error` and `isValidating` fields are.
  // It *only* runs the async function when the key changes (which only happens when the UUID changes).

  // The object here is problematic though because it is not memoized. This is definitely a mistake, looking at it now.
  return {
    currentVisit: offlineVisitSwr.data,
    isValidating: offlineVisitSwr.isValidating,
    error: offlineVisitSwr.error,
    mutate: offlineVisitSwr.mutate,
  };


  // Here is what should have been done instead of the above return:
  return useMemo(() => ({
    currentVisit: offlineVisitSwr.data,
    isValidating: offlineVisitSwr.isValidating,
    error: offlineVisitSwr.error,
    mutate: offlineVisitSwr.mutate,
  }), [offlineVisitSwr.data, offlineVisitSwr.isValidating, offlineVisitSwr.error, offlineVisitSwr.mutate]);
}
```

The extension slot state was also not memoized. Doing that alone might drastically reduce the rerenders. Then finally, if that doesn't help, there is still the option to memo the `<ExtensionSlot ... />` itself.

---

I don't want to claim that the changes above would entirely fix the extension rerendering issue (I haven't tested it), but I do think that there are ways to approach this problem without removing the useSWR calls.

I'd also strongly advocate to reintroduce the SWR hooks in the patient-chart. The reason for introducing SWR in the first place is that there is no longer any need for custom data fetching, plus, of course, all the other benefits like performance/deduplication/caching/...). By falling back to custom `useState`/`useEffect` again, we're missing out on all of those + we'll have to maintain much more code than necessary. If the usage of SWR causes rerenders then let's fix those, but ideally without removing the SWR call itself. I'd love to help here, I just need some pointers as to how you go about debugging the form rerenders.

---

> this is probably a bug in useSWR and not your code, but it is something we need to work around

If that is actually true, we should 100% investigate. We are using SWR in so many locations by now that a bug there would be an actual problem.

ZacButko (approved):
I'm hearing that @manuelroemer is volunteering to clean up extra renders due to this useSWR should they present themselves. @manuelroemer can you watch this on dev and verify it doesn't cause any adverse side effects?

manuelroemer:
Yes, that's fair. After all I'd be the culprit who will have introduced the SWRs then. 
Regarding the problems with the form engine rerendering (if that does happen again after this PR), I'd need some help first, ideally in the form of some description as to how those rerenders are best debugged. I never dove that deep into the extension system's internals, so any pointers here would be super helpful.

PR #519, opened 2022-08-30
(fix) Fixes 'openmrs develop' 

ZacButko:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Fixes deprecation warnings, resolves app shell, upgrades webpack.
On local this allows 'npx openmrs develop ...' to run properly again in `esm-patient-chart` repo

### DIFF
--- a/packages/tooling/openmrs/README.md
+++ b/packages/tooling/openmrs/README.md
@@ -4,7 +4,7 @@ The one stop CLI for using the OpenMRS 3.0 Frontend app.
 
 ## Prerequisites
 
-You need Node.js with NPM. We recommend using version 10.18 (LTS) or later.
+You need Node.js with NPM. We recommend using version 16.17 (LTS) or later.
 
 ## Installation
 
--- a/packages/tooling/openmrs/src/commands/develop.ts
+++ b/packages/tooling/openmrs/src/commands/develop.ts
@@ -31,7 +31,7 @@ export function runDevelop(args: DevelopArgs) {
   const source = resolve(
     require.resolve("@openmrs/esm-app-shell/package.json"),
     "..",
-    "lib"
+    "dist"
   );
   const index = resolve(source, "index.html");
   const indexContent = readFileSync(index, "utf8")
--- a/packages/tooling/openmrs/src/utils/debugger.ts
+++ b/packages/tooling/openmrs/src/utils/debugger.ts
@@ -19,24 +19,21 @@ function loadConfig(configPath: string) {
 }
 
 function debug(configPath: string, port: number) {
-  const webpack = require("webpack");
+  const Webpack = require("webpack");
   const WebpackDevServer = require("webpack-dev-server");
   const config = loadConfig(configPath);
 
-  const options = {
+  const compiler = Webpack(config);
+  const devServerOptions = {
     ...config.devServer,
     port,
     static: dirname(configPath),
   };
 
-  const server = new WebpackDevServer(webpack(config), options);
+  const server = new WebpackDevServer(devServerOptions, compiler);
 
-  server.listen(port, "localhost", (err?: Error) => {
-    if (err) {
-      logWarn(`Error: ${err}`);
-    } else {
-      logInfo(`Listening at http://localhost:${port}`);
-    }
+  server.startCallback(() => {
+    logInfo(`Listening at http://localhost:${port}`);
   });
 }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -13887,9 +13887,9 @@ __metadata:
     tar: ^6.0.5
     ts-loader: ^9.2.6
     typescript: ~4.5.2
-    webpack: ^5.64.0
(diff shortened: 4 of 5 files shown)


yarn.lock
```diff
@@ -13887,9 +13887,9 @@ __metadata:
     tar: ^6.0.5
     ts-loader: ^9.2.6
     typescript: ~4.5.2
-    webpack: ^5.64.0
-    webpack-cli: ^4.9.1
-    webpack-dev-server: ^4.6.0
+    webpack: ^5.74.0
+    webpack-cli: ^4.10.0
+    webpack-dev-server: ^4.10.1
```
ZacButko, line 13892: Bumping the versions was mostly just superstition, and didn't appear to have any real effect

packages/tooling/openmrs/src/utils/debugger.ts
```diff
@@ -19,24 +19,21 @@ function loadConfig(configPath: string) {
...
-  const options = {
+  const compiler = Webpack(config);
+  const devServerOptions = {
     ...config.devServer,
     port,
     static: dirname(configPath),
   };
 
-  const server = new WebpackDevServer(webpack(config), options);
+  const server = new WebpackDevServer(devServerOptions, compiler);
```
ZacButko, line 33: Order of args was switched since this was last touched 2 years ago. Now it's options first, then compiler

packages/tooling/openmrs/src/utils/debugger.ts
```diff
@@ -19,24 +19,21 @@ function loadConfig(configPath: string) {
...
 
-  server.listen(port, "localhost", (err?: Error) => {
-    if (err) {
-      logWarn(`Error: ${err}`);
-    } else {
-      logInfo(`Listening at http://localhost:${port}`);
-    }
+  server.startCallback(() => {
```
ZacButko, line 35: The other warning was to replace `server.listen` with `server.start` or `server.startCallback`. Fixing this makes resolving ports auto-magical again

packages/tooling/openmrs/src/commands/develop.ts
```diff
@@ -31,7 +31,7 @@ export function runDevelop(args: DevelopArgs) {
   const source = resolve(
     require.resolve("@openmrs/esm-app-shell/package.json"),
     "..",
-    "lib"
+    "dist"
```
ZacButko, line 34: For some reason the app shell wasn't resolving properly from node modules. Changing this to `dist` seems to work, but I'm wondering how the build step switched from `lib` to `dist` and if that's something that we should fix other places, or somehow revert

packages/tooling/openmrs/README.md
```diff
@@ -4,7 +4,7 @@ The one stop CLI for using the OpenMRS 3.0 Frontend app.
 
 ## Prerequisites
 
-You need Node.js with NPM. We recommend using version 10.18 (LTS) or later.
+You need Node.js with NPM. We recommend using version 16.17 (LTS) or later.
```
ZacButko, line 7: LTS is changing to 18.x in 2 months...

ZacButko:
Note, this seems to improve behavior of non-monorepo repos (esm-fast-data-entry-app) in that it will now load app shell without complaint, but for those repos it doesn't seem to add the locally served import map correctly.

For now that can be worked-around with `localStorage.setItem("import-map-override:@openmrs/esm-fast-data-entry-app", "http://localhost:8081/dist/openmrs-esm-fast-data-entry-app.js");`

packages/tooling/openmrs/src/commands/develop.ts
ibacher: It looks to me as though we're publishing to both `lib` and `dist`, but the one in `lib` should be preferred because that's the "production" verison. [Source](https://github.com/openmrs/openmrs-esm-core/blob/1bb3ee4d7988fe53aa09cf6b59e5fce712ff9c8a/packages/shell/esm-app-shell/webpack.config.js#L69).

ibacher (approved):
Aside from the lib / dist thing, this seems good. Nice work @ZacButko!

packages/tooling/openmrs/src/commands/develop.ts
ZacButko: Yeah.. not really sure what to do about that. The `lib` folder was just missing from node_modules. I think my preference would be to find some way to put that back, but not sure how or when it's generated.

ibacher: @ZacButko For which version of the @openmrs/esm-app-shell? It's included in `node_modules` for the repos I've looked at...

ZacButko: For me it is missing from `@openmrs/esm-app-shell` 4.0.1 
Screen Shot 2022-08-30 at 6 58 46 AM

I'm super worried if the same version results in different node_modules on different machines. Either that or we're about to learn something really interesting about yarn...

PR #523, opened 2022-09-07
(fix) Fix import map overrides panel

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR fixes an issue where the overrides list UI in the `import-map-overrides` panel wouldn't display following the recent 4.0 upgrade. It changes how we install the `import-map-overrides` package in esm-core by switching from installing it as an npm package to the [preferred approach](https://github.com/single-spa/import-map-overrides/blob/main/docs/installation.md) - installing it via a `script` tag in our HTML file.

Many thanks to @donaldkibet for pairing on this!

## Video

https://user-images.githubusercontent.com/8509731/188953227-c1e7d93d-4317-4127-b07c-384c80d39fda.mp4

### DIFF
--- a/packages/shell/esm-app-shell/src/index.ejs
+++ b/packages/shell/esm-app-shell/src/index.ejs
@@ -26,6 +26,10 @@
     >
     <script type="systemjs-importmap" src="<%= openmrsImportmapUrl %>"></script>
 <% } %>
+    <script
+    type="text/javascript"
+    src="https://cdn.jsdelivr.net/npm/import-map-overrides/dist/import-map-overrides.js"
+    ></script>
   </head>
   <body>
     <%= htmlWebpackPlugin.tags.bodyTags %>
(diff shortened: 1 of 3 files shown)


packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -26,6 +26,10 @@
     >
     <script type="systemjs-importmap" src="<%= openmrsImportmapUrl %>"></script>
 <% } %>
+    <script
+    type="text/javascript"
+    src="https://cdn.jsdelivr.net/npm/import-map-overrides/dist/import-map-overrides.js"
```
ibacher, line 31: Can we do this in a way that doesn't require a public CDN? This would make the feature not work in network-isolated environments.

ibacher (approved):

PR #526, opened 2022-09-08
Offline encryption POC with base64 encoding

hamza-kaizar:
## Requirements

## Summary

The following proof of concept aims to present how data being used in offline mode can be encrypted. Data in the syncQueue (being used for offline actions) and data in application cache (pre-cached when online) are both encrypted by a simple base64 encoding. Data is decrypted before it is accessed or has to be uploaded back to the server.


packages/framework/esm-offline/src/encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+  var encryptedJson = { content: encryptedData };
+
+  return JSON.stringify(encryptedJson);
+}
+
+/**
+ * Decrypts data for dynamic caching
+ * @param json response json data
+ */
+export function decryptCache(json) {
```
FlorianRappl, line 25: Great - but we need a type (e.g., `{ content: string }`)

packages/framework/esm-offline/src/encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+  var decryptedJson = JSON.parse(decryptedData);
+
+  return JSON.stringify(decryptedJson);
+}
+
+/**
+ * Encrypts data for offline sync
+ * @param json content json data
+ */
+export function encryptSyncData(json) {
```
FlorianRappl, line 37: What's the type? Unknown or of a known shape?

packages/framework/esm-offline/src/encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+  var data = JSON.stringify(json);
+
+  return encrypt(data);
+}
+
+/**
+ * Decrypts data for offline sync
+ * @param data content string data
+ */
+export function decryptSyncData(data) {
```
FlorianRappl, line 47: You guess it... Types :-).

packages/shell/esm-app-shell/src/service-worker/routing.ts
```diff
@@ -15,10 +15,58 @@ import {
...
+    return cachedResponse;
+  },
+  cacheWillUpdate: async ({ request, response, event, state }) => {
+    var responseClone = response.clone();
+    var contentType;
+    if (request.url.includes("fhir")) {
+      var resHeaders = new Headers(responseClone.headers);
+      contentType = resHeaders.get("content-type");
+      if (contentType == "application/fhir+json;charset=UTF-8") {
+        resHeaders.append("encryption", "true");
```
FlorianRappl, line 50: Right now this is always set to `true`. I thought it should be configurable (=> we carry on `encryption` pretty much always, making it seem configurable).

Can we have this parameterized to prepare for this configuration option?

FlorianRappl (commented):
Looks good in general.

Miss the option to provide an encryption / decryption key here (just as a meaningless parameter right now - would be later on filled with the user's choice).

Also just from a data perspective we'd need a special IndexDB entry that can read in the beginning to verify reliably that an entered passphrase (enc. / dec. key) is actually valid.

packages/framework/esm-offline/src/encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+function decrypt(data: string) {
+  return atob(data);
+}
+
+/**
+ * Encrypts data for dynamic caching
+ * @param json response json data
+ */
+export function encryptCache(json) {
+  var data = JSON.stringify(json);
```
ibacher, line 14: I think it's preferable to use `const` or `let` where practicable instead of `var`.

packages/shell/esm-app-shell/src/service-worker/routing.ts
```diff
@@ -15,10 +15,58 @@ import {
...
+        var resJson = await responseClone.json().then((json) => json);
```
ibacher, line 51:
```suggestion
        const resJson = await responseClone.json();
```

packages/framework/esm-offline/src/encryption.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Decrypts data for dynamic caching
+ * @param json response json data
+ */
+export function decryptCache(json) {
+  var data = json["content"];
+  var decryptedData = decrypt(data);
+  var decryptedJson = JSON.parse(decryptedData);
+
+  return JSON.stringify(decryptedJson);
```
ibacher, line 30: If `encrypt()` is `(unknown) => { content: string }`, `decrypt()` should be `({content: string}) => unknown` right?

packages/framework/esm-offline/src/sync.ts
```diff
@@ -187,8 +189,15 @@ async function processHandler(
   const userId = await getUserId();
 
   await db.syncQueue.where({ type, userId }).each((item, cursor) => {
-    items.push([cursor.primaryKey, item.content, item.descriptor]);
-    contents.push(item.content);
+    var content: any;
```
ibacher, line 192:
```suggestion
    let content: unknown;
```

packages/framework/esm-offline/src/encryption.ts
```diff
@@ -0,0 +1,53 @@
...
+
+function decrypt(data: string) {
+  return atob(data);
+}
+
+/**
+ * Encrypts data for dynamic caching
+ * @param json response json data
+ */
+export function encryptCache(json: JSON) {
```
FlorianRappl, line 15: As these functions are, they are not symmetric, i.e., I'd expect that `decryptCache(encryptCache(x)) = x`, but this is not the case.

Here we have:

- `encryptCache` expects a JSON and returns a `string`
- `decryptCache` expects a JSON and returns a `string`

Just from the signature I cannot distinguish.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -38,6 +38,13 @@ import { setupI18n } from "./locale";
 import { loadModules } from "./load-modules";
 import { appName, getCoreExtensions } from "./ui";
 import { registerModules, sharedDependencies } from "./dependencies";
+import { 
+  setCryptoKey, 
+  setPasswordData, 
+  clearPasswordData, 
+  isPasswordExpired, 
+  isPasswordCorrect
+} from "@openmrs/esm-offline/src/encryption";
```
FlorianRappl, line 47: Please export directly such that you can use `@openmrs/esm-offline` as import reference

packages/shell/esm-app-shell/src/service-worker/caching.ts
```diff
@@ -102,3 +102,14 @@ async function invalidateObsoleteCacheEntries(newImportMapUrls: Array<string>) {
 
   await Promise.all(urlsToInvalidate.map((url) => cache.delete(url)));
 }
+
+export async function clearEncryptedCacheEntries() {
+  const cache = await caches.open(omrsCacheName);
+  (await cache.keys())
+    .filter((request) => request.url.includes("fhir"))
+    .forEach(async (request) =>
```
FlorianRappl, line 110: This is pretty much useless. You are doing here `async` to use `await`, but there is no need to use `await`. It already returns a promise and `forEach` is not sensitive to the returned value.

If your intention here was to process the requests one-by-one (i.e., await the processing of the first before continuing with the second etc.) then prefer a `for-of` loop where you really await the individual parts. Otherwise, you can just ignore the promise as you don't depend on its result.

packages/shell/esm-app-shell/src/service-worker/message.ts
```diff
@@ -2,15 +2,19 @@ import type {
   MessageServiceWorkerResult,
   OnImportMapChangedMessage,
   RegisterDynamicRouteMessage,
+  UpdateEncryptionKeyMessage
 } from "@openmrs/esm-offline";
 import escapeRegExp from "lodash-es/escapeRegExp";
-import { cacheImportMapReferences } from "./caching";
+import { cacheImportMapReferences, clearEncryptedCacheEntries } from "./caching";
 import { DynamicRouteRegistration, ServiceWorkerDb } from "./storage";
+import { setCryptoKey } from "@openmrs/esm-offline/src/encryption";
```
FlorianRappl, line 10: Use import from `@openmrs/esm-offline` (see above).

PR #536, opened 2022-09-14
(feat) Tweak content switcher appearance

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Tweaks the appearance of the content switcher component, applying an inner `color-blue-30` border on either side of the middle content section. This gets the component in line with the [available designs](https://zpl.io/EmvXKX8).

## Screenshots

> Before

Screenshot 2022-09-15 at 00 04 42

Screenshot 2022-09-15 at 00 04 53

> After

Screenshot 2022-09-15 at 00 01 28

Screenshot 2022-09-15 at 00 01 36


packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -78,7 +78,9 @@
 }
 
 .cds--content-switcher-btn:before {
-  display: none;
+  background-color: $color-blue-30;
```
ZacButko, line 81: Instead, what do you think about `@use '@carbon/colors';` above then `colors.$blue-30` here? This would eliminate need to define custom colors in _vars.scss file below

ZacButko (approved):
What do you think about using the `@use` syntax instead?

packages/framework/esm-styleguide/src/_overrides.scss
denniskigen: Yep. Eventually, we shouldn't need to have any carbon colors in `vars.scss`.

ZacButko (approved):

PR #540, opened 2022-09-20
(chore) Disable manual workflow triggering

denniskigen:
## Requirements

## Summary

Removes the `workflow_dispatch` event from the workflow triggers event list. This was previously only set for debugging purposes.


ZacButko (commented):
I can't think of a good reason to get rid of this?

denniskigen:
Fair enough, closing based on the discussion we had on Slack.

PR #543, opened 2022-09-25
03-1533: Add an actionable notification component.

hadijahkyampeire:
## Requirements

#### For changes to apps

#### If applicable

## Summary
- Add an actionable notification component which will be our snack bar for notifications that require interactive actions

## Screenshots
Screenshot 2022-09-18 at 20 22 48

https://user-images.githubusercontent.com/30952856/196437552-f66fb44b-d6cb-4036-ad4d-a1d8eaa4113f.mov

## Related Issue
[Add an actionable notification component.](https://issues.openmrs.org/browse/O3-1533)

## Other
After merging this I will add it to the patient management repo in the appointments work and then delete the actionable component from patient management.

### DIFF
--- a/packages/framework/esm-styleguide/src/notifications/_actionable-notifications.scss
+++ b/packages/framework/esm-styleguide/src/notifications/_actionable-notifications.scss
@@ -1,16 +1,11 @@
-@use '@carbon/styles/scss/spacing';
-@use '@carbon/styles/scss/type';
-@import '../../root.scss';
-
 .omrs-actionable-notifications-container {
   z-index: 1000;
   position: fixed;
   width: 100%;
   display: flex;
   flex-direction: column;
-  align-items: center;
-  justify-content: center;
-  bottom: spacing.$spacing-07;
+  margin-left: 0.2rem;
+  bottom: 2rem;
 }
 
 .omrs-actionable-notification-mounting,
--- a/packages/framework/esm-styleguide/src/notifications/_actionable-notifications.scss
+++ b/packages/framework/esm-styleguide/src/notifications/_actionable-notifications.scss
@@ -1,3 +1,5 @@
+@use "@carbon/styles/scss/spacing";
+
 .omrs-actionable-notifications-container {
   z-index: 1000;
   position: fixed;
@@ -5,7 +7,7 @@
   display: flex;
   flex-direction: column;
   margin-left: 0.2rem;
-  bottom: 2rem;
+  bottom: spacing.$spacing-07;
 }
 
 .omrs-actionable-notification-mounting,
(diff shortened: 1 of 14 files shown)


packages/framework/esm-styleguide/src/notifications/actionable-notification.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+    title,
+    critical,
+    progressActionLabel = "Undoing",
+    ...props
+  } = notification;
+
+  return (
+    <ActionableNotification
+      kind={kind || "info"}
+      actionButtonLabel={actionButtonLabel || progressActionLabel}
```
ZacButko, line 50: Do we need both`actionButtonLabel` and `progressActionLabel` here?

ZacButko (commented):
Code looks ok, but unless you add `<div class="omrs-actionable-notifications-container"></div>` to `index.ejs` I don't think this will ever mount, and worse will probably throw an error when it tries to mount to an empty query selector.

packages/framework/esm-styleguide/src/notifications/actionable-notification.component.tsx
hadijahkyampeire: We only need the actionButtonLabel not to be empty but when it is clicked you might want to pass a progressActionLabel or else the actionLabel stays. Actually, I need to first test it because there is a way I was implementing it with a hook and I think it won't be the same.

ibacher:
When adding something to the styleguide, it would be nice to have some doc string comments that we can turn into automated documentation that describe how these are intended to be used. Bonus points if you can add some doc strings to some of the types too...

ZacButko:
Agree with @ibacher . The prerequisite for adding components to esm-framework is that we understand them very well - and therefore we should be able to fully document each feature

hadijahkyampeire:
@ZacButko @ibacher I added some documentation in some Md regarding how the component will be used but I might have not exhausted all places, I will check again.

brandones:
Has someone looked to see if this should just be added as a feature to the existing notification component?

hadijahkyampeire:
@brandones I looked at that but the existing inline notification no longer accepts actions, they created a new component, that's why I am adding it as a new component.

packages/framework/esm-styleguide/src/notifications/_actionable-notifications.scss
```diff
@@ -0,0 +1,19 @@
+.omrs-actionable-notifications-container {
+  z-index: 1000;
+  position: fixed;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  margin-left: 0.2rem;
+  bottom: 2rem;
```
vasharma05, line 8:
```suggestion
  bottom: spacing.$spacing-07;
```

vasharma05 (approved):
Great work @hadijahkyampeire!

PR #557, opened 2022-10-13
(fix): make app shell dependencies more self-contained

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Really, this is about making the app shell dependencies more explicit. Basically, we should be importing things from esm-framework rather than via it's component packages, except the CSS from esm-styleguide which doesn't appear to be directly exported...

### DIFF
--- a/packages/shell/esm-app-shell/package.json
+++ b/packages/shell/esm-app-shell/package.json
@@ -13,7 +13,7 @@
     "analyze": "webpack --mode=production --env analyze=true",
     "watch": "webpack serve --mode development",
     "lint": "eslint src --ext ts,tsx",
-    "version": "yarn add @openmrs/esm-framework@$npm_package_version --exact"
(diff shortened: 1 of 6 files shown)


denniskigen:
Are the broken CI builds to be expected with these changes, @ibacher?

ibacher:
No... This one worked on my machine and broke on CI... I haven't tracked down why yet.

ibacher:
Alright... I must not have tested this properly. It seems that the `workbox-webpack-plugin` spawns it's own copy of Webpack with a similar configuration but a different set of plugins. Once I figured that out, it was easy to just add the`MiniCssExtractorPlugin` to the workbox plugin list. However, doing this while changing the imports in the service worker blew-up the service worker size to over 1 MB. 

This leads me to believe that `esm-framework` itself is not tree-shakable or is setup incorrectly to do that, which might account for the absurd size of our bundles (it's probably ok that this also balloons `openmrs.js` for now since we're actually including the whole thing).

PR #562, opened 2022-11-01
(chore) Use esbuild-loader to transpile app shell code

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Switches from `swc-loader` to `esbuild-loader` for transpiling code in `esm-app-shell`. The primary motivation for this change is to provide a workaround for an esoteric bug with the `swc-loader` in the app shell where the `import-map-overrides` library is not getting properly transpiled. The immediate benefit this change confers upon frontend developers is that the import map overrides panel is back working. This should hopefully result in a better DX for folks working on features for O3.

### DIFF
--- a/packages/shell/esm-app-shell/webpack.config.js
+++ b/packages/shell/esm-app-shell/webpack.config.js
@@ -157,7 +157,7 @@ module.exports = (env, argv = {}) => {
           type: "asset/source",
         },
         {
-          test: /\.(js|jsx)$/,
+          test: /\.jsx?$/,
           use: [
             {
               loader: require.resolve("esbuild-loader"),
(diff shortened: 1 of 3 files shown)


packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -157,10 +157,24 @@ module.exports = (env, argv = {}) => {
           type: "asset/source",
         },
         {
-          test: /\.(j|t)sx?$/,
+          test: /\.(js|jsx)$/,
```
ibacher, line 160: Personally, I find this easier to read...

```suggestion
          test: /\.jsx?$/,
```

ibacher (approved):
LGTM

PR #565, opened 2022-11-04
(chore) Set `minimum-change-threshold` value

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Sets a minimum file size change threshold value of 10 kb for the bundle size reporter so the action doesn't report on tiny changes.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -15,4 +15,4 @@ jobs:
       - uses: preactjs/compressed-size-action@v2
         with:
           build-script: "turbo run build --color"
-          minimum-change-threshold: 10000 # 10 KB
+          minimum-change-threshold: 500 # bytes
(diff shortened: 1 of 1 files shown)


.github/workflows/bundle-size.yml
```diff
@@ -15,3 +15,4 @@ jobs:
       - uses: preactjs/compressed-size-action@v2
         with:
           build-script: "turbo run build --color"
+          minimum-change-threshold: 10000 # 10 KB
```
ibacher, line 18: Seems a bit large no? Since we don't fail the build for this, I'd probably go with, e.g. 1 kb or maybe evey 500 b... It's just informative after all.

denniskigen: Fair. Amending.

ibacher (approved):

PR #577, opened 2022-12-02
feat:(cli) Add the ability to generate a manifest of actually assembled versions

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

The idea here is to have, as part of the assembled frontend, a manifest file which declares all of there versions in the right format to be re-consumed by the assemble command. Basically, this allows the assemble command to be repeatably run.

### DIFF
--- a/packages/tooling/openmrs/src/commands/assemble.ts
+++ b/packages/tooling/openmrs/src/commands/assemble.ts
@@ -20,6 +20,7 @@ export interface AssembleArgs {
   config: string;
   registry: string;
   fresh: boolean;
+  manifest: boolean;
 }
 
 interface NpmSearchResult {
@@ -160,6 +161,7 @@ async function extractFiles(sourceFile: string, targetDir: string) {
   const packageJson = JSON.parse(
     files[`${packageRoot}/package.json`].toString("utf8")
   );
+  const version = packageJson.version ?? "unknown";
   const entryModule =
     packageJson.browser ?? packageJson.module ?? packageJson.main;
   const fileName = basename(entryModule);
@@ -177,13 +179,16 @@ async function extractFiles(sourceFile: string, targetDir: string) {
     });
 
   unlinkSync(sourceFile);
-  return fileName;
+  return [fileName, version];
 }
 
 export async function runAssemble(args: AssembleArgs) {
   const config = await readConfig(args.mode, args.config, args.registry);
   const importmap = {
     imports: {},
+  };
+  const versionManifest = {
+    frontendModules: {},
   };
 
   logInfo(`Assembling the importmap ...`);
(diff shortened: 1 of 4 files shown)


packages/tooling/openmrs/src/commands/assemble.ts
```diff
@@ -160,6 +161,7 @@ async function extractFiles(sourceFile: string, targetDir: string) {
   const packageJson = JSON.parse(
     files[`${packageRoot}/package.json`].toString("utf8")
   );
+  const version = packageJson.version ?? "unknown";
```
FlorianRappl, line 164: I'd rather prefer something like `0.0.0` as its still a valid version number, but indicating that something is missing.

With the given value we have an invalid version - i.e., some tools might panic here. If this is what we want I'd say let's throw an error right away instead of failing with these tools - or have a proper version as a fallback value here.

packages/tooling/openmrs/src/commands/assemble.ts
```diff
@@ -221,4 +227,12 @@ export async function runAssemble(args: AssembleArgs) {
     JSON.stringify(importmap, undefined, 2),
     "utf8"
   );
+
+  if (args.manifest) {
+    writeFileSync(
```
FlorianRappl, line 232: The rest of the function is async - should we make this async too?

Could be rather straight forward when using `fs/promises`, e.g., `import { writeFile } from 'fs/promises'` => `await writeFile(...)`.

packages/tooling/openmrs/src/commands/assemble.ts
```diff
@@ -177,14 +179,17 @@ async function extractFiles(sourceFile: string, targetDir: string) {
     });
 
   unlinkSync(sourceFile);
-  return fileName;
+  return [fileName, version];
```
FlorianRappl, line 182: Are we only using `extractFiles` here? Maybe we should not `export` it then, or? Not sure - anyway in general looks fine - I just did not see anything else being changed when the return type changed here. Not there is a silent (i.e., non-TS guarded) error awaiting here...

FlorianRappl (commented):
LGTM - great work!

packages/tooling/openmrs/src/commands/assemble.ts
ibacher: It's already not exported right? Just async.

packages/tooling/openmrs/src/commands/assemble.ts
ibacher: Cool... yeah... I think we only had the typings for Node 12, so quite a bit behind the times here.

packages/tooling/openmrs/src/commands/assemble.ts
FlorianRappl: Right, maybe I overlooked this.

PR #579, opened 2022-12-09
O3-1594: Added the event listener to record the origin page for the patient chart

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR adds the event listener to record the origin page for the patient-chart, so whenever we exit from the patient chart, we should be directed back to the origin page from where we jumped into the patient chart.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-1594

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import {
   LoggedInUser,
   useLayoutType,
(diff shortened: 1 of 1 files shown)


ibacher:
I don't love this living in core at all, but I particularly don't like it in this app. Perhaps we can add a listener to the framework that stores the URL in the before-routing event and then call that state inside the patient chart?

vasharma05:
Where will be a good place to put this code into the framework?
I was also hesistant to put it in the primary navigation app, but thought putting it in the app shell might increase the complexity maybe.

hadijahkyampeire:
@vasharma05 were you able to implement this as Ian suggested in the comments above?

brandones:
@vasharma05 This work is now quite stale, would you mind taking the lead on it?

vasharma05:
Yes @brandones , this is on my radar for long time now, but something or other came through. 
I'll be on it. 
Thanks!

PR #580, opened 2022-12-11
O3-1444: Expired login session keeps the location page stuck

jnsereko:
## Requirements

#### For changes to apps

#### If applicable

## Summary

The session is saved as a global store using developit/unistore (https://github.com/developit/unistore).  This session object is only updated either by clearing the session store (https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-api/src/shared-api-objects/current-user.ts#L188), or by refetching the current user (https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-api/src/shared-api-objects/current-user.ts#L163).  

However, clearing or refetching the session store depends on only timeout and if the store is loaded, but doesn't check for the availability the JSESSIONID cookie.  At the same time, this cookie is an HTTP-ONLY cookie so it ain't available via document.cookie. 

to make this work, i manually opted to verifying that the user is logged in twice.  As this is working, it might be not working wright.

cc @brandones @denniskigen @vasharma05 @ibacher @hadijahkyampeire 

## Screenshots

https://user-images.githubusercontent.com/58003327/206906938-c784ce8d-3223-417b-9a81-934395690ffe.mov

## Related Issue

https://issues.openmrs.org/browse/O3-1444

### DIFF
--- a/packages/apps/esm-login-app/src/login/login.component.tsx
+++ b/packages/apps/esm-login-app/src/login/login.component.tsx
@@ -9,7 +9,14 @@ import {
 } from "@carbon/react";
 import { ArrowLeft, ArrowRight } from "@carbon/react/icons";
 import { useTranslation } from "react-i18next";
-import { useConfig, interpolateUrl, useSession } from "@openmrs/esm-framework";
+import {
+  useConfig,
+  interpolateUrl,
+  useSession,
+  refetchCurrentUser,
+  clearCurrentUser,
+  getSessionStore,
+} from "@openmrs/esm-framework";
 import { performLogin } from "./login.resource";
 import styles from "./login.scss";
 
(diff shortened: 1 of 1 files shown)


jnsereko:
AFAIK, If there is any way we can add a check  at https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-api/src/shared-api-objects/current-user.ts#L107-L108 so that the session object is refetched with absence of the `JSESSIONID cookie`, it would really be the simplest alternative. 

```
  if (
    lastFetchTimeMillis < Date.now() - 1000 * 60 ||
    !sessionStore.getState().loaded  || 
    document.cookie.split('; ').find((row) => row.startsWith('JSESSIONID'))
  ) {
    refetchCurrentUser();
  }
```

hadijahkyampeire:
@jnsereko I just checked on dev3 and the behavior is the same, when I remove the session cookie, the user gets redirected to the login, it is confusing whether we still have this issue or not

jnsereko:
> @jnsereko I just checked on dev3 and the behavior is the same, when I remove the session cookie, the user gets redirected to the login, it is confusing whether we still have this issue or not

Do you mind trying again @hadijahkyampeire.  I know this might sound silly but this issue sometimes responds differently mostly basing on how fast or slow your internet is.  It happens so fast that the location page ain't displayed but **choose location** URL appears in the **Address Bar**    
So, while am testing this, i actually ignore the displayed components and focus on the Address bar's URL changes.  After deleting the cookie, it should only redirect to `/login` and not `/login/location` first, then `/login` later on

hadijahkyampeire (approved):
Works well, thanks @jnsereko

PR #588, opened 2023-01-04
(feat)  Added property to FHIR resource so it can support DateTime values

icrc-agomes:
## Requirements

#### For changes to apps

#### If applicable

## Summary
There is a need for the FHIRResource to have a property destined for the date values.

### DIFF
--- a/packages/framework/esm-api/src/types/fhir-resource.ts
+++ b/packages/framework/esm-api/src/types/fhir-resource.ts
@@ -10,7 +10,7 @@ export interface FHIRResource {
     issued: Date;
     referenceRange: any;
     resourceType: string;
-    status: string;
+    status: string;ß
     subject: {
       display: string;
       identifier: { id: string; system: string; use: string; value: string };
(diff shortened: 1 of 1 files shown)


packages/framework/esm-api/src/types/fhir-resource.ts
```diff
@@ -22,6 +22,7 @@ export interface FHIRResource {
     };
     valueString: string;
     valueCodeableConcept: { coding: Array<FHIRCode> };
+    valueDateTime: string;
```
vasharma05, line 25: This change will be made through the whole app, might raise issues at various places.

```suggestion
    valueDateTime?: string;
```

icrc-agomes: @vasharma05 Done!
Thanks

packages/framework/esm-api/src/types/fhir-resource.ts
```diff
@@ -10,7 +10,7 @@ export interface FHIRResource {
     issued: Date;
     referenceRange: any;
     resourceType: string;
-    status: string;
+    status: string;ß
```
denniskigen, line 13: Could you get rid of the ß character?

PR #589, opened 2023-01-04
(fix) Fix lodash import

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Fixes a lodash import, ensuring `isEqual` gets imported from `lodash-es` instead of `lodash`, as Is the case with other usages of the library in this package.  The corollary of this change is that the overall bundle size gets reduced by around 39 kB (2%).

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -628,7 +628,7 @@ ___
 
 #### Defined in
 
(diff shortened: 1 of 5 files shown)


ibacher (approved):

ibacher:
Surprised this actually has a noticeable size impact!

denniskigen:
I was not after seeing the bundle size analysis.

PR #599, opened 2023-01-31
Added prerequisite that an O3 instance needs to be created first

gracepotma:
Since we saw today that a team used this prereq guide first and got confused when the esm-template-app was all they got out of it (didn't yet have an O3 instance).


denniskigen (approved):

denniskigen (approved):
Thanks, Grace.

PR #602, opened 2023-02-01
(docs) Update an outdated FAQ about importmaps

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Since we no longer embed a version of the importmap in the app-shell the FAQ item here no longer applies. However, I suppose there's some utility in having an explicit note about how to set the import map, so I've re-worded things.


denniskigen (approved):
Thanks, Ian!

PR #603, opened 2023-02-03
(release) Release v4.1.0

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Bumps package versions to `v4.1.0`, which corresponds to the latest [minor release](https://github.com/openmrs/openmrs-esm-core/releases/tag/v4.1.0).

### DIFF
--- a/packages/tooling/openmrs/package.json
+++ b/packages/tooling/openmrs/package.json
@@ -65,6 +65,6 @@
     "@types/react": "^18.0.9",
     "@types/rimraf": "^2.0.2",
     "@types/systemjs": "^6.1.0",
-    "@types/tar": "^4.1.0"
+    "@types/tar": "^4.0.3"
   }
 }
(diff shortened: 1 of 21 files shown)


packages/tooling/openmrs/package.json
```diff
@@ -65,6 +65,6 @@
     "@types/react": "^18.0.9",
     "@types/rimraf": "^2.0.2",
     "@types/systemjs": "^6.1.0",
-    "@types/tar": "^4.0.3"
+    "@types/tar": "^4.1.0"
```
ibacher, line 68: I think this was a mistaken find and replace? Otherwise, we need an update to yarn.lock

ibacher: PS It should be possible to (I guess with a Yarn plugin) just run `yarn version 4.1.0` to get the right behaviour.

ibacher (approved):

PR #612, opened 2023-02-20
(feat) Added styles for overflow menus in data tables

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR fixes the data-table's overflow menu stylings. Currently the overflow menu takes up space in the data table, as shown in the image below:
image

As per these fixes, the tables will have the overflow menus, at the end of the tables, taking up less space and giving the other content in the table more space to expand.

image

image

## Screenshots
Attached above

## Related Issue
*None*

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -232,3 +232,33 @@
 .cds--pagination-nav__page:focus {
   outline: 2px solid var(--brand-03);
 }
+
+.cds--data-table--xs .cds--overflow-menu {
+  height: spacing.$spacing-07;
+}
+
+.cds--data-table--sm .cds--overflow-menu {
+  height: spacing.$spacing-08;
+}
+
+.cds--data-table--sm .cds--overflow-menu--sm {
+  height: spacing.$spacing-07;
+}
+
+.cds--data-table--xs .cds--overflow-menu {
+  max-height: spacing.$spacing-07;
+}
+
+.cds--data-table-container {
+  padding-top: 0;
+}
+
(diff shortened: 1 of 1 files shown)


hadijahkyampeire (approved):
code looks good, just wondering, were you able to test all the tables @vasharma05

packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -233,6 +233,36 @@
...
+}
+
+.cds--data-table--sm .cds--overflow-menu--sm {
+  height: spacing.$spacing-07;
+}
+
+.cds--data-table--xs .cds--overflow-menu {
+  max-height: spacing.$spacing-07;
+}
+
```
denniskigen, line 251:
```suggestion
.cds--data-table--lg .cds--overflow-menu {
  height: spacing.$spacing-08;
  width: spacing.$spacing-08;
}
```

An example of where this would get used is the `All Encounters` table on the `Visits` page (Note how the overflow menu button gets scrunched when viewing in tablet mode).

denniskigen (approved):
Overall, very good. One other suggestion would be to move this code block containing Overflow Menu style overrides higher up in the file (say, to just below the `Datatables with expansion` overrides and above the `Misc` section). Then you could add a comment that says `/* Overflow Menus */` above that section so someone can tell immediately what they're looking at when parsing that bit of the file.

packages/framework/esm-styleguide/src/_overrides.scss
vasharma05: Thanks a lot for the addition @denniskigen !

PR #615, opened 2023-02-22
(fix) Functions wrapped in an extension should have correct arguments

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Extensions take a `wrap` argument, which allows them to wrap the slot with some component if desired. This was deprecated to support just including the function as the body of the extension, but they aren't exactly equivalent. This basically solves that, so we can finally kill the `wrap` argument.


denniskigen (approved):

denniskigen:
Thanks, @ibacher.

PR #617, opened 2023-02-24
(refactor) Refactor primary navigation

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR refactors the primary navigation widget with the aim of improving performance. The significant change it introduces is how the session gets loaded. Specifically, instead of using `getSynchronizedCurrentUser` to fetch the session object, we use the `useSession` hook instead. In local testing, the navbar gets loaded faster each time compared to the previous implementation. The potential downside of this change is that changes made in offline mode might not sync as well as they did when using `getSynchronizedCurrentUser`.

Additionally, this PR moves some logic from the root component into the Navbar component. These include functions that fetch the current user session and log the user out.

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -1,11 +1,5 @@
-import React, { useCallback, useMemo, useState } from "react";
-import {
-  LoggedInUser,
-  useLayoutType,
-  ExtensionSlot,
-  ConfigurableLink,
-  useAssignedExtensions,
-} from "@openmrs/esm-framework";
+import React, { memo, useCallback, useMemo, useState } from "react";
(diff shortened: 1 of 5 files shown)


packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -66,127 +62,131 @@ const Navbar: React.FC<NavbarProps> = ({
...
+          allowedLocales={allowedLocales}
+          onLogout={logout}
+          hidePanel={hidePanel}
+        />
+      </Header>
+    </>
   );
 
-  return <div>{session && <HeaderContainer render={render} />}</div>;
+  if (user) {
```
donaldkibet, line 174: Could we also have a check to see the session location is also set because it's critical with the way we use o3

donaldkibet (approved):
LGTM Just one suggestion

PR #620, opened 2023-02-28
(fix) Initialize config schema for esm-offline-tools-app

icrc-jofrancisco:
## Requirements
- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.

#### For changes to apps
- [x] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).

#### If applicable
- [x] My work includes tests or is validated by existing tests.
- [x] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary

This pull request updates the  esm-offline-tooks-app schema initialization process to support the  "Display conditions" and "Translation overrides" features in the existing enhancedSchema configuration.

By initialising the schema with this update, the "Display conditions" and "Translation overrides" will be avaiable, which can be particularly useful to display the app based on privileges and multilingual projects.

Thanks,

## Screenshots
None

## Related Issue
None

PR #621, opened 2023-02-28
(feat) Add implicit config for all registered apps (O3-1933)

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Adds an implicit configuration for every application when it is registered with the framework. This configuration is empty, but should be overrideable by any app calling `defineConfigSchema()` themselves.

## Related Issue

https://issues.openmrs.org/browse/O3-1933

### DIFF
--- a/packages/shell/esm-app-shell/src/apps.ts
+++ b/packages/shell/esm-app-shell/src/apps.ts
@@ -7,6 +7,7 @@ import {
   registerExtension,
   ResourceLoader,
   Lifecycle,
+  defineConfigSchema,
 } from "@openmrs/esm-framework/src/internal";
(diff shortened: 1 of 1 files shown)


denniskigen (approved):
Will we have to undo per-app invocations such as https://github.com/openmrs/openmrs-esm-core/pull/620/files as a corollary of this change?

ibacher:
@denniskigen No... it should just mean that we no longer need to write changes like that... Calling `defineConfigSchema` multiple times is harmless (as long as, you know, you don't overwrite a config schema with an empty one... I put this before the call to `setupOpenmrs()` so that the app can define the schema there).

PR #625, opened 2023-03-03
(feat) Show more useful global errors

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR amends the global error handling logic so that it displays the actual error messages instead of the uninformative `Oops! An unexpected error occurred` line. For context, both screenshots below relate to a situation where the app couldn't get a session object from the backend. 

## Screenshots

> Before

Screenshot 2023-03-03 at 16 03 56

> After
Screenshot 2023-03-03 at 15 57 39

### DIFF
--- a/packages/framework/esm-error-handling/src/index.ts
+++ b/packages/framework/esm-error-handling/src/index.ts
@@ -4,17 +4,18 @@ import { dispatchNotificationShown } from "@openmrs/esm-globals";
 window.onerror = function (error) {
   console.error("Unexpected error: ", error);
   dispatchNotificationShown({
-    description: error,
+    description: error ?? "Oops! An unexpected error occurred.",
     kind: "error",
     title: "Error",
   });
   return false;
 };
 
-window.onunhandledrejection = function (error) {
-  console.error("Unhandled rejection: ", error);
+window.onunhandledrejection = function (event) {
+  console.error("Unhandled rejection: ", event.reason);
   dispatchNotificationShown({
-    description: error?.message,
+    description:
+      event.reason ?? "Oops! An unhandled promise rejection occurred.",
     kind: "error",
     title: "Error",
   });
--- a/packages/framework/esm-error-handling/src/index.ts
+++ b/packages/framework/esm-error-handling/src/index.ts
@@ -1,16 +1,22 @@
 /** @module @category Error Handling */
 import { dispatchNotificationShown } from "@openmrs/esm-globals";
 
-window.onerror = function () {
+window.onerror = function (error) {
+  console.error("Unexpected error: ", error);
   dispatchNotificationShown({
-    description: "Oops! An unexpected error occurred.",
+    description: error,
+    kind: "error",
+    title: "Error",
   });
   return false;
 };
 
-window.onunhandledrejection = function () {
+window.onunhandledrejection = function (error) {
+  console.error("Unhandled rejection: ", error);
   dispatchNotificationShown({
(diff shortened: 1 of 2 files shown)


packages/framework/esm-error-handling/src/index.ts
```diff
@@ -1,16 +1,22 @@
...
   });
   return false;
 };
 
-window.onunhandledrejection = function () {
+window.onunhandledrejection = function (error) {
+  console.error("Unhandled rejection: ", error);
   dispatchNotificationShown({
-    description: "Oops! An unexpected error occurred.",
+    description: error?.message,
```
ibacher, line 17: Maybe this, i.e., if we don't have a message, go back to the old thing?

```suggestion
    description: error?.message ?? "Oops! An unexpected error occurred.",
```

packages/framework/esm-error-handling/src/index.ts
```diff
@@ -1,16 +1,22 @@
 /** @module @category Error Handling */
 import { dispatchNotificationShown } from "@openmrs/esm-globals";
 
-window.onerror = function () {
+window.onerror = function (error) {
+  console.error("Unexpected error: ", error);
   dispatchNotificationShown({
-    description: "Oops! An unexpected error occurred.",
+    description: error,
```
ibacher, line 7: Any reason this is different from the other?

ibacher (approved):

packages/framework/esm-error-handling/src/index.ts
denniskigen: Different APIs for `onError` and `onUnhandledRejection`. 

`onunhandledrejection` takes an `event` object of type `PromiseRejectionEvent`. The value or Object indicating why the promise was rejected is contained in its `reason` property.  

`onerror`s signature includes five arguments: `event`, `source`, `lineno`, `colno`, and  `error`. The error message we care about is in `error.message` or `event.message`.

packages/framework/esm-error-handling/src/index.ts
```diff
@@ -1,16 +1,23 @@
...
   return false;
 };
 
-window.onunhandledrejection = function () {
+window.onunhandledrejection = function (event) {
+  console.error("Unhandled rejection: ", event.reason);
   dispatchNotificationShown({
-    description: "Oops! An unexpected error occurred.",
+    description:
+      event.reason ?? "Oops! An unhandled promise rejection occurred.",
```
denniskigen, line 18: Should I stick with `Oops! An unexpected error occurred` as above?

ibacher: I think so... I mean, I don't love having a non-descript error message, but it's better than having an error with no message at all...

denniskigen: I meant the default. We're currently showing `Oops! An unexpected error occurred` for both unhandled promise rejections and `window.onerror`. I've changed it up to show a slightly different string for unhandled promise rejections: `Oops! An unhandled promise rejection occurred.`

ibacher: Oh... I missed that... Sorry! I'm of two minds about this. "unhandled promise rejection" is more technically correct, but gibberish to end users. On the other hand, if they report this to a support team, the support team might have an idea what's going on, whereas "unexpected error" is likely always unhelpful.

I guess my preference would be to keep "unhandled promise rejection".

PR #626, opened 2023-03-03
(feat) Optional typings for useConfig()

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

In our code, we sometimes have types defined for config objects, which results in them being used like so:

```ts
const config = useConfig() as MyConfigType;
```

This PR allows us to write:

```ts
const config = useConfig<MyConfigType>();
```

Which could be easily wrapped as something like:

```ts
const useMyConfig = useConfig<MyConfigType>;
```


denniskigen (approved):
LGTM

PR #627, opened 2023-03-03
(fix) Better typings in loadModules()

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

A small collection of TS-related adjustments. The most important change here is some runtime type-checking when loading modules to ensure that the thing we're loading from the `window` object looks like a MF module.


denniskigen (approved):
Nice! Thanks

PR #632, opened 2023-03-07
BREAKING: Migrate from Unistore -> Zustand

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This migrates core from using Unistore to using Zustand. Zustand is very similar to Unistore and, indeed, their API's are largely identical (both heavily inspired by Redux / Flux). Unistore is pretty stable (and very simple) but quite a bit less popular, which suggests its a somewhat better target in terms of long-term supportability. It also has built-in support for React Hooks, although this PR does not try to convert us over to those hooks.

There are two main API differences:

1. Unistore supports registering actions with a store at any time. Zustand only supports actions registered on a store at creation time, but, these are generally easy to replace with setState() (and our only use of this functionality is [here](https://github.com/openmrs/openmrs-esm-core/blob/9c351f84e3a770996167ae4614670455590295f9/packages/apps/esm-implementer-tools-app/src/ui-editor/ui-editor.tsx#L73)).
2. Unistore returns a four function API, including an unsubscribe() function. Zustand returns an unsubscribe function, but only as the result of a call to subscribe.

While we are using stores in a few places outside of core, it's only in Core that we have a particularly strong dependency on the Unistore API differences AFAICT. I've marked this change as breaking, however, since there are now a couple of API functions on the store itself that are no longer available. Our wrapper APIs (which is what everyone should be using) remain unchanged (apart from Typescript adjustments for the now missing APIs).


ibacher:
@denniskigen I'm not 💯 this is a good idea, so I'd like your opinion on that too, in addition to any normal checks.

packages/framework/esm-react-utils/src/createUseStore.ts
```diff
@@ -1,40 +1,46 @@
...
     useEffect(() => store.subscribe((state) => set(state)), []);
-    let boundActions: BoundActions = {};
-
-    if (actions) {
-      boundActions = useMemo(
-        () => bindActions(store, actions),
-        [store, actions]
-      );
-    }
+    let boundActions: BoundActions = useMemo(
```
ibacher, line 40: This change (with its partner) is the not directly related to the migration to Zustand, but rather the rules of hooks

PR #641, opened 2023-03-13
(feat) Add support for caching the importmap

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Allows us to generate content hashes for importmaps so they can be cached.

Previously, we've relied on the pattern of running:
  openmrs build
  openmrs assemble

In that order. Since the build step needs to know the importmap URL, to take advantage of this you need to run:
  openmrs assemble --hash-importmap
  openmrs build

The build step should automatically detect the importmap (based on name).

This PR also changes our default hash function to xxhash64, which is slated to be the default hash function for Webpack 6 and should allow us to avoid having to use `--legacy-ssl-provider` when running the build and assemble steps (this is required because Webpack's default hash is MD4, which is very fast, but very broken; xxhash64 is intentionally **not** a cryptographic algorithm).

## Related Issue

https://issues.openmrs.org/browse/O3-1938

### DIFF
--- a/packages/shell/esm-app-shell/webpack.config.js
+++ b/packages/shell/esm-app-shell/webpack.config.js
@@ -52,6 +52,7 @@ module.exports = (env, argv = {}) => {
       chunkFilename: "[chunkhash].js",
       path: resolve(__dirname, outDir),
       publicPath: "",
+      hashFunction: "xxhash64",
     },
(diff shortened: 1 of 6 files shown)


denniskigen (approved):
LGTM. Thanks, Ian.

PR #644, opened 2023-03-14
(fix) O3-1492: Remove dangerous global selectors from stylesheets

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Sets up global overrides for the `SideNav` and `Tooltip` carbon components in the style guide. This is the canonical way to set up style overrides without polluting the global scope. 

## Related Issue
https://issues.openmrs.org/browse/O3-1492

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -32,6 +32,16 @@
   border-right: none;
 }
 
+/* Side nav */
+.cds--side-nav {
+  z-index: 0;
+}
+
+/* Tooltip */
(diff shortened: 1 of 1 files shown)


ibacher (approved):

packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -32,6 +32,16 @@
   border-right: none;
 }
 
+/* Side nav */
+.cds--side-nav {
```
ibacher, line 36: Although, I assume this is reasonable if we're using Carbon's UI shell layout

denniskigen: We are.

PR #646, opened 2023-03-15
(fix) Properly scope hyperlink styles for the left nav menu

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Fixes the appearance of hyperlinks in the app after I broke them by mistakenly applying [these styles globally](https://github.com/openmrs/openmrs-esm-core/blob/a98a7b15cc16b615cce4b19f9d3133faf5f2b74e/packages/framework/esm-styleguide/src/left-nav/left-nav.module.scss#L44). This has borked the display of things like breadcrumbs in the app.

This PR fixes that by properly scoping the styles affecting hyperlinks in the left nav menu under the `leftNav` class name.

## Screenshots

> Broken breadcrumbs menu due to height override

Screenshot 2023-03-15 at 12 11 44 PM

### DIFF
--- a/packages/framework/esm-styleguide/src/left-nav/left-nav.module.scss
+++ b/packages/framework/esm-styleguide/src/left-nav/left-nav.module.scss
@@ -22,6 +22,14 @@
     }
 
     a {
+      :global(.omrs-breakpoint-gt-tablet) {
+        height: 2rem;
+      }
+
+      :global(.omrs-breakpoint-lt-desktop) {
+        height: 3rem;
+      }
+
       color: $ui-04;
       @include type.type-style("heading-compact-01");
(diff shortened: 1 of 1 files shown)


hadijahkyampeire (approved):

vasharma05 (approved):
We can replace 2rem and 3rem with spacing variables.

denniskigen:
> We can replace 2rem and 3rem with spacing variables.

Wouldn't be overly worried about that. If necessary, we could run a codemod at some point in the future that would convert all these relative values to their corresponding spacing tokens. I prefer using relative values because they don't have the overhead of needing to resolve what each token corresponds to what value in length.

PR #649, opened 2023-03-22
(feat) Clear SWR cache on logout

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Does what it says, that is, clears the SWR cache when the user clicks the logout button. This hopefully prevents the user from seeing stale data that they shouldn't in the case where:

1. A more privileged user (e.g., System Developer) logs in and then logs out
2. Then, without reloading the page, a less privileged user logs in to the app

Without this change, the SWR cache is storing the results from session 1 as the "stale" data, which means both that the less privileged user in session 2 will see this stale data "flash".

Related Slack conversation: https://openmrs.slack.com/archives/CHP5QAE5R/p1678956705607029


vasharma05 (approved):
Thanks for working on this!

PR #653, opened 2023-03-28
Add generate Date Format function

nanfuka:
## Requirements
- [ ] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

#### For changes to apps

#### If applicable

## Summary
The date format used in datepickers  computed based on the locale

## Related Issue
https://issues.openmrs.org/browse/O3-998

### DIFF
--- a/packages/framework/esm-utils/src/omrs-dates.test.ts
+++ b/packages/framework/esm-utils/src/omrs-dates.test.ts
@@ -2,6 +2,7 @@ import {
   toOmrsIsoString,
   toDateObjectStrict,
   isOmrsDateStrict,
+  generateFormatting,
 } from "./omrs-dates";
 import dayjs from "dayjs";
 import timezoneMock from "timezone-mock";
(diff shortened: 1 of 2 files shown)


ibacher (commented):
The real thing this PR needs, IMO, is some tests. While we don't use unit tests extensively in the apps, relying on them being tested by e2e test, for functionality like this in core, we really should have a few tests using different locales to ensure this outputs the right thing.

nanfuka:
@ibacher I have added the tests. kindly review

brandones:
I think this has very little to do with the requirements. I don't know what `parse` and `format` are supposed to do or why they are present. `placeholder` is no longer used in [Carbon React DatePicker](https://react.carbondesignsystem.com/?path=/story/components-datepicker--simple) and therefore shouldn't be generated. `dateFormat`, is just a static format string which is not localized in any way. Generating a locale-specific date format string is basically the whole point of this ticket.

PR #655, opened 2023-03-30
BREAKING: Refactor `usePatient` to account for offline patients

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Draft PR for changes to `usePatient()` hook, which allow it take account of any offline-registered patients. Previously, this was only handled in the patient chart. In essence, this works by adding an optional, but default-on check of the offline registered patient queue for a patient that matches the UUID.

This PR is marked as breaking because it removes two APIs deprecated in 4.0.0:

* `useCurrentPatient()` (replaced by `usePatient()`)
* `fhir.read()` (replaced by nothing; `openmrsFetch()` works really well for this and `fhir.read()` has been implemented in terms of it for a while)

`useCurrentPatient()` isn't used anywhere I can find. `fhir.read()` is only used in the template app.

`fetchCurrentPatient()`'s return type is different, but more accurate. I've left this as a public API because it's useful in non-React contexts, but React apps should just be using `usePatient()`.

## Other

In draft because I haven't yet confirmed that this doesn't break offline patients.

I know I said that the `usePatientOrOfflineRegisteredPatient()` was a performance bottle-neck. So we're clear, the problem isn't that checking the offline patient queue is expensive, it's that the logic in that hook to determine whether it had actually resolved found a patient in either of those was convoluted and probably broken under some conditions.


packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -4,7 +4,11 @@ import isPlainObject from "lodash-es/isPlainObject";
 import { getConfig, navigate } from "@openmrs/esm-config";
 import { FetchResponse } from "./types";
 
-export const sessionEndpoint = "/ws/rest/v1/session";
+export const restBaseUrl = "/ws/rest/v1/";
+
+export const fhirBaseUrl = "/ws/fhir2/R4";
```
ibacher, line 9: These are inconsistent, but only because I didn't want to change `fhirBaseUrl`. The trailing `/` should be there, though.

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -126,10 +133,21 @@ export function openmrsFetch<T = any>(
...
    */
-  if (typeof fetchInit.headers["Disable-WWW-Authenticate"] === "undefined") {
+  if (
+    path.startsWith(restBaseUrl) &&
+    typeof fetchInit.headers["Disable-WWW-Authenticate"] === "undefined"
+  ) {
     fetchInit.headers["Disable-WWW-Authenticate"] = "true";
   }
 
+  if (path.startsWith(fhirBaseUrl)) {
```
ibacher, line 143: `_summary=data` skips generating the FHIR Narrative, which is a relatively expensive operation and we don't really use it anyway, so this ensures requests to the FHIR API always have `_summary=data` as part of the request.

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -303,7 +321,7 @@ export class OpenmrsFetchError extends Error {
   responseBody: string | FetchResponseJson | null;
 }
 
-interface FetchConfig extends Omit<Omit<RequestInit, "body">, "headers"> {
+export interface FetchConfig extends Omit<RequestInit, "body" | "headers"> {
```
ibacher, line 324: Exporting this because it's actually different from the `fetch()` API (we don't support the same "headers" formats)

denniskigen (approved):
LGTM. Thanks, Ian.

denniskigen:
Dunno why the pre-release is failing

PR #664, opened 2023-04-10
(fix) KH-126: Redirected to home page without selecting a location

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR fixes the issue of redirection to home page without selecting any location, mainly triggered when searching.

The cause of the issue was the below condition:
```
// Handle cases where the location picker is disabled, there is only one location, or there are no locations.
  useEffect(() => {
    if (!isLoading) {
      if (!config.chooseLocation.enabled || locations?.length === 1) {
        changeLocation(locations[0]?.resource.id);
      }
      if (!isLoading && !locations?.length) {
        changeLocation();
      }
    }
  }, [config.chooseLocation.enabled, isLoading, locations, changeLocation]);
```

Since the searching is done via backend's endpoint, cases where only 1 or no results were returned, the user was redirected to the home page without selecting any location.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/KH-126

### DIFF
--- a/packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
+++ b/packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
@@ -117,18 +117,6 @@ const LocationPicker: React.FC<LocationPickerProps> = ({
     return chooseLocation.numberToShow;
   });
 
-  // Handle cases where the location picker is disabled, there is only one location, or there are no locations.
-  useEffect(() => {
-    if (!isLoading) {
-      if (!config.chooseLocation.enabled || locations?.length === 1) {
-        changeLocation(locations[0]?.resource.id);
-      }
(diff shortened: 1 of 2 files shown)


donaldkibet:
@vasharma05 please provide a video of the changes you have mad. Secondly i think it was implemented such that if we have one location, there is no need of forcing the user to select it. I think we should support both cases, because in our use case having this feature is important. cc @CynthiaKamau @makombe for commet

vasharma05:
Got it, so what I can do is that I can make the useEffect only work when there's no searchString. Is that okay @donaldkibet ?

vasharma05:
Hi @donaldkibet , please re-review the changes.
Thanks!

donaldkibet (approved):
Thanks @vasharma05 LGTM

PR #665, opened 2023-04-11
(fix) User should be redirected to location picker if no location is found in session

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Currently, if the user by any chance skip the location picker and comes directly to the home page after login, there are many places where various APIs fails due to `undefined` location.

This fix will redirect the user to the location picker, if there's no location found in the session object.

SIde effect: The time to set location and moving to the home page increases due to the wait for updating the session object before moving ahead in the app.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/KH-125

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -171,8 +171,10 @@ const Navbar: React.FC = () => {
     </>
   );
 
-  if (user && session && !session.sessionLocation) {
-    return (
+  if (user && session) {
+    return session.sessionLocation ? (
+      <HeaderContainer render={memo(HeaderItems)}></HeaderContainer>
+    ) : (
       <Navigate
         to={`${openmrsSpaBase}login/location`}
         state={{
@@ -184,10 +186,6 @@ const Navbar: React.FC = () => {
         }}
       />
     );
-  }
-
-  if (user && session) {
-    return <HeaderContainer render={memo(HeaderItems)}></HeaderContainer>;
   }
 
   return (
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -171,6 +171,21 @@ const Navbar: React.FC = () => {
     </>
   );
 
+  if (user && session && !session.sessionLocation) {
+    return (
+      <Navigate
+        to={`${openmrsSpaBase}login/location`}
+        state={{
+          referrer: window.location.pathname.slice(
+            window.location.pathname.indexOf(openmrsSpaBase) +
+              openmrsSpaBase.length -
+              1
+          ),
+        }}
+      />
+    );
+  }
+
   if (user && session) {
     return <HeaderContainer render={memo(HeaderItems)}></HeaderContainer>;
   }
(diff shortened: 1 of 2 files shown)


packages/framework/esm-api/src/shared-api-objects/current-user.ts
```diff
@@ -160,8 +160,8 @@ function isSuperUser(user: { roles: Array<Role> }) {
  * refetchCurrentUser()
  * ```
  */
-export function refetchCurrentUser() {
-  return new Promise((resolve, reject) => {
+export async function refetchCurrentUser() {
```
ibacher, line 163: There's no good reason to write things this way. It still returns a promise...

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -171,6 +171,21 @@ const Navbar: React.FC = () => {
     </>
   );
 
+  if (user && session && !session.sessionLocation) {
```
ibacher, line 174: Why not factor this into the other `if` where we're already checking if `user` and `session` are set?

ibacher (changes requested):

packages/framework/esm-api/src/shared-api-objects/current-user.ts
vasharma05: I did this to await for the session store to be updated with the latest data before it returns the response.

ibacher: So `async` / `await` under the hood basically just does the same thing as `Promise.then()`... If you don't have anything that does in the `then()` block, there's no reason to use it, just like you'd omit the `then()` and just return the promise to the caller.

packages/framework/esm-api/src/shared-api-objects/current-user.ts
```diff
@@ -249,5 +249,5 @@ export async function setSessionLocation(
     signal: abortController.signal,
   });
 
-  refetchCurrentUser();
+  await refetchCurrentUser();
```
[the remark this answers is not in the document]
vasharma05: Actually, the issue comes here that the POST request is made above and hence the redirection happens. The correct point of redirection is when the POST request is completed, the store should be updated before direction to home page. Otherwise the flow becomes:

Location picker -> submits POST requests -> initiates update to the session store (async) -> (meanwhile) directed to home page -> primary navigation sees no location in session object since the store is not yet updated -> redirected to location picker.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
vasharma05: Sure!

packages/framework/esm-api/src/shared-api-objects/current-user.ts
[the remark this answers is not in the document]
ibacher: Ah... I've got a better solution for this I'm working on anyway, so it's NBD.

packages/framework/esm-api/src/shared-api-objects/current-user.ts
vasharma05: Got it, will revert this. Thanks!

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
vasharma05: Done!

PR #670, opened 2023-04-27
(feat) Add @openmrs/esm-dynamic-loading

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This is a draft of an idea for offering a centralised service for dynamically loading named exports from modules. Essentially, this is attempting to provide framework-level support for writing things like openmrs/openmrs-esm-patient-chart#772.

Basically, this adds one function to the framework, `loadDynamicExport()` that can be used like this:

```js
const someComponent = loadDynamicExport("@openmrs/esm-my-app", "someComponent");
```

This will get the `someComponent` export from the `@openmrs/esm-my-app`'s `index.ts` file. This is just straight-forwardly moving the relevant code from that PR to a new part of esm-framework.

---

Alternative implementation to consider:

```js
const { someComponent } = importDynamic("@openmrs/esm-my-app");
```

That's probably more in keeping with how ESM's work.

---

### Use Cases

* Supporting custom data sources in the form engine
* Supporting custom helper functions in the form engine
* Sharing components across MFs without relying on the extension system

### DIFF
--- a/packages/framework/esm-dynamic-loading/README.md
+++ b/packages/framework/esm-dynamic-loading/README.md
@@ -1,0 +1,4 @@
+# openmrs-esm-dynamic-loading
+
+openmrs-esm-dynamic-loading provides functions for dynamically loading JS files using Webpack Module Federation
+dynamic remotes.
--- a/packages/framework/esm-dynamic-loading/jest.config.js
+++ b/packages/framework/esm-dynamic-loading/jest.config.js
@@ -1,0 +1,9 @@
+module.exports = {
+  transform: {
+    "^.+\\.tsx?$": ["@swc/jest"],
+  },
+  testEnvironment: "jsdom",
+  testEnvironmentOptions: {
+    url: "http://localhost/",
+  },
+};
--- a/packages/framework/esm-dynamic-loading/package.json
+++ b/packages/framework/esm-dynamic-loading/package.json
@@ -1,12 +1,13 @@
 {
-  "name": "@openmrs/esm-framework",
+  "name": "@openmrs/esm-dynamic-loading",
   "version": "4.3.1",
   "license": "MPL-2.0",
-  "browser": "dist/openmrs-esm-framework.js",
+  "description": "Utilities for dynamically loading code in OpenMRS",
+  "browser": "dist/openmrs-esm-dynamic-loading.js",
   "main": "src/index.ts",
   "source": true,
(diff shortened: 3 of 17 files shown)


ibacher:
P.S. I'm not entirely sure how to write a test for this. I've tested it locally by swapping out the implementation of `loadModules()` for a version that uses this:

```ts
export async function loadModules(modules: Record<string, string>) {
  return Promise.all(
    Object.entries(modules).map(async ([name, url]) => {
      if (!url || !url.match(/\.js$/)) {
        return [name, {}];
      }

      try {
        const module = importDynamic(name);
        return [name, module];
      } catch {
        try {
          const module = await System.import(name);
          return [name, module];
        } catch (error) {
          console.error(
            `Failed to load module ${name} using either the supported mechanism or the legacy loading mechanism.`,
            error
          );
          return [name, {}];
        }
      }
    })
  );
}
```

Which appears to work locally, so I'm marking it as ready-to-review.

denniskigen (approved):
This looks very good. Thanks, Ian.

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -0,0 +1,125 @@
...
+      loadScript(url, resolve, reject);
+    });
+  }
+
+  const container = window[jsPackageSlug] as unknown;
+  if (!isFederatedModule(container)) {
+    const error = `The global variable ${jsPackageSlug}  does not refer to a federated module`;
+    console.error(error);
+    throw new Error(error);
+  }
```
samuelmale, line 65: Post-merge comment: @ibacher, are we planning to provide support for non-federated packages? I might be mistaken, but I believe "@openmrs/esm-form-entry-app" serves as a good example of such packages.

ibacher: @samuelmale Short version, no. There's not really a strong need for that, AFAICT. There's one module that doesn't use federation, and that's form-entry-app, but the equivalent API for doing what this function enables is:

```ts
const { someComponent } = await System.import("@openmrs/esm-my-app");
```

To be clear, if you call this code from a non-federated package, it should work (as long as it was built with Webpack 5). This is really just a patch for the fact that there's no simple API for importing from a dynamic remote using the module federation pattern.

PR #671, opened 2023-05-05
BREAKING: Switch to using JSON metadata

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR is a draft of how we might approach minimizing what needs to be downloaded, at least in production.

The goal is to be able to swap the part of the AppShell implementation that loads everything from the importmap for an implementation that loads modules as they are required.

In essence, all this does is move stuff from the `setupOpenMRS()` function of the module into a file called `routes.json`. This JSON file describes all of the pages and extensions supported by the apps and provides the information to allow them to be loaded.

Things to be done:

- [x] Provide routes.json for all core apps. This is also an exercise in trying to work out what we need to support for a MVP, since the core apps have the weirdest routing requirements.
- [x] Changes to the assemble command in the CLI to take `routes.json` files and produce a "compiled" JSON file like this:

```
{
  "@openmrs/esm-login-app": /* contents of metadata.json from the login-app */,
  "@openmrs/esm-primary-navigation-app": /* contents of metadata.json from the primary-navigation-app */
}
```
- [x] Changes to the app shell itself to support loading this document
- [x] Changes to the app shell support dynamically loading the apps and extensions as they are needed (this will probably be accomplished via [this mechanism](https://github.com/openmrs/openmrs-esm-core/pull/670)).
- [x] Support for "core" extensions (extensions from esm-app-shell itself)
- [x] Figuring out how we continue to support actual development using a process similar to the one we have.
- [x] Changes to the implementer tools so the compiled routes can be inspected
- [x] Ensure we still support esm-form-entry-app via SystemJS
- [x] Support online / offline boolean properties for extensions
- [x] <s>Figure out how to convince webpack to watch routes.json for changes</s> (deferring this)
- [x] Create a [migration guide](https://o3-docs.vercel.app/docs/migration-guide#case-study-login-frontend-module) for frontend developers


packages/apps/esm-login-app/src/metadata.json
```diff
@@ -0,0 +1,63 @@
+{
+  "version": "__VERSION__",
+  "backendDependencies": {
+    "webservices.rest": ">=2.24.0"
+  },
+  "pages": [
+    {
+      "component": "root",
+      "route": "login",
```
ibacher, line 9: `route` is replacing the `activityFunction` in the old design. This means that we support strings and (hopefully) JSON-encoded regexes, but not more complicated logic. This is technically a feature degradation, but I don't think it's a feature anyone is actively using.

packages/apps/esm-login-app/src/metadata.json
```diff
@@ -0,0 +1,63 @@
...
+      "online": {
+        "isLoginEnabled": true
+      },
+      "offline": {
+        "isLoginEnabled": false
+      }
```
ibacher, line 15: For demo purposes, I've left these `online` and `offline` properties in, but I actually think prop-passing like this is less useful. We should investigate just supporting `useConnectivity()` in the affected components (or a similar API).

packages/apps/esm-login-app/src/metadata.json
```diff
@@ -0,0 +1,63 @@
...
+    }
+  ],
+  "extensions": [
+    {
+      "name": "location-picker",
+      "slot": "location-picker",
```
ibacher, line 31: Extensions can still declare a default slot (and order), but this will remain optional.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -255,6 +252,28 @@ export default (
           return obj;
         }, {}),
       }),
+      new CopyWebpackPlugin({
```
ibacher, line 255: This is really the change needed for this support; specifically, it copies `src/metadata.json` to the `dist` directory, substitutes the string `__VERSION__` (if any) with the current version number and then minifies the JSON.

packages/apps/esm-primary-navigation-app/src/routes.json
```diff
@@ -0,0 +1,46 @@
+{
+  "version": "__VERSION__",
+  "backendDependencies": {
+    "webservices.rest": ">=2.2.0"
+  },
+  "pages": [
+    {
+      "component": "root",
+      "activator": "rootPageActivator",
```
ibacher, line 9: Adding support for primary-navigation-app brings some new challenges, specifically around supporting pages that have more complicated "should I render" logic.

Single-spa applications support an [`activityFn`](https://single-spa.js.org/docs/api#simple-arguments) property, which matches this signature: `(location: Location) => boolean`. This is used to determine whether the "page" should be rendered currently. We abstract over this by allowing a frontend module to specify the page using a `route` property that can be either: a string, a regex, an `activityFn`, or an array of any of those. We use the [`preproccessActivator()`](https://github.com/openmrs/openmrs-esm-core/blob/73cbd50c748494bcb8d2316fdc7d23cf8794693d/packages/shell/esm-app-shell/src/apps.ts#L24) function to turn those into an `activityFn`.

Moving to a JSON-based configuration makes all of this a little bit more complex. With login-app (and, I suspect, most apps that actually render a page), we were always using a string as the `route` property, so adding a `route` property to the JSON that takes a string and gets handled as it does today is straight-forward enough. To support regexes (like we use in [patient-chart-app](https://github.com/openmrs/openmrs-esm-patient-chart/blob/f378562afba25ec2bd4767dfd26ba4732e10bf59/packages/esm-patient-chart-app/src/index.ts#L48), we can probably just add a `routeRegex` property (since this needs to be written in JSON, the Regex will need to be properly encoded so it seems worth having two properties to determine how we should parse things). However, where we are using `activityFn`s, like in primary-navigation-app, things get more complicated.

In particular, primary-navigation-app provides two "pages": one the UI shell, which is displayed on every page except the login page and two a "redirect" page that sets the default home page to "/home".

There are a few implementation options here:

1. We could support both of these pages via `routeRegex` (though the regex for "not the login" page is a little complicated)
2. We could support a "special" syntax for routes so that `"!login"` matches anything that is not the login page and "/" matches only the root page.
3. We can keep using an `activityFn`, like we do here.

I've implemented this using option 3 to see what it looked like and gather opinions. Option 3 has a serious draw-back, which is that we need to load the frontend module to be able to get the `activityFn` and we were trying to defer needing to load as many of these as possible. Now, for the primary-navigation-app, this is probably a non-issue because we almost certainly need to pre-load this app as part of the start-up. However, for non-core applications, we would like to avoid loading the code if at all possible.

packages/apps/esm-primary-navigation-app/src/components/choose-locale/change-locale.component.tsx
```diff
@@ -15,6 +23,23 @@ export interface ChangeLocaleProps {
   postSessionLocale: PostSessionLocale;
 }
 
+const ChangeLocaleWrapper: React.FC<
+  Pick<ChangeLocaleProps, "allowedLocales" | "user">
+> = (props) => {
+  const isOnline = useConnectivity();
+  const [postUserProperties, postSessionLocale] = useMemo(
```
ibacher, line 30: This is really the more important kind of change. Previous `ChangeLocale` relied on `postUserProperties` and `postSessionLocale` being properties whose value was set in `index.ts` to point to the appropriate function. It seemed fine to keep this functionality inside the frontend module, so I simply added a wrapper component that will do the right thing.

ibacher: Another thing that came up in the call with Joel as [this block](https://github.com/openmrs/openmrs-esm-core/blob/73cbd50c748494bcb8d2316fdc7d23cf8794693d/packages/shell/esm-app-shell/src/run.ts#L78), which re-registers every application on every change in connectivity status. This seems like overkill, since everything gets re-rendered. Pushing logic like this into the frontend module instead of making it a framework feature seems like a smarter way to go, and now all the `online` / `offline` property is telling us is if the page / extension should be rendered at all in the specific mode.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -142,6 +148,8 @@ export default (
   const srcFile = resolve(root, browser ? main : types);
   const ident = makeIdent(name);
   const frameworkVersion = getFrameworkVersion();
+  const routes = resolve(root, "src/routes.json");
+  const hasRoutesDefined = fileExistsSync(routes);
```
ibacher, line 152: TODO: In the future, we should output some kind of warning if someone defines a frontend module without including a `routes.json`. For now, I'm just aiming to get more things working.

ibacher (commented):
A few new commits, so some new notes on things.

packages/apps/esm-primary-navigation-app/src/routes.json
```diff
@@ -0,0 +1,46 @@
+{
+  "version": "__VERSION__",
+  "backendDependencies": {
+    "webservices.rest": ">=2.2.0"
+  },
+  "pages": [
+    {
+      "component": "root",
+      "routeRegex": "(?!login/?)",
```
ibacher, line 9: Alright, for these use-cases, the `routeRegex` looks much better. I'll probably just go with that.

packages/apps/esm-devtools-app/src/routes.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "__VERSION__",
+  "pages": [
+    {
+      "component": "devtools",
+      "routeActivation": "devtoolsActivation",
```
ibacher, line 6: Unlike with primary-navigation-app, we don't have a choice with devtools. The activation function is:

```ts
() => window.spaEnv === "development" || !!localStorage.getItem("openmrs:devtools");
```

ibacher: Two things to consider:

1. In `setupOpenMRS()`, apparently we could just return a bare object; We'd need to give it some kind of key here and I used `pages` for that purpose because `pages` are, generally speaking, what we convert to single-spa application, but it may not be the obvious name for this feature.
2. We've now created properties `route`, `routeRegex` and `routeActivation`. Does it make sense to continue with this or should we support:

```ts
route: string | {
  regex?: string,
  activation: string
}
```

Basically allowing us to write:

```json
{
  "route": "myRoute"
  /* alternatively */
  "route": { "regex": "^myRoute" }
  /* alternatively */
  "route": { "activation": "myActivationFn" }
}
```

packages/apps/esm-devtools-app/src/index.ts
```diff
@@ -1,26 +1,16 @@
...
-    ),
-    activate: () =>
-      window.spaEnv === "development" ||
-      !!localStorage.getItem("openmrs:devtools"),
-  };
-}
-
-export { setupOpenMRS, importTranslation, version };
+export const devtoolsActivation = () =>
+  window.spaEnv === "development" || !!localStorage.getItem("openmrs:devtools");
```
denniskigen, line 16:
```suggestion
  window.spaEnv === "development" || Boolean(localStorage.getItem("openmrs:devtools"));
```

routes.schema.json
```diff
@@ -0,0 +1,114 @@
...
+  "title": "OpenMRS Routes Schema",
+  "description": "This schema defines the expected values of routes.json files provided by OpenMRS frontend modules. These files describe the pages, extensions, and requirements of the frontend module so they can be appropriately consumed by the app shell.",
+  "type": "object",
+  "properties": {
+    "version": {
+      "type": "string",
+      "description": "The version of the module. This should always be the constant __VERSION__.",
+      "default": "__VERSION__",
+      "pattern": "^__VERSION__$"
+    },
```
ibacher, line 13: It was after I wrote this out that I started questioning if we even need this element here. It might be better just to add the version element automatically as part of the Webpack tooling. Thoughts?

routes.schema.json
```diff
@@ -0,0 +1,114 @@
...
+      "description": "An array of all pages supported by this frontend module. Pages are automatically mounted based on a route.",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "component": {
+            "type": "string",
+            "description": "The name of the component exported by this frontend module."
+          },
+          "route": {
```
ibacher, line 32: `route` is a little weird because we allow a boolean (to accommodate the implementer tools, which are always rendered, but only visible if the user is a system developer)

packages/apps/esm-devtools-app/src/routes.json
ibacher: It occurs to me that _if_ we switched this to the same implementation as we use in the implementer-tools (always mounted, but not rendered unless specific conditions are met), we could get away with not needing to support custom activation functions at all, which is probably the right way to go. Just to re-iterate: if we allow custom functions to define whether or not a route is active, then we end up requiring that the code for the frontend module's index file is downloaded as part of the start-up (i.e., before anything renders).

packages/apps/esm-offline-tools-app/src/index.ts
```diff
@@ -11,28 +11,94 @@ import OfflineToolsNavLink from "./nav/offline-tools-nav-link.component";
...
+const options = {
+  featureName: "offline-tools",
+  moduleName,
 };
 
-function setupOpenMRS() {
-  const moduleName = "@openmrs/esm-offline-tools-app";
-  const options = {
-    featureName: "offline-tools",
+export const offlineTools = getAsyncLifecycle(
```
ibacher, line 27: One problem with this scheme is that all of these previously anonymous functions are now required to be named exports... This is a lot of quite verbose names.

routes.schema.json
```diff
@@ -0,0 +1,114 @@
...
+    "version": {
+      "type": "string",
+      "description": "The version of the module. This should always be the constant __VERSION__.",
+      "default": "__VERSION__",
+      "pattern": "^__VERSION__$"
+    },
+    "backendDependencies": {
+      "type": "object",
+      "description": "A list of backend modules necessary for this frontend module and the corresponding required versions.",
+      "additionalProperties": {
```
ibacher, line 17: It might be better to define some constraints around this, e.g., the key should probably just be an ASCII string (since it needs to match a backend module id) and the value should be anything that SemVer can parse as a range. The former is trivial, but didn't seem hugely important. The latter, I couldn't find a convenient way of expressing and, while you can probably do that with regex, it's probably not helpful to do so.

ibacher:
Oh! For now, if you want to try things with the JSON Schema enabled in VSCode, you can use the following workspace settings (assuming you have this branch checked out):

```json
{
  "json.schemas": [
  {
    "fileMatch": [
      "routes.json"
    ],
    "url": "./routes.schema.json"
  }
  ]
}
```

This can give you a sense of the code completions available. If any one has feedback on any of the description strings, that'd be great to hear!

routes.schema.json
samuelmale: +1 for auto-injecting the ESM's version as part of the webpack build process. I would like to suggest maintaining this pattern as you have implemented it.

packages/tooling/webpack-config/src/index.ts
samuelmale: Do we benefit from having this optional? Shouldn't we rather ensure that all esms define this file by throwing some kind of error?

packages/tooling/openmrs/src/commands/assemble.ts
```diff
@@ -243,6 +255,14 @@ export async function runAssemble(args: AssembleArgs) {
     "utf8"
   );
 
+  if (args.buildRoutes) {
```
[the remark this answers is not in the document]
ibacher: This is basically an automatic scan. However, I've left outputting to a file as optional (but, note, on by default). This is because there are use cases (e.g., CI builds) where we might want to be able to output an importmap but not really care about having a full routes file.

I also didn't want to make routes required because the assemble command should be able to create an importmap for versions of frontend modules that do not use routes.json (i.e., we still want the current version of the command to be able to output something usable for previous module versions in case someone is just running `npx openmrs assemble` to build an importmap for an existing install).

packages/apps/esm-login-app/src/index.ts
```diff
@@ -1,85 +1,37 @@
...
-
-function setupOpenMRS() {
-  const moduleName = "@openmrs/esm-login-app";
-
-  const options = {
-    featureName: "login",
-    moduleName,
-  };
-
+export function startupOpenMRS() {
```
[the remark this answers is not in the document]
mogoodrich: Does it make sense to call this "startupApp" or just "startup", since this is what this actually is about, not starting OpenMRS.  Though maintaining something similar to "setupOpenMRS" has some value... anyway, not a strong preference here...

packages/apps/esm-primary-navigation-app/src/routes.json
```diff
@@ -0,0 +1,46 @@
+{
```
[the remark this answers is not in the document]
mogoodrich: Agree... metadata makes me think concepts, encounter types, etc...

packages/apps/esm-login-app/src/index.ts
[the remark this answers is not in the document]
ibacher: I really like `startupApp()`! To me that makes it (reasonably) clear when and why it's called.

packages/apps/esm-primary-navigation-app/src/components/choose-locale/change-locale.resource.tsx
```diff
@@ -26,15 +26,29 @@ export async function postUserPropertiesOnline(
   window.location.reload();
 }
 
+export async function postUserPropertiesOffline(
+  userUuid: string,
+  userProperties: any
```
denniskigen, line 31:
```suggestion
  userProperties: Record<string, string>;
```

https://rest.openmrs.org/#create-a-user

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -95,6 +94,11 @@ function isFederatedModule(a: unknown): a is FederatedModule {
   );
 }
 
+// internals to track script loading
+// basically, if we're already loading a script, we should wait until the script is loaded
+// we use a global to track this
+const OPENMRS_SCRIPT_LOADING = "__openmrs_script_loading";
```
ibacher, line 100: I missed a case with this that's important to how we're intending to use this, specifically: what happens if the same module is being loaded twice, simultaneously. We guarantee that once a module has been loaded, it is not reloaded (HMR will refresh the page), but this guarantee only holds after the initial load is done. Prior to that, two asynchronously executing blocks of code could try to import the same module twice. Now this is caught and the resulting module is loaded only once.

packages/framework/esm-framework/webpack.config.js
```diff
@@ -33,12 +39,28 @@ module.exports = (env) => ({
     extensions: [".ts", ".js", ".tsx", ".jsx", ".scss"],
   },
   plugins: [
-    new SystemJSPublicPathWebpackPlugin(),
     new CleanWebpackPlugin(),
     new ForkTsCheckerWebpackPlugin(),
     new BundleAnalyzerPlugin({
       analyzerMode: env && env.analyze ? "static" : "disabled",
     }),
+    new ModuleFederationPlugin({
```
ibacher, line 47: This turns out to be necessary to get everything loaded at runtime. For now, I've just manually added a `peerDependencies` element to the `package.json` with all the peer dependencies from any sub-part of the framework that isn't part of the framework itself.

It might be nice to semi-automate this at some point, though I'm not sure the actual peer dependencies will be changing too much.

With the SystemJS format, we were using webpack externals, but I couldn't really figure out how to make that work, so this one module now just repackages the framework into a shared library... There may be a better way to do this, but I haven't found it yet.

packages/shell/esm-app-shell/src/index.ts
```diff
@@ -1,11 +1,35 @@
...
+import "systemjs/dist/system";
+import "systemjs/dist/extras/amd";
+import "systemjs/dist/extras/named-exports";
+import "systemjs/dist/extras/named-register";
+import "systemjs/dist/extras/use-default";
+import "@openmrs/esm-framework";
+import type { SpaConfig } from "@openmrs/esm-framework/src/internal";
 
-declare var __webpack_public_path__: string;
+export function setupPaths(config: SpaConfig) {
```
ibacher, line 11: These functions used to be declared in esm-framework. However, they are really tied to the app-shell functionality and there is quite literally no reason to ever call them from anywhere else, so I've just moved them into the app-shell.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -74,14 +45,47 @@ function registerCoreExtensions() {
...
-async function setupApps(modules: Array<[string, System.Module]>) {
-  modules.forEach(([appName, appExports]) => registerApp(appName, appExports));
-  subscribeConnectivity(async () => {
-    const appNames = getAppNames();
-    await Promise.all(appNames.map(unregisterApplication));
-    finishRegisteringAllApps();
-  });
-  window.installedModules = modules;
+async function setupApps() {
+  const scriptTags = document.querySelectorAll<HTMLScriptElement>(
```
ibacher, line 49: I've modelled this code for loading routes.json on the stuff that import-map-overrides does to load import maps, though it's not identical.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -74,14 +45,47 @@ function registerCoreExtensions() {
...
+  for (let i = 0; i < scriptTags.length; i++) {
+    promises.push(
+      (async (scriptTag) => {
+        const modules: typeof window.installedModules = [];
+        let routes: OpenmrsRoutes | undefined = undefined;
+        try {
+          if (scriptTag.textContent) {
+            routes = JSON.parse(scriptTag.textContent) as OpenmrsRoutes;
+          } else if (scriptTag.src) {
+            routes = (await openmrsFetch<OpenmrsRoutes>(scriptTag.src)).data;
```
ibacher, line 63: Importantly, and routes files loaded in the HTML _must_ be accessible prior to an OpenMRS session being established. Other forms of authentication like IP whitelists are perfectly fine.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -98,6 +102,9 @@ async function loadConfigs(configs: Array<{ name: string; value: Config }>) {
  */
 function connectivityChanged() {
   const online = navigator.onLine;
+  // NB We do not wait for this to be done; it is simply scheduled
+  triggerAppChange();
```
ibacher, line 106: One sub-goal of this is to push more work into the single-spa activator. Specifically, we have a bunch of separate code for:

1. Should the page render
2. Should the page render in offline mode? Online mode?
3. Should the page render given the current user's privileges?

In the current state, 1 is handled by single-spa and 2 & 3 are handled by wrappers in the loader that return `undefined` if the page / extension should not render.

At least for apps, all three of these can be done in the activator function. This means that instead of what we used to do previously (namely, unregister and re-register all single-spa applications when connectivity changed), we can do the much more light-weight thing of telling single-spa to re-check all mounted apps.

Since we are no longer passing properties through the `online` or `offline` settings this is simple, straight-forward, and way more performant.

I still need to check what happens when connectivity changes with extensions (since extensions also might need to change based on connectivity status). But, the worst case here is we add one function to esm-extensions that re-checks the activity function of all extensions.

packages/shell/esm-app-shell/src/types.ts
```diff
@@ -2,8 +2,129 @@ export interface Activator {
   (location: Location): boolean;
 }
 
-export type ActivatorDefinition = Activator | RegExp | string;
+export type ActivatorDefinition = RegExp | string | boolean;
 
-export interface ModuleResolver {
-  (): System.Module;
+export type OpenmrsRoutes = Record<string, OpenmrsAppRoutes>;
```
ibacher, line 7: Does anyone have strong opinions on where these types should go? I think because some of this (`ExtensionDefinition`) is re-used inside the framework, the corresponding types were in esm-global, but pages are purely an app-shell concern.

packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -32,6 +33,8 @@ const openmrsPageTitle = process.env.OMRS_PAGE_TITLE || "OpenMRS";
 const openmrsFavicon =
   process.env.OMRS_FAVICON || `${openmrsPublicPath}/favicon.ico`;
 const openmrsOffline = process.env.OMRS_OFFLINE !== "disable";
+const openmrsRoutes =
+  process.env.OMRS_ROUTES || `${openmrsPublicPath}/routes.json`;
```
ibacher, line 37: TODO Add relevant options to the CLI commands to override this.

packages/shell/esm-app-shell/src/run.ts
ibacher: NB If you look at `index.ejs` (the template for `index.html`), you might notice that we prefetch this source, so this is hopefully not really going to need to wait for a full fetch. This is also how we handle importmaps.

packages/shell/esm-app-shell/src/index.ts
```diff
@@ -29,7 +47,11 @@ function initializeSpa(config: SpaConfig) {
   setupUtils();
   setupPaths(config);
   wireSpaPaths();
-  return runSpa(config);
+  return Promise.resolve(__webpack_init_sharing__("default")).then(async () => {
```
ibacher, line 50: The add-on here starts WMF before we start trying to load any modules. Technically, it would be plausible to do this in the `run()` call as well, but this seemed cleaner.

ibacher:
Alright, so I've had to force-push this branch to incorporate upstream changes without making myself insane. If you've actually checked this out, you'll probably need to delete and rebuild. Sorry about that.

On to the exciting news: this works! Albeit for a very minimal use case. Specifically, in the app-shell's assets directory, I added a sample routes file and have wired things up to actually use that when you use `run:shell` (there are currently some bits that aren't ideal; don't reload when any sub-page `/openmrs/spa/`). The only routes in there are for the primary-navigation-app, which provides to pages, both of which appear to render. Basically, if you check this branch out, run `yarn`, `yarn turbo run build:development` and then `yarn run:shell`, you should be on a very exciting completely blank page at the url `/openmrs/spa/home`, which actually indicates that both the pages from primary-navigation-app are loaded and working.

I've gone down some really deep rabbits holes on this. However, the up-shot is that to make things work `@openmrs/esm-framework` (unlike it's constituent libraries) now uses WMF. This was necessary to ensure that, at runtime, all the stuff in `@openmrs/esm-framework` relies on the shared versions of things.

Among other things, I think this should squash a warning message that's been bothering me for a while where single-spa reports that multiple copies of single-spa are loaded. Hopefully with this bugbear out of the way, I can resume a more normal velocity on this.

denniskigen:
Really tremendous progress, @ibacher. How this hasn’t got you going stark raving mad is quite the feat.

mogoodrich:
Yes, thanks @ibacher !

packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -134,27 +139,50 @@ module.exports = (env, argv = {}) => {
...
         {
-          context: [
-            `${openmrsApiUrl}/**`,
-            `${openmrsPublicPath}/**`,
-            `!${openmrsPublicPath}/index.html`,
-          ],
+          /**
+           * @param {String} path
+           */
+          context(path) {
```
ibacher, line 152: Turns out this can take a function and the rules are much easier to describe this way. There were probably some weird dev edge cases that this fixes too.

routes.schema.json
```diff
@@ -0,0 +1,115 @@
...
+          },
+          "component": {
+            "type": "string",
+            "description": "The name of the component exported by this frontend module."
+          },
+          "slot": {
+            "type": "string",
+            "description": "If supplied, the slot that this extension is rendered into by default."
+          },
+          "privilege": {
```
ibacher, line 82: We currently only support privileges for extensions. Should we also support them for pages?

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -313,11 +313,13 @@ function getAssignedExtensionsFromSlotData(
     );
     const name = getExtensionNameFromId(id);
     const extension = internalState.extensions[name];
+
     // if the extension has not been registered yet, do not include it
     if (extension) {
       const requiredPrivileges =
         extensionConfig?.["Display conditions"]?.privileges ??
```
ibacher, line 320: Eons ago, I added this `Display conditions` functionality, but couldn't really figure out how to get it to be able to override the extension privileges, since those were evaluated in the loader. I've tried to remove that logic from the loader and move it to here, which should mean that it's possible, in configuration, to override an extensions privileges (and, in fact, that's the standard behaviour). That's a new breaking change, but I think useful functionality (it also seems to have broken a test).

ibacher:
This PR is probably now at a point where it makes sense for anyone who is interested to actually try it out. The process is a bit convoluted, so please bear with me:

1. In a copy of openmrs-esm-core, checkout this PR and run `yarn && yarn turbo run build` to build everything.
2. Inside the repo you would like to work on, run `yarn link ../<path to esm-core>/packages/tooling/openmrs` (assumes you are using Yarn v3).
3. Start the app you want to work on using `yarn start` in the usual way. This should give you a blank page.
4. Inside openmrs-esm-core, start the devtools-app, login-app and the primary-navigation-app (these are the essential apps). You can do this with the commands `yarn workspace @openmrs/esm-devtools-app serve`, `yarn workspace @openmrs/esm-login-app serve` and `yarn workspace @openmrs/esm-primary-navigation-app serve` from the root of the project. Note the ports these launch on, as you will need to add them to the import map overrides.
5. Manually add an override to point to the devtools import map. You do this by typing (at the browser console): `localStorage.setItem("import-map-override:@openmrs/esm-devtools-app", "http://localhost:8082/openmrs-esm-devtools-app.js")` (substituting 8082 for whatever port the devtools started on).
6. Enable the devtools in the usual way (`localStorage.setItem("openmrs:devtools", true)`) and reload the page. You should now see the devtools icon.
7. Add the overrides for login-app and primary-navigation-app in the usual way. Reload the page and you should hopefully see the login screen.

At this point, you want to start modifying the `routes.json` file at `<esm-core>/packages/shell/esm-app-shell/dist/routes.json` to add the routes for the new app. You'll also want to change the `index.ts` file to match the new format (see the notes on this PR). You probably need to refresh the page for your changes to the routes.json file to take effect.

If you do try this out, please report back.

**NB** If you try this on the patient-chart, please remove the esm-form-entry-app from the workspace before trying any of this. It is a special snowflake and we'll deal with it later. The easiest way to do this is to rename it's `package.json` file to pretty much anything else (I used `_package.json`).

ibacher:
Small update: I've now published the JSON schema for routes definitions to https://json.openmrs.org/routes.schema.json (this is just GH Pages hosting [this repo](https://github.com/openmrs/openmrs-contrib-json-schemas)). The idea is to make it easier to consume this in your IDE. For example, using VSCode, you can now use the snippet:

```
{
  "json.schemas": [
  {
    "fileMatch": [
      "src/routes.json"
    ],
    "url": "https://json.openmrs.org/routes.schema.json"
  }
  ]
}
```

The schema should work with your IDE's autocompletions to help ensure you are entering the schema in the expected format. Eventually we will also have linting setup to help with this.

ibacher:
Feedback solicitation: So for the implementer tools bit, originally I was going to add a new tab, but I realised this could just sit in the existing "Frontend Modules" view. I've dummied-up something that looks like this:

Screenshot 2023-05-19 at 14 42 58

Screenshot 2023-05-19 at 14 43 13

But I'm not 100% certain this is right, so if anyone has any opinions, let me know. (Of particular concern is displaying something sensible for pages. Extensions have names, since we need to refer to them in configs, but pages currently do not. Really, the only thing you need to define a page is a route and a component function; I tried rendering the route in one cell and the component name in another, but it doesn't really look great because routes can be defined as a boolean, string, or regex. I'm considering just dropping them altogether, though, because I don't know that the component name—which is intended to just point to a named export—is meaningful). Personally, I think the list of extensions is really useful and I don't know why we didn't already have this—extensions are things you need to know about to use them in configurations!

denniskigen:
> Feedback solicitation: for the implementer tools bit, originally I was going to add a new tab, but I realised this could just sit in the existing "Frontend Modules" view. I've dummied-up something that looks like this:
> 
> Screenshot 2023-05-19 at 14 42 58 Screenshot 2023-05-19 at 14 43 13
> So, basically, concrete questions here:
> 
> 1. Is it too noisy having this in the Frontend Modules tab?
> 2. Can you understand which pages and extensions are included in each Module?
> 3. Is the display of pages useful?
> 4. Is the display of extensions useful?
> 5. Is there information on pages and extensions you would like to have displayed here that isn't?

This looks decent enough. Using a [fixed heading type set](https://carbondesignsystem.com/guidelines/typography/type-sets#fixed-heading-styles) for the different categories (pages, extensions etc) would make for a better visual hierarchy.

packages/apps/esm-implementer-tools-app/src/popup/popup.component.tsx
```diff
@@ -31,9 +31,9 @@ export default function Popup({
     } else if (activeTab == 1) {
       return <FrontendModules frontendModules={frontendModules} />;
     } else {
-      return <ModuleDiagnostics frontendModules={backendDependencies} />;
+      return <BackendDependencies frontendModules={backendDependencies} />;
```
denniskigen, line 34: Nit: this prop should be just `modules`.

ibacher:
> Ensure we still support esm-form-entry-app via SystemJS

Thanks to @denniskigen [in this PR](https://github.com/openmrs/openmrs-esm-patient-chart/commit/3d1e771de4584aa97a3559dd87d4f4026fb0bb24), this is no longer a concern, so going forward, we will be dropping support for SystemJS altogether. This should also mean that esm-form-entry-app can be supported basically the same way we support everything else.

packages/apps/esm-implementer-tools-app/src/popup/popup.component.tsx
ibacher: Yeah... that really should've been called `backendDependencies` originally, but I only noticed after it was committed.

ibacher:
@denniskigen I suppose this is growing on me a bit
Screenshot of frontend modules with new headings

I should probably change the other views to use the same heading style.

ibacher:
So out-of-the-box, this will support routes via basically the same mechanism as importmaps in development except that there's no equivalent to import-map-overrides. Basically, what that means is that as long as you're running `openmrs develop` (so `yarn start`), things should work as expected. If you run into a situation where you need import map overrides, everything will work as expected as long as you don't need to add a new page or extensions to the thing loaded from the override.

Because using import-map-overrides should be the exception rather than the rule, I don't want to hold up the rest of the progress on that for a feature we can add later.

Inevitably, I need to clean-up whatever is failing in the build, but otherwise, this should be ready to review. I _think_ I've addressed all the comments so far except my last one about styling the implementer tools.

packages/apps/esm-implementer-tools-app/src/frontend-modules/frontend-modules.component.tsx
```diff
@@ -57,12 +53,32 @@ export const FrontendModules: React.FC<FrontendModulesProps> = ({
...
+                      </TableCell>
+                      <TableCell>
+                        {esm.version ?? t("unknownVersion", "unknown")}
+                      </TableCell>
+                    </TableRow>
+                    {Boolean(esm.extensions) ? (
+                      <TableRow key={`${esm.name}-extensions-header`}>
+                        <TableCell>
+                          <span className={styles.moduleComponentHeader}>
+                            Extensions
```
denniskigen, line 69:
```suggestion
                            {t('extensions', 'Extensions')}
```

ibacher:
Ok, so using this in other repos is a little more annoying:

1. In the other repo (patient-chart / patient-management / etc.) link to the `openmrs` package (`yarn link <path>/<to>/<esm-core>/packages/tooling/openmrs`)
2. In the `package.json` add resolutions for `webpack`, resolving it to 5.77.0 and `typescript`, resolving it to 4.8.4.
3. Run `yarn`. Shouldn't be any warnings about `openmrs`
4. In place of `yarn start`, us this dozy:

```
node --preserve-symlinks ./node_modules/openmrs/dist/cli.js develop --sources <path>/<to>/<esm-core>/packages/apps/esm-primary-navigation-app --sources <path>/<to>/<esm-core>/packages/apps/esm-login-app --sources <path>/<to>/<esm-core>/packages/apps/esm-implementer-tools-app --sources <normal sources for the app(s) you want to work on>
```

E.g., on my computer (where all MF repos are in one directory), this looks like:

```
node --preserve-symlinks ./node_modules/openmrs/dist/cli.js develop --sources ../openmrs-esm-core/packages/apps/esm-primary-navigation-app --sources ../openmrs-esm-core/packages/apps/esm-login-app --sources ../openmrs-esm-core/packages/apps/esm-implementer-tools-app --sources packages/esm-patient-search-app
```

These steps are only temporary until this PR is merged, at which point things should be more normal.

ibacher:
In other news, here's this running with the patient-search-app (I've only converted the search page, not the extensions):

Screenshot of very minimal patient search app

denniskigen:
> In other news, here's this running with the patient-search-app (I've only converted the search page, not the extensions):
> 
> Screenshot of very minimal patient search app

Sweet!

packages/shell/esm-app-shell/src/apps.ts
```diff
@@ -1,142 +1,186 @@
...
-          load: getLoader(result.lifecycle, result.resources),
-          route: result.activate,
-          offline: result.offline,
-          online: result.online,
-          privilege: result.privilege,
-          order: result.order || 1,
-        });
-      }
+    routes.pages?.forEach((p) => {
+      pages.push({ ...p, order: p.order ?? Number.MAX_SAFE_INTEGER, appName });
```
denniskigen, line 167: Why `Number.MAX_SAFE_INTEGER` versus 1 here, @ibacher?

ibacher: So, the `MAX_SAFE_INTEGER` is used as a default for pages that don't define an order, which I'm assuming are the pages for which we don't care about order. If the default is `1` than only pages with an order of `0` will be loaded before a page with no default order defined; however, if we have, say, two pages that need to be loaded in a particular order before any default pages, they would need to be `-1` and `0` to make that work. I thought allowing us to use positive integers was preferable. 

We could, of course, move this to some other, reasonable value (maybe 20?) to allow us to explicitly define pages that are rendered after any other pages. (In general, the only thing we're using order for is to ensure that pages that need to be rendered before other pages are rendered first).

denniskigen: Makes sense. Thanks for the explainer, @ibacher.

denniskigen:
What changes would we need to make elsewhere, at a minimum, to get this working post-merge? Should we begin to set those in motion across the board?

ibacher:
So, the key thing is to have a branch on each repo that’s tracking the changes to the `index.ts` file and that adds a `routes.json` file. That’s what I’m going to start working on next.

ibacher:
On the devtools side of things, Webpack will currently rebuild a module if the routes.json is updated (good); however these changes won't be reflected in the locally running versions unless it's stopped and restarted (not good). The limiting factor here is that we actually merge all the `routes.json` into a big `routes.registry.json` in the CLI app, which is not responsive to changes.

While this is not an ideal DevX, I'd like to push solving that to a later date.

Remaining task is "Support online / offline boolean properties for extensions"; this is actually a little deceptive. What we need is an equivalent to single-spa's `triggerAppChange()` (which allows us to manually trigger a re-check of which apps should be mounted) for the extension system. Then we just need to call that when connectivity changes, and it should work.

ibacher:
Actually, on reflection, I don't think there's anything to do with the extension system. It's worth noting that while the extension system is somewhat framework agnostic, the `<ExtensionSlot/>` (and `<Extension/>`) components are React-components, and the only real way to render an extension. Since `<ExtensionSlot />` already transitively depends on the connectivity status for which extensions to render, it means that extensions will already be recalculated when changing between online and offline mode, which is the effect we needed.

So, for all it's imperfections, I think this PR is now substantially complete.

I still need to sort out the test failure, assuming I can reproduce it (it seems to fail on CI, but not locally).

Jexsie:
> Actually, on reflection, I don't think there's anything to do with the extension system. It's worth noting that while the extension system is somewhat framework agnostic, the `<ExtensionSlot/>` (and `<Extension/>`) components are React-components, and the only real way to render an extension. Since `<ExtensionSlot />` already transitively depends on the connectivity status for which extensions to render, it means that extensions will already be recalculated when changing between online and offline mode, which is the effect we needed.
> 
> So, for all it's imperfections, I think this PR is now substantially complete.
> 
> I still need to sort out the test failure, assuming I can reproduce it (it seems to fail on CI, but not locally).

Congs Ian. Am more excited than you with the little follow-up I've done.

packages/apps/esm-login-app/src/index.ts
```diff
@@ -1,85 +1,37 @@
...
-          () => import("./logout/logout.component"),
-          options
-        ),
-        online: true,
-        offline: false,
-      },
-      {
-        name: "location-changer",
-        slot: "user-panel-slot",
-        order: 1,
```
denniskigen, line 74: @ibacher Looks like this `order` property didn't make it into [routes.json](#diff-c69d5bbbced7cc85fb9bcbb0ea2a893f160fe51e4c397ce912c7ce94aa97b1f8).

packages/apps/esm-implementer-tools-app/src/configuration/interactive-editor/value-editor.tsx
```diff
@@ -32,20 +32,23 @@ export function ValueEditor({
...
   const valueType = customType ?? element._type;
 
-  const keyListener = (e: KeyboardEvent) => {
-    if (e.key === "Escape") {
-      handleClose();
-    } else if (e.key === "Enter") {
-      handleSave(JSON.stringify(tmpValue));
-    }
-  };
+  const keyListener = useCallback(
```
joeldenning, line 35: this is a nit and mostly just preference, but moving keyListener inside of the useEffect on line 46 might allow you to avoid `useCallback` and managing all the dependency arrays with react hooks.

packages/apps/esm-login-app/src/root.component.tsx
```diff
@@ -4,30 +4,16 @@ import LocationPicker from "./location-picker/location-picker.component";
...
+export interface RootProps {}
 
-const Root: React.FC<RootProps> = ({ isLoginEnabled }) => {
+const Root: React.FC<RootProps> = () => {
   return (
     <BrowserRouter basename={window.spaBase}>
       <Routes>
-        <Route
-          path="/login"
-          element={<Login isLoginEnabled={isLoginEnabled} />}
```
joeldenning, line 17: 👍 switching away from `isLoginEnabled` prop seems like a good thing. The useConnectivity hook seems a bit cleaner

packages/framework/esm-api/webpack.config.js
```diff
@@ -11,30 +10,28 @@ module.exports = (env) => ({
...
-        use: "swc-loader",
+        use: "@swc-node/loader",
       },
     ],
   },
   resolve: {
     extensions: [".ts", ".js", ".tsx", ".jsx"],
   },
   plugins: [
-    new SystemJSPublicPathWebpackPlugin(),
```
joeldenning, line 30: `systemjs-webpack-interop` can probably be uninstalled from esm-api now, I think?

packages/framework/esm-framework/webpack.config.js
joeldenning: As discussed on the call, I think module federation is a good technology, but it's hard to ensure that it's always the same version of every shared dependency being loaded, since each microfrontend produces shared dependency from its node_modules/package.json. The problems with module federation are solvable and I don't think it's a big mistake to switch to it, just there are a few things to be solved.

packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -25,6 +22,18 @@
       crossorigin="anonymous"
     >
     <script type="systemjs-importmap" src="<%= openmrsImportmapUrl %>"></script>
+<% } %>
+<% if (openmrsCoreImportmap) { %>
+  <script type="systemjs-importmap"><%= openmrsCoreImportmap %></script>
```
joeldenning, line 27: Why was the order of the systemjs-importmaps changed?

packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -25,6 +22,18 @@
...
+<% } %>
+<% if (openmrsRoutes) { %>
+  <link
+    rel="preload"
+    href="<%= openmrsRoutes %>"
+    as="fetch"
+    crossorigin="anonymous"
+  >
+  <script type="openmrs-routes" src="<%= openmrsRoutes %>"></script>
```
joeldenning, line 36: Could the routes be embedded as json inside this script tag, rather than a `src` attribute? To avoid a blocking network call at the beginning of every page load

packages/shell/esm-app-shell/src/run.ts
joeldenning: Assuming the routes json can't be inlined into the HTML, it's important to make sure that this line of code runs as soon as possible in the initialization of the page. The html preload is nice to get the network request going, but

packages/shell/esm-app-shell/src/run.ts
joeldenning: Are you calling triggerAppChange since the single-spa activity functions change behavior based on connectivity?

joeldenning (approved):
This is definitely a performance improvement - frontloading all microfrontends was unscalable. Extracting the routes/pages to JSON is a good way of making this something the app-shell can process without having to download every module. Nice work

packages/apps/esm-devtools-app/src/index.ts
```diff
@@ -1,26 +1,13 @@
...
-    featureName: "devtools",
-    moduleName: "@openmrs/esm-devtools-app",
-  };
-
-  return {
-    lifecycle: getAsyncLifecycle(
-      () => import("./devtools/devtools.component"),
-      options
-    ),
-    activate: () =>
```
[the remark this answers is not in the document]
ibacher: Yes. While this does need to be communicated, I checked through the major community repos and repos for known implementations. Every "page" I found, aside from the apps in this monorepo, uses a string URL to determine whether or not it's visible.

packages/apps/esm-login-app/package.json
```diff
@@ -51,12 +51,14 @@
   "devDependencies": {
     "@openmrs/esm-framework": "^4.5.0",
     "@openmrs/webpack-config": "^4.5.0",
+    "history": "^5.3.0",
```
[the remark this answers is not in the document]
ibacher: Apparently this was used in a test utility (to populate the initial history for a MemoryRouter). Easy enough to get rid of though. Weirdly, although `history` is dependency of the monorepo, it's not actually a dependency of any of the packages we're using, so `import history from "history"` was failing. This must've been something that got glossed-over when we moved from React Router 5 to 6.

packages/shell/esm-app-shell/src/index.ejs
ibacher: The gist is that if `openmrsCoreImportmap` is defined (and it only is in one special case), I want it to override the base import map.

`openmrsCoreImportmap` is only populated when running the app shell in development mode, and it basically ensures that the import-map is pointing to the user's local copies of the apps in the `packages/apps` part of the monorepo. While working on this PR, I wanted to be able to edit the core apps and the app shell simultaneously, and moving this in the DOM allowed me to do that. There's history to why this exists; the app-shell used to also bundle the core apps and then use this variable to provide the actual import map for those core apps at runtime, but this had some disadvantages (unnecessary coupling; no ability for a distribution to opt-out of, say, having the dev tools or implementer tools running in their production environment), so we moved away from it.

packages/shell/esm-app-shell/src/index.ejs
ibacher: This is actually a really good point and something I hadn't considered at all. Currently, we're also not embedding the importmap either, though we probably should for exactly the same reason.

packages/shell/esm-app-shell/src/run.ts
ibacher: Yes. Specifically, we support two properties for apps called "online" and "offline" which are booleans that indicate whether the app should render when there is connectivity and when there isn't connectivity. These are added to the activity function in the `wrapActivityFn()` function in `src/apps.ts`.

The previous solution to this had the connectivity check inside the `load()` function (which is basically used to lazy-load the single-spa application) and on connectivity change, de-registered and then re-registered all single-spa applications. `triggerAppChange()` and a wrapper for the activity function seemed like a much cleaner solution.

However, if you've got a better way to do this, I'm open to ti!

denniskigen (approved):

PR #687, opened 2023-06-07
(feat) O3-2155: Using the global system default_locale by default for the O3 

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR allows using the global system's default locale for the O3 application.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-2155

### DIFF
--- a/packages/shell/esm-app-shell/src/locale.ts
+++ b/packages/shell/esm-app-shell/src/locale.ts
@@ -3,7 +3,10 @@ import ICU from "i18next-icu";
 import LanguageDetector from "i18next-browser-languagedetector";
 import { initReactI18next } from "react-i18next";
 import merge from "lodash-es/merge";
-import { getConfigInternal } from "@openmrs/esm-framework/src/internal";
+import {
+  getConfigInternal,
+  openmrsFetch,
+} from "@openmrs/esm-framework/src/internal";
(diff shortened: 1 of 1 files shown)


ibacher:
Strictly speaking, no, there's no guarantee the backend will be available. Thinking about this for a second, I actually think we want this to be a build-time property of the app shell rather than something we fetch from the backend. A couple of considerations here:

1. Even if the backend is unavailable, users will want the UI in their language (at some point we need to handle the "backend is down").
2. While this would end up duplicating a second, the default locale of a district is unlikely to be a property we actually need to handle switching at runtime and, at the very least, a think we can defer implementing that until a later point.

Implementation-wise, take a look at `index.ejs` and how we inject values from webpack.

vasharma05:
How about we don't do this during the build time and instead do it on  login app, either fetching it from system setting or using the naviogator.language?
Thanks!

vasharma05:
> I'll get this implemented as a build step tomorrow.
How will it be taken as a build step? Still not clear on this.
Thanks!

vasharma05:
Please note the following PR from @ibacher here: https://github.com/openmrs/openmrs-esm-core/pull/695

PR #695, opened 2023-06-09
(feat) Add support for setting default language via html lang attribute

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Adds support for configuring a default language via the build config or command line arguments.

While this is technically duplicative of the `default_locale` system setting from the backend, we have good reason to ensure that some kind of default language is set in the frontend in case the backend is, for whatever reason unavailable.

@vasharma05 To support Mekom's use-case, once merged, I'll add an overridable environment variable to the O3 RefApp so that this can be set on a per-instance basis.

## Related Issue
https://issues.openmrs.org/browse/O3-2155

### DIFF
--- a/packages/tooling/openmrs/src/commands/build.ts
+++ b/packages/tooling/openmrs/src/commands/build.ts
@@ -14,7 +14,7 @@ import { basename, join, parse, resolve } from "path";
 export interface BuildArgs {
   target: string;
   registry: string;
-  defaultLanguage: string;
+  defaultLocale: string;
   importmap: string;
   spaPath: string;
   fresh?: boolean;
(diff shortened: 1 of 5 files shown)


vasharma05 (commented):
Can we change `default-language` with default-locale to match with the global setting `default-locale`?

Thanks for working on this @ibacher ! Really grateful 🙏

ibacher:
@vasharma05 Good call!

packages/tooling/openmrs/src/commands/build.ts
```diff
@@ -87,6 +89,7 @@ export async function runBuild(args: BuildArgs) {
     env: "production",
     apiUrl: buildConfig.apiUrl || args.apiUrl,
     configUrls: configUrls,
+    defaultLocale: args.defaultLocale || buildConfig.defaultLocale,
```
ibacher, line 92: Note: we should update the other properties to follow this pattern, i.e., explicitly CLI args should always override the configuration file.

denniskigen (approved):
LGTM

PR #704, opened 2023-06-20
(fix)O3-2207:openmrs build script exits with exit code 0 even if there is a build failure

jwnasambu:
## Requirements

#### For changes to apps

#### If applicable

## Summary
I added a condition for the build script to exit  with code 0 when the build is successful and code 1 if there is a build failure.

## Related Issue
https://issues.openmrs.org/browse/O3-2207

### DIFF
--- a/packages/tooling/openmrs/src/cli.ts
+++ b/packages/tooling/openmrs/src/cli.ts
@@ -30,7 +30,13 @@ function runCommand<T extends CommandNames>(
     args,
   });
 
-  ps.on("exit", (code) => process.exit(code || 0));
+  ps.on("exit", (code) => {
+    if (code === 0) {
+      process.exit(0); // Build success, exit with code 0
+    } else {
+      process.exit(1); // Build failure, exit with code 1
+    }
+  });
 }
 
 yargs.command(


packages/tooling/openmrs/src/cli.ts
```diff
@@ -30,7 +30,13 @@ function runCommand<T extends CommandNames>(
     args,
   });
 
-  ps.on("exit", (code) => process.exit(code || 0));
+  ps.on("exit", (code) => {
```
ibacher, line 33: We should preserve the existing behaviour for non-zero exit codes, i.e., if there is one, use it. If there isn't try to return something that won't have been returned by webpack. But also, is this where the error occurs? Do you have a test case where the build fails, but the exit code is 0, because we really need that reproduction before coding.

jwnasambu: @ibacher to be honest I don't know how to reproduce the error though I have tried without success.

jwnasambu: @ibacher for the good advice!

denniskigen:
Closing because of missing repro steps. Thanks @ibacher, @jwnasambu.

PR #707, opened 2023-06-23
(fix) Various fixes for compatibility of new versions

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

* esm-framework itself should use System-style output
* Properly handle multiple routes overriding each other
* Properly handle ordering of pages across apps and app indices based on page order

### DIFF
--- a/packages/shell/esm-app-shell/src/apps.ts
+++ b/packages/shell/esm-app-shell/src/apps.ts
@@ -182,22 +182,26 @@ export function registerApp(appName: string, routes: OpenmrsAppRoutes) {
  * Each page is rendered into a div with an appropriate name.
  */
 export function finishRegisteringAllApps() {
-  pages.sort((a, b) => a.order - b.order);
+  pages.sort((a, b) => {
+    let sort = a.order - b.order;
+    if (sort != 0) {
+      return sort;
+    }
+    return a.appName.localeCompare(b.appName);
+  });
 
   // Create a div for each page. This ensures their DOM order.
   // If we don't do this, Single-SPA 5 will create the DOM element only once
   // the page becomes active, which makes it impossible to guarantee order.
-
-  let index = 0;
-  let lastAppName: string | undefined = undefined;
-
+  let appIndices = new Map();
   for (let page of pages) {
-    if (page.appName !== lastAppName) {
-      index = 0;
-      lastAppName = page.appName;
+    if (!appIndices.has(page.appName)) {
+      appIndices.set(page.appName, 0);
     } else {
-      index++;
+      appIndices.set(page.appName, appIndices.get(page.appName) + 1);
     }
+    const index = appIndices.get(page.appName);
+
     const name = `${page.appName}-page-${index}`;
     const div = document.createElement("div");
     div.id = `single-spa-application:${name}`;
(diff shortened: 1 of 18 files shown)


packages/shell/esm-app-shell/src/apps.ts
```diff
@@ -182,22 +182,26 @@ export function registerApp(appName: string, routes: OpenmrsAppRoutes) {
  * Each page is rendered into a div with an appropriate name.
  */
 export function finishRegisteringAllApps() {
-  pages.sort((a, b) => a.order - b.order);
```
ibacher, line 185: So, I think we were implicitly relying on sorting pages by app and then by page order. Pages are now sorted _purely_ by order (and then appName according to English rules) so we're not implicitly relying on anything.

This is also why the index-counting behaviour needed to be changed to track an index-per-app.

packages/shell/esm-app-shell/src/apps.ts
```diff
@@ -182,22 +182,26 @@ export function registerApp(appName: string, routes: OpenmrsAppRoutes) {
...
+  });
 
   // Create a div for each page. This ensures their DOM order.
   // If we don't do this, Single-SPA 5 will create the DOM element only once
   // the page becomes active, which makes it impossible to guarantee order.
-
-  let index = 0;
-  let lastAppName: string | undefined = undefined;
-
+  let appIndices = new Map();
```
ibacher, line 196: We probably have a lot of places where we're using Objects and it would make sense to use Maps. That said, the syntax for objects is much more comfortable, i.e., I don't think there's as ergonomic a version of `{...base, ...overrides}` for merging maps as for objects, though I suppose:

```
function mapMerge(base: Map, overrides: Map): Map
```

wouldn't be hard to write.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -38,21 +38,19 @@ import { appName, getCoreExtensions } from "./ui";
...
+ * list of apps from the routes.registry.json file, which serves
+ * as the registry of all apps in the application.
  */
 async function setupApps() {
   const scriptTags = document.querySelectorAll<HTMLScriptElement>(
     "script[type='openmrs-routes']"
   );
 
-  const promises: Array<Promise<typeof window.installedModules>> = [];
+  const promises: Array<Promise<OpenmrsRoutes>> = [];
```
ibacher, line 50: We process each script tag asynchronously, but we need to merge them into a single consistent view of routes before we do anything else. We were accidentally double-loading some routes in development, which was not great.

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -68,22 +66,26 @@ async function setupApps() {
...
   }
 
-  window.installedModules = (await Promise.all(promises)).flat();
+  const routes = (await Promise.all(promises)).reduce(
+    (accumulatedRoutes, routes) => ({ ...accumulatedRoutes, ...routes }),
+    {}
+  );
+
+  const modules: typeof window.installedModules = [];
+  Object.entries(routes).forEach(async ([module, routes]) => {
```
ibacher, line 83: Technically, this doesn't guarantee that all modules will be in `installedModules`, i.e., the array can continue to grow asynchronously. I don't know if it's worth implementing the kind of blocking that would be required to ensure that once it was set it's value was not changed.

denniskigen (approved):
LGTM

PR #710, opened 2023-06-26
(feat) Load scripts for non-visible modules asynchronously after starting application

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This is something I was deferring to "phase 2" because it was a little hard to test without the full routes setup. The idea here is that we pre-load the script (and the initial chunk containing the contents of `index.ts`), but not call any code at this point.

Hopefully, this should make things feel more responsive.

NB It is quite trivial for this design to be less than optimal; however, further optimising this likely involves prioritising the preloading (i.e., ensure that scripts are preloaded in the order they are likely needed, but that requires some way of declaring that prioritisation).

### DIFF
--- a/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
+++ b/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
@@ -37,7 +37,7 @@ export async function importDynamic<T = any>(
 
   const container = window[jsPackageSlug] as unknown;
   if (!isFederatedModule(container)) {
-    const error = `The global variable ${jsPackageSlug}  does not refer to a federated module`;
+    const error = `The global variable ${jsPackageSlug} does not refer to a federated module`;
     console.error(error);
     throw new Error(error);
   }
--- a/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
+++ b/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
@@ -1,7 +1,7 @@
 /** @module @category Dynamic Loading */
 "use strict";
 // hack to make the types defined in esm-globals available here
-import type {} from "@openmrs/esm-globals";
+import type { ImportMap } from "@openmrs/esm-globals";
 
 /**
  * @internal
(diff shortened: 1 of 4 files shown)


packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -28,8 +28,46 @@ export function slugify(name: string) {
...
+  share: string = "./start",
+  options?: { importMap?: ImportMap }
 ): Promise<T> {
+  await preloadImport(jsPackage, options?.importMap);
+
+  const jsPackageSlug = slugify(jsPackage);
+
+  const container = window[jsPackageSlug] as unknown;
+  if (!isFederatedModule(container)) {
+    const error = `The global variable ${jsPackageSlug}  does not refer to a federated module`;
```
denniskigen, line 40:
```suggestion
    const error = `The global variable ${jsPackageSlug} does not refer to a federated module`;
```

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -28,8 +28,46 @@ export function slugify(name: string) {
...
+    throw new Error(error);
+  }
+
+  container.init(__webpack_share_scopes__.default);
+
+  const factory = await container.get(share);
+  const module = factory();
+
+  if (!(typeof module === "object") || module === null) {
+    const error = `Container for ${jsPackage} did not return an ESM module as expected`;
```
denniskigen, line 51:
```suggestion
    const error = `Container for ${jsPackage} did not return an ES module as expected`;
```

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -360,5 +381,6 @@ export function run(configUrls: Array<string>, offline: boolean) {
     .then(runShell)
     .catch(handleInitFailure)
     .then(closeLoading)
-    .then(() => (offline ? setupOffline() : undefined));
+    .then(offline ? setupOffline : undefined)
```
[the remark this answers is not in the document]
ibacher: It gets called by `.then()` implicitly. If I explicitly add the parentheses, it will be called before the `.then()` happens. (When it was wrapped in a function, we had to explicitly call it; this basically removes an unnecessary function wrapper.

denniskigen (approved):
Thanks, @ibacher.

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
ibacher: ES = "ECMAScript" (the semi-official name of Javascript)
ESM = "ECMAScript Module"

Hence

"ES module" == "ESM"

PR #708, opened 2023-06-24
(fix) Extensions should have meaningful names

ibacher:
## Requirements
- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

#### For changes to apps
- [x]  My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://zeroheight.com/23a080e38/p/880723-introduction).

#### If applicable
- [ ] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary

Extensions now are internally named "${extensionName} in ${extensionSlotName}"

Additionally:

* reduce the number of times we ask single-spa to render an extension
* some internal typing improvements

## Screenshots

A couple of places this is visible:

#### Errors

Screenshot showing error occurred in parcel-5

Screenshot showing error occured in system-admininstration-app-menu-link in app-menu-slot

#### Developer Tools

Screenshot of React developer tools with parcel-n style names

Screenshot of React developer tools with reasonable names

PR #720, opened 2023-07-06
(fix) Fix issues with single-spa caused by naming extensions

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

#708 accidentally broke our use of single-spa. Specifically, single-spa uses parcel names in order to track the parcel internally. However, we support rendering each extension basically any number of times. With the new names, this meant that if the same extension was loaded more than once (which happens relatively frequently, likely due to re-renders), this caused single-spa to improperly track the parcel's state resulting in a number of single-spa errors.

This shifts us to using the convention: `${slotName}/${extensionName}-${uniqueNumber}`, which avoids these errors.

As future work, we should look into better memoisation of rendered extensions and extension-slots to reduce the number of parcels we are generating.

### DIFF
--- a/packages/apps/esm-implementer-tools-app/translations/am.json
+++ b/packages/apps/esm-implementer-tools-app/translations/am.json
@@ -6,6 +6,7 @@
   "downloadConfig": "Download Config",
   "edit": "Edit",
   "editValueButtonText": "Edit",
+  "extensions": "Extensions",
   "frontendModules": "Frontend Modules",
   "installedVersion": "Installed Version",
   "itemDescriptionSourceDefaultText": "The current value is the default.",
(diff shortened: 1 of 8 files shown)


denniskigen (approved):
Thanks, @ibacher!

brandones:
Might ticket the parcel re-rendering in https://issues.openmrs.org/browse/O3-769 @ibacher . Also I am reminded of https://issues.openmrs.org/browse/O3-1418, which we may want to resurrect if we think excess parcels cause performance problems. I suspect we are better off focusing on the loaders à la core v5.

ibacher:
@brandones Yeah, definitely should ticket it. I'm not too concerned about it performance-wise, but I'm a little worried that there's a gap between the mental model of an extension and the actual implementation, since it's easy to think about "extension x in slot y (with hash z)" as a single, unique identity for the extension, but, in fact, the actual instance that extension creates can be different, depending on the application state.

Basically, it would seem cleaner if e.g., `patient-chart-dashboard-slot/charts-summary-dashboard#` was a unique identifier for the parcel, but the parcel in that combination of slot / extension / distinguishing hash can actually be different instances of the extension, even without any user navigation.

As a practical matter, though, this doesn't seem to really be affecting anything, so I'm not sure it's the highest priority.

PR #725, opened 2023-07-11
(feat) Simplify the openmrsComponentDecorator

ibacher:
## Requirements
- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

#### For changes to apps
- [x]  My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://zeroheight.com/23a080e38/p/880723-introduction).

#### If applicable
- [ ] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary

This removes the React `Suspense` component that attempts to preload translations, which seems to be non-functional.

I'm not 100% certain of the consequences of this, but it seems to work?

PR #731, opened 2023-07-13
(fix) Restore suspense wrapper to openmrsComponentDecorator

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Accidentally broke the implementer tools without realising in #725 by removing the `<Suspense />` component. We still need it for `React.lazy()` and friends which are used by the implementer tools.


denniskigen (approved):
Just tested it locally, and it LGTM! Much appreciated, @ibacher!

brandones:
I think it would be reasonable for us to expect users of `React.lazy` to wrap their own components in `<Suspense>`. That is, I think we should again remove Suspense from this component and just add it to the Implementer Tools directly. Thoughts?

ibacher:
@brandones Yeah, that makes sense, especially as we aren't really able to set a sensible fallback here, i.e., further down in the tree, we have much better control over this.

PR #733, opened 2023-07-15
(feat) O3-2258: Extension system support for feature flags

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This makes it so you can add `featureFlag: "xyz"` to an extension registration in a routes file, and then that extension will be togglable using the feature flag.

## Related Issue

https://issues.openmrs.org/browse/O3-2258

### DIFF
--- a/packages/framework/esm-extensions/src/extensions.ts
+++ b/packages/framework/esm-extensions/src/extensions.ts
@@ -19,6 +19,7 @@ import {
   getExtensionSlotConfigFromStore,
   getExtensionSlotsConfigStore,
 } from "@openmrs/esm-config";
+import { featureFlagsStore } from "@openmrs/esm-feature-flags";
 import isEqual from "lodash-es/isEqual";
 import isUndefined from "lodash-es/isUndefined";
 import {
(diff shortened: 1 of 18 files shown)


packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -282,16 +283,27 @@ function getOrder(
...
  * @param online Whether the app is currently online. If `null`, uses `navigator.onLine`.
+ * @param enabledFeatureFlags The names of all enabled feature flags. If `null`, looks
+ *    up the feature flags using the feature flags API.
  * @returns A list of extensions that should be rendered
  */
 export function getConnectedExtensions(
   assignedExtensions: Array<AssignedExtension>,
-  online: boolean | null = null
+  online: boolean | null = null,
+  enabledFeatureFlags: Array<string> | null = null
```
brandones, line 293: I wasn't sure whether the featureFlag check should happen in the step where we determine assigned extensions or connected extensions. But if I had done it in the assigned extensions phase, I'd have needed to create a subscription on the feature flags store which updated the extension output store. I opted for the less complex option, which unfortunately creates a bit of ambiguity around the meaning of the word "connected."

brandones:
This is ready for review @ibacher @denniskigen

packages/framework/esm-extensions/src/extensions.ts
ibacher: At some point, we should probably look into refactoring these anyways (I've never been a big fan of having hooks explicitly named after "connection status" rather than, say, some sort of lifecycle phase).

ibacher (approved):
LGTM. Thanks @brandones!

PR #738, opened 2023-07-26
(feat) Core support for O3-2242: Separate 'active visit' and 'current visit'

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Creates separate "active visit" and "current visit" returns from `useVisit`. This will allow us to have the user go between retrospective entry and entry on an ongoing visit.

## Screenshots
Screencast from 2023-07-25 20-09-28

## Related Issue
https://issues.openmrs.org/browse/O3-2242

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -1401,9 +1401,21 @@ ___
 
 ▸ **useVisit**(`patientUuid`): [`VisitReturnType`](interfaces/VisitReturnType.md)
 
-This React hook returns a visit object. If the `patientUuid` is provided
-as a parameter, then the currentVisit, error and mutate function
-for that patient visit is returned.
+This React hook returns visit information if the patient UUID is not null. There are
+potentially two relevant visits at a time: "active" and "current".
+
+The active visit is the most recent visit without an end date. The presence of an active
+visit generally means that the patient is in the facility.
+
+The current visit is the active visit, unless a retrospective visit has been selected.
+If there is no active visit and no selected retrospective visit, then there is no
+current visit. If there is no active visit but there is a retrospective visit, then
+the retrospective visit is the current visit. `currentVisitIsRetrospective` tells you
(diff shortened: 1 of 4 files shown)


ibacher:
So, technically, the code here is great. I've got a conceptual niggle with this, though. Technically, it's a configurable option whether overlapping visits are allowed or not (and, by default in the RefApp, we allow overlapping visits). That means, its perfectly plausible for a patient to have two "active" visits and a user might need to switch between them.

The use-case for multiple active visits is cases where a user switches between services; e.g., a patient on a mental health ward may have a medical issue and be admitted to a clinical care unit; they haven't ended their mental health visit, but also the clinical care visit is separately managed care because when they are discharged from the clinical care unit, they are being discharged back to the mental health ward.

The simplicity of having a "current" visit kind of allows us to hide those details, but it does mean that when existing "retrospective" mode, we need a way for the user to be put back in the state they were in before. Could we do something with, e.g., a stack of visits maybe so when you exit retrospective mode it pops the retrospective visit, and the current visit now points to an active visit? Or alternatively, prompting users?

Because so many OpenMRS facilities are run as HIV clinics, this isn't a common case across the community, but there are implementations that provide mental health services that have exactly the concern I outlined above.

brandones:
Yeah that makes sense @ibacher . However we don't really have UI support for multiple active visits and so I'm not inclined to build anything specific for that in Core until we have some vague idea of how we actually expect users to interact with them. Right now there's not even a way to enter RDE mode (that I'm aware of) when the patient has an active visit. The content of this PR seems like a safe bet to me though. Current visit sets the user context. Active visit seems likely to become an array "active visits" in the future. But we can jump that hurdle when we get to it.

Does that seem like a fine approach, or are there changes you would like to see that you think are reasonable to include in the scope of this problem?

ibacher (approved):

ibacher:
It's a good point. We should keep the scope here as limited as possible.

PR #742, opened 2023-08-04
(fix) Resolve some errors that appear when an app has empty routes

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

In one of the recent releases, the form builder would fail to load because a `publish` step wasn't properly defined, causing an empty entry in the `route.registry.json`. This PR at least prevents that specific set of circumstances from happening by ensuring that:

1. The record only appears in the `routes.registry.json` file if the `routes.json` file exists
2. If we get an error while trying to preload a module, we swallow the error.

### DIFF
--- a/packages/shell/esm-app-shell/src/run.ts
+++ b/packages/shell/esm-app-shell/src/run.ts
@@ -142,7 +142,8 @@ async function preloadScripts() {
   ]);
 
   window.installedModules.map(async ([module]) => {
-    importDynamic(module, undefined, { importMap });
+    // we simply swallow the error here since this is only a preload
+    importDynamic(module, undefined, { importMap }).catch();
   });
 }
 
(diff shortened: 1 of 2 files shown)


brandones:
We swallow the error because a more verbose and useful error happens later?

ibacher:
@brandones In this particular case, we should actually throw a better and more useful error before we get to this point, i.e., during the assemble command.

However, the reason for not throwing an error here is that it’s not in response to a user action; in the loop I’ve added the catch to, we’re trying to pre-cache the container scripts for apps the user hasn’t loaded yet to speed things up when they load them. At the point where the user would perform an action that would require the script, they would still get basically the same error.

brandones (approved):

brandones:
Sorry about the long turnaround; LGTM

PR #746, opened 2023-08-11
Introduce locale sensitive date-picker

samuelmale:
## Requirements
- [X] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

#### For changes to apps
- [X]  My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://zeroheight.com/23a080e38/p/880723-introduction).

#### If applicable

## Summary

This PR basically introduces a locale-sensitive date-picker to the style-guide. So far it's a mix of Carbon and Adobe React Spectrum's (ARS) date-picker.

#### How does it work?
ARS' date-picker has more concrete support for Internationalization out of the box. The fact that it supports non-gregorian calendars was the most compelling factor for choosing this library. We simply render ARS's date-picker if the current locale is [supported by ARS](https://github.com/openmrs/openmrs-esm-core/blob/2e6772d722b2dc7c7b07b39cbe9b5349740703ef/packages/framework/esm-styleguide/src/datepicker/react-spectrum/locales.ts), otherwise, we default to Carbon's date-picker. Take note that all the "English" specific locales have been excluded from the ARS' supported locales.
 
#### What next?
- Carbonize the ARS's date-picker for more consistency 
- Get rid of Carbon's date-picker
## Screenshots

Screenshot 2023-08-11 at 14 53 53
<br/>
Screenshot 2023-08-11 at 14 55 36
<br/>
Screenshot 2023-08-11 at 14 56 11
<br/>
Screenshot 2023-08-11 at 14 56 42

## Related Issue

N/A

### Resources
- https://talk.openmrs.org/t/o3-embedding-international-calendars-datepicker-and-date-converter-to-o3/38226
- https://talk.openmrs.org/t/o3-supporting-non-western-calendars-ethiopia-and-beyond/36566
### Shouts 
Kudos to @sinteco for working on this


samuelmale:
Early comments:

React Spectrum's picker has a white (rgb(248, 248, 248)) background color which makes it look a little weird if the container doesn't explicitly define boundaries. Maybe we should use the same color as Carbon's?

packages/framework/esm-styleguide/src/datepicker/react-spectrum/adobe-react-spectrum-date-wrapper.component.tsx
```diff
@@ -0,0 +1,88 @@
+import React from "react";
+import { CalendarDate, parseDate } from "@internationalized/date";
+import { DatePicker } from "@react-spectrum/datepicker";
+import { Provider } from "@react-spectrum/provider";
+import { theme as defaultTheme } from "@react-spectrum/theme-default";
+import { parseDate as parseArbitraryDateString } from "@openmrs/esm-framework";
+
+const prefferedEthiopicLocaleCode = "am-AM-u-ca-ethiopic";
```
ibacher, line 8: 1. This should be configurable
2. We need to support more than just the Ethiopic calendar

packages/framework/esm-styleguide/src/datepicker/react-spectrum/locales.ts
```diff
@@ -0,0 +1,40 @@
...
+  "zh-CN": "Chinese (Simplified)",
+  "zh-TW": "Chinese (Traditional)",
+  "hr-HR": "Croatian (Croatia)",
+  "cs-CZ": "Czech (Czech Republic)",
+  "da-DK": "Danish (Denmark)",
+  "nl-NL": "Dutch (Netherlands)",
+  // English specific locales will be only supported by Carbon
+  //   "en": "English",
+  //   "en-GB": "English (Great Britain)",
+  //   "en-US": "English (United States)",
```
ibacher, line 14: It would be best to also support English through this component, so we can just use one.

ibacher (commented):
Thanks for the nice work on this! A few things (this whole thing is a bit tricky), but hey.

mseaton:
I added @mogoodrich as a reviewer, as I know he is interested in this functionality.

packages/framework/esm-styleguide/src/datepicker/react-spectrum/adobe-react-spectrum-date-wrapper.component.tsx
```diff
@@ -0,0 +1,88 @@
...
+  maxDate,
+  readonly,
+  isDisabled,
+  invalid,
+  invalidText,
+  onChange,
+}) => {
+  return (
+    <Provider
+      locale={locale === "am" ? prefferedEthiopicLocaleCode : locale}
```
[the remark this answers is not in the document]
samuelmale: Nice catch, though it's: `preferredEthiopicLocaleCode`

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  return (
+    <Provider locale={locale} colorScheme="light" theme={defaultTheme}>
+      <DatePicker
+        id={id}
+        label={labelText}
+        value={parseToCalendarDate(value)}
+        onChange={(calendarDate) => {
+          onChange?.(calendarDate.toDate("UTC"));
```
mogoodrich, line 96: I haven't researched how this calendar widget works, but, just to confirm, what does this do?  My concern, based on the the behavior I was seeing in the Form Engine app, was that this might change the date?  We don't want necessarily to convert the date to UTC.  For instance, if the user is in one time zone ahead of UTC,  can we confirm that onChange, the calendarDate *doesn't* get set to "2023‐08‐18T00:00:00+01:00" and then *incorrectly* shifted to "023‐08‐17T23:00:00Z".  (ie the day before).  When working with pure dates, we really want the "birthdate" / "LocalDate" approach as definied here: https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html   which seems surprisingly tricky to do.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  onChange,
+}) => {
+  const [preferredEthiopicLocale, setPreferredEthiopicLocale] = useState();
+
+  const locale = useMemo(() => {
+    if (preferredLocale) {
+      return preferredLocale;
+    }
+    const currentLocale = getLocale();
+    if (currentLocale === "am" && preferredEthiopicLocale) {
```
mogoodrich, line 75: Seems odd/unfortunate to hard-code locale references here, but if we need to, can we add a comment documenting what this does?

mogoodrich (commented):
I'm not familiar with this widget, so didn't do a full review, but please see my comments @samuelmale  @ibacher @mseaton , especially the one about setting UTC date.

packages/framework/esm-styleguide/src/datepicker/index.tsx
samuelmale: Valid point @mogoodrich

packages/framework/esm-styleguide/src/datepicker/index.tsx
samuelmale: I've very limited knowledge about this but I believe the ICAP (Ethiopian) team has done the necessary calibrations and chose  "am-AM-u-ca-ethiopic" locale code over the "am". I've tested both locales:

1. "am": Within this locale, the year timeline is similar to the Gregorian calendar timeline (The current year is 2023)
2. "am-AM-u-ca-ethiopic": This locale has a custom timeline and the current year is 2015 and I believe that's what the ICAP team expects.

I've traced in the date utils and found a "similar" override [here](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/omrs-dates.ts#L273-L275). Take note that this doesn't overwrite the locale provided by the component consumer. 

cc:  @sinteco, @aman-icap

ibacher: Those are slightly different. The `ht` thing is about having a fallback language that's different from the base language (`ht-HT` -> `fr-HT` -> `fr`, i.e., Kreyòl as spoken in Haiti to French as spoken in Haiti to general french—technically this has been fixed in the Unicode CLDR, so we may not need the hack). `am-AM-u-ca-ethiopic` is the `am-AM` (Amharic) locale with the extension (`u`) tell us to use the Ge'ez (`ethiopic`) calendar (`ca`).

Concretely, I think the ask here is that instead of having a `preferredEthiopicLocale` property, we have a `preferredCalendar` configuration option which could take an object like:

```js
{
  "am": "ethiopic",
  "en-US": "buddhist",
}
```

To support both the needs of the ICAP team and the weird edge-case where I want the app in US English, but the calendars to use the Thai Buddhist calendar.

Incidentally, you can use that data to construct the string programmatically like this:

```js
let locale = new Intl.Locale("am", {calendar: "ethiopic"}) ;
locale.toString(); // am-u-ca-ethiopic
```

samuelmale: This was a lot of learning for me. Addressed in my latest commit.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const currentLocale = preferredLocale || getLocale();
+    if (preferredCalendar?.[currentLocale]) {
+      return new Intl.Locale(currentLocale, {
+        calendar: preferredCalendar[currentLocale],
+      }).toString();
+    }
+    return currentLocale;
+  }, [preferredLocale, preferredCalendar]);
+
+  getConfigStore("@openmrs/esm-styleguide").subscribe((store) => {
```
ibacher, line 80: You can now just use `useConfig()` to do this, which has a number of advantages (inter alia, never call `subscribe()` outside of a `useEffect()` hook. The subscription will never be dropped, which leads to a memory leak).

samuelmale: That was my original approach but using `useConfig()` in this context autowires the config belonging to the esm consuming the styleguide. Let's keep in mind the style guide is a library.

ibacher: I meant you can do:

```js
useConfig("@openmrs/esm-styleguide");
```

This is a [new feature](https://github.com/openmrs/openmrs-esm-core/pull/751).

samuelmale: Woow, nice one!

ebambo:
@samuelmale thank you so much for working on this.  @mseaton @ibacher @mogoodrich I agree and value everyone's comments and ideas shared for this PR but I have a special request: can we revert (temporarily) to the initial solution Samuel has proposed, of having the carbon date picker being returned if the timezone is English otherwise return the spectrum data picker. Am asking this because the Ethiopia Team needs this feature incorporated as soon as possible, they are rolling out O3 next week.  They had initially forked everything but we managed to move them away from that. The plan is not to leave this as is but work on carbonizing the component as recommended.
CC @gracepotma

mogoodrich:
@ebambo  I haven't been following this super closely, but I'm generally okay with doing something temporary in order for you to move forward, as long as there is a plan to move away from the temporary solution.  Defer to @samuelmale and @ibacher  on the specifics.

ibacher:
@ebambo In principle, I think it's fine to integrate something like the earlier version. The comments Mark and I made specifically around date calculations will need to be fixed. It would also be important to swap the call to `getConfigStore()` for a call to `useConfig()` because, as is, this is a memory leak. Finally, the last version of this PR that uses the Carbon component incorrectly assigns any `value` as a property of the `DatePickerInput`, instead of the `DatePicker` component. 

As long as those minimal changes are addressed and we have some assurance that changing from `DatePicker` -> `OpenmrsDatePicker` will be tested with both the Ge'ez and Gregorian calendars, we can go ahead and merge the earlier version of this.

ebambo:
@ibacher @mogoodrich thank you so much for prompt response. @ibacher sure thing, we will address the points you raised and merge the PR.

packages/framework/esm-styleguide/src/datepicker/react-spectrum/locales.ts
```diff
@@ -0,0 +1,5 @@
+export const supportedLocales = {
+  am: "Amharic",
+  am_ET: "Amharic (Ethiopia)",
+  ti_ET: "Tigrinya (Ethiopia)",
+};
```
samuelmale, line 5: To minimize potential chaos, I have restricted this to locales specific to Ethiopia only.

PR #747, opened 2023-08-11
(fix) Bug where store actions don't receive correct arguments; get rid of appState

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary
[(fix) Improve typing of store actions](https://github.com/openmrs/openmrs-esm-core/commit/0851add3bd97cc355c6dd6ab1536de397a92aa16). I had provided an invalid action and there was no type error. It would have been easier to debug the problem had there been a type error.

[(fix) Bug where store actions don't receive correct arguments](https://github.com/openmrs/openmrs-esm-core/commit/5a115aec3219fd9f63c993473e7e05cc3be1ba61). Once I was providing correct actions, it still didn't work, because of a bug.

[(fix) Get rid of appState](https://github.com/openmrs/openmrs-esm-core/commit/baeeda333c47912718ac2be5c0553f059c8a0c52). No one uses it, which is good, because we probably should avoid having state this global unless we make a very clear and well-considered decision that we need it.

I've also made it so that esm-state exposes all the app's stores on `window.stores` when in development mode. This should make debugging certain kinds of problems a little easier.


ibacher (approved):
Thanks @brandones! This looks great!

PR #751, opened 2023-08-16
(feat) Allow useConfig() to load configuration from other modules

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Occasionally, we want a module to be able to load the configuration from a different module; this is possible in the underlying configuration API, but not via the `useConfig()` hook. This PR just extends the `useConfig()` hook to allow you to pass an _optional_ module name, in which case `useConfig()` will return the configuration of that module. Note that the usage of this feature is _discouraged_ as it may add unnecessary coupling between modules.

### DIFF
--- a/packages/framework/esm-react-utils/src/useConfig.ts
+++ b/packages/framework/esm-react-utils/src/useConfig.ts
@@ -157,31 +157,37 @@ function useNormalConfig(moduleName: string) {
   return state;
 }
 
+interface UseConfigOptions {
+  /** An external module to load the configuration from. This option should only be used if
+      absolutely necessary as it can end up making frontend modules coupled to one another. */
+  externalModuleName?: string;
+}
+
 /**
  * Use this React Hook to obtain your module's configuration.
+ *
+ * @param options Additional options that can be passed to useConfig()
  */
-export function useConfig<T = Record<string, any>>(
-  moduleName: string | null | undefined = undefined
-) {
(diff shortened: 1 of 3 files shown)


packages/framework/esm-react-utils/src/useConfig.ts
```diff
@@ -160,26 +160,28 @@ function useNormalConfig(moduleName: string) {
 /**
  * Use this React Hook to obtain your module's configuration.
  */
-export function useConfig<
-  T = Omit<ConfigObject, "Display conditions" | "Translation overrides">
->() {
+export function useConfig<T = Record<string, any>>(
+  moduleName: string | null | undefined = undefined
```
brandones, line 164: Why not

```suggestion
  moduleName?: string
```

?

packages/framework/esm-react-utils/src/useConfig.ts
```diff
@@ -160,26 +160,28 @@ function useNormalConfig(moduleName: string) {
 /**
  * Use this React Hook to obtain your module's configuration.
```
brandones, line 161: I suggest renaming `moduleName` to something like `foreignModuleName` and adding a docstring comment explaining that it is not needed for normal use and indeed should not be used if it can be avoided. Making `useConfig` take an `options` object as its argument so that users have to type out `foreignModuleName` or whatever in order to pass one might be even better.

brandones (approved):
Thanks Ian! Nice test, too.

PR #756, opened 2023-08-24
(feat) Adding useDebounce to the react utils

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary

We use SWR everywhere to fetch information and control the number of requests, especially for searching. We also use lodash in many places for this, which is fine. However, one downside is that the search component is no longer a controlled input. This is because the state set by the search input will be set after a delay, and if that same state is passed to the search box, it will not work correctly in the UI. In some cases, having an uncontrolled input is acceptable, but in cases where there is an external "Clear results" button that clears the search input, it becomes tedious to handle.

Instead of all these workarounds for a single use case, why not have a hook that handles debouncing for us without requiring an uncontrolled input?

For a similar example, you can refer: https://github.com/openmrs/openmrs-esm-patient-chart/pull/1322

## Example
```tsx
import { useDebounceValue } from "@openmrs/esm-react-utils";

function MyComponent() {
  const [searchTerm, setSearchTerm] = useState('');
  const debouncedSearchTerm = useDebounce(searchTerm);
  const swrResult = useSWR(`/api/search?q=${debouncedSearchTerm}`) 

  return (
    <Search
      labelText={t('search', 'Search')}
      onChange={(e) => setSearchTerm(e.target.value)}
      value={searchTerm}
    />
  )
}
```


packages/framework/esm-react-utils/src/useDebounce.ts
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useState } from "react";
+
+/**
+ * This hook will help us out in getting the debounced value of a state variable,
+ * which then makes a request to the backend.
```
brandones, line 5:
```suggestion
 * This hook debounces a state variable. That state variable can then be used as the
 * value of a controlled input, while the return value of this hook is used for making
 * a request.
```

packages/framework/esm-react-utils/src/useDebounce.ts
```diff
@@ -0,0 +1,43 @@
...
+ * 
+ * For example,
+ *
```
brandones, line 8:
```suggestion
 * @example
```

packages/framework/esm-react-utils/src/useDebounce.ts
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useState } from "react";
+
+/**
+ * This hook will help us out in getting the debounced value of a state variable,
+ * which then makes a request to the backend.
+ * 
+ * For example,
+ * 
+ * ```tsx
+ * import { useDebounceValue } from "@openmrs/esm-react-utils";
```
brandones, line 10:
```suggestion
 * import { useDebounce } from "@openmrs/esm-react-utils";
```

packages/framework/esm-react-utils/src/useDebounce.ts
```diff
@@ -0,0 +1,43 @@
...
+ *   const debouncedSearchTerm = useDebounce(searchTerm);
+ *   const swrResult = useSWR(`/api/search?q=${debouncedSearchTerm}`) 
+ * 
+ *  return (
+ *    <Search
+        labelText={t('search', 'Search')}
+        onChange={(e) => setSearchTerm(e.target.value)}
+        value={searchTerm}
+      />
+ *  )
```
brandones, line 23:
```suggestion
 *    <Search
 *      labelText={t('search', 'Search')}
 *      onChange={(e) => setSearchTerm(e.target.value)}
 *      value={searchTerm}
 *    />
 *  )
```

packages/framework/esm-react-utils/src/useDebounce.ts
```diff
@@ -0,0 +1,43 @@
...
+ * }
+ * 
+ * ```
+ * 
+ * @param value
+ * @param delay number = 300
+ * @returns debounceValue
+ */
```
brandones, line 31:
```suggestion
 * ```
 * 
 * @param value: The value that will be used to set `debounceValue`
 * @param delay number = 300: The number of milliseconds to wait before updating `debounceValue`
 * @returns debounceValue: The debounced value
 */
```

brandones:
You'll need to add the export to `public.ts` for it to be importable via `@openmrs/esm-framework`.

Please add this to the esm-framework mocks as well. You can just do `export { useDebounce } from '....wherever';`.

Please don't just check all the boxes in the PR. Only check them if they're both relevant and done. Styleguide: irrelevant. Tests: not done but optional for a hook this simple. Mock: required, not done.

brandones:
Applied suggestions. You still need to add to public.ts.

vasharma05:
I thought I did. Pushing it now

vasharma05:
I did the changes but just had it locally only 🤦 

Please re-review
Thanks!

packages/framework/esm-framework/docs/API.md
```diff
@@ -154,6 +154,7 @@
 ### Other Functions
 
 - [ExtensionSlot](API.md#extensionslot)
+- [useDebounce](API.md#usedebounce)
```
ibacher, line 157: Could we categorise this as an API function?

packages/framework/esm-framework/docs/API.md
```diff
@@ -3751,6 +3752,60 @@ Passing a function as children
...
+| Name | Type | Default value | Description |
+| :------ | :------ | :------ | :------ |
+| `value` | `T` | `undefined` | - |
+| `delay` | `number` | `300` | number = 300: The number of milliseconds to wait before updating `debounceValue` |
+
+#### Returns
+
+`T`
+
+debounceValue: The debounced value
```
ibacher, line 3801: It looks like this comment isn't formatted correctly (note the "Default value" of `undefined` and the lack of description for `value` above. Also, I'm not sure the return docs are right.

brandones (approved):

PR #765, opened 2023-09-11
(feat) Age should be represented in internationalized translated way

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
The age function currently returns the age in english strings, i.e. "4 years", "4 weeks". The age function will now return the age according to the locale of the application.

### Infant born on the same day
image

image

### Infant smaller than 1 month
image

image

### Infant less than 1 year
image

image

### Child less than 2 years
image

image

### Age in years
image

image


ibacher (changes requested):
We can't accept this as-is, I think. It's common to refer to ages under 2 years (and sometimes three) by months, that is "18 months" rather than "1 year 6 months", even though these refer to the same period of time. If we want to make that English-specific, I'd accept it with the endorsement of a non-English-speaking pediatrician, but absent that, we should maintain the same output as is already here.

Also, it would be best if we didn't do all of the calculations of every possible string every time, but only when they were relevant.

packages/framework/esm-utils/src/age-helpers.tsx
```diff
@@ -67,24 +67,64 @@ export function age(dateString: string): string {
   // The "remainder" number of days after the weeksAgo number of weeks
   const remainderDaysInWeek = totalDaysAgo - weeksAgo * 7;
 
+  const rtf = new Intl.RelativeTimeFormat(getLocale(), { numeric: "auto" });
+
+  const totalDaysAgoStr =
```
ibacher, line 72: Sorry, a couple more small things:

1. Can we reduce the `getLocale()` calls to a single call?
2. It would be nice if we only calculated the strings we're actually using instead of all of them every time.
3. For weeks ago, I'd probably leave of the "days". Although this is relevant, I'm worried that the word order isn't always going to render properly, e.g., for Hebrew and Arabic, and we can fix that later.

vasharma05:
Hi @ibacher !
I've made the changes and updated the screenshots too.

ibacher (approved):

PR #755, opened 2023-08-23
(fix) Tabs should have the max-width of their content

vasharma05:
## Requirements
- [ ] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

#### For changes to apps
- [ ]  My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and **design documentation**.

#### If applicable
- [ ] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary
The tabs have a max-width of 10rem, but it should be expanded to their max-content

### Before
image

### After
image

PR #770, opened 2023-09-14
(chore) Bump @carbon/react and @carbon/charts

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

In this PR, I bumped the `@carbon/charts` and `@carbon/react` packages to their latest versions. In https://github.com/openmrs/openmrs-esm-core/pull/755, some of the changes included are obviated by upstream changes to @carbon/react. This upgrade allows us to leverage some [potential bundle size reductions](https://github.com/carbon-design-system/carbon-charts/pull/1652) and improved functionality per the associated release notes for each package. The Carbon team are working on adding type support to individual components incrementally. This has led to some known issues with typings, which is why I've elected to add ambient type declarations from `@carbon/react` to each individual package's declarations file and omit some type declarations (for now). 

## Related Issue

https://github.com/openmrs/openmrs-esm-core/pull/755

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/declarations.d.ts
+++ b/packages/apps/esm-primary-navigation-app/src/declarations.d.ts
@@ -1,0 +1,7 @@
+declare module "*.css";
+declare module "*.scss";
+declare module "@carbon/react";
+declare type HeaderPanelProps = {};
+declare type SideNavProps = {
+  isChildOfHeader: boolean;
+};
--- a/packages/apps/esm-devtools-app/package.json
+++ b/packages/apps/esm-devtools-app/package.json
@@ -50,6 +50,6 @@
     "webpack": "^5.88.0"
   },
   "dependencies": {
-    "@carbon/react": "^1.12.0"
+    "@carbon/react": "^1.37.0"
   }
 }
--- a/packages/apps/esm-devtools-app/src/declarations.d.ts
+++ b/packages/apps/esm-devtools-app/src/declarations.d.ts
@@ -1,3 +1,5 @@
+declare module "@carbon/react";
+
 declare namespace JSX {
   interface IntrinsicElements {
     "import-map-overrides-list": any;
@@ -13,3 +15,5 @@ declare module "*.scss" {
   const styles: any;
   export default styles;
 }
+
+declare type SideNavProps = {};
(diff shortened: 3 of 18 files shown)


ibacher (approved):

packages/apps/esm-primary-navigation-app/src/declarations.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module "*.css";
+declare module "*.scss";
+declare module "@carbon/react";
+declare type HeaderPanelProps = {};
+declare type SideNavProps = {
```
ibacher, line 5: Worryingly, this one seems to [actually be defined](https://github.com/carbon-design-system/carbon/blob/5078eb5204b21763244768af3b0b283551561d47/packages/react/src/components/UIShell/SideNav.tsx#L31C8-L31C8).

denniskigen: Yeah. Dunno why it's not getting picked up.

ibacher: Yes, especially because we seem to be trying directly use the type name, but the type isn't exported. Even doing the really hacky `Parameters<typeof SideNav>[0]` gives me "Exported variable 'LeftNavMenu' has or is using name 'SideNavProps' from external module ... but cannot be named.ts(4023)
"

ibacher: I guess we'll see how [this](https://github.com/carbon-design-system/carbon/pull/14657) goes...

denniskigen: Thanks, Ian.

denniskigen: PR got merged, @ibacher! Do we need to wait for a release?

ibacher: Unfortunately yes. Unlike what we do Carbon [doesn’t appear to publish on every commit](https://www.npmjs.com/package/@carbon/react?activeTab=versions), so the only way to get those changes right now would be to publish our own version from the `master` branch which sounds messy.

Fortunately, Carbon appears to do a release every 2 weeks or so. I guess we can either hold off on updating until the next release or just aim to fix this as soon as there’s a RC. I’d probably go for the “fix it in a couple of weeks” part.

PR #772, opened 2023-09-21
Adding configuration to allow loading css loaders for react-html5-camera-photo

arodidev:
…era-photo

## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR adds configurations to the base webpack config file to allow adding of css loaders that enable styles for react-html5-camera-photo component, which is currently being bottlenecked due to lack of configuration. This is in relation work being done to add file-attachment rendering to the react form engine, where files can be uploaded and also captured via camera. The camera component being imported has issues with loading the shutter button out of the gate, hence the need to import a css file to add styles to the camera component. 

The camera component though is unresponsive to css being loaded directly from the imported file and a webpack configuration has to be added to the consuming application to allow for the styles to apply to the camera component, so that the shutter button is visible out of the box. 

Housing the configuration in the base webpack config file will allow for the configurations to be applied to any esm utilizing react form engine without the need of further adding configuration to each esm.

### DIFF
--- a/packages/tooling/webpack-config/src/index.ts
+++ b/packages/tooling/webpack-config/src/index.ts
@@ -314,3 +314,18 @@ export default (
   };
   return mergeWith(baseConfig, additionalConfig, mergeFunction);
 };
+
+const cfgCreator = require('openmrs/default-webpack-config');
+
+module.exports = (env, argv) => {
+  const config = cfgCreator(env, argv);
+  const cssLoaderMode = resourcePath => {
+    if (/.*react-html5-camera-photo\/build\/css\/index.css/i.test(resourcePath) || /styles.css$/i.test(resourcePath)) {
+      return 'global';
+    }
+    return 'local';
+  };
+  config.module.rules[1].use[1].options.modules['mode'] = cssLoaderMode;
+  config.module.rules[2].use[1].options.modules['mode'] = cssLoaderMode;
+  return config;
+};


packages/tooling/webpack-config/src/index.ts
```diff
@@ -314,3 +314,18 @@ export default (
...
+  const config = cfgCreator(env, argv);
+  const cssLoaderMode = resourcePath => {
+    if (/.*react-html5-camera-photo\/build\/css\/index.css/i.test(resourcePath) || /styles.css$/i.test(resourcePath)) {
+      return 'global';
+    }
+    return 'local';
+  };
+  config.module.rules[1].use[1].options.modules['mode'] = cssLoaderMode;
+  config.module.rules[2].use[1].options.modules['mode'] = cssLoaderMode;
+  return config;
```
samuelmale, line 330: @arodidev have you actually tested this out locally? There is no way this works.

denniskigen: @samuelmale Could you work with Jamie to come up with something that works

vasharma05:
We are using the html5 camera photo package in the attachments too. Please do confirm the attachements widget before merging.
Thanks!

ibacher:
Yeah, one consideration for this is what happens if we mark the CSS as global (so non-module scoped), but end up with multiple copies of it. Rather than just hacking around with how the CSS is loaded via webpack, we probably want to make it into a shared module (`peerDependency`).

arodidev:
Did you mean something like [this](https://github.com/openmrs/openmrs-esm-core/blob/45d267e14fb62160b0878033c662bbd00b7e31f6/packages/tooling/webpack-config/src/index.ts#L266)

ibacher:
@arodidev What I was suggesting is that each app that uses the react-html5-camera-photo should have that library listed in their `peerDependencies`. That part shouldn’t require any modifications to the default webpack configuration.

arodidev:
@ibacher I attemped adding 'react-html5-camera-photo' as a `peerDependency`  to the app utilising the react-html5-camera-photo (which is the react form engine) and this does not seem to solve the issue. Could this be resolved in another way?

PR #774, opened 2023-09-27
(feat) Use SWR as a shared library

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

So these should be the code changes necessary to load SWR as a ModuleFederation shared. Using SWR this way is hard due to two glaring bugs, specifically https://github.com/webpack/webpack/issues/16125 and https://github.com/vercel/swr/issues/2356.

In essence the way this works is just what's suggested in https://github.com/webpack/webpack/issues/16125, specifically, only the SWR sub-modules are actually marked as webpack shares while SWR core itself is not. However, because of how SWR's internals works, this actually works (i.e., the cache is shared across apps).

This PR also does the following:

1. Forces the primary navigation to not render the default component if we're on a login page. Previously this was handled by logic inside the component, which lead to a bunch of unnecessary re-rendering. Now, while the app is mounted always, the primary nav component only render on URLs that do not start with `${openmrsSpaBase}login` or `${openmrsSpaBase}logout`.
2. Webpack shares shared from the app shell now actually declare a "requiredVersion", which basically just brings the entries in line with how app Webpack shares work.
3. The infinite scroll for login locations now actually uses the FHIR `next` links instead of sending a new request. This should be more efficient.

Once this is merged, for packages that we want to use the shared SWR cache (which should be most of them), we need to update the version of `openmrs` and add `swr` to the peer dependencies.


denniskigen (approved):
LGTM, @ibacher. Keeping tabs on the related issues.

brandones:
Wow. Awesome work, Ian. Thanks so much for taking this on.

PR #781, opened 2023-10-05
(fix) Fix the appearance of Carbon icon-only buttons

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR fixes a regression affecting the icon-only variant of the Carbon Button component. Before the recent Carbon version bump, the icon-only variant of the Button component would adjust its size according to whatever value passed to it via the `size` prop. This is owing to a new `--cds-layout-size-height-local` class that got added after an upstream [change](https://github.com/carbon-design-system/carbon/pull/13287) to the Button component that added [contextual layout tokens](https://github.com/carbon-design-system/carbon/issues/13923). 

It does so by setting the default size of the button to `sm` and removing some extraneous padding.

## Screenshots

> After 
CleanShot 2023-10-05 at 9  31 58@2x

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -33,6 +33,14 @@
   border-right: none;
 }
 
+/* Icon button */
+.cds--btn--icon-only {
+  // Apply a size layout token that sets the default size of the icon-only button to `sm`
+  @include layout.use("size", $default: "sm");
+  // Remove extraneous top padding from the icon-only button
+  padding-block-start: 0;
+}
+
 /* Side nav */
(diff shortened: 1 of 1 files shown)


ibacher (approved):

packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -33,6 +33,14 @@
   border-right: none;
 }
 
+/* Icon button */
+.cds--btn--icon-only {
+  // Apply a size layout token that sets the default size of the icon-only button to `sm`
+  @include layout.use("size", $default: "sm");
```
vasharma05, line 39: Will these work for tablet views as well?

PR #782, opened 2023-10-06
(feat) O3-2484: The application should revert to default locale on logout

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
For an application having the default locale other than `en` (say 'km'), if the logged in user changes the locale to `en`, after the logout, the application's locale is set to `en`, instead of `km`. This PR resets the application to the default locale on logout.

## Screenshots
https://github.com/openmrs/openmrs-esm-core/assets/51502471/af84fc74-5568-44d7-97f6-98e936c97761

## Related Issue
https://issues.openmrs.org/projects/O3/issues/O3-2484


packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+      authenticated: true,
+      sessionId: "xyz",
+    } as Session);
+    (useConfig as jest.Mock).mockReturnValue({
+      provider: {
+        type: "",
+      },
+    });
+  });
+  it("should render Logout and redirect to login page ig provider.type !== oauth2", async () => {
```
brandones, line 55:
```suggestion
  it("should render Logout and redirect to login page if provider.type !== oauth2", async () => {
```

packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="<%= openmrsDefaultLocale %>">
+<html lang="<%= openmrsDefaultLocale %>" defaultLang="<%= openmrsDefaultLocale %>">
```
brandones, line 2:
```suggestion
<html lang="<%= openmrsDefaultLocale %>" data-default-lang="<%= openmrsDefaultLocale %>">
```

Custom attributes must be prefixed.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
```diff
@@ -19,6 +20,13 @@ const RedirectLogout: React.FC<RedirectLogoutProps> = () => {
       navigate({ to: "${openmrsSpaBase}/login" });
     } else {
       performLogout().then(() => {
+        const defaultLang =
+          document.documentElement.getAttribute("defaultLang");
```
brandones, line 24:
```suggestion
          document.documentElement.getAttribute("data-default-lang");
```

brandones (changes requested):

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
ibacher: Shouldn't we just be using the `lang` attribute for this?

brandones: The trouble is that the `lang` attribute gets updated with the user preference. That attribute is supposed to be set to the current language of that page. We need a way of storing the default language, which is passed in to the app shell and becomes an ejs parameter. I imagine there is also a way of obtaining the default without rendering it into the HTML. But I think this way of doing it is fine.

ibacher: We have an `openmrsDefaultLocale`, although right now, that's just setting the [default value of the lang attribute](https://github.com/openmrs/openmrs-esm-core/blob/03c1200de4b1a69f58cad696e3ea39c5128a2e91/packages/shell/esm-app-shell/src/index.ejs#L2). I suppose having it also set `data-default-lang` would be a good way to go.

ibacher: Yeah, ok, that makes sense.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
ibacher: I don't think the `if provider.type !== oauth2` is informative or useful here. If we need something here, it might be better to say `by default`.

Basically, I don't think our test descriptions should hard-code the assumption that the only `provider.types` are "default" or "oauth2".

brandones: @ibacher Could you take another look at this part and see if it makes sense to you? I don't know anything about provider type and oauth.

brandones:
Everything LGTM other than the `provider.type` thing @ibacher  commented on; I will let Ian take a look at the latter and give the approval.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
ibacher: I'm happy with the new test name.

How I'm thinking about this is there: there are various forms of SSO protocols (Kerberos, SAML, OIDC—which is what we're calling OAuth2) with slightly different mechanisms for communicating between the authentication server and the resource server (which is the role OpenMRS plays in any configuration). The `provider` attribute allows us to customize the way the login app works based on the expected authentication server type.

Currently, we only really have two modes: OpenMRS base authentication and OAuth2 (which is, at least OIDC-like), but it's helpful to have it be treated as more than just a binary value.

ibacher (approved):

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
```diff
@@ -19,6 +20,13 @@ const RedirectLogout: React.FC<RedirectLogoutProps> = () => {
       navigate({ to: "${openmrsSpaBase}/login" });
     } else {
       performLogout().then(() => {
+        const defaultLang =
+          document.documentElement.getAttribute("data-default-lang");
+        setUserLanguage({
+          locale: defaultLang,
+          authenticated: false,
+          sessionId: "",
+        });
```
ibacher, line 29: It might be better to make this part of the `performLogout()` or even the logic that happens [here](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-api/src/shared-api-objects/current-user.ts#L183), which might solve similar issues (for instance, when a user's session times out, should the language be reset? If yes, it probably belongs in the `refetchCurrentUser()` logic), but this is only a suggestion.

vasharma05: Right @ibacher , I'll look into that too.
Thanks!

vasharma05:
@brandones , the changes are requested by you, can you approve the PR too?
Thanks!

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
brandones: Nice, and thanks for the explanation!

brandones (approved):

PR #757, opened 2023-08-28
(fix) Locale switcher in the primary navigation should show session locale if no default locale found

vasharma05:
## Requirements
- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

#### For changes to apps
- [ ]  My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide) and **design documentation**.

#### If applicable
- [x] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary
This issue is found on the Ozone distro, where the default locale for the application is `km`. For the first time logged in users, the `defaultLocale` property is not found in the userProperties and hence the locale switcher shows the `en` as the selected language, even though the system is in Khmer.

image

## Related Issue
None

PR #783, opened 2023-10-06
(fix) Locale switcher should default to user's defaultLocale or the session locale

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
Continuation to #757 
The locale switcher was getting the session properties from the props, but props no props were passed into the component.

## Screenshots
None


packages/apps/esm-primary-navigation-app/src/components/choose-locale/change-locale.component.tsx
```diff
@@ -57,26 +57,25 @@ export const ChangeLocale: React.FC<ChangeLocaleProps> = ({
...
+        const ac = new AbortController();
+        postUserProperties(
+          user.uuid,
+          {
+            ...(user.userProperties ?? {}),
+            defaultLocale: newLocale,
+          },
+          ac
+        );
+        postSessionLocale(newLocale, ac);
```
ibacher, line 74: We'd still want to change the session locale if only that value changed, right?

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.component.tsx
```diff
@@ -43,6 +43,7 @@ const UserMenuPanel: React.FC<UserMenuPanelProps> = ({
           onLogout: onLogout,
           referer: window.location.pathname,
           currentLocation: session?.sessionLocation?.display,
+          locale: session?.locale,
```
ibacher, line 46: If we're now passing in the session locale, we should use that as a default rather than the user's default locale...

ibacher (commented):
Thanks @vasharma05! Couple of small things to adjust.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.component.tsx
vasharma05: Actually, we are prioritizing the the user's defaultLocale prop than the session locale.

Taking the example of Cambodia project, the locale ofthe application is `km` and hence it is the session locale, but if the user wants to use the app in `en`, then we'll be prioritizing the user's prop, than the session locale.

vasharma05: Do correct me if I'm wrong

packages/apps/esm-primary-navigation-app/src/components/choose-locale/change-locale.component.tsx
vasharma05: What do you recommend?

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.component.tsx
ibacher: The sessionLocale already defaults to the user's default locale once the user logs in. [See here](https://github.com/openmrs/openmrs-core/blob/04e3921cbcf3bec7b915a119214b47072cf56aa2/api/src/main/java/org/openmrs/api/context/UserContext.java#L463). My point was more we don't need the duplicative logic on the frontend.

vasharma05: Ohhkay!
Got it.
Let me push the changes then
Thanks!

vasharma05: @ibacher, so should we even call the `postSessionLocale`, since we are updating the user's `defaultLocale`?
Thanks!

ibacher:
@vasharma05 I don't think that last commit works. We still need to update the user's session locale.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.component.tsx
ibacher: Yes. By design the session locale is independent of the default locale, and should be the source of truth for the user's locale. The default locale posting is an add-on. The feature should be built around the session locale.

ibacher: To understand this: historically, it's been possible for users to have a "default locale", managed through a small administrative page, but the "session locale" is what locale things are currently displayed in. These are independent to allow, e.g., demoing or testing things in different languages without affecting the user's default display.

Somewhere along the line with O3, someone decided that we should also update the default locale when changing the session locale.

vasharma05: So, what should be my next step on this?

I'm a bit confused. Can you please help me clearing things out, what should be chosen as the locale, session locale or defaultLocale, and when it comes to update locale, which one should be updated, session locale or user defaultLocale?
Thanks!

ibacher:
I guess if it works, that's great. Saves us a network request.

vasharma05:
Hi @ibacher @denniskigen !
Can you please re-review the changes?
Thanks!

ibacher (approved):

PR #786, opened 2023-10-11
(feat) Added custom date formatter for Ethiopian calendar

aman-icap:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR adds a customized date conversion for Ethiopian calendar use cases. Any date that is displayed throughout the platform will first be converted to Ethiopian calendar format whenever the locale is set to 'am'. The Ethiopian date formatter handles all format options that the default formatter handles, such as standard or wide mode, with or without year, with or without day.

## Screenshots
scrnsht

### DIFF
--- a/packages/framework/esm-utils/package.json
+++ b/packages/framework/esm-utils/package.json
@@ -48,6 +48,7 @@
     "rxjs": "6.x"
   },
   "dependencies": {
+    "@adobe/react-spectrum": "^3.29.0",
     "semver": "7.3.2"
   }
 }
--- a/packages/framework/esm-utils/package.json
+++ b/packages/framework/esm-utils/package.json
@@ -48,7 +48,7 @@
     "rxjs": "6.x"
   },
   "dependencies": {
-    "@adobe/react-spectrum": "^3.29.0",
+    "@internationalized/date": "^3.5.0",
     "semver": "7.3.2"
   }
 }
--- a/packages/framework/esm-utils/src/ethiopian-calendar-conversion.ts
+++ b/packages/framework/esm-utils/src/ethiopian-calendar-conversion.ts
@@ -23,15 +23,16 @@ const ethiopianCalendarMonths = {
 };
 
 export function convertToEthiopianDate(date, options: FormatDateOptions) {
-  const splittedDate = dayjs(date).format("MM/DD/YYYY").split("/");
-  if (splittedDate.length === 3) {
-    let yearValue = parseInt(splittedDate[2]);
-    let monthValue = parseInt(splittedDate[0]);
-    let dayValue = parseInt(splittedDate[1]);
+  const parsedDate = dayjs(date);
+  if (parsedDate != null) {
+    let yearValue = parsedDate.get("year");
+    let monthValue = parsedDate.get("month") + 1;
+    let dayValue = parsedDate.get("date");
     let gregorianDate = new CalendarDate(yearValue, monthValue, dayValue);
     let ethiopianDate = toCalendar(gregorianDate, new EthiopicCalendar());
     return formatDate(ethiopianDate, options);
-  } else return null;
+  }
+  return null;
 }
 
 function formatDate(ethiopianDate, options) {
--- a/packages/framework/esm-utils/src/ethiopian-calendar-conversion.ts
+++ b/packages/framework/esm-utils/src/ethiopian-calendar-conversion.ts
@@ -1,0 +1,69 @@
+import {
+  CalendarDate,
+  EthiopicCalendar,
+  toCalendar,
+} from "@internationalized/date";
+import dayjs from "dayjs";
+import { FormatDateOptions } from "./omrs-dates";
+
+const ethiopianCalendarMonths = {
+  1: "Meskerem",
+  2: "Tikimt",
+  3: "Hidar",
+  4: "Tahsas",
+  5: "Tir",
+  6: "Yekatit",
+  7: "Megabit",
+  8: "Meyazya",
+  9: "Ginbot",
+  10: "Sene",
+  11: "Hamle",
+  12: "Nehase",
+  13: "Puagme",
+};
+
+export function convertToEthiopianDate(date, options: FormatDateOptions) {
+  const splittedDate = dayjs(date).format("MM/DD/YYYY").split("/");
+  if (splittedDate.length === 3) {
+    let yearValue = parseInt(splittedDate[2]);
+    let monthValue = parseInt(splittedDate[0]);
+    let dayValue = parseInt(splittedDate[1]);
+    let gregorianDate = new CalendarDate(yearValue, monthValue, dayValue);
+    let ethiopianDate = toCalendar(gregorianDate, new EthiopicCalendar());
+    return formatDate(ethiopianDate, options);
+  } else return null;
+}
+
+function formatDate(ethiopianDate, options) {
+  let formattedDate =
+    ethiopianDate.day +
+    "-" +
+    ethiopianCalendarMonths[ethiopianDate.month] +
+    "-" +
+    ethiopianDate.year;
+
+  if (options.day === false) {
+    formattedDate =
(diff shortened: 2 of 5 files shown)


packages/framework/esm-utils/src/omrs-dates.ts
```diff
@@ -205,6 +206,17 @@ export function formatDate(date: Date, options?: Partial<FormatDateOptions>) {
   let locale = getLocale();
   let localeString: string;
   const isToday = dayjs(date).isToday();
+  if (locale == "am") {
```
ebambo, line 209: I would suggest that the registry pattern is applied here. You can have a look at [this](https://github.com/openmrs/openmrs-form-engine-lib/blob/main/src/registry/registry.ts) class and [this](https://github.com/openmrs/openmrs-form-engine-lib/tree/main/src/datasources) package to have an idea. 

I believe the existing code is for the en locale (default).... I believe this approach would be beneficial for maintenance and extendability (for instance the locale in Mozambique os pt-PT, same as Portugal but we usually represent our dates differently or I may want to do other customisations). 

Any thoughts @ibacher @denniskigen @kajambiya

packages/framework/esm-utils/src/ethiopian-calendar-conversion.ts
```diff
@@ -0,0 +1,69 @@
...
+export function convertToEthiopianDate(date, options: FormatDateOptions) {
+  const splittedDate = dayjs(date).format("MM/DD/YYYY").split("/");
+  if (splittedDate.length === 3) {
+    let yearValue = parseInt(splittedDate[2]);
+    let monthValue = parseInt(splittedDate[0]);
+    let dayValue = parseInt(splittedDate[1]);
+    let gregorianDate = new CalendarDate(yearValue, monthValue, dayValue);
+    let ethiopianDate = toCalendar(gregorianDate, new EthiopicCalendar());
+    return formatDate(ethiopianDate, options);
+  } else return null;
```
ebambo, line 34: I think you can just return null, no need to add the else here.

packages/framework/esm-utils/package.json
```diff
@@ -48,6 +48,7 @@
     "rxjs": "6.x"
   },
   "dependencies": {
+    "@adobe/react-spectrum": "^3.29.0",
```
ibacher, line 51: It would be fine to depend on the `@internationalized` library, but we cannot depend on anything React-based here.

packages/framework/esm-utils/src/ethiopian-calendar-conversion.ts
```diff
@@ -0,0 +1,69 @@
...
+  8: "Meyazya",
+  9: "Ginbot",
+  10: "Sene",
+  11: "Hamle",
+  12: "Nehase",
+  13: "Puagme",
+};
+
+export function convertToEthiopianDate(date, options: FormatDateOptions) {
+  const splittedDate = dayjs(date).format("MM/DD/YYYY").split("/");
```
ibacher, line 26: Please use dayjs’s API instead of formatting the date to a string a splitting it.

packages/framework/esm-utils/src/omrs-dates.ts
ibacher: The biggest problem here, regardless of what mechanism for integrating this we use, is that `convertToEthiopianDate()` also ignores the `Today` stuff, and the time handling. These aren’t just calendar differences, and it means that we’ll have a harder time ensuring that this functions correctly in the AM locale.

@ebambo The existing code is intended to be locale-neutral with some hacks for `en` specific things the designs called for. However the conversion of the date to a format string does assume a Gregorian calendar. I’m not sure I like the idea of a per-locale registry; the changing variable is the locale calendar rather than the locale…

ebambo:
@ibacher I don't know why I can't reply directly to your comment but what would you then suggest @aman-icap approaches this if not via the registry pattern? I know they need this ASAP because of the deployments they need to do soon. Do you think he can resolve the other comments and and we merge this then we create a ticket to make this code better?

ibacher:
@ebambo I don’t have any issue with the registry pattern per we; it’s fine, but its overkill here. My issue is first, that whatever mechanism selects the implementation needs to be based not on language but on a requested calendar type (which is part of a locale); second that the implementations should (all other things being equal) actually do the same thing.

In my ideal implementation of this, we wouldn’t really have two implementations of this, though, we’d just have one implementation that relies on the user-preferred calendar and the @internationalized/date package to format the date.

packages/framework/esm-utils/src/omrs-dates.ts
aman-icap: @ebambo @ibacher The comments have been addressed on the latest commit except for the `Today` / time handling. The Ethiopian date conversion exits before the `isToday` logic because it's not needed on the Ethiopian use case. Same for the time formater, we do not insert date-time on any forms or display them anywhere. But like @ibacher said, it would be nice to support all possible date conversions through the ` @internationalized/date` package and not have multiple implemenations for each locale. But I feel like that will need it's own ticket to improve the code, as we really need this date formatter integrated as soon as possible for our upcoming deployments.

vasharma05:
Hi team, 
Please correct me if I'm wrong, but all the dates returned by the server is in the ISO Format, right?
If yes, we get the ISO string, can't we just compare the 2 ISO strings to sort dates, whatever the locale might be.
Thanks!

packages/framework/esm-utils/src/omrs-dates.ts
ebambo: @aman-icap for me it is fine. Lets merge this one and create a ticket to improve the code.

vasharma05: Let's cover all scenarios before merging in, since we are near the new release.
Thanks!

gracepotma:
@aman-icap chiming in just to say, so great to see a PR here from you :)

packages/framework/esm-utils/src/omrs-dates.ts
ebambo: @vasharma05 and @ibacher the Ethiopia team needs this code merge and @aman-icap has made a commitment to create a new ticket to work on the improvements. This is a huge blocker for them and it is clear they don't have the time now to cover all the scenarios we are mentioning and the code he have implemented does not change the original behaviour of this code. Can we please merge this?
cc: @gracepotma

gracepotma: So, after some thought: we strongly encourage following up on the above review suggestions; we're worried about merging this without those being addressed, as the one commit since hasn't really addressed these concerns: i.e. use the locale calendar rather than locale name; implement the today stuff in the Ge’ez calendar mode. The concern is we've let other PRs through in the past as a temporary measure that did not get followed up on. 

@aman-icap what timeline would you be looking at to implement those change suggestions?

gracepotma:
So, after some thought: we strongly encourage following up on the above review suggestions; we're worried about merging this without those being addressed, as the one commit since hasn't really addressed these concerns: i.e. use the locale calendar rather than locale name; implement the today stuff in the Ge’ez calendar mode. The concern is we've let other PRs through in the past as a temporary measure that did not get followed up on.

@aman-icap what timeline would you be looking at to implement those change suggestions? If there's some kind of deployment timeline constraining you, that information would be helpful too :)

gracepotma:
OK so @ibacher the justification here is that this would only impact implementers specifically using Amharic.

gracepotma:
Update for everyone: Please see Ian's new PR here, which has a proposal for a better way of achieving the goal of this PR. We hope this is helpful 😄  https://github.com/openmrs/openmrs-esm-core/pull/790

gracepotma:
@aman-icap did you have feedback on the proposed approach in #790 ?

gracepotma:
Update from Dagim: We can disregard this PR for now; in the meantime the ICAP-Eth team is testing the work Ian did on the separate branch in https://github.com/openmrs/openmrs-esm-core/pull/790

PR #790, opened 2023-10-18
(feat) Adds support for non-Gregorian calendars, with special work-arounds for Amharic

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Basically, this has the same intention as #786, but with a somewhat cleaner implementation, utilising the standard JS library. It also adds support for registering default custom calendars on a per-locale basis, which is currently only used to use the Ethiopian calendar when the locale is Amharic.

See the added tests for some of what this PR allows.

## Screenshots

Date Formatting

## Other

There's a hack in here I don't love: if the locale is set to Amharic, we format the date in English, but using the Ethiopic calendar, but this was done to produce the same results as #786.

#786 also basically stops the display of time strings, but this is a configurable option for the call to `formatDate()`, and we should probably have a mechanism to handle the implementation-specific defaults rather than just relying on the workaround of having a separate formatter that doesn't display time in some circumstances.

### DIFF
--- a/packages/framework/esm-utils/src/omrs-dates.test.ts
+++ b/packages/framework/esm-utils/src/omrs-dates.test.ts
@@ -88,6 +88,46 @@ describe("Openmrs Dates", () => {
     );
   });
 
+  it("formats dates with respect to the active calendar", () => {
+    timezoneMock.register("UTC");
+    const testDate = new Date("2021-12-09T13:15:33");
+    window.i18next.language = "am";
+    expect(formatDate(testDate)).toEqual("30-Hedar-2014");
+    expect(formatDate(testDate, { day: false })).toEqual("Hedar 2014");
+    expect(formatDate(testDate, { year: false })).toEqual("30 Hedar");
+    expect(formatDate(testDate, { mode: "wide" })).toEqual("30 — Hedar — 2014");
+    expect(formatDate(testDate, { mode: "wide", year: false })).toEqual(
+      "30 — Hedar"
+    );
+
+    window.i18next.language = "am-ET";
+    expect(formatDate(testDate)).toEqual("30-Hedar-2014");
+    expect(formatDate(testDate, { day: false })).toEqual("Hedar 2014");
+    expect(formatDate(testDate, { year: false })).toEqual("30 Hedar");
+    expect(formatDate(testDate, { mode: "wide" })).toEqual("30 — Hedar — 2014");
+    expect(formatDate(testDate, { mode: "wide", year: false })).toEqual(
+      "30 — Hedar"
+    );
+
+    window.i18next.language = "en-u-ca-ethiopic";
+    expect(formatDate(testDate)).toEqual("30-Hedar-2014");
+    expect(formatDate(testDate, { day: false })).toEqual("Hedar 2014");
+    expect(formatDate(testDate, { year: false })).toEqual("30 Hedar");
+    expect(formatDate(testDate, { mode: "wide" })).toEqual("30 — Hedar — 2014");
+    expect(formatDate(testDate, { mode: "wide", year: false })).toEqual(
+      "30 — Hedar"
+    );
+
+    window.i18next.language = "th-TH";
+    expect(formatDate(testDate)).toEqual("09 ธ.ค. 2564");
+    expect(formatDate(testDate, { day: false })).toEqual("ธ.ค. 2564");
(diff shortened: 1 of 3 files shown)


ibacher:
cc: @amn-icap @gracepotma This is my proposal for a better way of achieving the goal of #786.

gracepotma:
Wow - bravo Ian for pushing this through, and so fast. CC @aman-icap

packages/framework/esm-utils/src/omrs-dates.test.ts
```diff
@@ -88,6 +88,46 @@ describe("Openmrs Dates", () => {
...
+    window.i18next.language = "th-TH";
+    expect(formatDate(testDate)).toEqual("09 ธ.ค. 2564");
+    expect(formatDate(testDate, { day: false })).toEqual("ธ.ค. 2564");
+    expect(formatDate(testDate, { year: false })).toEqual("09 ธ.ค.");
+    expect(formatDate(testDate, { mode: "wide" })).toEqual("09 — ธ.ค. — 2564");
+    expect(formatDate(testDate, { mode: "wide", year: false })).toEqual(
+      "09 — ธ.ค."
+    );
+  });
+
```
denniskigen, line 130: Might this be a useful thing to add, @ibacher? 

```ts
  it("handles leap years correctly in Ethiopian locales", () => {
    timezoneMock.register("UTC");
    const testDate = new Date("2021-12-09T13:15:33");
    window.i18next.language = "en";
    expect(formatDate(testDate)).toEqual("09-Dec-2021");

    const dateBeforeLeapYear = new Date("2023-09-11T09:00:00");
    window.i18next.language = "am";
    expect(formatDate(dateBeforeLeapYear)).toEqual("06-Pagumen-2015");

    const dateAfterLeapYear = new Date("2023-09-12T09:00:00");
    expect(formatDate(dateAfterLeapYear)).toEqual("01-Meskerem-2016");
  });
```

packages/framework/esm-utils/src/omrs-dates.ts
```diff
@@ -212,33 +287,77 @@ export function formatDate(date: Date, options?: Partial<FormatDateOptions>) {
...
+
     if (mode == "wide") {
-      localeString = localeString.replace(/ /g, " — "); // space-emdash-space
-      if (/ru.*/.test(locale)) {
-        // Remove the extra em-dash that gets added between the year and the suffix 'r.'
-        const len = localeString.length;
-        localeString =
-          localeString.slice(0, len - 5) +
-          localeString.slice(len - 5).replace(" — ", " ");
-      }
```
denniskigen, line 233: Do we need to add this logic for `ru` back in?

ibacher: No, because it's handled in the `formatParts()` call (specifically, the separator is no longer changed).

packages/framework/esm-utils/src/omrs-dates.ts
```diff
@@ -212,33 +287,77 @@ export function formatDate(date: Date, options?: Partial<FormatDateOptions>) {
...
       // This locale override is here rather than in `getLocale`
       // because Americans should see AM/PM for times.
       locale = "en-GB";
     }
-    localeString = date.toLocaleDateString(locale, formatterOptions);
-    if (locale == "en-GB" && mode == "standard" && year && day) {
-      // Custom formatting for English. Use hyphens instead of spaces.
-      localeString = localeString.replace(/ /g, "-");
+
+    if (_locale.language === "am") {
```
ebambo, line 296: @ibacher why do we need to keep the month names in English. Wouldn't it better to keep them in Amharic ? What do you think @aman-icap

ibacher: I did this because I was trying to keep this PR in line with the functionality in #786, which [also uses English names for months in Amharic](https://github.com/openmrs/openmrs-esm-core/pull/786/files#diff-1ffa29066d8c6a510e6de6209347b4cf8d6328b91fda7cfab96e643cc26e9f80R9). I mentioned that in the PR description.

denniskigen:
Any updates here?

ibacher:
In the absence of any updates, approvals, etc. I'm just going to merge this in as it should be harmless for existing use-cases. Future work needed here can be addressed as additional tickets.

gracepotma:
Update from Dagim: the ICAP-Eth team is testing the work Ian done on this branch and recommended we disregard the other/original #786 PR for now.

PR #793, opened 2023-10-20
(feat) O3-2504: Allow saving user preference for login location as user property

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR allows the user to save preference for login location either as a user property or in localStorage.

### Retrieving Default Location
The login location preference can be found with the key `defaultLocation`.

```
{
  "defaultLocation": <locationUuid>
}
```

### Workflows
1. If there is no `defaultLocation` property set, when the user lands on the login location page, if they check the checkbox for saving preference and save it, the selected location will be saved as the `defaultLocation`.
2. If the `defaultLocation` is present, then when they land on the login location screen, the `defaultLocation` will be verified and then the user will be redirected to the home page, with the session location being the same as `defaultLocation`.
3. If the user wants to update the preference, they can go to the user panel on the primary navigation bar and then update the location.
3. a. If the user keeps the checkbox checked and submits, the `defaultLocation` will be updated.
3. b. If the user unchecks the checkbox and submits, the property `defaultLocation` will be deleted for the user.

### Login
If the user property `defaultLocation` is set for a user, then on login, the `sessionLocation` will be set to the user property saved and the user will be directed from login to home page.

## Screenshots

https://github.com/openmrs/openmrs-esm-core/assets/51502471/6bc0dee4-9e1a-4033-8d71-fb0b83ef77e3

image

image

## Related Issue
https://issues.openmrs.org/browse/O3-2504


vasharma05:
>  In particular, using a userProperty assumes the user logs in before the login page, which is the current flow of the O3 RefApp, but we're seeing a number of implementations that want something more similar to the O2 page where locations are shown on the same page as logins, where using a userProperty wouldn't work.
I understand, I'll start a thread on the #openmrs3 channel.
Thanks!

ibacher:
> But for the 1st iteration, can we move ahead with this change, as this is a requirement.

Short version: The more we "move ahead" with things like this without discussion the more technical debt we accrue.

Some high-level things:

1. The name `defaultLoginLocation` is inappropriate. Semantically, we have properties already defined on the user, `defaultLocale` = user's locale when they login, if defined; `defaultLocation` = user's location when they login if defined. `preferredLoginLocation` is better.
2. Is the user the right level at which to be making this decision? Or is it better to have per-device information, e.g., something from either localStorage or a cookie or similar? (If, for example, computers are stored in physical locations, then it's very likely that clinicians using the computer want to be logged into the location the computer is located at rather than whatever their "default" clinic might be).
3. There are existing implementations of exactly this feature. For example, [module-authentication](https://github.com/openmrs/openmrs-module-authentication/blob/d7ea506bb28aaf9edb352a7e58080ba6d8542346/omod/src/main/java/org/openmrs/module/authentication/web/BasicWithLocationAuthenticationScheme.java#L88) supports this through the `sessionLocation` parameter; the O2 reference application [uses a global property to track the user attribute which stores a comma-separated list of user locations](https://github.com/openmrs/openmrs-module-referenceapplication/blob/a6dda613b882161b30df45176149e28f4ff3e665/omod/src/main/java/org/openmrs/module/referenceapplication/page/controller/LoginPageController.java#L378). So why are we inventing yet another mechanism for doing the same thing here?

vasharma05:
Hi @ibacher @mseaton !
Based on the suggestions on [O3-2504](https://issues.openmrs.org/browse/O3-2504) and the above suggestions from @ibacher , I have created configuration points to give the implementer the freedom to choose w/e suits them better.

Please read the updated description for better understanding.
Thanks!

packages/apps/esm-login-app/src/change-location-link/change-location-link.scss
```diff
@@ -19,15 +19,16 @@
...
   align-items: center;
 }
 
-.changeLocationLinkContainer button {
-  @include brand-02(background-color);
-  padding: 0rem 0.75rem;
+.changeLocationLinkContainer button,
+.changeLocationLinkContainer button:hover {
+  // @include brand-02(background-color);
+  // padding: 0rem 0.75rem;
```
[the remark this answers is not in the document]
vasharma05: Removed.

packages/apps/esm-login-app/src/config-schema.ts
```diff
@@ -48,6 +48,24 @@ export const configSchema = {
...
+        _type: Type.Boolean,
+        _default: true,
+        _description: "Whether to save preference on the system only.",
+      },
+      savePreferenceAsUserProp: {
+        _type: Type.Boolean,
+        _default: true,
+        _description: "Whether to save preference as a user property.",
+      },
+    },
```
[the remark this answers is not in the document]
vasharma05: The configuration was totally removed.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -230,7 +234,7 @@ const LocationPicker: React.FC<LocationPickerProps> = ({
                     <RadioButtonGroup
                       valueSelected={activeLocation}
                       orientation="vertical"
-                      name={activeLocation}
+                      name="Login locations"
```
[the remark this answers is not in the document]
vasharma05: It's not visible on the screen, it is there in the code only.

packages/apps/esm-login-app/src/change-location-link/change-location-link.component.tsx
```diff
@@ -27,7 +27,9 @@ const ChangeLocationLink: React.FC<ChangeLocationLinkProps> = ({
       <Location size={20} />
       <div>
         {currentLocation}
-        <Button onClick={changeLocation}>{t("change", "Change")}</Button>
+        <Button kind="ghost" onClick={changeLocation}>
```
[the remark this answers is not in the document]
vasharma05: Added to the PR description

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -20,59 +26,77 @@ import {
...
   hideWelcomeMessage,
   currentLocationUuid,
 }) => {
   const { t } = useTranslation();
-  const searchTimeout = 300;
-  const inputRef = useRef();
-  const { user } = useSession();
-  const currentUser = user?.display;
-  const config = useConfig();
+  const config: ConfigSchema = useConfig();
```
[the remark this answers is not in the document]
vasharma05: Done

packages/apps/esm-login-app/src/change-location-link/change-location-link.component.tsx
```diff
@@ -18,7 +18,7 @@ const ChangeLocationLink: React.FC<ChangeLocationLinkProps> = ({
 
   const changeLocation = () => {
     navigate({
-      to: `\${openmrsSpaBase}/login/location?returnToUrl=${referer}`,
+      to: `\${openmrsSpaBase}/login/location?returnToUrl=${referer}&update=true`,
```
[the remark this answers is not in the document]
vasharma05: Added

vasharma05:
Hi @ibacher !
Please re-review the changes, some reviews of yours were 5 days old, which I believe wasn't added in the review back then, but was worked.

I've pushed the review changes. Thanks for your time!

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -0,0 +1,119 @@
...
+  const { user } = useSession();
+  const { userUuid, userProperties } = useMemo(
+    () => ({
+      userUuid: user?.uuid,
+      userProperties: user?.userProperties,
+    }),
+    [user]
+  );
+
+  const defaultLocation = useMemo(() => {
```
ibacher, line 24: There's really no good reason to have `useMemo` here

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -0,0 +1,119 @@
...
+  }, [userProperties?.defaultLocation]);
+
+  const { isLocationValid, defaultLocation: defaultLocationFhir } =
+    useValidateLocationUuid(defaultLocation);
+
+  useEffect(() => {
+    if (defaultLocation) {
+      setSavePreference(true);
+    }
+  }, [setSavePreference, defaultLocation]);
```
ibacher, line 35: This seems like a weird concern to have in this hook... Isn't whether we're saving the preference fundamentally tied to whether the user checked the checkbox on the screen? I don't understand why the logic for this is part of this hook.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  const updateUserPropsWithPreference = useCallback(
+    async (locationUuid: string, saveUserPreference: boolean) => {
```
ibacher, line 38: I think it's bad to call this a "preference" because it's a "default" not a "preference" per se, i.e., if this is checked, next time they login it will go to this location.

```suggestion
    async (locationUuid: string, saveDefaultLocation: boolean) => {
```

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -0,0 +1,119 @@
...
+  useSession,
+} from "@openmrs/esm-framework";
+import { useCallback, useEffect, useMemo, useState } from "react";
+import { useTranslation } from "react-i18next";
+import { useValidateLocationUuid } from "../login.resource";
+import { useSearchParams } from "react-router-dom";
+
+export function useDefaultLocation() {
+  const { t } = useTranslation();
+  const [searchParams] = useSearchParams();
```
ibacher, line 13: I think having `useSearchParams()` here is a clear sign that this hook has too many responsibilities. The component can know about the search parameters, but I'm really unclear why we need this in the hook. The entirety of the `updateUserPreference()` function fundamentally is part of a component and shouldn't be done in a hook like this.

packages/apps/esm-login-app/src/login.resource.ts
```diff
@@ -47,6 +49,7 @@ export function useLoginLocations(
   searchQuery: string = ""
 ): LoginLocationData {
   const { t } = useTranslation();
+  const debouncedSearchQuery = useDebounce(searchQuery);
```
ibacher, line 52: This also strikes me as a concern boundary that shouldn't be crossed inside this hook. The debouncing should be done in the component that uses the hook.

packages/apps/esm-login-app/src/login.resource.ts
```diff
@@ -119,3 +125,24 @@ export function useLoginLocations(
...
 }
+
+export function useValidateLocationUuid(userPreferredLocationUuid: string) {
+  const url = userPreferredLocationUuid
+    ? `/ws/fhir2/R4/Location?_id=${userPreferredLocationUuid}`
+    : null;
+  const { data, error, isLoading } = useSwrImmutable<
+    FetchResponse<LocationResponse>
+  >(url, openmrsFetch, {
+    errorRetryCount: 0,
```
ibacher, line 136: I don't love using this setting; this should at least be responsive to the type of error returned. A 4XX, for example, means don't retry (but you'll never get that here), a 5XX response probably should be retried...

ibacher (changes requested):
Thanks @vasharma05! I think it's important to keep UI-concerns out of hooks, or at least out of hooks that are primarily about data-fetching. This leads to a cleaner and more re-usable API.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
ibacher: The slightly lower-level `updateUserPropsWithPreference()` is reasonable to put in the hook because it varies with the user preferences, and that way we can minimize reloading of them.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
vasharma05: The reason I added this hook was because the `useSession` after logging in, first returns `null` and then the session object, and by that time, the initial value of the state `savePreference` is set to false, which should've been true, hence using the useEffect.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
vasharma05: The `defaultLocation` is dependency in 2 useEffects and other useCallback functions.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
vasharma05: I understand, let me fix it.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
ibacher: Right, but that's only useful if we're dealing with an object ([see the React docs](https://react.dev/reference/react/useMemo#memoizing-a-dependency-of-another-hook)). In this case, `defaultLocation` will always be: `undefined`, `null`, or some string value. Specifically, `useMemo()` helps with things like this:

```ts
let a = { a: 'Value' };
let b = { a: 'Value' };

a === b; // always false
```

But with this:

```ts
let a = 'Value';
let b = 'Value';

a === b; // always true
```

It's just overhead.

packages/apps/esm-login-app/src/login.resource.ts
vasharma05: The value returned by `useDebounce` is the debounced value and is consumed by the useSWR hook, hence it is in this hook.

ibacher: Right, my point is that the `useDebounce` should be in the component, not the hook. I get why it's here.

ibacher (approved):

PR #795, opened 2023-10-24
(feat) Add a no-op SW when running with offline disabled

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Does what it says in the commit message.

### DIFF
--- a/packages/shell/esm-app-shell/src/service-worker/noop.ts
+++ b/packages/shell/esm-app-shell/src/service-worker/noop.ts
@@ -1,0 +1,27 @@
+self.__WB_MANIFEST = [];
+self.__WB_DISABLE_DEV_LOGS = true;
+
+self.addEventListener("message", () => {});
+
+self.addEventListener("install", () => {
+  // Skip over the "waiting" lifecycle state, to ensure that our
+  // new service worker is activated immediately, even if there's
+  // another tab open controlled by our older service worker code.
+  self.skipWaiting();
+});
+
+self.addEventListener("activate", () => {
+  // Optional: Get a list of all the current open windows/tabs under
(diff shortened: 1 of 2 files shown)


ibacher:
@denniskigen When you get a chance (no rush), could you try this out? It seems to work for me, but I'm worried that my browser isn't as scarred by service workers as yours is.

denniskigen:
Seems to work fine on Firefox, but not on Chrome. I'm getting some weird results on Chrome.

ibacher:
Ok, the underlying issue here appears to be that Chrome is re-installing the service-worker.js file, but seems to be using a cached copy, which is fixable, but hard to undo for broken browsers.

ibacher:
@denniskigen Ok... Looked at this again and it was embarrassingly, stupidly easy to fix. Basically, we're using the `InjectManifest` plugin from Workbox that allows a lot of control over how the manifest is processed. It does this by injecting the manifest (the fixed files that need to be cached) into the variable named `self.__WB_MANIFEST`, basically in the same manner as Webpack's Define Plugin does (i.e., replace a matching string with the contents. I had this:

```ts
self.__WB_MANIFEST = [];
```

Which resulted in something like:

```ts
[{/* some JS object */}, {/* some JS object */}, {/* some JS object */}, /* ... */] = [];
```

Which fails because JS reads that as a destructuring assignment. Since that lead to an exception being thrown when the SW was loaded, the new service worker was never actually installed.

I've fixed things so that an empty manifest is generated and assigned to a normal JS variable.

denniskigen:
LGTM! Thanks, @ibacher!

denniskigen (approved):

PR #797, opened 2023-10-26
(fix) Adds two React-hooks, shallowEqual implementation and some small fixes

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Adds two beta-quality React hooks aimed at better data-fetching hygiene. Specifically:

`useAbortController()` - to get an abort controller that resets every abort and aborts when the hook is unmounted; useful to ensure that long-running fetches are stopped if the component that made the request is no longer mounted.

`useOpenmrsSWR()` - very experimental, but an attempt to make an easy-to-use drop in for:

```ts
const { data, error } = useSWR('/some/url', openmrsFetch);
```

```ts
const { data, error } = useOpenmrsSWR('/some/url');
```

However, this hook will also ensure that the fetch request is made with an abort controller, either one passed in or one created inside the hook (using `useAbortController`). While this doesn't result in a shorter API, the idea is to make it easier to remember some basic house-keeping.

Also adds an implementation of the `shallowEqual()` algorithm that React uses to determine whether a component needs to be re-rendered, because it's sometimes a useful utility.

There are also some very minor refactorings, mostly renaming `.tsx` -> `.ts` where there are no actual JSX elements.

## Screenshots
*None*

## Related Issue
*None*

## Other
*None*


packages/framework/esm-react-utils/src/useDebounce.ts
```diff
@@ -8,7 +8,7 @@ import { useEffect, useState } from "react";
  *
  * @example
  * ```tsx
- * import { useDebounce } from "@openmrs/esm-react-utils";
+ * import { useDebounce } from "@openmrs/esm-framework";
```
denniskigen, line 11: Thanks for catching this!

denniskigen:
Pretty cool, thanks @ibacher. Do we need to update the mock?

denniskigen (approved):

ibacher:
Yeah, let me do that before merging it in.

denniskigen (approved):

PR #799, opened 2023-10-27
BREAKING: Removing i18next-icu

vasharma05:
## Requirements

#### For changes to apps

#### If applicable

## Summary
This PR remove the `i18next-icu` to be used in the i18next initialization because it is supported till i18next's `v19`. Hence the pluralization translations weren't working in the application.

Basically `i18next-icu` was helping us to use `{` in the variables instead of `{{`. The `{{` interpolation is what i18next's default configuration.

(@jayasanka-sack I finally got the answer to why we were using `{` instead of `{{`)

## Detailed Explaination
`i18next-icu` uses the [Intl Message Format](https://formatjs.io/docs/intl-messageformat/). If we look into the i18next-icu's description [here](https://github.com/i18next/i18next-icu#advice), it says:
```
When using this module, only the ICU message format is respected, this means the i18next format interpolation will not work. So for example instead of Hy {{name}}! it is Hy {name}!
```

For pluralization, i18next uses the [Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/PluralRules), which is a browser API. As said above, the i18next-icu overrides the i18next plural API with the Intl message formatter's rules and you can have a look at the Intl message formatter's plural support [here](https://formatjs.io/docs/intl-messageformat/), which is very different than the `Intl.PluralRules`. 

So what was happening is that the translations for keys `key_one` or `key_other` wasn't working for `t('key', {count: 2})`, because the it wasn't finding the exact key, and as mentioned above, the ICU message formatter is different than the Intl.PluralRules, hence the no proper working for i18next pluralization.

## Screenshots
None

## Related Issue
None


denniskigen:
Hey Vineet, could you clear up the PR description and the title some more?

ibacher:
@vasharma05 I've re-written the description. Please feel free to change it, though.

ibacher (approved):
LGTM

packages/apps/esm-offline-tools-app/src/offline-patients/empty-state.component.tsx
```diff
@@ -25,9 +26,9 @@ const EmptyState: React.FC<EmptyStateProps> = ({
         </div>
         <EmptyDataIllustration />
         <p className={styles.content}>
-          <Trans i18nKey="emptyStateText" values={{ displayText: displayText }}>
-            There are no {displayText} to display
-          </Trans>
+          {t("emptyStateText", "There are no {{displayText}} to display", {
```
ibacher, line 29: As a general rule, I'd probably actually favour the `<Trans />` component where its used like this.

vasharma05:
Sorry @ibacher, I think I overrided your description with mine, you submitted before me and I didn't see that. 
Please review my description and feel free to add.

@denniskigen please feel free to ask any doubts.
Thanks!

packages/apps/esm-offline-tools-app/src/offline-patients/empty-state.component.tsx
vasharma05: Actually, the `{{displayText}}` was getting changed to `{{ displayText }}` by prettier, hence I made the change

vasharma05:
I am creating PRs in another repos as well, we'll merge them altogether.
Thanks!

packages/apps/esm-offline-tools-app/src/offline-patients/empty-state.component.tsx
ibacher: That makes perfect sense!

PR #804, opened 2023-11-02
(fix) O3-2543 useSession not always receiving updated session object

brandones:
## Requirements

#### For changes to apps

#### If applicable

## Summary
The problem was that the `unsubscribe` variable was at module level, so only one instance of `useSession` was actually able to get its subscription set up properly.

## Related Issue
https://issues.openmrs.org/browse/O3-2543

## Other
Unblocks https://github.com/openmrs/openmrs-esm-patient-management/pull/775

### DIFF
--- a/packages/framework/esm-react-utils/src/useSession.test.tsx
+++ b/packages/framework/esm-react-utils/src/useSession.test.tsx
@@ -81,4 +81,25 @@ describe("useSession", () => {
     );
     expect(screen.getByText(/"authenticated":false/)).toBeInTheDocument();
   });
+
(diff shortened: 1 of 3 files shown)


ibacher (approved):
Excellent work here! Is it ok for the promise to still be held at module level?

brandones:
Thanks @ibacher ! Yes, the promise manages the network response, and is intended to be shared between instances. Though actually since the network request is made and managed by the module with the session store, each instance of useSession could just as well have its own promise. So either way is fine.

PR #806, opened 2023-11-03
(feat) Synchronously load most extensions and pages

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This ensures that all components are in the main chunk of the apps in esm-core. This reduces the number of JS requests by about 13, though I think there are more significant gains to be had by adopting this in the patient-chart and patient-management repos, especially.

At least for me, the primary navigation is, expected, a little snappier under this change.

## Other

Added bonus: this adds types for the Webpack config.

### DIFF
--- a/packages/tooling/webpack-config/src/index.ts
+++ b/packages/tooling/webpack-config/src/index.ts
@@ -40,13 +40,23 @@ import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
 import { resolve, dirname, basename } from "path";
 import { CleanWebpackPlugin } from "clean-webpack-plugin";
 import CopyWebpackPlugin from "copy-webpack-plugin";
-import { DefinePlugin, container } from "webpack";
+import {
+  DefinePlugin,
+  container,
+  type ModuleOptions,
+  type WebpackOptionsNormalized as WebpackConfiguration,
+  type RuleSetRule,
+} from "webpack";
 import { BundleAnalyzerPlugin } from "webpack-bundle-analyzer";
 import { StatsWriterPlugin } from "webpack-stats-plugin";
 // eslint-disable-next-line no-restricted-imports
 import { merge, mergeWith, isArray } from "lodash";
 import { existsSync, statSync } from "fs";
 import { inc } from "semver";
+
+type OpenmrsWebpackConfig = Omit<Partial<WebpackConfiguration>, "module"> & {
+  module: ModuleOptions;
+};
 
 const production = "production";
 const { ModuleFederationPlugin } = container;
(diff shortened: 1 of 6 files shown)


packages/tooling/webpack-config/src/index.ts
```diff
@@ -245,6 +266,17 @@ export default (
     performance: {
       hints: mode === production && "warning",
     },
+    optimization: merge(
+      {
+        // The defaults for both of these are 30; however, due to the modular nature of
+        // the frontend, we want each app to produce substantially
```
denniskigen, line 272: Looks like this comment isn't complete.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -148,7 +166,10 @@ export default (
     main,
     types,
   } = require(resolve(root, "package.json"));
-  const mode = argv.mode || process.env.NODE_ENV || "development";
+  // this typing is provably incorrect, but actually works
```
denniskigen, line 169:
```suggestion
  // this typing is probably incorrect, but actually works
```

packages/tooling/webpack-config/src/index.ts
ibacher: Right I got involved with writing an override for it and forgot to go back.

packages/tooling/webpack-config/src/index.ts
ibacher: I actually meant "provably" here. The value isn't guaranteed to be one of "none", "production", or "development", but I'd rather leave that validation to Webpack than figure it out here, so I'm lying to Typescript 😁.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -40,14 +40,24 @@ import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
...
+  DefinePlugin,
+  container,
+  type ModuleOptions,
+  type WebpackOptionsNormalized as WebpackConfiguration,
+  type RuleSetRule,
+} from "webpack";
 import { BundleAnalyzerPlugin } from "webpack-bundle-analyzer";
 import { StatsWriterPlugin } from "webpack-stats-plugin";
 // eslint-disable-next-line no-restricted-imports
 import { merge, mergeWith, isArray } from "lodash";
```
denniskigen, line 53: Worth fixing this?

```suggestion
import merge from "lodash-es/merge";
import mergeWith from "lodash-es/mergeWith";
import isArray from "lodash-es/isArray";
```

denniskigen (approved):
Great work, @ibacher! Thanks!

packages/tooling/webpack-config/src/index.ts
ibacher: In the webpack-config? Nah... We don't actually include it in a bundle, so the fact that `lodash` doesn't really tree-shake here is a non-issue. We're basically just using it as a NodeJS library.

PR #816, opened 2023-11-17
(feat) Support for automatically overriding routes and Carbonize devtools

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This is the first in a planned series of commits to build out better dev tooling for handling stuff related to routes.json files, used by Core V5. The goal of this PR is pretty straight-forward: when adding an app to the import-map-overrides, we should load the routes, if any along-side that.

This does several things:

1. The DevTools button upstream has been changed a little bit to make it more useful, so with active overrides it now looks like this:
    Screenshot 2023-11-17 at 12 48 10 PM
2. I Carbonized the import-map-overrides. This is somewhat unnecessary, but I'm planning on adding a routes-specific panel to the devtools eventually. This is basically functionally-equivalent to import-map-overrides.
    Screenshot 2023-11-17 at 12 48 03 PM
3. When adding a import map override, it also adds a routes override. Similar to import map overrides, these are only processed on page refreshes, so refreshing the page is necessary to see the change.
4. You can also manually override routes by providing an entry in local storage with the key `openmrs-routes:<name of module>`, e.g., `openmrs-routes:@openmrs/esm-patient-chart-app`. The value for this can be either a JSON string (so double-quotes) that points to a URL that resolves to a custom `routes.json` (this is how 3 above is implemented) or a stringified JSON object that conforms to the routes schema. There is currently no UI for this other than your browser devtools.
5. I added some sanity checks to the code that registers pages and extensions to ensure that we don't accidentally try to register an invalid page or extension object.

## Other

Future plans:

1. Add a tab to display information about the current active routes files and routes overrides; this will also have some capability to edit the routes overrides directly.
2. Add a tab to show all registered pages and extensions as well as which of those are active on the page.
3. Semi-related, it's possible (though no one uses this) to specify an inline import map on the CLI when running `develop` and some other commands. I thought that I wrote the code to also support that for routes, but it doesn't work, so I'm going to fix that up.


packages/framework/esm-routes/README.md
```diff
@@ -0,0 +1,3 @@
+# openmrs-esm-utils
+
+openmrs-esm-utils provides helper functions.
```
denniskigen, line 3: Shouldn't the content here be about routes?

ibacher: Yes, it should. Let me fix that!

packages/framework/esm-routes/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@openmrs/esm-routes",
+  "version": "5.3.0",
+  "license": "MPL-2.0",
+  "description": "Helper utilities for OpenMRS",
```
denniskigen, line 5: Pedantic, but I imagine this should also relate to routes.

denniskigen (commented):
Nice work, Ian! Code looks good on first reading. I haven't tested locally yet (dunno if I can get it to build with the issues I've been running into lately)

denniskigen:
@ibacher, it looks like when I add a module override, and then subsequently remove it, the `routesOverride` doesn't get removed automatically from localStorage:

https://github.com/openmrs/openmrs-esm-core/assets/8509731/c5eff929-7686-46bc-b755-5ca290f46b2f

ibacher:
@denniskigen I see... So, clicking on the "Reset all overrides" was, stupidly, not something I added something for (it should do the right thing if you individually disable the override).

ibacher:
Hopefully that's fixed now!

denniskigen (approved):
Looks good to me. I've left a few comments.

denniskigen:
One nice thing to support that the library supports is the ability to specify a port number and have the tooling fill out the override URL for you automatically:

https://github.com/openmrs/openmrs-esm-core/assets/8509731/09a10927-f708-45e0-9587-15ba5d50d709

ibacher:
> One nice thing to support that the library supports is the ability to specify a port number and have the tooling fill out the override URL for you automatically:

Yeah, I've found that to be kinda flaky personally, but it's pretty easy to implement. I'll add it in.

denniskigen:
> > One nice thing to support that the library supports is the ability to specify a port number and have the tooling fill out the override URL for you automatically:
> 
> Yeah, I've found that to be kinda flaky personally, but it's pretty easy to implement. I'll add it in.

In that case, probably best to leave it out.

denniskigen (approved):

PR #819, opened 2023-11-21
(fix) Remove background color from overflow menu buttons

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR fixes an issue with the background color of OverflowMenu buttons.

## Screenshots

> Before

CleanShot 2023-11-21 at 11  55 22@2x

> After

CleanShot 2023-11-21 at 11  59 31@2x

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -292,6 +292,11 @@
   width: spacing.$spacing-09;
 }
 
+.cds--overflow-menu {
+  // overflow menu buttons shouldn't have a background color
+  background: none;
+}
+
 /* Search inputs */
 .cds--search-input:focus {
   outline: 2px solid colors.$orange-40;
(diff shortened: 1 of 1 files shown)


packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -292,6 +292,11 @@
   width: spacing.$spacing-09;
 }
 
+.cds--overflow-menu {
+  // overflow menu buttons shouldn't have a background color
+  background: none;
```
ibacher, line 297:
```suggestion
  background-color: none;
```

ibacher (approved):

packages/framework/esm-styleguide/src/_overrides.scss
ibacher: I think the above would also remove the need for the comment.

denniskigen: `background-color: none` isn't a valid CSS property iirc.

ibacher: https://developer.mozilla.org/en-US/docs/Web/CSS/background-color?

ibacher: Good point: `background-color: unset;`

PR #822, opened 2023-11-22
(feat) O3-2546: Add the ability to use actions in the existing toast

hadijahkyampeire:
## Requirements

#### For changes to apps

#### If applicable

## Summary
- This PR modifies the existing toast to all actions.

## Screenshots
Recording

https://github.com/openmrs/openmrs-esm-core/assets/30952856/058946d4-ff85-4713-8789-17de9bf4765f

## Related Issue
- https://issues.openmrs.org/browse/O3-2546


hadijahkyampeire:
@ibacher @denniskigen this PR is ready for review, please take a look.

packages/apps/esm-implementer-tools-app/src/implementer-tools.component.tsx
```diff
@@ -35,8 +35,8 @@ function PopupHandler() {
   useEffect(() => {
     // only show notification max. 1 time
     if (shouldShowNotification) {
-      showActionableNotification({
-        critical: false,
+      showActionableToast({
+        isLowContrast: false,
```
ibacher, line 39: I tend to think `critical` is a better name for this property because it semantically says "I want this to appear as a critical notification" and then the component can interpret what that means (even if that just means directly passing this value to the `lowContrast` property).

Alternatively, though I think this shouldn't be preferred, the property should just be `lowContrast` since it corresponds directly to the `lowConstrast` of the Carbon component. In either case `isLowContrast` neither tells me when I should set this nor is easily correlatable to the Carbon API.

packages/framework/esm-react-utils/__mocks__/openmrs-esm-styleguide.mock.tsx
```diff
@@ -2,10 +2,12 @@ export const showNotification = jest.fn();
...
 export const showSnackbar = jest.fn();
 export const showToast = jest.fn();
+export const showActionableToast = jest.fn();
 
 beforeEach(() => {
   showNotification.mockReset();
   showActionableNotification.mockReset();
   showToast.mockReset();
   showSnackbar.mockReset();
+  showActionableToast.mockReset();
```
ibacher, line 12: Shouldn't this go with `showToast()` above?

packages/framework/esm-styleguide/src/actionable-toasts/actionable-toast.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+    onActionButtonClick = () => {},
+    progressActionLabel,
+    subtitle,
+    title,
+    ...props
+  } = actionableToast;
+
+  const [actionText, setActionText] = useState(actionButtonLabel);
+
+  const handleActionClick = () => {
```
ibacher, line 50: This should be wrapped in `useCallback()`, since it's passed as a property to the child component, to avoid unnecessary re-renders.

packages/framework/esm-styleguide/src/actionable-toasts/actionable-toast.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+    onActionButtonClick();
+    closeActionableToast();
+    progressActionLabel && setActionText(progressActionLabel);
```
ibacher, line 53: I'm not quite sure what this is accomplishing. Is it possible to include a video of the component in action?

packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -44,6 +44,7 @@
   <body>
     <%= htmlWebpackPlugin.tags.bodyTags %>
     <div class="omrs-toasts-container"></div>
+    <div class="omrs-actionable-toasts-container"></div>
```
ibacher, line 47: Have we looked into how things look when there are both toasts and actionable toasts?

ibacher (commented):
Thanks @hadijahkyampeire! A few comments and questions here.

packages/apps/esm-implementer-tools-app/src/implementer-tools.component.tsx
denniskigen: `critical` fits the mental model for consumers of notifications in O3 better, so I agree with keeping the prop as is.

packages/framework/esm-react-utils/__mocks__/openmrs-esm-styleguide.mock.tsx
hadijahkyampeire: @ibacher this is a different Toast component and the function is different from the showToast that's why I am calling it differently but I think eventually we will remove the previous toast after confirming that it is no longer being used.

packages/shell/esm-app-shell/src/index.ejs
hadijahkyampeire: Actually not yet but I am going to test and see.

hadijahkyampeire: Screenshot 2023-12-07 at 02 24 44
This shows they stuck on top of each other but since we are going to replace the toasts with the snackbar and then get rid of the Toast component, maybe we do not need to worry about that. cc @denniskigen @ibacher

hadijahkyampeire: @ibacher would you propose we just modify the existing toast to accept actions or let new actionable toast use the same container for toast.

hadijahkyampeire:
@ibacher @denniskigen I have updated this PR to add the implementation to allow the use of an action, so it won't break anything as far as the existing toast works and for the parts where we want to add an action it can also allow. Please see the attached recording and review the changes.

packages/framework/esm-styleguide/src/toasts/toast.component.tsx
```diff
@@ -26,26 +28,37 @@ export interface ToastNotificationMeta extends ToastDescriptor {
...
-    if (waitingForTime) {
+    if (millis) {
       const timeoutId = setTimeout(closeToast, millis);
       return () => clearTimeout(timeoutId);
     }
-  }, [waitingForTime]);
+  }, [closeToast, millis]);
 
   return (
-    <div onMouseEnter={() => setWaitingForTime(false)} onMouseLeave={() => setWaitingForTime(true)}>
```
ibacher, line 41: It seems like we should keep this "keep the notification available if the user is hovering over it" functionality right?

hadijahkyampeire:
@ibacher I have implemented the changes, please take another look.

denniskigen:
Is this good to go, @ibacher.

packages/shell/esm-app-shell/src/index.ejs
ibacher: Missed this since it got marked as resolved. I'd favour having all toasts in a single container.

ibacher (approved):

ibacher:
Yes, I think so.

hadijahkyampeire:
Thanks @ibacher

PR #826, opened 2023-11-30
(chore) Switch to using SwcMinifyWebpackPlugin

ibacher:
## Requirements

#### For changes to apps

#### If applicable

## Summary

Switches from using Terser (default) to SwcMinifyWebpackPlugin, which runs SWC's minifier.

The main goal here is to reduce memory consumption in builds.

### DIFF
--- a/packages/framework/esm-framework/package.json
+++ b/packages/framework/esm-framework/package.json
@@ -66,6 +66,7 @@
     "jest": "^29.7.0",
     "jest-cli": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
+    "swc-minify-webpack-plugin": "^2.1.1",
     "webpack": "^5.88.0"
   }
 }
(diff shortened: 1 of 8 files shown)


denniskigen:
Might it be worthwhile to try and use TerserWebpackPlugin's [swc](https://webpack.js.org/plugins/terser-webpack-plugin/#swc) mode first before switching from Terser altogether, @ibacher?

ibacher:
@denniskigen Well, I have no real desire to switch from using Terser per se. This was just to see if we could pare back the memory usage. Since that was the goal, it made sense to drop Terser altogether, because Terser spawns a [worker per available CPU](https://github.com/webpack-contrib/terser-webpack-plugin/blob/1831a49183faadf11f647064aea8e8e2c1e8a671/src/index.js#L325), which would lead to an increased number of Node processes, and hence, memory consumption. It doesn't seem like this switch was meaningfully different though.

PR #837, opened 2023-12-06
(chore) Extend ESLint configuration

denniskigen:
## Requirements

#### For changes to apps

#### If applicable

## Summary

This PR tweaks our ESLint configuration, adding a rule [consistent type imports](https://typescript-eslint.io/rules/consistent-type-imports/) and other useful TypeScript-related rules. 
*None*

## Related Issue
*None*

## Other
*None*


.eslintrc
```diff
@@ -1,28 +1,54 @@
...
   "ignorePatterns": ["**/*.test.tsx"],
   "rules": {
+    // Disabling these rules for now just to keep the diff small. I'll enable them in a future PR that fixes lint issues.
+    "@typescript-eslint/ban-ts-comment": "off",
+    "@typescript-eslint/ban-types": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/triple-slash-reference": "off",
+    // Use `import type` instead of `import` for type imports https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how
```
ibacher, line 17: Personally, I'm a little iffy on this rule. Specifically, I like the idea of _requiring type qualification for type imports_, but the style I prefer is not one of the "consistent" options, specifically, the style I prefer is:

```ts
import type { Type1, Type2, Type3, Type4 } from "./types";
import {
   type Patient,
   type Visit,
   saveVisit
} from "@openmrs/esm-framework";
```

Basically, the idea is use `import type ...` if everything is a type and `import {type ...}` otherwise. This is primarily for aesthetics, i.e.,

```ts
import type { Type1, Type2, Type3, Type4 } from "./types";
```

Is (in my view) nicer than:

```ts
import {
  type Type1,
  type Type2,
  type Type3,
  type Type4,
} from "./types";
```

But also because of the transformation mentioned [here](https://typescript-eslint.io/rules/no-import-type-side-effects/) (`import {} from "./types"` might have side-effects that `tsc` cannot reason about).

ibacher (changes requested):
Thanks @denniskigen! Looks great. I'm not really sure what to do about my longer concern, so feel free to ignore that one.

packages/tooling/openmrs/src/utils/importmap.ts
```diff
@@ -15,7 +15,9 @@ async function readImportmap(path: string, backend?: string, spaPath?: string) {
     if (backend && spaPath) {
       try {
         return await fetchRemoteImportmap(`${backend}${spaPath}importmap.json`);
-      } catch {}
+      } catch (e) {
+        console.error('Error reading importmap file: ', e);
+      }
```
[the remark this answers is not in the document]
ibacher: This does three things:

1. Makes the message source a little clearer by outputing `[openmrs] Cannot read importmap...`
2. Describes what's happening more accurately
3. Uses `console.warn()`, because by itself, a failure here does not cause the tool to fail (and this was why there was originally no message).

.eslintrc
denniskigen: Fair points, well argued. Unfortunately, it doesn't seem like the rule's configuration options gives us the desired flexibility. Glancing through the ESLint configs of popular projects, it looks like most use the [defaults](https://github.com/openmrs/openmrs-esm-core/pull/849/files). I think inlining the type import is better than separate type imports, e.g.

I find this:

```tsx
import { loadWebpackConfig, logInfo, logWarn, type ImportMapDeclaration } from '../utils';
```

Preferable to:

```tsx
import type { ImportmapDeclaration } from '../utils';
import { loadWebpackConfig, logInfo, logWarn } from '../utils';
```

Ultimately, I don't know what the preferred compromise is.

ibacher (approved):

PR #852, opened 2023-12-15
(feat) Add support for specifying overrides using only a port

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

Hopefully simple feature: restores the ability to add an importmap override by specifying only the port, rather than the full URL.

This makes slightly different assumptions from the upstream version:

1. Port numbers are checked a more strictly (ports range from 0-65535)
2. We assume that the file name is whatever the file name already is in the import map, not that it's the result of the transformation on line 17 (which requires, e.g., `@openmrs/esm-login-app` produces a main file called `openmrs-esm-login-app.js`). I made this change even though we follow this naming convention.
3. We use a full URL rather than `//`, as I didn't want to re-write the routes logic, though it may make sense to switch to using `//` eventually.


brandones (approved):
LGTM, thanks Ian!

PR #629, opened 2023-03-07
feat: O3-310: Add an application-writable config.json file in frontends/

jnsereko:
## Requirements
- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.

#### For changes to apps
- [x] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).

#### If applicable
- [x] My work includes tests or is validated by existing tests.
- [x] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary

Added a simple button to push the `config.json` file and show a success or error message to the user respectively 
@vasharma05 @hadijahkyampeire @denniskigen

## Screenshots

https://github.com/openmrs/openmrs-esm-core/assets/58003327/293abeeb-664f-4334-80d4-95b57c21e132

PR #865, opened 2023-12-28
feat: O3-310: Add an application-writable config.json file in frontends/ (attempt #2)

brandones:
Reverts openmrs/openmrs-esm-core#864

This restores @jnsereko 's https://github.com/openmrs/openmrs-esm-core/pull/629.

Sorry about the hasty work. Let's get this stuff in though.


ibacher:
This actually depends on openmrs/openmrs-module-webservices.rest#595, which is the active version of that PR.

packages/apps/esm-implementer-tools-app/src/backend-dependencies/openmrs-backend-dependencies.ts
```diff
@@ -155,3 +156,21 @@ export async function checkModules(): Promise<Array<ResolvedDependenciesModule>>
...
   return frontendModules.some((m) => m.dependencies.some((n) => n.type !== 'okay'));
 }
+
+export function useBackendDependencyCheck(moduleName: string) {
+  const [backendDependencies, setBackendDependencies] = useState<Array<BackendModule>>([]);
+
+  useMemo(async () => {
+    const dependencies = await initInstalledBackendModules();
+    setBackendDependencies(dependencies);
+  }, []);
```
samuelmale, line 166: Can we use a `useEffect` instead?

packages/apps/esm-implementer-tools-app/src/backend-dependencies/openmrs-backend-dependencies.ts
```diff
@@ -155,3 +156,21 @@ export async function checkModules(): Promise<Array<ResolvedDependenciesModule>>
 export function hasInvalidDependencies(frontendModules: Array<ResolvedDependenciesModule>) {
   return frontendModules.some((m) => m.dependencies.some((n) => n.type !== 'okay'));
 }
+
+export function useBackendDependencyCheck(moduleName: string) {
```
samuelmale, line 160: I would rename the hook to something like: `useBackendModuleExists(moduleName: string)`. What do you think of adding an optional parameter that validates the module versions:

```js
export function useBackendModuleExists(moduleName: string, moduleVersion?: string) {}

```

brandones: Agree with the rename, but don't complicate code for no reason. You can add that parameter when you need it.

PR #872, opened 2024-01-03
(fix) Amend ContentSwitcher style overrides

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR updates style overrides affecting the Carbon [ContentSwitcher](https://react.carbondesignsystem.com/?path=/docs/components-contentswitcher--overview) component to align with the available [design guidelines](https://zeroheight.com/23a080e38/p/55274c-content-switcher/b/17b091). These changes were prompted by fixes in https://github.com/openmrs/openmrs-esm-patient-chart/pull/1553/files#diff-817781f77acfc04bd6794ae62d233f09be5f7cab6cf7b19ee80d568933d877c3 which updated the icon-only ContentSwitcher variant used in vitals and biometrics cards.

Specifically, that change switched to using the [IconSwitch](https://react.carbondesignsystem.com/?path=/story/components-contentswitcher--icon-only) instead of Switch component, enabling us to set an aria label via the text prop for improved accessibility. Because we use a [customized version](https://zeroheight.com/23a080e38/p/55274c-content-switcher/b/17b091) of the ContentSwitcher whose styling deviates from the default Carbon component, changes to the underlying elements necessitate amendments to the style overrides we apply.

## Screenshots

> Before (broken styles for Icon-only variant of the ContentSwitcher)

https://github.com/openmrs/openmrs-esm-core/assets/8509731/09acd0cf-8e4f-4f3c-bdd1-2c4c1f821dc5

> After (fixed styles for both the default ContentSwitcher and the Icon-only variant)

https://github.com/openmrs/openmrs-esm-core/assets/8509731/f2d9b508-41c4-40f1-94c9-cb7272d52dbd

## Other

TODO: Check that styles are not broken for RTL languages.

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -56,6 +56,26 @@
 /* Content Switcher */
 .cds--content-switcher {
   @include layout.use('size', $default: 'sm');
+
+  & > :first-child {
+    border-bottom-left-radius: 0.25rem;
+    border-right: unset !important;
+    border-top-left-radius: 0.25rem;
+
+    [aria-selected='false'] {
+      border-left: 0.0625rem solid colors.$blue-30 !important;
+    }
+  }
+
+  & > :last-child {
+    border-bottom-right-radius: 0.25rem;
+    border-left: unset !important;
+    border-top-right-radius: 0.25rem;
+
+    [aria-selected='false'] {
+      border-right: 0.0625rem solid colors.$blue-30 !important;
+    }
+  }
 }
 
 .cds--content-switcher-btn {
@@ -68,16 +88,12 @@
     background-color: transparent !important;
   }
 
-  &:first-child {
-    border-bottom-left-radius: 0.25rem;
-    border-left: 0.0625rem solid colors.$blue-30 !important;
-    border-top-left-radius: 0.25rem;
-  }
+  &.cds--btn--primary {
+    background-color: unset;
 
-  &:last-child {
-    border-bottom-right-radius: 0.25rem;
-    border-right: 0.0625rem solid colors.$blue-30 !important;
-    border-top-right-radius: 0.25rem;
+    &:hover {
+      background-color: colors.$gray-10-hover;
+    }
   }
 
   &::before {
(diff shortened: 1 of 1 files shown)


packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -56,6 +56,26 @@
 /* Content Switcher */
 .cds--content-switcher {
   @include layout.use('size', $default: 'sm');
+
+  & > :first-child {
+    border-bottom-left-radius: 0.25rem;
+    border-right: unset !important;
+    border-top-left-radius: 0.25rem;
```
ibacher, line 63:
```suggestion
		border-top-left-radius: 0.25rem;
    border-bottom-left-radius: 0.25rem;
    border-right: unset !important;
```

ibacher (approved):
LGTM! 

This is probably slightly broken in RTL since (I think) the order of the tabs should be reversed, and consequently, the `border-left` and `border-right` rules probably need to be inverted.

denniskigen:
@ibacher, looks like getting things right for RTL locales isn't as straightforward as it seemed. Here's the best I've managed thus far:

https://github.com/openmrs/openmrs-esm-core/assets/8509731/2ebe5dfb-9df3-495e-83ef-fe6938b57619

And here's how things look for non-RTL locales:

https://github.com/openmrs/openmrs-esm-core/assets/8509731/d2eda6b2-aa21-4b9d-a23c-eea384763c96

ibacher:
@denniskigen I agree that we should deal with the RTL stuff in a separate PR, because, yeah, it gets complicated. One thing to keep in mind is that because CSS is a cascade, the rules applied to non-RTL stuff will also get applied to the elements in RTL, so I'd think the RTL codes would look more like overrides (which is one of the reasons that `!important` is very annoying). That's mostly the code we already have.

denniskigen:
Yeah, @ibacher, I suspect I'll have to gut everything and start over again. I don't like the way things are looking in that overrides file, but it needs a fresh pair of eyes. Should we proceed to merge?

denniskigen:
Thanks, @ibacher

PR #873, opened 2024-01-04
(fix) O3-2694: fixed form display when RTL language is selected

elimm:
# Requirements

## For changes to apps

## If applicable

## Summary
Fixed form display when RTL language is selected

## Related Issue
[O3-2694](https://openmrs.atlassian.net/browse/O3-2694)

### DIFF
--- a/packages/framework/esm-styleguide/src/_overrides.scss
+++ b/packages/framework/esm-styleguide/src/_overrides.scss
@@ -306,6 +306,10 @@ html[dir='rtl'] {
     }
   }
 
+  .cds--header__name {
+    padding-inline: 1rem 2rem;
+  }
+
   .cds--header-panel--expanded {
     right: calc(100vw - var(--omrs-sidenav-width));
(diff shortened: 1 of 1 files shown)


packages/framework/esm-styleguide/src/_overrides.scss
```diff
@@ -483,14 +487,33 @@ html[dir='rtl'] {
         }
       }
     }
-    .tab-content {
-      margin-left: unset;
```
sherrif10, line 487: You would add this fix  in this https://github.com/openmrs/openmrs-esm-patient-chart/pull/1561.

elimm: I don't think so, to rewrite from there I would need to disable the Angular scope `encapsulation: ViewEncapsulation.None` which is error prone and all the RTL related styles are already posted here

ibacher (approved):

PR #874, opened 2024-01-04
(fix) Make navbar not re-render like crazy (fixes O3-2692)

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

You know how the right side of the navbar has been blinky for a long time and we all kind of assumed it was a tricky extension thing?

https://github.com/openmrs/openmrs-esm-core/assets/1031876/cfef9b9b-16e8-4998-a183-efa6c7a2938a

It was actually just a classic React "don't define a component within another component" issue.

This gets rid of all the visual glitching, unmounting/remounting, etc. shown above.

It also fixes https://openmrs.atlassian.net/browse/O3-2692 "Two errors on logout: person is undefined and Language ID should be string or object," which was caused by an extension within the user panel not getting unmounted when it should, which somehow was being caused by the erratic re-rendering behavior of the navbar, in which its slot was contained.

NB: Yes, this does indicate that there is a subtle bug in the extension system that can cause extensions to fail to be unmounted if there is a chaotic amount of churn. I haven't solved that. But hopefully we just don't have churn.

## Screenshots

https://github.com/openmrs/openmrs-esm-core/assets/1031876/76df5822-e810-4a3c-8889-061d8bf5d20c

## Related Issue
https://openmrs.atlassian.net/browse/O3-2692

### DIFF
--- a/packages/apps/esm-login-app/src/logout/logout.scss
+++ b/packages/apps/esm-login-app/src/logout/logout.scss
@@ -1,8 +1,7 @@
 @import '../root.scss';
 
-/* Repeat selector for specificity */
-button.logout.logout {
-  color: $field-01;
+button.logout {
+  color: $field-01 !important;
   font-size: 1rem;
   width: 16rem;
 
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.scss
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.scss
@@ -1,6 +1,10 @@
 @import '~@openmrs/esm-styleguide/src/vars';
 @import '../../root.scss';
 
+.headerPanel {
+  transition: none;
+}
+
 .userPanelSwitcher {
   padding-top: 1rem;
   align-items: start;
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
@@ -1,9 +1,8 @@
-import React, { memo, useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { Navigate } from 'react-router-dom';
 import classNames from 'classnames';
 import { HeaderContainer, Header, HeaderMenuButton, HeaderGlobalBar, HeaderGlobalAction } from '@carbon/react';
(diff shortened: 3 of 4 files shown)


packages/apps/esm-login-app/src/logout/logout.scss
```diff
@@ -1,8 +1,7 @@
 @import '../root.scss';
 
-/* Repeat selector for specificity */
-button.logout.logout {
-  color: $field-01;
+button.logout {
+  color: $field-01 !important;
```
brandones, line 4: The specificity hack wasn't working reliably.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.scss
```diff
@@ -1,6 +1,10 @@
 @import '~@openmrs/esm-styleguide/src/vars';
 @import '../../root.scss';
 
+.headerPanel {
+  transition: none;
```
brandones, line 5: Turns out that when it's working as intended, the HeaderPanel component slides in from the right. That looks kind of weird, since we're using it as a drop-down menu.

packages/apps/esm-primary-navigation-app/src/components/user-panel-switcher-item/user-panel-switcher.component.tsx
```diff
@@ -5,11 +5,10 @@ import { useSession } from '@openmrs/esm-framework';
 
 const UserPanelSwitcher: React.FC = () => {
   const session = useSession();
-  const user = session?.user;
   return (
     <SwitcherItem aria-label="User">
       <UserAvatarFilledAlt size={20} />
-      <p>{user.person.display}</p>
+      <p>{session?.user?.person?.display}</p>
```
brandones, line 11: This is not the fix.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -123,10 +119,15 @@ const Navbar: React.FC = () => {
...
     </>
   );
+};
+
+const Navbar: React.FC = () => {
+  const session = useSession();
+  const openmrsSpaBase = window['getOpenmrsSpaBase']();
 
-  if (user && session) {
+  if (session?.user?.person) {
```
brandones, line 128: This is not the fix.

denniskigen (approved):

PR #877, opened 2024-01-09
(feat) Create a global app history utility

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

I got into this in order to solve https://openmrs.atlassian.net/browse/O3-1594 . The trouble is that there is at present no good way for the close button to know where to go back to. `document.referrer` is often not what we want, `window.location.back()` will go back to the previous dashboard within the patient chart, and hacks involving localStorage are unpleasant.

This creates a global app history that is preserved across the session. This way, the close button can look backwards through the history in order to determine where it should send the user back to. This will also allow us to make the breadcrumb actually correctly reflect where the user came from—right now the top level breadcrumb in the chart is always "home."

I created a new library to house this, called `esm-navigation`, which I think can serve as a more natural home for some things. I moved the navigation utilities out of `esm-config` (kind of a strange place for them) and into this new library, along with the whole of `esm-breadcrumbs`. `@openmrs/esm-breadcrumbs` is now deprecated. We should delete it in the next core major version.

I tried really hard to deprecate direct imports of `navigate`, `interpolateString`, and `interpolateUrl` from esm-config. Unfortunately it made things with tests and mocks and types *very* complex, and I wasn't able to get the whole thing working. So I've simply removed these from `esm-config`. This is a breaking change for anyone who was doing `import { navigate } from '@openmrs/esm-config'` (or likewise with the other functions). I doubt anyone was doing that though, which is why I have not attached the BREAKING label.

Only vaguely related, `navigate` is now a little bit smarter and will do a SPA navigation if possible even if the caller has included the origin in the URL. Previously, `navigate({ to: 'https://dev3.openmrs.org/openmrs/spa/home' })` would do a `location.assign`.

## Screenshots

Will include screenshots in the patient chart PR.

## Related Issue

Related to https://openmrs.atlassian.net/browse/O3-1594


.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "env": {
-    "node": true
+    "node": true,
+    "browser": true
```
brandones, line 4: I was getting eslint errors in `esm-navigation` telling me there's no such thing as `window`. This fixed it, of course, but I have no idea why we don't get those errors in all the other modules that use `window`, and I imagine there must be another way to fix it.

packages/framework/esm-navigation/src/types.ts
```diff
@@ -0,0 +1,35 @@
+/** @module @category Breadcrumb */
```
brandones, line 1: Copy-pasted from `esm-breadcrumbs`

packages/framework/esm-navigation/src/history/history.ts
```diff
@@ -0,0 +1,65 @@
...
+  history.push(newLocation);
+  const maxSize = 50;
+  if (history.length > maxSize) {
+    history = history.slice(-maxSize);
+  }
+  sessionStorage.setItem(historyKey, JSON.stringify(history));
+}
+
+// Initialize history  from sessionStorage. If history  is empty, add document.referrer
+// if available.
```
ibacher, line 17: Should this be a JSDoc comment? (I realize it's not exported, so not part of the API docs)

.eslintrc
ibacher: There might be, but there's no reason not to tell eslint that this code is intended to be run in the browser.

ibacher (commented):
Thanks @brandones! A couple of really minor points, but this looks great!

packages/framework/esm-navigation/src/history/history.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+initializeHistory();
+
+window.addEventListener('single-spa:routing-event', (evt) => {
+  const history = getHistory();
+  if (history[history.length - 1] !== window.location.href) {
+    addToHistory(window.location.href);
+  }
+});
```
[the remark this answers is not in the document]
brandones: Agreed, that's a nice idea.

packages/framework/esm-breadcrumbs/src/index.ts
```diff
@@ -1,3 +1,17 @@
-export * from './db';
-export * from './filter';
 export * from './types';
+
+import { registerBreadcrumb as rb, registerBreadcrumbs as rbs, getBreadcrumbs as gb } from './db';
```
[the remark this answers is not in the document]
brandones: I like that philosophy. Will clean this up.

ibacher (approved):

PR #884, opened 2024-01-11
(fix) Make ConfigurableLink call onBeforeNavigate for alternative clicks

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

It was the case that `onBeforeNavigate` wouldn't be called when the user ctrl-clicks something. Since ctrl-click still navigates the user to the target page, albeit in a different tab, it seems like the app should probably behave in the same way as if the user had clicked the link normally. I'm not aware of any bugs that this fixes, but it seems like it should produce more consistent behavior.

### DIFF
--- a/packages/framework/esm-react-utils/src/ConfigurableLink.test.tsx
+++ b/packages/framework/esm-react-utils/src/ConfigurableLink.test.tsx
@@ -56,7 +56,7 @@ describe(`ConfigurableLink`, () => {
     expect(navigate).toHaveBeenCalledWith({ to: path });
   });
 
-  it('executes onBeforeNavigate callback if provided', async () => {
+  it('executes onBeforeNavigate callback on any click that would open the page (including ctrl-click etc)', async () => {
     const onBeforeNavigate = jest.fn();
     render(
       <ConfigurableLink to={path} onBeforeNavigate={onBeforeNavigate}>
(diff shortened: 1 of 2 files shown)


denniskigen (approved):
LGTM! Thanks, @brandones. One of the places in the app where alternative clicks are used is the Admin tools app. Here's a video of it in action (for posterity):

https://github.com/openmrs/openmrs-esm-core/assets/8509731/f274f112-8d31-4d23-8dec-d2c5598bc5de

I'm torn about whether the default behaviour should be to CTRL-click to a separate tab. On the one hand, the things getting linked to here are arguably non-core EMR features, so you can kind of see the rationale for that. On the other hand, that the navigation pattern isn't consistent feels a bit incongruous to me.

brandones:
@denniskigen alternative clicks? Isn't it just that in the admin app, normal clicks open new tabs? That would be unrelated to this change.

Where does the admin page code live? I couldn't find it. What I'm wondering is how those links work, are they passing `target="_blank"` to a ConfigurableLink, or what?

denniskigen:
Yes, they are. Which is why they open in a new tab. This is in the [admin tools](https://github.com/openmrs/openmrs-esm-admin-tools) app. And yeah, it's unrelated to what you've worked on here, I just wanted to get your opinion on whether you think its appropriate for those links to open in a new tab.

brandones:
Nice, thanks for the link! Hmmm actually it looks like they are using a `ClickableTile` and passing `target="_blank"` to that. Which makes sense; I don't think `ConfigurableLink` would do that right thing for SPA links if you did that.

I do think this is a reasonable thing to do. It might also be good to at some point ensure that `ConfigurableLink` also can be used in this way (to always open a new tab or a new window).

denniskigen:
Cool, thanks for the sanity check!

PR #888, opened 2024-01-12
(chore) Only trigger e2e tests when relevant code has changed

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This adds a step to the e2e test workflow that will skip running the actual tests unless some relevant file has been modified (basically, relevant files are any source code that is not simply in a Jest test). I think I've got everything and got the picomatch syntax right, but we should verify that. Note that since this PR updates the e2e workflow file itself, it _should_ trigger a full e2e test run.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,7 +14,7 @@ jobs:
     permissions:
       pull-requests: read
     outputs:
-      changes: ${{ steps.filter.outputs.changes.test-sources }}
+      test-changes: ${{ steps.changes.outputs.test-sources }}
 
     steps: 
       - name: Checkout repo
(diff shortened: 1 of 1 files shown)


ibacher:
Well, this has pretty conclusively proven that this can stop the e2e tests from being triggered. Let's see if it can trigger them...

denniskigen:
Did you get it working reliably, @ibacher?

ibacher:
Yes. With the first couple of commits the e2e tests didn't run at all (hence taking seconds), but now the e2e tests were successfully triggered because the PR updates the workflow file.

denniskigen (approved):
LGTM! Thanks, @ibacher.

kdaud:
> Well, this has pretty conclusively proven that this can stop the e2e tests from being triggered. Let's see if it can trigger them...

Positive feedback → Yes, the logic works pretty well. Thanks, @ibacher !

@ibacher @denniskigen @jayasanka-sack additionally I suggest that when there is a change in the library package (ie upgrade/downgrade) the e2e tests should be as well triggered. Why? Ideally a change in a library version may introduce a breaking change in the E2E tests so having them triggered can help to flag out the issue at an earlier stage.

So if the idea is embraced, then we can push [this](https://github.com/openmrs/openmrs-esm-core/pull/951) change into main. Happy merging....

PR #889, opened 2024-01-15
(fix) Simplify history mock, add more mocks

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

I regret overcomplicating the history mock. If people need those things to work in more meaningful ways they can provide implementations.

Just musing: I think it is helpful to have the much smarter mock for config because conformant config objects are needed in order for most components to render at all. Whereas few (if any) components actually rely themselves on anything from navigation or history for their internal (i.e. unit-testable) logic. This difference is what I have learned from this exercise.

### DIFF
--- a/packages/framework/esm-framework/mock.tsx
+++ b/packages/framework/esm-framework/mock.tsx
@@ -3,7 +3,6 @@ import type {} from '@openmrs/esm-globals';
 import { createStore, type StoreApi } from 'zustand';
 import { NEVER, of } from 'rxjs';
 import { type SessionStore } from '@openmrs/esm-api';
-import { getDefaultsFromConfigSchema } from '@openmrs/esm-utils';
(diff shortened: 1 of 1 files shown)


packages/framework/esm-framework/mock.tsx
```diff
@@ -4,7 +4,7 @@ import { createStore, type StoreApi } from 'zustand';
 import { NEVER, of } from 'rxjs';
 import { type SessionStore } from '@openmrs/esm-api';
 import { getDefaultsFromConfigSchema } from '@openmrs/esm-utils';
```
denniskigen, line 6: Should this be derived from the `utils` import below?

denniskigen (approved):
Just one concern about imports, but looks good to me. Thanks, @brandones!

PR #898, opened 2024-01-22
(fix) O3-2287: Config system reports unknown config keys for modules that haven't loaded yet

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

This makes a clean demarcation between computing configs based on implicit schemas (provided at module registration time in the app shell), versus based on the actually defined schema provided by a loaded module. We need to be able to do the first prior to the second because translation loading will block module loading. Having these two phases distinguished, we can avoid running validations for schemas we don't actually have access to yet.

## Related Issue

https://openmrs.atlassian.net/browse/O3-2287

### DIFF
--- a/packages/framework/esm-config/src/module-config/module-config.ts
+++ b/packages/framework/esm-config/src/module-config/module-config.ts
@@ -3,6 +3,7 @@ import { clone, reduce, mergeDeepRight, equals, omit } from 'ramda';
 import type { Config, ConfigObject, ConfigSchema, ExtensionSlotConfig, ExtensionSlotConfigObject } from '../types';
 import { Type } from '../types';
 import { isArray, isBoolean, isUuid, isNumber, isObject, isString } from '../validators/type-validators';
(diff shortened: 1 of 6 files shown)


packages/framework/esm-config/src/module-config/module-config.ts
```diff
@@ -214,13 +254,13 @@ export function getConfig<T = Record<string, any>>(moduleName: string): Promise<
 }
 
 /** @internal */
-export function getConfigInternal(moduleName: string): Promise<Config> {
```
ibacher, line 217: I still like the idea of having the `getConfigInternal()` function. It's sometimes useful when debugging.

ibacher (approved):

packages/framework/esm-config/src/module-config/module-config.ts
brandones: Beyond what you can do with `window.stores`?

PR #901, opened 2024-01-24
(fix) Fix a regression in package version numbers

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This fixes a small regression introduced by #895. Previously, the directory the file was extracted into depended on the name of the tarball containing the sources, which meant that in the most common case, it was formatted as `${tarManifest.name}-${tarManifest.version}`, basically using the manifest metadata from the NPM-compatible registry. In #895, I changed the naming structure to uniformly be `${esmName}-${esmVersion}`. This results in the string that's drawn from `spa-assemble-config.json`. For the normal case, this basically results in directories like `openmrs-esm-login-app-next` instead of `openmrs-esm-login-app-5.2.1-pre.1234`. While this change is fairly inconsequential, in this case, it's very bad if, instead of specifying a version, someone were to use a URL (e.g., something like: `openmrs-esm-login-app-https:--registry.npmjs.org--openmrs-esm-login-app---esm-login-app-5.3.3-pre.1431.tgz`).

This patches things so that the version number suffix is added to the directory name _at the point where we are extracting the tar archive_, which means that we should always have a version number, because we always have a package.json.

### DIFF
--- a/packages/tooling/openmrs/src/commands/assemble.ts
+++ b/packages/tooling/openmrs/src/commands/assemble.ts
@@ -189,7 +189,6 @@ async function downloadPackage(
 }
 
 async function extractFiles(buffer: Buffer, targetDir: string) {
-  await mkdir(targetDir, { recursive: true });
   const packageRoot = 'package';
   const rs = Readable.from(buffer);
   const files = await untar(rs);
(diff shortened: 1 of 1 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #906, opened 2024-01-28
(fix) O3-2768: Display error in case there are no locations

trevor-james-nangosha:
# Requirements

## For changes to apps

## If applicable

## Summary
This issue makes sure that an error is shown when no locations are returned from the backend on the location-selection screen

## Screenshots

This is a small video that shows what happens in the aforementioned case. I would like to know if I am on the right track, or if I could possibly make some changes. @denniskigen 

https://github.com/openmrs/openmrs-esm-core/assets/75243400/798d385e-045d-4b5f-86da-ee59262fbd37

## Related Issue

[Issue is here](https://openmrs.atlassian.net/browse/O3-2768)


trevor-james-nangosha:
@ibacher I have applied the changes as requested and also updated the video above to show the current state of things.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -108,6 +113,11 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
         changeLocation(locations[0]?.resource.id, false);
       }
       if (!locations?.length) {
+        showNotification({
```
ibacher, line 116: I think it's appropriate to use a Toast in this instance. This is a non-recoverable error that the user can't do anything about.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -77,22 +76,28 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
-      setIsSubmitting(true);
+      if (locationUuid) {
+        setIsSubmitting(true);
+      }
 
       const referrer = state?.referrer;
       const returnToUrl = new URLSearchParams(location?.search).get('returnToUrl');
 
-      const sessionDefined = locationUuid ? setSessionLocation(locationUuid, new AbortController()) : Promise.resolve();
+      const sessionDefined = locationUuid ? setSessionLocation(locationUuid, new AbortController()) : Promise.resolve({"redirectToHome": false});
```
ibacher, line 86: Instead of all this, it seems like all we need here is:

```suggestion
      const sessionDefined = locationUuid ? setSessionLocation(locationUuid, new AbortController()) : Promise.reject();
```

And then:

```
sessionDefined.then(/* old stuff */).catch(/* not an error, rejected because no location is set */);

File: packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx [hunk 3]
```diff
@@ -164,96 +174,99 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
-            <p className={styles.welcomeTitle}>
-              {t('welcome', 'Welcome')} {currentUser}
-            </p>
-            <p className={styles.welcomeMessage}>
-              {t(
-                'selectYourLocation',
-                'Select your location from the list below. Use the search bar to find your location.',
-              )}
-            </p>
+  <div className={styles.locationPickerContainer}>
```
ibacher, line 177: Why is this code here reformatted? The formatting here looks wrong...

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -164,96 +174,99 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+          {t('welcome', 'Welcome')} {currentUser}
+        </p>
+        <p className={styles.welcomeMessage}>
+          {t(
+            'selectYourLocation',
+            'Select your location from the list below. Use the search bar to find your location.',
+          )}
+        </p>
+      </div>
+      <Search
```
ibacher, line 191: There's not much point in rendering a search box that doesn't do anything. Why is this moved up here?

ibacher (dismissed):

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -10,7 +10,15 @@ import {
   RadioButtonGroup,
   RadioButtonSkeleton,
 } from '@carbon/react';
-import { navigate, setSessionLocation, useConfig, useConnectivity, useSession } from '@openmrs/esm-framework';
+import {
+  navigate,
+  setSessionLocation,
+  showNotification,
```
ibacher, line 16: I think this is an unused import?

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -108,6 +117,11 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
         changeLocation(locations[0]?.resource.id, false);
       }
       if (!locations?.length) {
+        showToast({
+          title: t('Error', 'Error'),
```
ibacher, line 121:
```suggestion
          title: t('error', 'Error'),
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -226,11 +242,12 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
                     </div>
                   )}
                 </div>
-                {hasMore && (
-                  <div className={styles.loadingIcon}>
-                    <InlineLoading description={t('loading', 'Loading')} />
-                  </div>
-                )}
+                {!locations?.length ||
```
ibacher, line 245: I think this added check is redundant since `hasMore` will only be truthy if there are locations.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -77,12 +84,14 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
 
   const changeLocation = useCallback(
     (locationUuid?: string, saveUserPreference?: boolean) => {
-      setIsSubmitting(true);
+      if (locationUuid) {
+        setIsSubmitting(true);
+      }
```
ibacher, line 89: The more I look at this, the more I wonder why we don't just bail (`return`) if `locationUuid` is `null` or `undefined`?

Basically:

```ts
if (!locationUuid) {
  return;
}

// proceed as things were before this PR.
```

trevor-james-nangosha: I had wished to make minimal changes to the code as I found it, but the more I think about it, the more it makes sense.
Let me make these changes.

trevor-james-nangosha:
Done with the changes as requested @ibacher

denniskigen:
Sorry I'm getting to this late @trevor-james-nangosha @ibacher, but this is a mockup that Ciaran's thrown together for what this error notification could look like:

Screenshot 2024-02-15 at 11 50 18

You could imagine this appearing in place of the search bar in case no locations are configured. Should we just show a subtitle instead of the reload button with something along the lines of `Please contact your administrator` @ibacher?

ibacher:
> Should we just show a subtitle instead of the reload button with something along the lines of Please contact your administrator @ibacher?

Maybe have a subtitle that says "If the issue persists please contact your administrator" and leave the refresh button?

denniskigen:
Would clicking the `Reload` button just invoke `window.location.reload` or would it do something else?

ibacher:
> Would clicking the Reload button just invoke window.location.reload() or would it do something else?

Presumably we could just call the mutator for SWR rather than doing a full page reload? This isn't a "component is broken" error, just a "data is missing error".

@trevor-james-nangosha To answer you're question from Slack, the image is just a sketch to show the inline notification. Leave the rest of the elements on the page.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -108,6 +116,11 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
         changeLocation(locations[0]?.resource.id, false);
       }
       if (!locations?.length) {
+        showToast({
+          title: t('error', 'Error'),
+          kind: 'error',
+          description: 'No locations were found for this system. Please contact your administrator',
```
ibacher, line 122: The description here should also be run through the translation system.
```suggestion
          description: 'No locations were found for this system. Please contact your administrator.',
```

jayasanka-sack:
Isn't the fallback for no locations is to Login without a location? As I can remember that's how the location picker was written.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -108,6 +116,11 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+        });
         changeLocation();
```
jayasanka-sack, line 124: We don't need to call this function, right?

ibacher:
> Isn't the fallback for no locations is to Login without a location?

I think that was what was originally implemented just to get something in. Most of the code in various places assumes that there's a session location and, while the backend technically supports logins without locations, the RefApp hasn't since O2.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
samuelmale: @ibacher Just a minor comment: I would suggest adjusting the message to something like: 'No login locations were found for this user or system. Please contact your administrator.' This is because the suggested message seems to imply that there are no locations found in the system, yet it could be either of the following scenarios:

1. No login locations are configured in the system.
3. The current user is not assigned any of the available login locations (relevant in cases where tools like the data-filter module are used. This used to work well with the 2.x login not sure if the filters apply to the way we load locations in O3)

ibacher: @samuelmale That's technically true, but is a user going to understand the difference between "location" and "login location"? The text on the screen always just describes them as "locations".

And, yes, datafilter still works with O3.

mseaton:
> Looking at this from a slightly different angle: we can either stop the user up-front with no locations or we can write tests for every place that accesses the session location to ensure they continue to function if there is no location available

I do think this may be worth a follow-up discussion to confirm that we definitely want to make session location required in esm-core and the location picker, rather than just configured to be required within the refapp implementation of O3.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
trevor-james-nangosha: Yea, did not see that. Will remove it!! Thanks.

trevor-james-nangosha:
Screenshot showing behavior.
Screenshot (22)

suubi-joshua:
> Screenshot showing behavior. Screenshot (22)

Great job @trevor-james-nangosha but I think I think if the locations don't load the confirm button should not show neither should the checker for remember my location.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -16,6 +16,7 @@ import { useLoginLocations } from '../login.resource';
 import styles from './location-picker.scss';
 import { useDefaultLocation } from './location-picker.resource';
 import type { ConfigSchema } from '../config-schema';
+import { Snackbar } from '@openmrs/esm-styleguide/src/snackbars/snackbar.component';
```
vasharma05, line 19:
```
import { showSnackbar } from "@openmrs/esm-framework"
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -77,12 +78,14 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
-      setIsSubmitting(true);
+      if (!locationUuid) {
+        return;
+      }
 
       const referrer = state?.referrer;
       const returnToUrl = new URLSearchParams(location?.search).get('returnToUrl');
 
-      const sessionDefined = locationUuid ? setSessionLocation(locationUuid, new AbortController()) : Promise.resolve();
+      const sessionDefined = locationUuid ? setSessionLocation(locationUuid, new AbortController()) : Promise.reject();
```
vasharma05, line 88: Since we have the check above for undefined/null locationUuid, why add this unnecessary promise here?

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -178,15 +178,32 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+                onActionButtonClick: () => mutate(),
+                kind: 'error',
+                title: t('locationsFailedToLoad', 'Locations failed to load'),
+                subtitle: t(
+                  'locationIssueContactAdministrator',
+                  'If the issue persists please contact your administrator',
+                ),
+              }}
+              closeSnackbar={() => {}}
+            />
```
vasharma05, line 205: Snackbar is a notification, it's not a component to be shown in place of search.
Please read [here](https://zeroheight.com/23a080e38/p/683580-notifications) for more info

vasharma05: We can use an ActionableSnackbar, using the props `actionButtonLabel` and `onActionButtonClick` in the `showSnackbar`. Learn more about `showSnackbar` here: https://github.com/openmrs/openmrs-esm-core/blob/13067210cba73a461e65345388a887186d9da2e5/packages/framework/esm-styleguide/src/snackbars/index.tsx#L32

vasharma05 (dismissed):

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -107,9 +110,6 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
       if (!config.chooseLocation.enabled || locations?.length === 1) {
         changeLocation(locations[0]?.resource.id, false);
       }
-      if (!locations?.length) {
-        changeLocation();
-      }
```
vasharma05, line 112:
```
      if (!error && !locations?.length) {
        changeLocation();
      }
```

vasharma05: @ibacher, do we have any implementation at all where there are no locations?
Thanks!

vasharma05:
Updated error screen:
image

vasharma05:
Re-requesting review from @ibacher @samuelmale @denniskigen.
Thanks!

packages/framework/esm-translations/src/translations.ts
```diff
@@ -20,6 +20,7 @@ export const coreTranslations = {
   change: 'Change',
   close: 'Close',
   confirm: 'Confirm',
+  contactAdministratorIfIssuePersists: 'If the problem persists contact your system administrator.',
```
brandones, line 23:
```suggestion
  contactAdministratorIfIssuePersists: 'Contact your system administrator. if the problem persists.',
```

brandones:
Based on the designs it looks like we can do away with the top "Welcome" block.

image

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
ibacher: It's not impossible, but I would say that that's a configuration error. Every functioning OMRS system should have at least one location.

ibacher:
> Based on the designs it looks like we can do away with the top "Welcome" block.

I thought the design there was just a sketch of the notification, not a full screen preview. I kind of prefer it with the "Welcome" block because it makes it look more like what it is (an unexpected error in one part of the system) rather than just a basically blank page.

brandones:
That makes sense, thanks @ibacher . I think you're right.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
brandones: @jayasanka-sack @vasharma05 What's up with this, should we remove this function call?

brandones (approved):
I have only minor feedback; feel free to ship this when you feel happy with it @vasharma05 .

PR #914, opened 2024-02-01
(fix) useOnClickOutside should not use the click event

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This reworks `useOnClickOutside()` so that it binds only to the `mousedown` or `touchstart` events rather than the `click` event.

A rationale for this with context is [here](https://openmrs.slack.com/archives/CHP5QAE5R/p1706810370264969), but, in essence, a React component can render in response to the React synthetic `click` event before the native DOM `click` has completed bubbling. This means that a component rendered as the result of a `onClick` event that uses this hook is likely to have its `OnClickOutside` handler fired before the component has actually appeared to the user, which is very undesirable. By binding to the `mousedown` or `touchstart` events, we bypass this issue, since the synthetic `click` event is only fired _after_ the native `mousedown` or `touchstart` event has completed.

Because of this fix, I've also reverted [9afd854](https://github.com/openmrs/openmrs-esm-core/commit/9afd85492ee53a60776b05fa984514dcdee0f2e5) and made a couple of small improvements in related code.

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/app-menu-panel.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/app-menu-panel.component.tsx
@@ -8,11 +8,11 @@ import styles from './app-menu-panel.scss';
 
 interface AppMenuProps {
   expanded: boolean;
-  hidePanel: () => void;
+  hidePanel: Parameters<typeof useOnClickOutside>[0];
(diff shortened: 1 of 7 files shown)


denniskigen:
Very good! Thanks, @ibacher!

denniskigen (approved):

denniskigen (approved):

PR #915, opened 2024-02-02
(feat) Extend `createAttachment` to handle descriptions of image uploads

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

Related to https://github.com/openmrs/openmrs-esm-patient-chart/pull/1626.

https://github.com/openmrs/openmrs-module-attachments/commit/613f3b01fcb2c55faf98f49bccced48fa722643b introduced the ability to persist a description for an image upload using the `fileCaption` property of the payload. The API response contains a `comment` property that returns the description. This PR amends the `createAttachment` function and type annotations for attachments to incorporate these new changes.

### DIFF
--- a/packages/framework/esm-api/src/attachments.ts
+++ b/packages/framework/esm-api/src/attachments.ts
@@ -19,7 +19,8 @@ export function getAttachments(patientUuid: string, includeEncounterless: boolea
 export async function createAttachment(patientUuid: string, fileToUpload: UploadedFile) {
   const formData = new FormData();
 
-  formData.append('fileCaption', fileToUpload.fileName);
+  formData.append('fileName', fileToUpload.fileName);
+  formData.append('fileCaption', fileToUpload.fileDescription);
   formData.append('patient', patientUuid);
 
   if (fileToUpload.file) {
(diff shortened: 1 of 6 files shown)


packages/framework/esm-api/src/attachments.ts
```diff
@@ -19,7 +19,8 @@ export function getAttachments(patientUuid: string, includeEncounterless: boolea
 export async function createAttachment(patientUuid: string, fileToUpload: UploadedFile) {
   const formData = new FormData();
 
-  formData.append('fileCaption', fileToUpload.fileName);
+  formData.append('fileName', fileToUpload.fileName);
```
ibacher, line 22: Actually submitting the filename is not required. The file name we use is part of the entry for the `file` itself, if that makes sense (i.e., line 29 already covers this).

ibacher (commented):
One small change, but LGTM otherwise.

packages/framework/esm-api/src/attachments.ts
denniskigen: It does, thanks.

ibacher (approved):

denniskigen:
Are we doing any special handling for webcam uploads? The filename isn't taking for those.

ibacher:
Kind of. The data there is being sent in the `base64Content` parameter and doesn't have a file name, so the file name is set (on the backend) as `cameracaptureXXXXXXXXX.png` where `XXXXXXXXXX` is a random number.

ibacher:
We should probably change the backend logic to use the file name if it's provided.

PR #916, opened 2024-02-05
(feat) O3-2808: Snackbars should be dismissed after 5 seconds

usamaidrsk:
# Requirements

## For changes to apps

## If applicable

## Summary

Snackbars should be dismissed after a 5 seconds. i.e  this is for non error snackbars. 
This PR also add `autoClose` prop that can be used to close error snackbars if needed.

## Screenshots

https://github.com/openmrs/openmrs-esm-core/assets/53287480/1c07671b-7677-4718-bb9d-f0ab95c1ae27

## Related Issue

[O3-2808](https://openmrs.atlassian.net/browse/O3-2808)

### DIFF
--- a/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
+++ b/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
@@ -30,19 +30,19 @@ export type SnackbarType = 'error' | 'info' | 'info-square' | 'success' | 'warni
 
 export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar: removeSnackBarFromDom }) => {
   const {
-    actionButtonLabel,
-    isLowContrast,
-    kind,
+    actionButtonLabel = '',
+    isLowContrast = true,
+    kind = 'info',
     onActionButtonClick = () => {},
     progressActionLabel,
-    subtitle,
+    subtitle = '',
     timeoutInMs = 5000,
     autoClose = kind !== 'error',
     title,
     ...props
   } = snackbar;
 
-  const [actionText, setActionText] = useState(actionButtonLabel || '');
+  const [actionText, setActionText] = useState(actionButtonLabel);
   const [applyAnimation, setApplyAnimation] = useState(true);
 
   const [isClosing, setIsClosing] = useState(false);
@@ -88,12 +88,12 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar: rem
         [styles.slideOut]: isClosing,
       })}
       inline
-      kind={kind || 'info'}
+      kind={kind}
       lowContrast={isLowContrast}
       onActionButtonClick={handleActionClick}
       onClose={closeSnackbar}
       statusIconDescription="Snackbar notification"
-      subtitle={subtitle || ''}
+      subtitle={subtitle}
       title={title}
       {...props}
     />
--- a/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
+++ b/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
@@ -1,5 +1,5 @@
 /** @module @category UI */
-import React, { useState, useEffect, useCallback, useMemo } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import { ActionableNotification } from '@carbon/react';
 import classnames from 'classnames';
 import styles from './snackbar.module.scss';
--- a/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
+++ b/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
@@ -1,5 +1,5 @@
 /** @module @category UI */
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect, useCallback, useMemo } from 'react';
 import { ActionableNotification } from '@carbon/react';
 import classnames from 'classnames';
 import styles from './snackbar.module.scss';
@@ -18,6 +18,7 @@ export interface SnackbarDescriptor {
   progressActionLabel?: string;
   subtitle?: string;
   timeoutInMs?: number;
+  userDismissible?: boolean;
   title: string;
 }
 
@@ -36,11 +37,12 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar }) =
     progressActionLabel,
     subtitle,
     timeoutInMs,
+    userDismissible,
     title,
     ...props
   } = snackbar;
 
-  const [actionText, setActionText] = useState(actionButtonLabel);
+  const [actionText, setActionText] = useState(actionButtonLabel || '');
   const [applyAnimation, setApplyAnimation] = useState(true);
 
   const [isClosing, setIsClosing] = useState(false);
@@ -57,11 +59,14 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar }) =
   };
 
   useEffect(() => {
-    if (timeoutInMs) {
-      const timeoutId = setTimeout(onCloseSnackbar, timeoutInMs);
+    // Snackbar can auto clear after 5s if it's a success or if it's warning and not user dismissible
+    const canAutoClear = kind === 'success' || (!userDismissible && (kind === 'warning' || kind === 'warning-alt'));
+
+    if (timeoutInMs || canAutoClear) {
+      const timeoutId = setTimeout(onCloseSnackbar, timeoutInMs || 5000);
       return () => clearTimeout(timeoutId);
     }
-  }, [timeoutInMs, onCloseSnackbar]);
+  }, [timeoutInMs, userDismissible, kind, onCloseSnackbar]);
 
   useEffect(() => {
     setApplyAnimation(false);
@@ -73,7 +78,7 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar }) =
 
   return (
     <ActionableNotification
-      actionButtonLabel={actionText || ''}
+      actionButtonLabel={actionText}
       ariaLabel="Close snackbar"
       className={classnames(styles.slideIn, {
         [styles.animated]: applyAnimation,
--- a/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
+++ b/packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
@@ -18,7 +18,7 @@ export interface SnackbarDescriptor {
   progressActionLabel?: string;
   subtitle?: string;
   timeoutInMs?: number;
-  userDismissible?: boolean;
+  autoClose?: boolean;
   title: string;
 }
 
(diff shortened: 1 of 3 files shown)


vasharma05:
Hi @usamaidrsk!
Please add a screenshot.
Thanks!

usamaidrsk:
@vasharma05 , @denniskigen  I have added a video to demo the changes
Thanks

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -1,5 +1,5 @@
 /** @module @category UI */
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect, useCallback, useMemo } from 'react';
```
jayasanka-sack, line 2: `useMemo` isn't used anywhere here, right?

jayasanka-sack:
Thanks for the PR, Usama! Could you please add a unit test for this?

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -57,11 +59,14 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar }) =
   };
 
   useEffect(() => {
-    if (timeoutInMs) {
-      const timeoutId = setTimeout(onCloseSnackbar, timeoutInMs);
+    // Snackbar can auto clear after 5s if it's a success or if it's warning and not user dismissible
+    const canAutoClear = kind === 'success' || (!userDismissible && (kind === 'warning' || kind === 'warning-alt'));
+
+    if (timeoutInMs || canAutoClear) {
```
jayasanka-sack, line 65: Why do we check for `timeoutInMs` here?

usamaidrsk: `timeoutInMs` value at some point it won't be passed on calling `showSnackbar`, thus it's checked and if it has a value, then the `snackbar` will clear after the time defined,  else clear after 5s if the `snackbar` has auto clear

Thanks @jayasanka-sack

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -57,11 +59,14 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar }) =
   };
 
   useEffect(() => {
-    if (timeoutInMs) {
-      const timeoutId = setTimeout(onCloseSnackbar, timeoutInMs);
+    // Snackbar can auto clear after 5s if it's a success or if it's warning and not user dismissible
+    const canAutoClear = kind === 'success' || (!userDismissible && (kind === 'warning' || kind === 'warning-alt'));
```
ibacher, line 63: The rule here should basically be `kind !== 'error'`. This leaves out `info`.

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -36,11 +37,12 @@ export const Snackbar: React.FC<SnackbarProps> = ({ snackbar, closeSnackbar }) =
     progressActionLabel,
     subtitle,
     timeoutInMs,
+    userDismissible,
```
ibacher, line 40: I don't really get why we need a `userDismissable` property here?

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
ibacher: TBH, I'm not even 100% on this. I think an acceptable solution to this would be:

```
const {
 // ... snip
 timeoutInMs = 5000,
 // ... snip
} = snackbar;
```

ibacher (changes requested):

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
usamaidrsk: I added `userDismissable` so as to bringout what the description mentioned about warning snackbars on the ticket you filed @ibacher 
That when on creating a warning snackbar it can be created with `userDismissable` true, so as not to auto clear.

"Snackbar notifications for warnings should default to clearing with the option to make them remain until cleared."

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
usamaidrsk: I think I can change to have that, and test it out

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
jayasanka-sack: Given that there's a default timeout set for 5 seconds, it's necessary to close it regardless of whether the `timeoutInMs` parameter has a value or not, correct?

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
jayasanka-sack: in that case, how about having a prop names something like `autoClose`?

the behaviour would be,
- if `autoClose` value is provided:  
    autoClose based on the provided value regardless the notification kind.
- if not provided,
    the `autoClose`s default value should be (feel free to use default value or suitable approach)
    - `true` for `success`
    - `false` for warnings and errors

usamaidrsk: Okay, thanks @jayasanka-sack

usamaidrsk: cc @jayasanka-sack I have pushed some updates
Thanks

denniskigen:
@ibacher could you re-review?

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -27,24 +28,31 @@ export interface SnackbarMeta extends SnackbarDescriptor {
...
     ...props
   } = snackbar;
 
-  const [actionText, setActionText] = useState(actionButtonLabel);
+  const [actionText, setActionText] = useState(actionButtonLabel || '');
   const [applyAnimation, setApplyAnimation] = useState(true);
 
   const [isClosing, setIsClosing] = useState(false);
 
+  const closeSnackbar = () => {
```
ibacher, line 50: General rule: any function created inside a functional component and passed to a child should use `useCallback()`. This is because the function body gets run on each render cycle, so each time, `closeSnackbar` is a new function, which means that the child component will re-render every time this component re-renders even if none of it's other props have changed.

ibacher (commented):
Aside from one small nit, LGTM. Thanks @usamaidrsk!

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
usamaidrsk: Thanks for the notes @ibacher

denniskigen (approved):

denniskigen:
Good to go, @ibacher?

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -27,24 +28,31 @@ export interface SnackbarMeta extends SnackbarDescriptor {
...
     subtitle,
-    timeoutInMs,
+    timeoutInMs = 5000,
+    autoClose = kind !== 'error',
     title,
     ...props
   } = snackbar;
 
-  const [actionText, setActionText] = useState(actionButtonLabel);
+  const [actionText, setActionText] = useState(actionButtonLabel || '');
```
jayasanka-sack, line 45: Why do we need a state?

jayasanka-sack: Any specific reason for this?

We can use it directly, right? 
```
 <ActionableNotification
      actionButtonLabel={actionButtonLabel}
```

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -27,24 +28,31 @@ export interface SnackbarMeta extends SnackbarDescriptor {
...
   const [applyAnimation, setApplyAnimation] = useState(true);
 
   const [isClosing, setIsClosing] = useState(false);
 
+  const closeSnackbar = useCallback(() => {
+    // This is to add a slide out animation before closing the snackbar
+    // The animation lasts for 250ms, thus the timeout
+    setTimeout(removeSnackBarFromDom, 250);
```
jayasanka-sack, line 53: Where the implementation for `removeSnackBarFromDom` comes from? Do we need to pass it as a prop?

jayasanka-sack (changes requested):
Usama, can you please write a unit test for this?

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
ibacher: It's the second prop already...

ibacher (approved):

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
jayasanka-sack: yes, but do we need to pass how to remove it on everywhere we use the snackbar? Can't we implement it within the component?

jayasanka-sack: Oh. I read the implementation. Makes sense now.

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
vasharma05: I agree, I don't know why this was resolved.

`setActionText` is not being used, then why are we using a state?

usamaidrsk:
```typescript
const handleActionClick = () => {
    onActionButtonClick();
    onCloseSnackbar();
    progressActionLabel && setActionText(progressActionLabel);
  };
``` 

@vasharma05 it is used here, Thanks

vasharma05: Hi @hadijahkyampeire!
Any reason why we use progressActionLabel since it is being set after the snack bar is closed?
Can we remove it all?
Thanks!

vasharma05:
@jayasanka-sack, please re-review the PR.
Thanks!

packages/framework/esm-styleguide/src/snackbars/snackbar.component.tsx
```diff
@@ -27,24 +28,30 @@ export interface SnackbarMeta extends SnackbarDescriptor {
...
+    actionButtonLabel = '',
+    kind = 'success',
     onActionButtonClick = () => {},
+    isLowContrast = kind !== 'error',
     progressActionLabel,
-    subtitle,
-    timeoutInMs,
+    subtitle = '',
+    timeoutInMs = 5000,
+    autoClose = kind !== 'error',
```
denniskigen, line 40: Ultimately, we'll likely want to revisit this logic. Consider actionable success notifications for example. Do we want them to stick around until the user dismisses the snackbar manually? If so, this logic won't work. Good enough to go for now imo, but worth thinking about nonetheless.

usamaidrsk: Nice catch. I think we can still add that check too.
Given that what makes a snackbar actionable, is when `actionButtonLabel` and `onActionButtonClick` are provided, if I am right.

cc @denniskigen

denniskigen: Yeah, it'll require some tweaking, and would probably be best done in a separate follow-up PR rather than trying to anticipate all the possible situations here.

jayasanka-sack (approved):

ibacher:
Thanks @usamaidrsk and everyone!

PR #919, opened 2024-02-08
(feat) O3-2831: Move patient banner into styleguide

brandones:
See https://github.com/openmrs/openmrs-esm-patient-chart/pull/1645 and https://github.com/openmrs/openmrs-esm-patient-management/pull/1018

# Requirements

## For changes to apps

## If applicable

- [x] My work includes tests or is validated by existing tests. _(well, I just moved the existing ones over)_

## Summary

This creates styleguide components for the building blocks of the patient banner. These are:
- CustomOverflowMenu
- PatientBannerActionsMenu
- PatientBannerContactDetails
- PatientBannerPatientInfo
- PatientBannerToggleContactDetailsButton

`CustomOverflowMenu` has some usage outside the patient banner, so it has been made more independent.

Later we should create a simple `<PatientBanner>` component that brings it all together in "the default" way, which will replace all current usages of `patient-banner-slot`. We can then give the patient banner extension a deprecation notice and eventually remove it.

Having these decomposed banner-parts makes it easy to construct a non-default banner, such as is needed for different types of patient searches. Thus there should be little pressure to add features to PatientBanner, since custom behavior can be implemented easily using the individual parts.

I also removed `OmrsContentSwitcher`, which was dead code.

## Screenshots

https://github.com/openmrs/openmrs-esm-core/assets/1031876/b4fbd852-cc5e-4463-a7e0-88a36956937f

## Related Issue
https://openmrs.atlassian.net/browse/O3-2831


packages/framework/esm-styleguide/src/config-schema.ts
```diff
@@ -15,8 +15,35 @@ export const esmStyleGuideSchema = {
   },
   preferredCalendar: {
     _type: Type.Object,
+    _description:
```
brandones, line 18: Unrelated, this was lacking a description and validation.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+              ))
+            : ''}
+        </div>
+      </div>
+    </div>
+  );
+}
+
+const getGender = (gender: string): string => {
+  // TODO: We ought to have some kind of "core translations."
```
ibacher, line 69: Yes!

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return (
+    <div className={styles.patientInfo}>
+      <div className={classNames(styles.row, styles.patientNameRow)}>
+        <div className={styles.flexRow}>
+          <span className={styles.patientName}>{name}</span>
+          {/* TODO: The fhir Patient passed in the state to the patient-banner-tags-slot is
+          totally unused other than the death date, which is used as a boolean to
+          tell if a patient is dead. We should eventually try to get rid of it and
+          just provide `isDeceased` as state. */}
```
ibacher, line 38: I think, in principle, we want the `patient-banner-tags-slot` to be able to load implementation-defined tags. It's true that we only actively use this for `isDeceased`, but an implementation-specific tag may have additional uses for the patient record.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+export function PatientBannerPatientInfo({
+  patientUuid,
+  name,
+  gender,
+  identifiers,
+  birthDate,
+  deathDate,
+}: PatientBannerPatientInfoProps) {
+  const { excludePatientIdentifierCodeTypes } = useConfig();
```
ibacher, line 25: What configuration is this going to pick up? This might be better passed in a prop to this component?

ibacher (commented):
In principle, I like the approach and this is a very reasonable component to move into the styleguide.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
brandones: Yeah, the trouble is that the patient banner doesn't always have access to the `fhir.Patient`. Specifically in the patient-search component, what is available is a SearchedPatient object, so the state passed to the tags slot ends up being this [terrible hack](https://github.com/openmrs/openmrs-esm-patient-management/blob/90ca8beba97580524d5ccfed10eb885e834589f1/packages/esm-patient-search-app/src/patient-search-page/patient-banner/banner/patient-banner.component.tsx#L84-L88). It would be better if the API could be satisfied by all the slots.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
brandones: This will be the styleguide config, which already exists. We'd want to pass it as props if we want it to be possible to configure differently in different apps, and I'm not sure why we'd want that.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
ibacher: Ah... that's awful. I had some designs on re-working the search to use the FHIR API, now that there's actually an endpoint that's responsive to our actual needs, but we haven't made the switch yet.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
brandones: My idea is to define a new core app just called like `@openmrs/esm-core-translations` that just serves translations, and then add a new function to the framework (maybe `esm-utils`, unfortunately) that basically just does

```ts
function getCoreTranslation(key, default) {
  return translateFrom('@openmrs/esm-core-translations', key, default)
}
```

We could even maintain a tuple type for `key` so clients know via typescript exactly what translation keys are available in core.

brandones:
> In principle, I like the approach and this is a very reasonable component to move into the styleguide.

Should I do these components one at a time, or expand this PR to do all four patient banner components?

ibacher:
> Should I do these components one at a time, or expand this PR to do all four patient banner components?

I'm happy to just have one PR that does everything.

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
ibacher: `useConfig()` by default uses the `ComponentContext` to figure out which module to use the config for, so in the current chart PR, it works because it's still resolving to "@openmrs/esm-patient-banner-app". I guess I'm really asking: should we have some kind of pattern for modifying the `ComponentContext` for components from the styleguide or use `useConfig({ externalModuleName: "@openmrs/esm-styleguide" })` (this syntax is intentionally clunky to avoid people over-relying on it).

brandones: Oh that's right! Good thinking. I'm fine with either, I suppose, as long as it's an implementation detail handled by esm-styleguide code so that app developers don't have to think about it. I am inclined to use `externalModuleName` just because it seems simpler to me (and I'm not averse to using core's more questionable features within core itself).

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
brandones: Oh! If we do that then the interface to `PatientBannerPatientInfo` could also be simplified just to accept the patient object. I'm actually inclined to wait on this until we change patient search to use the fhir patient. It looks like all other usages of the patient banner use the fhir patient.

brandones: Is there a ticket for that?

packages/framework/esm-styleguide/src/patient-banner/patient-banner-info/patient-banner-info.component.tsx
brandones: Filed it https://openmrs.atlassian.net/browse/O3-2882

packages/framework/esm-styleguide/src/custom-overflow-menu/custom-overflow-menu.component.tsx
```diff
@@ -0,0 +1,57 @@
+/** @module @category UI */
+import React, { useCallback } from 'react';
+import classNames from 'classnames';
+import { useLayoutType, useOnClickOutside } from '@openmrs/esm-framework';
+import styles from './custom-overflow-menu.module.scss';
+
+interface CustomOverflowMenuProps {
+  menuTitle: React.ReactNode;
+  children: React.ReactNode;
+  deceased?: boolean;
```
raghavawasthi24, line 10: decreased must not be optional because you are using it below

packages/framework/esm-styleguide/src/custom-overflow-menu/custom-overflow-menu.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <div data-overflow-menu className={classNames('cds--overflow-menu', styles.container)} ref={ref}>
+      <button
+        className={classNames(
+          'cds--btn',
+          'cds--btn--ghost',
+          'cds--overflow-menu__trigger',
+          { 'cds--overflow-menu--open': menuIsOpen },
+          { [styles.deceased]: deceased },
```
raghavawasthi24, line 27: If decreased is optional so whenever decreased is not passed as props , it will throw an error since it will be `undefined` in that case.

raghavawasthi24 (changes requested):

packages/framework/esm-framework/mock.tsx
```diff
@@ -279,6 +279,25 @@ export const unsetLeftNav = jest.fn();
 export const ResponsiveWrapper = jest.fn(({ children }) => <>{children}</>);
 export const OpenmrsDatePicker = jest.fn(() => <div>OpenMRS DatePicker</div>);
 
+export const CustomOverflowMenu = jest.fn(({ children }) => (
+  <div>
+    <button>Custom Overflow Menu</button>
+    {children}
+  </div>
```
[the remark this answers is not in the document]
ibacher: And what happens to the `{children}`?

packages/framework/esm-styleguide/src/custom-overflow-menu/custom-overflow-menu.component.tsx
ibacher: Did you test your theory or look at what the `classnames()` library does?

PR #923, opened 2024-02-09
(fix) Require fewer requests to the session endpoint

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

So, by my count, our current login cycle makes 8 requests to the /session endpoint:

Screenshot 2024-02-09 at 3 29 57 PM

The goal of this PR is to reduce the number of requests, specifically to three (the initial session check, the actual authentication, and then setting the session location). I've also rationalised things in `current-user` so that functions that return a promise now return a promise with a defined data type, specifically the session state.


denniskigen (approved):
Huge improvement! Thanks, @ibacher.

denniskigen:
Do you have the bandwidth to fix things here, @ibacher?

ibacher:
Yes, I'm going to fix this up next week, when I've got more time to devote to OpenMRS coding.

ibacher:
Hmmm... can't figure out this E2E test thing. Run locally, it works—I can't get the Docker config to work, though. Obviously, run here it doesn't.

denniskigen:
Can confirm that it works locally for me as well.

ibacher:
Alright, I've worked out what's happening with the e2e tests and it is something we need to solve, but not something that should block this PR right now. Specifically:

* The PR relies on a series of changes to the framework, specifically, `refetchCurrentUser()` and `setSessionLocation()` are both modified to skip the need for the additional session calls that were happening.
* The e2e test scripts only load the copies of the apps. The app shell remains the `next` version.
* `esm-framework` is always the version bundled with the app-shell because it is (correctly) a singleton and the app-shell (correctly) depends on functions inside the framework before any other apps are loaded (among other things, the actual code loading mechanism for all frontend modules is defined in the framework).
* Because of this, on the e2e tests, the tests are being run with the previous version of the framework, without the changes that make this PR work.

PR #933, opened 2024-02-18
(feat) O3-2882 Create core translations library

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

Creates a core translations library.

Copying from the issue description: This solves two problems. First, there is text that belongs to elements that we would like to move into esm-styleguide. esm-styleguide can’t really have its own translations without changing some things about how translations are loaded (the loader looks for them in apps). Second, there are a number of strings that are used many many times throughout O3, and currently have to be translated separately in each app.

The dirty work of adding a good set of common translations and switching to `getCoreTranslation` everywhere (in this repository) I am leaving until it's clear that this is the direction we want to go.

## Screenshots

image

## Related Issue

https://issues.openmrs.org/browse/O3-2882


brandones:
Oops, gotta fix CI & E2E...

packages/framework/esm-translations/src/index.ts
```diff
@@ -0,0 +1,27 @@
+/** @module @category Translation */
+import _i18n from 'i18next';
+import { coreTranslations } from './translations';
+
+export const importCoreTranslation = require.context('../translations', false, /.json$/, 'lazy');
+
+export function startupApp() {}
```
samuelmale, line 7: We don't need the "startupApp func" for libraries.

packages/framework/esm-translations/src/translations.ts
```diff
@@ -0,0 +1,10 @@
+export const coreTranslations = {
+  cancel: 'Cancel',
+  change: 'Change',
+  confirm: 'Confirm',
+  close: 'Close',
+  error: 'Error',
+  errorCopy:
+    'Sorry, there was a problem displaying this information. You can try to reload this page, or contact the site administrator and quote the error code above.',
+  loading: 'Loading',
```
samuelmale, line 9: I understand that this contributes to the idea of ensuring that the getCoreTranslation is strongly typed. The tradeoff here is that we may have to maintain two translation files, which could lead to potential issues along the way.

packages/framework/esm-translations/webpack.config.js
```diff
@@ -0,0 +1,42 @@
...
+    new BundleAnalyzerPlugin({
+      analyzerMode: env && env.analyze ? 'static' : 'disabled',
+    }),
+  ],
+  devServer: {
+    disableHostCheck: true,
+    headers: {
+      'Access-Control-Allow-Origin': '*',
+    },
+  },
```
samuelmale, line 41: I have seen this across webpack configs for libraries but do we need it?

packages/framework/esm-translations/src/translations.ts
ibacher: This is basically replacing the various `t('key', 'default')` or `// t('key', 'default')` lines we'd have since we're using a custom lexer to provide translation keys based solely on this file. The `en.json` translations are driven from this the same way they'd usually be driven from `t('key', 'default')` calls.

package.json
```diff
@@ -19,7 +19,7 @@
     "setup": "yarn install && turbo run build",
     "start": "openmrs develop",
     "verify": "turbo run lint && turbo run test && turbo run typescript",
-    "prettier": "prettier --config prettier.config.js --write \"packages/**/*.{ts,tsx,css,scss}\" \"e2e/**/*.ts\"",
+    "prettier": "prettier --config prettier.config.js --write \"packages/**/*.{ts,tsx,css,scss}\" \"e2e/**/*.ts\" | grep -v unchanged",
```
ibacher, line 22: This won't work on Windows.

```suggestion
    "prettier": "prettier --config prettier.config.js --write \"packages/**/*.{ts,tsx,css,scss}\" \"e2e/**/*.ts\" --list-different",
```

Accomplishes basically the same thing.

packages/shell/esm-app-shell/src/locale.ts
```diff
@@ -31,6 +31,20 @@ export function setupI18n() {
           callback(Error("can't handle translation namespace"), null);
         } else if (namespace === undefined || language === undefined) {
           callback(Error(), null);
+        } else if (namespace === 'core') {
```
ibacher, line 34: I think this is where we have the major technical design decision here: do we need to support special handling for both the core and `@openmrs/esm-app-shell` namespaces or can we get away with just one? The `@openmrs/esm-app-shell` namespace is currently not used for translations, so there's no harm in removing it and it seems like if we actually needed to use it (e.g., for some feature in the sole "core extension", the breadcrumbs), we could just as harmlessly use the `core` namespace for that?

packages/framework/esm-translations/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  const i18n: typeof _i18n = (_i18n as unknown as { default: typeof _i18n }).default || _i18n;
+
+  return i18n.t(key, {
+    ns: moduleName,
+    defaultValue: fallback,
+    ...options,
+  });
+}
+
+export function getCoreTranslation(key: CoreTranslationKey, defaultText?: string): string {
```
ibacher, line 21: I think `getFrameworkTranslation()` might be a better name? "Core" is a bit nebulous and "core translations" starts to sound like where I'd get translations like "Patient" from.

ibacher (approved):
Couple of small things that might be worth discussing, but I think the overall implementation looks good.

packages/framework/esm-translations/src/index.ts
brandones: Well I mean, things like "Patient" I think would fall under what I'm calling the second usage of this library: "Second, there are a number of strings that are used many many times throughout O3, and currently have to be translated separately in each app." Is this not a useful purpose for this feature?

packages/framework/esm-translations/src/translations.ts
samuelmale: @ibacher Just to be sure I understood you correctly, question: does this fit well with our current Transifex integration? By config, [en.json seems like the canonical source of truth](https://github.com/openmrs/openmrs-esm-core/blob/43fb48cdbb9b390e89f16513beb5940d25c8378e/.tx/config#L38), what happens if the two files get out of sync ie. the `en.json` file missing some keys that probably exist in this file?

samuelmale: Thanks @ibacher for the detailed explanation. 

>Remember: we do not manually edit the en.json file at any point.

We should update our coding conventions to enforce this. I've in many scenarios bypassed the parser because I've always manually updated the translation files as I add a translatable string. In as much the script "extract-translations" exists (can be consumed manually) independent of the pre-commit hook, not all devs make good use of this feature and by the time the parsing happens after the pre-commit hook runs, you're are in a post-commit state and you just push the changes bypassing the parser. This compromised my discernment here. Thanks!!

mseaton (commented):
I added some comments and thoughts on the ticket.  Would be good to discuss before merging.

packages/shell/esm-app-shell/src/locale.ts
brandones: @ibacher ping

packages/framework/esm-translations/src/index.ts
brandones: @ibacher ?

ibacher: > Is this not a useful purpose for this feature?

That's a good point.

packages/shell/esm-app-shell/src/locale.ts
ibacher: Yes.

brandones:
Turns out it wasn't working at all; I needed to manually load the namespace (which `react-i18next` does for you, behind the scenes) and ensure that the translation overrides could be parsed. Now it's working!



@ibacher, I'd appreciate a quick re-review if you don't mind. Not much code but definitely some choices made.

ibacher:
It looks like the build failures are down to the internal parts of the framework that used to import `translateFrom()` from `esm-utils`. I think those just need to be updated to import from `esm-translation`.

brandones:
Oh yeah, thanks @ibacher ! Actually I went ahead and added them all as core translations and totally got rid of all the translateFrom usages. Sound good?

brandones:
Sorry I keep throwing more stuff into this PR. I decided to fix an issue with the tests by moving mocks into the library they belong to, and that sort of set off a cascade... nevertheless I'm happy with the results.

ibacher:
> Sound good?

Sounds excellent!

ibacher (approved):

brandones:
[21ad8df0eb678414e712ff1fea1f918f4543d35f.webm](https://github.com/openmrs/openmrs-esm-core/assets/1031876/8c5896fa-3979-4698-96c9-128a1502188e)

Well, this is a new one for me. It works fine on my dev server, `yarn verify` passing, but it seems like the E2E server app shell is using a version of esm-core that doesn't have the latest change? Seems like maybe a problem with the E2E setup. @jayasanka-sack @ibacher does that seem right, should I go ahead and merge?

jayasanka-sack:
@brandones Yes the E2E setup can only have changes of apps from the PR. For other utils, it uses from the `next` version. We need to fix this later. I'll merge the PR for now. 
https://github.com/openmrs/openmrs-esm-core/blob/8435603550789f2fdfd4dfd66ad03a71c526af28/e2e/support/github/run-e2e-docker-env.sh#L8

brandones:
Ok, thanks @jayasanka-sack !

denniskigen:
It'd be superb to have something in the docs related to this, @brandones!

brandones:
Docs PR: https://github.com/denniskigen/o3-docs/pull/27

PR #937, opened 2024-02-20
(feat) O3-2883: Use toast notifications for system-generated error messages

hadijahkyampeire:
# Requirements

## For changes to apps

## If applicable

## Summary

This makes it so that system-generated error messages render toast notifications instead of inline notifications per the [notifications design guidelines](https://zeroheight.com/23a080e38/p/683580-notifications).

### Before

CleanShot 2024-02-20 at 11  50 33@2x

### After

CleanShot 2024-02-20 at 11  06 52@2x

## Related Issue
https://openmrs.atlassian.net/browse/O3-2883

### DIFF
--- a/packages/framework/esm-error-handling/src/index.ts
+++ b/packages/framework/esm-error-handling/src/index.ts
@@ -1,9 +1,9 @@
 /** @module @category Error Handling */
-import { dispatchNotificationShown } from '@openmrs/esm-globals';
+import { dispatchToastShown } from '@openmrs/esm-globals';
 
 window.onerror = function (error) {
   console.error('Unexpected error: ', error);
-  dispatchNotificationShown({
+  dispatchToastShown({
     description: error ?? 'Oops! An unexpected error occurred.',
     kind: 'error',
     title: 'Error',
@@ -13,7 +13,7 @@ window.onerror = function (error) {
 
 window.onunhandledrejection = function (event) {
   console.error('Unhandled rejection: ', event.reason);
-  dispatchNotificationShown({
+  dispatchToastShown({
     description: event.reason ?? 'Oops! An unhandled promise rejection occurred.',
     kind: 'error',
     title: 'Error',
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -690,7 +690,7 @@ ___
 
 #### Defined in
 
-[packages/framework/esm-styleguide/src/toasts/toast.component.tsx:28](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/toasts/toast.component.tsx#L28)
+[packages/framework/esm-styleguide/src/toasts/toast.component.tsx:23](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/toasts/toast.component.tsx#L23)
 
 ___
 
@@ -4629,7 +4629,7 @@ ___
 
 #### Defined in
 
-[packages/framework/esm-globals/src/events.ts:104](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-globals/src/events.ts#L104)
+[packages/framework/esm-globals/src/events.ts:107](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-globals/src/events.ts#L107)
 
 ___
 
@@ -4655,7 +4655,7 @@ ___
 
 #### Defined in
 
-[packages/framework/esm-globals/src/events.ts:97](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-globals/src/events.ts#L97)
+[packages/framework/esm-globals/src/events.ts:100](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-globals/src/events.ts#L100)
 
 ___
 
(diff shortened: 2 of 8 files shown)


packages/framework/esm-error-handling/src/index.ts
```diff
@@ -1,9 +1,9 @@
 /** @module @category Error Handling */
-import { dispatchNotificationShown } from '@openmrs/esm-globals';
+import { dispatchToastShown } from '@openmrs/esm-globals';
 
 window.onerror = function (error) {
   console.error('Unexpected error: ', error);
-  dispatchNotificationShown({
+  dispatchToastShown({
```
denniskigen, line 6: Per the [docs](https://zeroheight.com/23a080e38/p/683580-notifications/t/37c75c), system generated toast notifications should not use an auto-dismissable timer:

> Behaviour
>
> We do not recommend using the auto-dismissible timer for toast notification, due to their higher 
> level of importance and their text density. They can, however, offer the close button so the user 
> can dismiss them if they want.

Now's as good a time as ever to effect that change.

hadijahkyampeire: So they are not actionable and still not auto-dismissable? How are we gonna handle that, because the useEffect checks for the actionabelLabel and mouseOver the toast which means the toast will auto dismiss if the mouse is not over it.

hadijahkyampeire: So we should probably do what was done with the snackbars and add an autoClose boolean which is true unless specified otherwise.

denniskigen: Toasts can be actionable (see the example in the behaviour section of the toast docs in zeroheight and also the missing backend dependencies [notification](https://github.com/openmrs/openmrs-esm-core/pull/758)). I don't expect that they'll need to be auto-dismissable based on their priority. Removing the default `millis` prop or whatever gets passed out to the `timeout` [prop](https://react.carbondesignsystem.com/?path=/docs/components-notifications-toast--overview) should do the trick.

packages/framework/esm-error-handling/src/index.ts
```diff
@@ -13,7 +13,7 @@ window.onerror = function (error) {
 
 window.onunhandledrejection = function (event) {
   console.error('Unhandled rejection: ', event.reason);
-  dispatchNotificationShown({
+  dispatchToastShown({
     description: event.reason ?? 'Oops! An unhandled promise rejection occurred.',
```
usamaidrsk, line 17: @hadijahkyampeire a simple inquiry, we don't have translations setup here!

packages/framework/esm-error-handling/src/index.ts
usamaidrsk: I think the default millis would be these defined [here](https://github.com/openmrs/openmrs-esm-core/blob/1f24fc5ba7a9437fa6d7676e13a965d97d8f63a7/packages/framework/esm-styleguide/src/toasts/toast.component.tsx#L5-L7).

```typescript
const defaultOptions = {
  millis: 5000,
};
``` 
And used at this [point](https://github.com/openmrs/openmrs-esm-core/blob/1f24fc5ba7a9437fa6d7676e13a965d97d8f63a7/packages/framework/esm-styleguide/src/toasts/toast.component.tsx#L38)

```typescript


export const Toast: React.FC<ToastProps> = ({ toast, closeToast }) => {
  const {
    description,
    kind,
    critical,
    title,
    actionButtonLabel,
    onActionButtonClick = () => {},
    millis = defaultOptions.millis,
  } = toast;
```

packages/framework/esm-error-handling/src/index.ts
ibacher: That's a good point, but we can move that into a separate ticket, since I think it's dependent on [this](https://github.com/openmrs/openmrs-esm-core/pull/933).

packages/framework/esm-error-handling/src/index.ts
ibacher: > So they are not actionable and still not auto-dismissable?

Toasts should _never_ auto-dismiss by default, though it seems ok to have an API that allows the code generating the Toast to set the toast up to be auto-dismissable. Anything that should automatically be dismissed should use the Snackbar component.

packages/framework/esm-error-handling/src/index.ts
hadijahkyampeire: okay I will create a follow up ticket for the translation

packages/framework/esm-error-handling/src/index.ts
hadijahkyampeire: I am going to remove `millis` from the props so that we never allow Toasts to auto dismiss, I also think we don't need the MouseOver and MouseOut effect because it simulates the auto dismiss as well.

hadijahkyampeire:
@denniskigen @ibacher @usamaidrsk I have pushed the changes, please take another look.

denniskigen (approved):
Thanks, @hadijahkyampeire!

packages/apps/esm-offline-tools-app/translations/en.json
denniskigen, line 1: Not sure about these translation strings? Should we remove them from the diff, @ibacher?

denniskigen: I've removed them because it looks like they shouldn't be deleted based on https://github.com/openmrs/openmrs-esm-core/pull/938/files.

PR #942, opened 2024-02-22
(feat) Add loud error messages if script loading fails

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This adds some new features to `importDynamic()` primarily aimed at making the application noisier in case things like the recent issue with OHRI's PMTCT module re-appears. This adds a few related features:

1. `importDynamic()` now supports a `maxLoadingTime` optional parameter that will cause the promise returned by `importDynamic()` to reject if the loading isn't finished in the specified period of time. This defaults to the rather absurdly long 10 minutes.
2. A console error is now generated if it takes longer than 5 seconds for any added script tag to either load or error out. This is separate from and does not cause the `importDynamic()` call to reject; it is more or less intended as a warning to give developers some insight.
3. I cleaned up the various event listeners, which didn't always get removed.
4. I toned down the logging that can happen if there was an error loading the script. Previously, all places that were waiting for the script to finish loading would write an error message to the console and reject the promise. Now the console error message is only written by the initial request for a specific script.

## Other

I went back and forth quite a bit on the timing before landing on these defaults. I'm not entirely sure that 5 seconds is the right amount of time, but usually `importDynamic()` is only responsible for loading the entry point of a module, and subsequent chunks are loaded via the Webpack runtime. These initial chunks are usually less than 10kB each, so 5 seconds feels generous. Even throttling my connection down to "Good 2G", I only get one error message (though the app is _painfully_ slow).

Note that the two timeouts are not directly connected, and we do not make any attempt to stop the browser from downloading the script (it's done asynchronously anyways).

It would be possible to have the 5 second console error message configurable through `importDynamic()`. I just didn't do that because it's unlikely to be that useful (there are many use-cases for `importDynamic()` outside of the shell itself.

### DIFF
--- a/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
+++ b/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
@@ -22,16 +22,41 @@ export function slugify(name: string) {
  * const { someComponent } = importDynamic("@openmrs/esm-template-app")
  * ```
  *
- * @param jsPackage The package to load the export from
+ * @param jsPackage The package to load the export from.
  * @param share Indicates the name of the shared module; this is an advanced feature if the package you are loading
- *   doesn't use the default OpenMRS shared module name "./start"
+ *   doesn't use the default OpenMRS shared module name "./start".
+ * @param options Additional options to control loading this script.
+ * @param options.importMap The import map to use to load the script. This is useful for situations where you're
+ *   loading multiple scripts at a time, since it allows the calling code to supply an importMap, saving multiple
+ *   calls to `getCurrentImportMap()`.
+ * @param options.maxLoadingTime A positive integer representing the maximum number of milliseconds to wait for the
+ *   script to load before the promise returned from this function will be rejected. Defaults to 600,000 milliseconds,
+ *   i.e., 10 minutes.
  */
 export async function importDynamic<T = any>(
   jsPackage: string,
   share: string = './start',
-  options?: { importMap?: ImportMap },
+  options?: {
+    importMap?: ImportMap;
+    maxLoadingTime?: number;
+  },
 ): Promise<T> {
-  await preloadImport(jsPackage, options?.importMap);
+  // default to 10 minutes
+  const maxLoadingTime = !options?.maxLoadingTime || options.maxLoadingTime <= 0 ? 600_000 : options.maxLoadingTime;
+
+  let timeout: ReturnType<typeof setTimeout> | undefined = undefined;
+  await Promise.race([
+    preloadImport(jsPackage, options?.importMap),
+    new Promise((_, reject) => {
+      timeout = setTimeout(() => {
+        reject(
+          new Error(`Could not resolve requested script, ${jsPackage}, within ${humanReadableMs(maxLoadingTime)}.`),
+        );
+      }, maxLoadingTime);
+    }),
+  ]);
+
+  timeout && clearTimeout(timeout);
 
   const jsPackageSlug = slugify(jsPackage);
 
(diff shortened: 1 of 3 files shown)


packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -22,16 +22,41 @@ export function slugify(name: string) {
...
+    importMap?: ImportMap;
+    maxLoadingTime?: number;
+  },
 ): Promise<T> {
-  await preloadImport(jsPackage, options?.importMap);
+  // default to 10 minutes
+  const maxLoadingTime = !options?.maxLoadingTime || options.maxLoadingTime <= 0 ? 600_000 : options.maxLoadingTime;
+
+  let timeout: ReturnType<typeof setTimeout> | undefined = undefined;
+  await Promise.race([
```
ibacher, line 48: This is basically what `Promise.race()` exists for, timeouts for other promises.

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -56,6 +81,27 @@ export async function importDynamic<T = any>(
...
 
+/**
+ * Utility function to convert milliseconds into human-readable strings with rather absurd
+ * levels of precision.
+ *
+ * @param ms Number of milliseconds
+ * @returns A human-readable string useful only for error logging, where we can assume English
+ */
+function humanReadableMs(ms: number) {
+  if (ms < 1_000) {
```
ibacher, line 92: We can probably and should consider translating these (and the other strings) eventually. For now the target audience for these messages is really developers.

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -143,38 +195,65 @@ function loadScript(url: string, resolve: (value: unknown) => void, reject: (rea
     element.src = url;
     element.type = 'text/javascript';
     element.async = true;
-    const loadFn = () => {
+
+    // loadTime() displays an error if a script takes more than 5 seconds to load
+    const loadTimeout = setTimeout(() => {
+      console.error(
+        `The script at ${url} did not load within 5 seconds. This may indicate an issue with the imports in the script's entry-point file or with the script's bundler configuration.`,
```
ibacher, line 202: This is, at best, a guess.

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
```diff
@@ -143,38 +195,65 @@ function loadScript(url: string, resolve: (value: unknown) => void, reject: (rea
...
+      );
+    }, 5_000); // 5 seconds; this is arbitrary
+
+    let loadFn: () => void, errFn: (ev: ErrorEvent) => void, finishScriptLoading: () => void;
+
+    finishScriptLoading = () => {
```
ibacher, line 208: Our linter doesn't like inline functions, but is ok with lamba-style inline functions

ibacher:
WTH?

This saved 50% of the size of the framework????

denniskigen:
> WTH?
> 
> This saved 50% of the size of the framework????

Really?

packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
denniskigen: Reads fine to me

denniskigen (approved):
Approach LTGM! Thanks, @ibacher!

ibacher:
FWIW, I did actually test this out in ways that ensured this works as expected. To do this, I added a dummy line to the app shell like this:

```ts
importDynamic("@openmrs/esm-fake-app", undefined, {
  importMap: {
    imports: {
      "@openmrs/esm-fake-app": "https://httpstat.us/504?sleep=300000" // URL that won't resolve for a while
    }
  },
  maxLoadingTime: 10, // pretty much guarantee to _not_ be long enough
})
```

PR #946, opened 2024-02-29
(feat) O3-2261:  Add back button when changing location

usamaidrsk:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR adds a back button on change location screen, for a user to go back to previous page if they want to cancel the operation

## Screenshots

https://github.com/openmrs/openmrs-esm-core/assets/53287480/c7391785-fe02-459d-8c1d-566a5bb8f82e

## Related Issue

[O3-2261](https://openmrs.atlassian.net/browse/O3-2261)


denniskigen (changes requested):

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -166,6 +170,19 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
   return (
     <div className={styles.locationPickerContainer}>
       <form onSubmit={handleSubmit}>
+        {isUpdateFlow ? (
+          <div className={styles['back-button-div']}>
+            <Button
+              className={styles['back-button']}
+              iconDescription="Back to username"
```
denniskigen, line 177: This should be translated.

denniskigen (approved):
Nicely done, @usamaidrsk. LGTM.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -75,30 +76,33 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
 
+  const handleNavigation = useCallback(() => {
+    const referrer = state?.referrer;
+    const returnToUrl = new URLSearchParams(location?.search).get('returnToUrl');
+
+    if (referrer && !['/', '/login', '/login/location'].includes(referrer)) {
+      navigate({ to: '${openmrsSpaBase}' + referrer });
+      return;
+    }
+    if (returnToUrl && returnToUrl !== '/') {
```
ibacher, line 87: Even though this sort of is implied, I always think it's better to be explicit about this. This also allows us to skip the `return` on line 85.

```suggestion
    } else if (returnToUrl && returnToUrl !== '/') {
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -75,30 +76,33 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
     (locationUuid?: string, saveUserPreference?: boolean) => {
       setIsSubmitting(true);
-
-      const referrer = state?.referrer;
-      const returnToUrl = new URLSearchParams(location?.search).get('returnToUrl');
-
       const sessionDefined = locationUuid ? setSessionLocation(locationUuid, new AbortController()) : Promise.resolve();
 
       updateDefaultLocation(locationUuid, saveUserPreference);
       sessionDefined.then(() => {
```
ibacher, line 100:
```suggestion
      sessionDefined.then(handleNavigation);
```

packages/apps/esm-login-app/src/root.scss
```diff
@@ -29,3 +29,17 @@
 .caption01 {
   @include type.type-style('legal-01');
 }
+
+.back-button {
```
ibacher, line 33: Are these actually applied? It seems like the point where they are intended to be used `location-picker.component.tsx`, `styles` refers to `location-picker.styles.scss` and not this file.

ibacher (changes requested):
Thanks @usamaidrsk. Couple of small things, but I think this basically looks good.

mseaton (commented):
Functionally this looks good in that it provides what is missing.  Is this really the UI guideline that carbon advocates for this?  I have the same question about the existing "Back" link on the password page that takes you back to the login page, with this odd link at the top of the url for this.

I would have thought things would look at lot more consistent to do this via a button alongside the main save button - eg. a Cancel button that would serve to take you back to the previous step.

For example, we have this pattern on the registration form:

image

We have this pattern on the cancel dialog:

image

Aside from these clear inconsistencies, it does seem like a styled button is a better UI than an unstyled link thrown at the top of the form.

Anyone else feel this way?

packages/apps/esm-login-app/src/root.scss
usamaidrsk: Yes they're used, as the `root` is importend inside `location-picker.styles.scss` as well as the `login.scss`

usamaidrsk:
https://github.com/openmrs/openmrs-esm-core/assets/53287480/a8c40c8f-3712-4f08-a056-995315527851

packages/apps/esm-login-app/src/root.scss
ibacher: Oh... that's a little silly. In either case, what I'm not really getting is why this should be moved here if we're _only_ using this class in one component _and_ that component has it's own style sheet?

rbuisson:
I see what @mseaton  means here.

I think the `<- Back` button works well when you're within a multi-step flow of a "single page" (where screens come directly after one another though you don't feel you're moving away). Which is the case for the login page.

However, in the Location picker here, when opened outside of the login flow, the Back button is less intuitive. We're just leaving the home page to open a new 'full size' page. Therefore, a 'Discard/Confirm' button might make more sense.

@usamaidrsk , would you be able to go through both the OpenMRS style guide and designs to find examples or guidelines, as well as Carbon official documentation?

usamaidrsk:
@rbuisson I do get it now.
I think I can as well consult from [ux-design-advisory](https://openmrs.slack.com/archives/CKS32D55G/p1709803687480519)

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -166,6 +165,19 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+            <Button
+              className={styles['back-button']}
+              iconDescription={t('backToPreviousPage', 'Back to previous page')}
+              kind="ghost"
+              onClick={handleNavigation}
+              renderIcon={(props) => <ArrowLeft size={24} {...props} />}
+            >
+              <span>{t('back', 'Back')}</span>
+            </Button>
+          </div>
```
raghavawasthi24, line 179: The `div` wrapping `Button` is unnecessary as you too can apply css directly to` Button`.

raghavawasthi24 (changes requested):

packages/apps/esm-login-app/src/root.scss
vasharma05: Hi @usamaidrsk!
The class names should follow camelCasing.
Thanks!

vasharma05:
Hi @mseaton @rbuisson,
I see @usamaidrsk didn't update the PR, we had a connect with Paul and he shared the new designs for the location picker (ticket [O3-2951](https://openmrs.atlassian.net/browse/O3-2951)). Since the new implementation might take time, can we move ahead with the changes from @usamaidrsk or wait for the new location picker implementation only?
Thanks!

rbuisson:
@vasharma05 @usamaidrsk  , from my message on Slack, I still suggest:
>In that case, if we are already differentiating the 2 cases of (1) location selection right after login VS (2) location selection from the home page, then I would suggest to replace the Back button with a standard discard/confirm button for case 2.

So let's just change to a standard button and merge that, while we can work on the new location picker later.

usamaidrsk:
> So let's just change to a standard button and merge that, while we can work on the new location picker later.

@rbuisson @vasharma05 @denniskigen I have updated this PR as per the suggestions.
Thanks

rbuisson:
@usamaidrsk , you did put Confirm and Cancel one under the other, as opposed to one next the other.
Is this taken from an existing similar design?

Carbon [suggests the latter for modals](https://carbondesignsystem.com/components/modal/usage/) (I know it's not exactly a modal here, but it looks like one)

vasharma05 (approved):
A small nit, but all good to go.
Thanks @usamaidrsk!

vasharma05:
Hi @ibacher!
Can you please re-review this PR?
Thanks!

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -23,6 +23,7 @@ import { useLoginLocations } from '../login.resource';
 import styles from './location-picker.scss';
 import { useDefaultLocation } from './location-picker.resource';
 import type { ConfigSchema } from '../config-schema';
+import { ArrowLeft } from '@carbon/react/icons';
```
ibacher, line 26: Please import the ArrowLeft component from `@openmrs/esm-framework`

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -82,30 +83,28 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+    const returnToUrl = new URLSearchParams(location?.search).get('returnToUrl');
+
+    if (referrer && !['/', '/login', '/login/location'].includes(referrer)) {
+      navigate({ to: '${openmrsSpaBase}' + referrer });
+    } else if (returnToUrl && returnToUrl !== '/') {
+      navigate({ to: returnToUrl });
+    } else {
+      navigate({ to: config.links.loginSuccess });
+    }
+  }, [state?.referrer, location, config.links.loginSuccess]);
```
ibacher, line 97:
```suggestion
  }, [state?.referrer, location?.search, config.links.loginSuccess]);
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -249,13 +248,26 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
               ) : (
-                <span>{getCoreTranslation('confirm')}</span>
+                null
               )}
-            </Button>
+              <Button kind="primary" type="submit" disabled={!activeLocation || !isLoginEnabled || isSubmitting}>
+                {isSubmitting ? (
+                  <InlineLoading className={styles.loader} description={t('submitting', 'Submitting')} />
+                ) : (
+                  <span>{t('confirm', 'Confirm')}</span>
```
ibacher, line 267: Use `getCoreTranslations()` in preference to app-local translations here and for the `Cancel` button.

ibacher: @usamaidrsk ^^

denniskigen:
Good to go, @ibacher?

PR #947, opened 2024-03-01
(feat) O3-2724: Added `showWorkspace` API to display overlays

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR adds a new API inside the **@openmrs/esm-framework** to display overlays in the app.
Added a new `omrs--overlay--container` inside the `index.ejs`.

### registerOverlay
We can register an overlay by calling the `registerOverlay` with the following structure inside the ESM's `index.ts` under `startupApp`.

```
import { registerOverlay } from '@openmrs/esm-framework';

export function startupApp() {
  registerOverlay({
    name: string,
    title: string,
    load: getAsyncLifecycle(() => import('path-to-component-to-be-rendered'), options),
  });
}
```

### showOverlay
Params:
1. name: string
2. props : Object
2.a. overlayTitle: string // To override the overlay title defined in the overlay registration.

```
showOverlay('overlay-name')
```

## Screenshots
For demo purpose, I have created a sample overlay content.

### OverlayContent
```
import React from 'react';

const OverlayContent = () => {
  return <h1>Hello World!</h1>;
};

export default OverlayContent;
```

### index.ts
```
export function startupApp() {
  registerOverlay({
    name: 'overlay',
    load: getAsyncLifecycle(() => import('./components/overlay/overlay.component'), options),
    title: 'Overlay title',
  });
}
```

### Overlay launching button
```
import { Button } from '@carbon/react';
import { showOverlay } from '@openmrs/esm-framework';
import React from 'react';

const OverlayLauncher = () => {
  const launchOverlay = () => {
    showOverlay('overlay');
  };
  return <Button onClick={launchOverlay}>Launch Overlay</Button>;
};

export default OverlayLauncher;
```

https://github.com/openmrs/openmrs-esm-core/assets/51502471/82129151-029a-4a03-ba98-6007ea3d9342

## Related Issue
https://openmrs.atlassian.net/browse/O3-2724


vasharma05:
Requesting @chibongho @brandones @ibacher @denniskigen for review here.
Thanks!

vasharma05:
We will require a similar implementation of the overlays as that of workspaces. I am thinking of creating a common implementation with both: overlays and workspaces, which can be utilized. I believe we can take that in another iteration.
Thanks!

ibacher:
Thanks @vasharma05! I'm not sure that this pattern (basically the same thing that we're using for notifications) is the right pattern to use for this. Specifically, as an end goal, registrations of overlays / workspaces should be done in the apps routes.json file rather than via an explicit call to `registerOverlay()`. This gives us the opportunity to actually load the relevant scripts and configuration before an overlay is displayed (which is a flaw in the current workspace system).

Basically, I think we need to do this the same way the shell handles extensions.

ibacher (changes requested):
See earlier comment.

vasharma05:
Hi @ibacher!
I wanted to know where is the translations for the strings in the esm-styleguide be registered to which namespace?

For reference, the translations in the patient-common-lib are mapped to esm-patient-chart-app.
Thanks!

ibacher:
@vasharma05 For right now, we don't have a means of supporting translations in the styleguide itself. Translations have to be provided either directly as props or by the rendering app providing a `t` function with access to the necessary keys.

packages/framework/esm-styleguide/src/workspaces/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  type CloseWorkspaceOptions,
+  closeWorkspace,
+  promptBeforeClosing,
+  type WorkspaceInstance,
+  getPromptBeforeClosingFcn,
+  type Prompt,
+} from './workspaces.resource';
+import { translateFrom } from '@openmrs/esm-framework';
+
+export function renderWorkspaces(workspaceContainer: HTMLElement | null) {
```
brandones, line 16: Just chiming in that this will definitely need a robust docstring explaining how and why to use this function. All public functions in core should have docstrings.

packages/framework/esm-styleguide/src/workspaces/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  if (workspaceContainer) {
+    const root = createRoot(workspaceContainer);
```
brandones, line 18: Will be interesting to see if this works without memory leaks or anything like that. I would normally expect a Single-SPA parcel here. It will be nice if this just works though.

brandones:
I'm doing this as part of a larger refactor of the workspace system, building off the good work Vineet has done here.

PR #952, opened 2024-03-06
(feat) Allow user to reset the importmap and reload if overridden script fails to load

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

Displays an error toast if we fail to load a script that is overridden on the import map with the opportunity to reset and overload it.

## Screenshots

Screenshot of example notification

## Other

Future work:

* This should be translatable, but because we don't have a good pattern for translating things in the framework, I've left that aside for now.
* Currently the `unhandledrejection` handler is getting called when this fires as well. This needs a little care, because we fundamentally do want most of the promises to reject. The issue is that [this line](https://github.com/openmrs/openmrs-esm-core/blob/25ca4080f3b0963c8b5ea7d3bc8e5528ea8fa71c/packages/shell/esm-app-shell/src/run.ts#L201) isn't doing anything. In any case, the additional toast is a somewhat separate issue from this PR.
* If you've overridden a lot of modules, you will get one error-per-module. It would probably be best to simplify this case so only a single error gets displayed, but that probably requires working out the above point.

### DIFF
--- a/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
+++ b/packages/framework/esm-dynamic-loading/src/dynamic-loading.ts
@@ -1,7 +1,7 @@
 /** @module @category Dynamic Loading */
 'use strict';
 // hack to make the types defined in esm-globals available here
-import type { ImportMap } from '@openmrs/esm-globals';
+import { dispatchToastShown, type ImportMap } from '@openmrs/esm-globals';
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


ibacher:
@denniskigen I believe this is something like the feature we discussed on Slack the other day?

denniskigen (approved):
Thanks, @ibacher!

PR #954, opened 2024-03-12
(feat) O3-2262 & O3-2263: Record previously logged in locations of the users and show them at the top of the locations' list

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR now saves the last 10 logged in locations of the users in the userProperties under the key `previousLoggedInLocations` as a string joined by `,`. The previous logged in locations are then shown at the top of the list before the rest of the locations.

The order of the locations in the list is:
1. Default location
2. Previously logged in locations
3. Rest of the locations

```
# To get a list of previous logged in locations' UUIDs
user?.userProperties?.previousLoggedInLocations?.split(',') ?? []

#To save the location UUIDs:
const userProperties = {
	previousLoggedInLocations: previousLoggedInLocations.join(',')
}
```

## Screenshots
Uploading soon.

## Related Issue
https://openmrs.atlassian.net/browse/O3-2262
https://openmrs.atlassian.net/browse/O3-2263


packages/apps/esm-login-app/src/login.resource.ts
```diff
@@ -1,121 +1,17 @@
...
-    }
-
-    if (page) {
-      urlSearchParameters.append('_getpagesoffset', '' + page * count);
-    }
-
-    if (useLoginLocationTag) {
-      urlSearchParameters.append('_tag', 'Login Location');
-    }
+export async function performLogin(username: string, password: string): Promise<{ data: Session }> {
```
ibacher, line 3: This whole function can be replaced with a call to `refetchCurrentUser()`, so do that instead.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -11,38 +11,42 @@ import {
...
-    useDefaultLocation(isUpdateFlow);
-
-  const [searchTerm, setSearchTerm] = useState(null);
+  const { state } = useLocation() as unknown as Omit<Location, 'state'> & {
+    state: LoginReferrer;
+  };
 
   const { user, sessionLocation } = useSession();
-  const { currentUser, userProperties } = useMemo(
+  const { currentUser } = useMemo(
```
ibacher, line 49:
```suggestion
  const currentUser = user?.display;
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -52,35 +56,58 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
-  const locations = useMemo(() => {
-    if (!defaultLocationFhir?.length || !fetchedLocations) {
-      return fetchedLocations;
-    }
-    return [
-      ...(defaultLocationFhir ?? []),
-      ...fetchedLocations?.filter(({ resource }) => resource.id !== defaultLocationFhir?.[0].resource.id),
-    ];
-  }, [defaultLocationFhir, fetchedLocations]);
+  const updateUserProperties = useCallback(
```
ibacher, line 69: This function could use a better name.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -52,35 +56,58 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+    (locationUuid: string) => {
+      if (!locationUuid) return;
```
ibacher, line 71:
```suggestion
      if (!locationUuid) {
        return;
      }
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -52,35 +56,58 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+        user?.userProperties?.previousLoggedInLocations,
+      );
 
-  const [activeLocation, setActiveLocation] = useState(() => {
-    if (currentLocationUuid && hideWelcomeMessage) {
-      return currentLocationUuid;
-    }
-    return sessionLocation?.uuid ?? defaultLocation;
-  });
+      const prevUserProperties = user?.userProperties;
```
ibacher, line 77: Seems like this could be moved up and then the previous references to `user?.userProperties` changed to use this variable.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -52,35 +56,58 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
 
-  const [isSubmitting, setIsSubmitting] = useState(false);
+      const updatedUserProperties = {
+        ...(user?.userProperties ?? {}),
```
ibacher, line 80:
```suggestion
        ...(prevUserProperties ?? {}),
```

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -105,49 +131,58 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
+  }, [lastLoggedInLocation, activeLocation, setActiveLocation]);
+
+  useEffect(() => {
+    if (isDefaultLocationValid) {
+      setSaveDefaultLocation(true);
+    }
+  }, [setSaveDefaultLocation, isDefaultLocationValid]);
+
+  useEffect(() => {
+    // Handle cases where the login location is present in the userProperties.
```
ibacher, line 162: What circumstance is this necessary in?

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
```diff
@@ -105,49 +131,58 @@ const LocationPicker: React.FC<LocationPickerProps> = ({ hideWelcomeMessage, cur
...
-
-      changeLocation(activeLocation, savePreference);
+      changeLocation(activeLocation, saveDefaultLocation);
     },
-    [activeLocation, changeLocation, savePreference],
+    [activeLocation, changeLocation, saveDefaultLocation],
   );
 
-  // Infinite scroll
+  const handleFetchNextSet = useCallback(() => {
```
ibacher, line 182: What's the value in adding this as a separate `useCallback()` hook?

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -1,83 +1,16 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
-import { useTranslation } from 'react-i18next';
-import { setUserProperties, showSnackbar, useSession } from '@openmrs/esm-framework';
-import { useValidateLocationUuid } from '../login.resource';
+import { type LoggedInUser } from '@openmrs/esm-framework';
```
ibacher, line 1: Is this used?

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -1,83 +1,16 @@
...
+  return updatedLoggedInLocations.join(',');
+}
 
-  return {
-    defaultLocationFhir,
-    defaultLocation: isLocationValid ? defaultLocation : null,
-    updateDefaultLocation,
-    savePreference,
-    setSavePreference,
+export function getUserPropertiesWithDefaultAndLogInLocation(locationUuid: string, previousLoggedInLocations: string) {
```
ibacher, line 9: This is also badly named, since it doesn't actually return the user properties in any meaningful sense. Maybe `getDefaultLocationUserProperties()`or something.

packages/framework/esm-react-utils/src/useLocations.ts
```diff
@@ -0,0 +1,236 @@
...
+      defaultLocationArr: isDefaultLocationValid ? data?.data?.entry : [],
+      error,
+      isLoadingDefaultLocation: isLoading,
+    }),
+    [isDefaultLocationValid, data?.data?.entry, error, isLoading],
+  );
+  return results;
+}
+
+export function useLocations(useLoginLocationTag: boolean, count: number = 0, searchQuery: string = '') {
```
ibacher, line 188: Since this only pertains to login locations. I'm also unclear on why we need this in the framework itself and not as part of the login app, which is where I think it belongs (there are other uses to locations; these should not be default restricted to login locations and they should not have the same logic that we're using here to surface locations; in other cases where we need locations, it makes sense to surface the location the user is logged into, but not the _previous_ locations they have logged into).

```suggestion
export function useLoginLocations(useLoginLocationTag: boolean, count: number = 0, searchQuery: string = '') {
```

ibacher (dismissed):
Some initial comments. Nice work!

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
vasharma05: If there is a default login location present in user properties

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
vasharma05: Yes @ibacher , it is used on line 11 to define the user properties types

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
```diff
@@ -1,83 +1,16 @@
...
-            subtitle: t('locationPreferenceRemovedMessage', 'You will need to select a location on each login'),
-            kind: 'success',
-            isLowContrast: true,
-          });
-        }
-      });
-    },
-    [savePreference, defaultLocation, updateUserPropsWithDefaultLocation, t, isUpdateFlow],
-  );
+export function getUpdatedLoggedInLocations(previousLoggedInLocations: string, locationUuid: string): string {
```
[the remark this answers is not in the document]
vasharma05: Hi @ibacher!
Actually this function stores all the logged in locations joined with `,`

packages/framework/esm-react-utils/src/useLocations.ts
vasharma05: > it makes sense to surface the location the user is logged into, but not the previous locations they have logged into

I actually thought that this case you mentioned above would have to handled in other places, say location dropdown somewhere else in UI, to use both default(logged in) and previous logged in locations to show at the top, hence using the same hook everywhere else rather than creating this again and again.

Should I move this from framework to login app again?

Also, we can pass the `useLoginLocationTag` as true/false, hence I believe `useLocations` is a better name for handling both the cases. Please confirm on this as well.
Thanks!

packages/tooling/openmrs/src/cli.ts
```diff
@@ -233,7 +233,7 @@ yargs.command(
       }),
   async (args) =>
     runCommand('runBuild', {
-      configUrls: args['config-url'],
+      configUrls: args['config-url'] as Array<string>,
```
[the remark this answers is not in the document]
vasharma05: Actually I was getting a TS error.

packages/framework/esm-react-utils/src/useLocations.ts
ibacher: So a couple of things here:

1. Locations have an entire tagging system, which are intended for a sort of "folksonomy" of locations. That is, it might make sense (for other applications) to tag all pharmacy locations as "Pharmacy" and then, say, in the dispensing app, load those "Pharmacy" locations as the list of valid locations a user can dispense from. The upshot of this, I think, is that a framework-wide `useLocations()` hook would take an arbitrary `string` for tags rather than a boolean to indicate whether to use a single tag or not.
2. I'm not convinced that previous logged in locations are the right thing to display for anything other than previous logged-in locations. For example, if I am logged into Clinic 1 which has the sub-locations Room A, Room B, and Room C, it might make more sense for a form (like say the start visit form) to present me first with Room A, Room B, and Room C, since those are the sub-locations of the place I am currently working.

Both of these kind of speak against this being a useful enough function to include in the framework, especially with the name `useLocations()`.

ibacher: What I'm getting at isn't that we shouldn't have a `useLocations()` hook in the framework. It's that that hook should have different properties from the hook that is specific to login locations. It would be possible to implement a `useLoginLocations()` hook in terms of a generic `useLocations()` hook, but implementing a generic `useLocations()` hook cannot be easily done from a hook that implements logic that mostly makes sense in the context of login locations.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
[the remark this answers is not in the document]
ibacher: I get what the function does. My point was that the name here is both quite long and not very descriptive. `getDefaultLoginLocations()` is based on my suggestion below to renamed `getUserPropertiesWithDefaultAndLogInLocation()` to `getDefaultLocationUserProperties()`.

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
ibacher: Right, but the backend core handles that situation already or at least it should, so it shouldn't be necessary to do so here.

packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts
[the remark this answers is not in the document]
ibacher: Also, shouldn't this function ensure the invariant that no more than 10 locations are stored in the `updatedLoggedInLocations()` (less a concern in newer versions of the backend where user properties can be longer, but there's always a maximum length and truncation will almost certainly lead to invalid UUIDs in the list).

packages/apps/esm-login-app/src/location-picker/location-picker.component.tsx
vasharma05: Right!
Let me test this again.
Thanks!

vasharma05:
Hi @brandones !
Awaiting @ibacher 's review, pushed prev review changes.
Thanks!

ibacher:
@vasharma05 Rename `useLocations()` to `useLoginLocations()` and I'm happy.

brandones:
@vasharma05 Thanks for making the requested changes. Can you resolve the conflicts and we'll get this in?

brandones:
Ping @vasharma05 , it's so close, just have to resolve the conflicts...

gracepotma:
@vasharma05 any reason this is still open?

vasharma05:
Hi Grace!
The work done here was done a long time ago, and I didn't get time to see if it's still working the same or not with the recent changes. I'll recommend closing this PR and then we can go ahead and open a new one referencing this PR.
Thanks!

PR #962, opened 2024-03-28
(feat) O3-3010: Disable triggering offline mode when offline disabled

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

As reported on the squad call today, on a local machine with no internet connectivity, we end up triggering offline mode, which is unfortunate, since we do need to support single-machine installs with no internet connectivity.

This PR probably doesn't fully address the issue; however, what I've done is to move the event handlers that respond to connectivity changes so that they are only registered if offline mode is enabled. This should enable installations on a single machine with no network connectivity provided that offline mode is disabled in the build configuration, which is the default. Since such installations without LAN or internet connectivity have no real reason to use offline mode in any fashion, this seems to be a way of addressing the issue with the minimal amount of change.

## Related Issue

https://issues.openmrs.org/browse/O3-3010

## Other

A more thorough-going process would probably use something like a websocket to subscribe to the OpenMRS backend, which could be used to detect when the backend is down. Such a solution, however, requires additions to the backend as well as frontend and needs to be carefully considered, since it may cause some issues for clinics with many devices connecting to a server and a relatively small server instance.

### DIFF
--- a/packages/shell/esm-app-shell/src/run.ts
+++ b/packages/shell/esm-app-shell/src/run.ts
@@ -186,8 +186,6 @@ function connectivityChanged() {
  * Runs the shell by importing the translations and starting single SPA.
  */
 function runShell() {
-  window.addEventListener('offline', connectivityChanged);
-  window.addEventListener('online', connectivityChanged);
   return setupI18n()
     .catch((err) => console.error(`Failed to initialize translations`, err))
     .then(() => start());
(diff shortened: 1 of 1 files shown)


ibacher:
cc: @slubwama

denniskigen (approved):
Thanks, @ibacher. Let's see how this goes.

denniskigen:
Do you know how to fix the failing docs workflow? Something related to esm-translations...

ibacher:
@denniskigen I think we just need to add an exclusion for `package/` links [here](https://github.com/openmrs/openmrs-esm-core/blob/c94ac5b7c08720a64ef91410b2c0513e4f90b57a/markdown_link_check_config.json#L10). They work on GitHub (I think), but not in any other context.

PR #976, opened 2024-04-18
(feat) Create an application-wide context for shared state

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This is currently a draft and more of a sketch of an idea than an actual implementation.

In essence, this actually ends up restoring something like the `AppState` that we used to have but did not use: a global state called the "context". The idea here is that contextually-important shared state can be pushed here and accessed from here. So, for example, the current session data could be stored and periodically refreshed or, for example, the current patient in a chart context (or current visit) could be stored here and hooks could defer to using it.

There are a few differences between this iteration and the `AppState`:

1. All data is bound to a namespace, which effectively functions as a selector (with the ability to add sub-selectors)
2. `useAppContext()` basically returns read-only copies of the data. The idea here is that one component defines and maintains the context and other components can leverage it, but not update it.
3. `useDefineAppContextNamespace` and the `OpenmrsAppContext` React component will clear the managed namespace from the context when the managing component is removed from the vDOM.

The main usage of this from a client standpoint looks something like this, having one component that creates and provides the state and then any number of consumers.

In the provider, you might have something like:

```ts
const patient = usePatient();
useDefineAppContextNamespace<PatientContext>('patient', patient);
```

And then in the consumer, something like:

```ts
const { patientUuid } = useAppContext<PatientContext>(('patient');
```

Right now, this is pretty useless, but the goal is to enable some level of app-level caching beyond just relying on SWR de-duplicating requests.

## Other

This setup may be too Redux-like to be a good idea in practice. While it doesn't explicitly require the use of reducers or state transitions, it does try to share a similar single point of responsibility for data updating, albeit in a much cruder way.


brandones:
Cool idea, thanks for this. Right now it seems like we generally solve this kind of problem using hooks like `usePatient` and relying on the SWR cache. What problem is this addressing, or what improvement would this allow?

ibacher:
> What problem is this addressing, or what improvement would this allow?

So it's aimed at addressing of few things:

1. The shared SWR cache is not actually de-duplicating things as much as we would like. The Patient Chart on dev3 makes multiple requests to the `/fhir2/R4/Patient/<uuid>` endpoint (I count 11 on initial load). While it's true that SWR will ensure that the component is rendered with the first response and so the refetches don't _necessarily_ mess with the rendering of the chart components, it's still got the potential to make the app feel slower, especially on, e.g., low-powered tablets. (Technically, this is SWR performing per-spec, since SWR really just means "returns the stale data, makes the request and then updates stale data with the new data"). Obviously cache invalidation is a very hard problem, but currently we don't even have an accessible cache to invalidate.
2. I'd like to use this to replace a few of the stores that we're using, like the visit store, since that's both contextual data and the context in which the current visit is valid is a little narrower than a global store (a current visit makes the most sense in the scope of the patient chart, but less sense for most of the patient management, home page, and admin tools stuff). The React components added here are designed to enable temporarily setting values in the store while the parent component is active, giving us better control over the lifecycle of that contextual data. I think a similar structure is also useful for working with the "current encounter" concept we need to add to the chart.
3. The real driver here is enabling context-specific dynamic rules, especially for things like extension rendering, so that we can support things like "only render the pregnancy dashboard if the patient is capable of being pregnant" via a config-driven approach (basically, optionally rendering extensions based on JS expressions). The idea here is to be able to have a feature similar to O2's [`require`](https://github.com/openmrs/openmrs-module-coreapps/blob/1bfb07b8a8712e8c0c60ab1f91db472356a79042/omod/src/main/resources/apps/overallActions_extension.json#L10) expressions.
3b. Loading the context into a framework-agnostic store allows us to operate on the state _outside_ of React, which is desirable for use-cases like 3, and also for reworking parts of the framework that operate outside React to interact with that state.

None of the 3 things I outlined there are specifically solved by this PR, but it's intended as a building block to solve them.

brandones:
Love it. Those are great reasons. This PR seems like a good way to start; I look forward to seeing how this evolves.

As a side note, when I was first designing the extension system, I imagined that extensions would receive props based on the context. So `patientUuid`, `visitUuid`, etc. would be passed automatically as props, and extensions would even make explicit what kind of context they required, so they could only be attached in a valid context. Maybe that's still "not gonna happen," but this at least lays the technical foundation for it.

> The shared SWR cache is not actually de-duplicating things as much as we would like. The Patient Chart on dev3 makes multiple requests to the /fhir2/R4/Patient/<uuid> endpoint (I count 11 on initial load).

!!! Wow I had no idea, that's quite bad

packages/framework/esm-context/README.md
```diff
@@ -0,0 +1,4 @@
+# openmrs-esm-dynamic-loading
+
+openmrs-esm-dynamic-loading provides functions for dynamically loading JS files using Webpack Module Federation
+dynamic remotes.
```
denniskigen, line 4: Shouldn't this be stuff about esm-context?

ibacher:
Alright, I'm now happy this works as intended, albeit it's pretty minimally functional. The only current difference between this and a global store is that the value should only ever be a read-only copy of the value.

denniskigen (approved):
LGTM, @ibacher! Thanks.

FelixKiprotich350:
hello @ibacher in light of this implementation, just on the laboratory, i realize we are making two calls to get the metrics and workload items.I thought it could be better we make a one call from the child component which has filters , fetch functions  and displays the items then use something like a state to store the metrics (_counts maybe by status_) which can then be displayed on top of the tabs or alongside every tab. 
In regards, i have tried to do a useDefineContext which can then be shared between the components but seems it has conflict with the laboratory-date-filter context or rather keeps triggering refresh continuously. 
@ibacher Would you advise how i can go about this?. I feel its more efficient and reduces the work by just making one call and sharing the metrics instead of pulling them separately.
Thanks

FelixKiprotich350:
> Sharing metrics seems like the right approach. Without seeing code, it's hard to be concrete, but ideally, the `useDefineAppContext()` call is in some component that's a parent of any component displaying the metrics, which might help with re-renders. 

Sure, using this idea will it cause any glitch if i do two useDefineAppContext in the same parent component....or rather does useDefineAppContext allow more than two properties?

ibacher:
> Sure, using this idea will it cause any glitch if i do two useDefineAppContext in the same parent component....or rather does useDefineAppContext allow more than two properties?

As long as you use separate namespaces, no glitch. `useDefineAppContext()` can be used to associate a namespace with an object, though, which would give you the multiple properties.

FelixKiprotich350:
> > Sure, using this idea will it cause any glitch if i do two useDefineAppContext in the same parent component....or rather does useDefineAppContext allow more than two properties?
> 
> As long as you use separate namespaces, no glitch. `useDefineAppContext()` can be used to associate a namespace with an object, though, which would give you the multiple properties. 
> 

Thank you....i am giving it a try .... just to see if we can reduce the number of requests made to make the app a bit lighter

PR #978, opened 2024-04-18
(BREAKING) O3-3191: Move LeftNav into primary navigation app

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This extracts the LeftNav component from the styleguide into the primary navigation app, with the idea being that we have a single left nav that changes contents as the user browses.

This means that the API functions (now migrated to esm-state instead of esm-styleguide) `setLeftNav()` and `unsetLeftNav()` are quite important. Essentially a page opts into using the LeftNav by calling `setLeftNav()` to declare the slot. The LeftNav is only shown when there is at least one extension bound to the declared slot. So although the left nav is now a single component, it can contextually change which links are shown based on the extension slot name.

NB This _also_ means that the hamburger menu is only shown if there are any extensions mounted in the nav.

TODO: Figure out when it makes sense to fire an automated call to `unsetLeftNav()` this probably needs to be done in the page `mount()` call. I'll play with that.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3191

## Other

PRs are coming for Home and Patient Chart. Patient Chart, in particular, will require some work to make this happen.

Also, if you test out this PR right now, you'll probably see no real changes (other than a weird rendering of an empty bar on the home page); it does appear to work, but the left nav items in Home and Chart are currently rendered over the nav bar.

Related PRs: https://github.com/openmrs/openmrs-contrib-json-schemas/pull/10


packages/apps/esm-primary-navigation-app/src/components/left-nav/index.tsx
```diff
@@ -1,31 +1,10 @@
 /** @module @category UI */
 import React from 'react';
-import { ExtensionSlot, useStore } from '@openmrs/esm-react-utils';
-import { createGlobalStore } from '@openmrs/esm-state';
+import { ExtensionSlot, useStore, leftNavStore } from '@openmrs/esm-framework/src/internal';
```
brandones, line 3: I've run into very esoteric problems with this. What happened in my case was that it seemed that Jest would come across this line, load the `internal` library, which loads the entire `styleguide/index`, which at some point loads some SVGs, and Jest sees the SVGs and crashes with syntax errors. Importing from the sub-libraries directly avoids that problem.

package.json
```diff
@@ -69,7 +69,7 @@
     "webpack": "^5.88.0"
   },
   "lint-staged": {
-    "*.{ts,tsx}": "eslint --cache --fix --max-warnings 0",
+    "!(*test).{ts,tsx}": "eslint --cache --fix --max-warnings 0",
```
brandones, line 72: In https://github.com/openmrs/openmrs-esm-core/pull/979 I make it so eslint runs on the tests (I don't know why it shouldn't). Feel free to leave this in if you think this will be in before https://github.com/openmrs/openmrs-esm-core/pull/979.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -20,14 +22,16 @@ import UserMenuPanel from '../navbar-header-panels/user-menu-panel.component';
...
 
 const HeaderItems: React.FC = () => {
   const { t } = useTranslation();
   const config = useConfig();
   const [activeHeaderPanel, setActiveHeaderPanel] = useState<string>(null);
   const layout = useLayoutType();
-  const navMenuItems = useConnectedExtensions('patient-chart-dashboard-slot').map((e) => e.id);
   const appMenuItems = useConnectedExtensions('app-menu-slot');
+  const { slotName: leftNavSlot } = useStore(leftNavStore);
+  const leftNavItems = useConnectedExtensions(leftNavSlot);
```
brandones, line 34: Another pattern to consider is one I introduce in https://github.com/openmrs/openmrs-esm-core/pull/979: using the `featureName` of the root component for the slot name. So like the feature name for patient chart is `patient-chart`. If we add `featureName` to `ComponentContext` then you can just get that context and make the slot name like `left-nav-${featureName}-slot`.

I have no opinion on whether that pattern is better or worse for this use case, I just want to put it out there that it's an option.

ibacher: Actually, I'm going to play around with that. Right now, the major problem with this PR is the need to manually call `setLeftNav()` and `unsetLeftNav()`, and this would give us a clean way of having the left nav just "magically" supported.

brandones: I love that after like five years of extremely marginal existence `featureName` is suddenly becoming very useful

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -42,7 +46,8 @@ const HeaderItems: React.FC = () => {
     [],
   );
 
-  const showHamburger = useMemo(() => !isDesktop(layout) && navMenuItems.length > 0, [navMenuItems.length, layout]);
+  const showLeftNav = useMemo(() => isDesktop(layout) && leftNavItems.length > 0, [leftNavItems.length, layout]);
+  const showHamburger = useMemo(() => !isDesktop(layout) && leftNavItems.length > 0, [leftNavItems.length, layout]);
```
denniskigen, line 50:
```suggestion
  const showHamburger = useMemo(() => !isDesktop(layout) && leftNavItems.length > 0, [leftNavItems.length, layout]);
```

nit: `showHamburgerMenu` reads better here.

packages/apps/esm-primary-navigation-app/src/components/left-nav/index.tsx
```diff
@@ -1,31 +1,10 @@
...
-
-export function unsetLeftNav(name) {
-  if (leftNavStore.getState().slotName == name) {
-    leftNavStore.setState({ slotName: null });
-  }
-}
-
 type LeftNavMenuProps = SideNavProps;
 
 export const LeftNavMenu = React.forwardRef<HTMLElement, LeftNavMenuProps>((props, ref) => {
```
denniskigen, line 10: Is it possible to use the `SideNavProps` annotation directly instead of aliasing it to `LeftNavMenuProps`? Would that obviate the need to define it every [ambient type declaration file](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/apps/esm-login-app/src/declarations.d.ts#L6C14-L6C26)?

brandones: I don't think it makes a difference whether there's an alias or not, but it makes the intention a little clearer—the `forwardRef` call becomes a little more idiomatic with the alias. The ambient type declaration file is unfortunately necessary because of Carbon's poor type support, alias or no alias doesn't matter.

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
mseaton: @ibacher It would be great if this also supported some sort of default-left-nav-slot, which would render if and only if no specific left nav is set.  This would be different from the existing global nav slot, which always renders.  This way, implementations could basically have a left nave that is consistent across all routes in the application, which is then replaced for specific routes (eg. patient chart).

ibacher:
So, a question on this:

I was looking into the `featureName` thing @brandones suggested and it's not quite workable here because there's no guaranteed extension to pull from. That said, I came up with something similar which just uses the active page name to determine the default slot name. This matches what we already do in the chart (`patient-chart-dashboard-slot`) and, with a slight adjustment, the home page (`home-dashboard-slot` instead of `homepage-dashboard-slot`). I like this a lot because it eliminate the need to figure out how to call `setNavSlot` and `unsetNavSlot` appropriately.

However, this has two disadvantages:

1. This results in a per-app left nav, meaning all pages in the same app would have the same left nav. Currently, this is a non-issue, since most apps have at most one page. The only exception (outside of stuff in core) is the registration app, and it makes sense that if there was a left nav for registration, they would share the same left nav. (The two pages in question are the "create a patient" and "edit a patient" pages).
2. It's possible to have multiple active pages at the same time. This is effectively what we do for the implementer tools, dev tools, and primary navigation itself. Currently, my solution is just to hard-code ignoring those pages. This would mean, though, that the system wouldn't work correctly if anyone tried to add any similar custom pages.

It's possible to work around these. For example, I thought we could address 1 by adding a "name" property for pages, with some sort of fallback for legacy code. 2 could be addressed by associating a "type" with each page, maybe "Main Screen" and "Component" or something, so that we could remove the hard-coding and only look for "Main Screen" components. But I'm also not sure if the added properties are worth it to solve what are likely corner-cases.

Any opinions on this?

mseaton:
What about using the overall route to determine what nav extension slot to render?  Similar to how the activation function works generally for pages, but applied to slots / extensions?  And if multiple are matched, then the most specific wins?  

Let's assume an extension slot base (suffix) of `left-nav-slot`.

So, when the url starts with `<spaBasePath>/home`, then you'd load extensions registered with `home-left-nav-slot`, and when the url starts with `<spaBasePath>/patient` you'd load extensions registered with `patient-left-nav-slot`, etc.

But if someone wanted a custom left nav for, say, service-queues, which would by default match up with the `home` slot at `<spaBasePath>/home/service-queues`, then one could register extensions at `home-service-queues-left-nav-slot` and this would take precedence.  This avoids the "one nav per app" issue an also avoids the "multiple pages per route" issue.

This would also inherently allow for defining a global, default slot by registering extensions for the root url at `<spaBasePath>/`, which would load extensions from the `left-nav-slot`, which solves what I think will be the primary pattern we will want, which is to have one default / global nav (not something tied to "home"), which can then be overridden by one in patient chart.

brandones:
Yeah, definitely an interesting idea, @mseaton . It might require a little more infrastructure around how we relate the app state to the URL, but that might be a good thing. I kind of prefer that to a solution which misbehaves if there are multiple apps per page.

> I was looking into the featureName thing @brandones suggested and it's not quite workable here because there's no guaranteed extension to pull from.

Can you expand a bit on that? I don't understand.

ibacher:
> Can you expand a bit on that @ibacher ? I don't understand.

Well, I was trying to do away with the need to call `setLeftNav()`. Without that call, there's no single local extension I could pick the feature name from, so we'd be stuck with a single slot for the left nav. Maybe that's desirable, but then need to rework how we do a different slot in the patient chart. (For the most part, I kind of imagine that we'd have separate navs for at least home, the patient chart, and maybe things that are sort-of external applications, like dispensing.

> So, when the url starts with <spaBasePath>/home, then you'd load extensions registered with home-left-nav-slot, and when the url starts with <spaBasePath>/patient you'd load extensions registered with patient-left-nav-slot, etc.

So, I generally agree that using the URL makes sense. I don't totally love defining multiple possible slots for nested URLs. E.g., for `<spaBasePath>/home/service-queues`, we have to check three slots; for the chart, we currently use URLs like `<spaBasePath>/patient/:patientUuid/chart/:view`, which could imply something four-slots deep. We'd also have to have a strategy to skip over the `:patientUuid` token, since I don't imagine there's any need for a per patient left nav slot.

Using just the first part of the URL might be tempting to avoid checking numerous possible slots when trying to figure out the left nav slot, but we'd have to figure out how we distinguish between the chart (`<spaBasePath>/patient/:patientUuid/chart/:view`) and the "Edit Patient Details" page (`<spaBasePath>/patient/:patientUuid/edit`).

The idea of using the mounted apps was to basically be able to leverage apps `activeWhen()` functions to determine when the slot needed to change.

brandones:
Ohhhh right, of course, if the left nav is part of primary-navigation-app the feature name it receives would always be primary-navigation-app. Why do we want to move LeftNav into the primary navigation app, actually?

One solution is to leave the left nav as it is, exported by the styleguide, rather than moving it into the primary navigation app. And then the LeftNav component would get the feature name of the app which has used it. To render the "default" left nav it could take an optional prop like `<LeftNav default />`. OR instead of using the feature name it could take an optional prop `<LeftNav name="patient-chart" />` which, when left blank, renders the default left nav. Either way we can get rid of the `setLeftNav` stuff and state management.

denniskigen:
@chibongho, what's the impact of your most recent refactors to the left nav on this?

chibongho:
Hm... I wish I saw this PR before I merged [this one](https://github.com/openmrs/openmrs-esm-core/pull/1279) in.

Yeah, it makes sense to close this one now. With our goal to provide collapsible left nav, and to support rtl language properly, I think we should work towards deprecating the `<LeftNav>` component and just let the primary navigation app render it for us.

Moving `setLeftNav()` and `unsetLeftNav()` to esm-api makes sense. Not that important, but this is how we're using them in `esm-home` and `patient-chart` now:

    useEffect(() => {
       setLeftNav('slot-name');
       return () => unsetLeftNav('slot-name');
    })

I don't love this pattern, and we should be able to refactor that to `useLeftNav('slot-name')`

PR #991, opened 2024-05-11
(fix) O3-3188 make modal body vertical-scroll properly

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

The `showModal()` function adds an unnecessary `<div>` wrapping the `<ModalHeader>` and `<ModalBody>`. The div causes a CSS issue with `<ModalBody>` with long content unable to scroll. Removing the div fixes that.

Testing done:
- manually inspected the HTML DOM to verify that the extra `<div>` is no longer there.
- opened up a modal with long content to verify that its scrollable
- open up several modals with less content (not enough to use the modal to hit its max-height limit) to verify that it still tenders fine.
## Screenshots

Before:
modal_before

After:
modal_after

## Other

This change should only be noticeable for modals that require vertical-scrolling for its body. Also, when calling `showModal()`, the input modal component should not have a `<div>` wrapping its `<ModalHeader>` and `<ModalBody>` (rather, it should be wrapped in a `<React.Fragment>`). Unfortunately, that is done in many places in `patient-chart` and `patient-management`.

### DIFF
--- a/packages/framework/esm-styleguide/src/modals/index.tsx
+++ b/packages/framework/esm-styleguide/src/modals/index.tsx
@@ -43,14 +43,12 @@ function createModalFrame() {
   ) as HTMLButtonElement;
 
   closeButton.addEventListener('click', closeHighestInstance);
-  const outer = document.createElement('div');
-  outer.className = 'cds--modal-container';
-  const contentContainer = document.createElement('div');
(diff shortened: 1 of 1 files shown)


ibacher (approved):
Thanks @chibongho!

PR #992, opened 2024-05-13
(fix) Styleguide should not depend on framework

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

So the slowness I was experiencing building core was, I think, an accidental side-effect of the commit that moved the workspace into core. The issue is that the files implementing the workspace directly referenced `@openmrs/esm-framework` inside the styleguide; this is an issue because the styleguide is a dependency of the framework. Basically, what was happening was that a "fresh" build (with no `dist`) would take a reasonable time (about 8 seconds for the styleguide and framework each on my machine), but subsequent builds with the now existing `dist` folder would begin to grow in length following an exponential curve (12 seconds, then 24 seconds, up to, at one point 30 minutes).

Removing this dependency cycle doesn't seem to harm the workspace behaviour and resolves the growing build time.

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -5169,7 +5169,9 @@ ___
 Shows a modal dialog.
 
 The modal must have been registered by name. This should be done in the `routes.json` file of the
-app that defines the modal.
+app that defines the modal. Note that both the `<ModelHeader>` and `<ModalBody>` should be at the
+top level of the modal component (wrapped in a React.Fragment), or else the content of the modal
+body might not vertical-scroll properly.
 
 #### Parameters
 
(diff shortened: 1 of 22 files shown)


packages/framework/esm-framework/docs/API.md
```diff
@@ -5169,7 +5169,9 @@ ___
 Shows a modal dialog.
 
 The modal must have been registered by name. This should be done in the `routes.json` file of the
-app that defines the modal.
+app that defines the modal. Note that both the `<ModelHeader>` and `<ModalBody>` should be at the
```
denniskigen, line 5172:
```suggestion
app that defines the modal. Note that both the `<ModalHeader>` and `<ModalBody>` should be at the
```

denniskigen (approved):
LGTM. Thanks, Ian.

packages/framework/esm-framework/docs/API.md
ibacher: Oh, yeah, we should fix that in the comment (unconnected to this PR).

denniskigen:
I don't know whether this is related, but there's a new problem with installing the `openmrs` CLI globally. Running `npm i -g openmrs@next` fails with an error related to `esm-translations`. Let me get the screenshot and post it on how.

ibacher:
I think that's because we haven't had a "real" release of core since the esm-translations sub-library was added. Could you try running with `--legacy-peer-deps`?

denniskigen:
Makes sense, thanks!

PR #994, opened 2024-05-14
(fix) O3-2724: Fix workspace URL detection

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

The workspace tries to check that the "context" it has been given is correct for the current page URL. However, the regex used is designed for testing against the pathname and not the full href, which may contain hashes (`#`) or query strings (`?`). This just ensures we're only running the regex against the pathname. I've also made some small adjustments, like escaping and regex metachars in th `contextKey` and using the single-spa provided types.


denniskigen (approved):
LGTM. Thanks, @ibacher!.

PR #997, opened 2024-05-15
(feat) O3-3168: O3 should work without requiring a network connection

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This adds a new global `offlineEnabled` which just records whether or not the app was setup with offline mode enabled or disabled. This global is now used everywhere in the framework that we check the online / offline status. If `offlineEnabled` is `false` (which is now the default), offline mode will never be triggered and every extension should function as though it were in online mode.

The underlying assumption here is that any distro built for a no-network use-case has no reason to enable offline mode. This may entail supporting something like our reference envsubst usage to be able to maintain a single build that can support both offline mode and no-network use cases. I'll be updating the RefApp so that offline mode is controlled by an environment variable.

### DIFF
--- a/packages/framework/esm-extensions/package.json
+++ b/packages/framework/esm-extensions/package.json
@@ -42,6 +42,7 @@
     "@openmrs/esm-config": "5.x",
     "@openmrs/esm-feature-flags": "5.x",
     "@openmrs/esm-state": "5.x",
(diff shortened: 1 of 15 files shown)


denniskigen:
FYI @donaldkibet @slubwama @ojwanganto

PR #1001, opened 2024-05-17
(chore) - add common typings for patient and user

chibongho:
# Requirements

## For changes to apps

## If applicable


packages/framework/esm-api/src/types/person-resource.ts
```diff
@@ -0,0 +1,69 @@
+import { type Concept } from './concept-resource';
+import { type OpenmrsResource } from './openmrs-resource';
+
+export interface PersonAttribute {
+  attributeType: OpenmrsResource;
+  display: string;
+  uuid: string;
+  value: string | number;
```
mseaton, line 8: I don't think this is necessarily accurate.  Can you check on this typing for value?

mseaton (commented):
Generally LGTM.  Are these mostly copied out of the work we did in service queues?  I'd want to validate that these interfaces exactly match the REST responses - can you confirm?

packages/framework/esm-api/src/types/person-resource.ts
chibongho: you're right. This should be a string.

chibongho:
> Generally LGTM. Are these mostly copied out of the work we did in service queues? I'd want to validate that these interfaces exactly match the REST responses - can you confirm?

Yeah, most of these were from when we were making these types together in the service queues module months ago.

ibacher:
One issue here is how we handle the difference between different representations. I'm not even necessarily thinking about custom representations, but, e.g., the default REF format for a person (which is returned in some circumstances):

```ts
interface Person {
  uuid: string;
  display: string;
  links: Array<{ rel: string; uri: string; resourceAlias?: string; }>;
}
```

packages/framework/esm-api/src/types/person-resource.ts
```diff
@@ -0,0 +1,69 @@
...
+  deathDate: string;
+  causeOfDeath: Concept;
+  preferredName: PersonName;
+  preferredAddress: PersonAddress;
+  names: Array<PersonName>;
+  addresses: Array<PersonAddress>;
+  attributes: Array<PersonAttribute>;
+  birthtime: string;
+  deathdateEstimated: boolean;
+  causeOfDeathNonCoded: string;
```
ibacher, line 28: This one is also nullable.

ibacher (commented):
Couple of quick thoughts here to make the types a little more reflective of the API.

chibongho:
> Generally LGTM. Are these mostly copied out of the work we did in service queues? I'd want to validate that these interfaces exactly match the REST responses - can you confirm?

@mseaton these interfaces are definitely incomplete (for example, Patient has a `voided` field with v=full rep). They are probably good enough for most use cases, but I can add those missing fields.

> One issue here is how we handle the difference between different representations. I'm not even necessarily thinking about custom representations, but, e.g., the default REF format for a person (which is returned in some circumstances):
> 
@ibacher Thanks for the suggestions. Unless a field's type can change based on the representation, this should be ok right? I think it'll still helpful to see what fields are available for each type while having the expectation that it's up to the implementor to verify that the fields they need are indeed there.

ibacher:
> Thanks for the suggestions. Unless a field's type can change based on the representation, this should be ok right?

I'm not really too concerned about that (it's fully possible with `custom` and nested objects). The real thing is I'm wondering if the types themselves should force the implementer to verify that the fields are there, i.e., something like:

```ts
interface Person extends Pick<FullPerson, 'uuid' | 'display'>, Partial<Omit<FullPerson, 'uuid' | 'display'>> {}
```

Because that would kind of force using something like:

```ts
const patientName: string = patient.person.preferredName ?? patient.person.name?[0] ?? 'UNKNOWN';
```

chibongho:
> > Thanks for the suggestions. Unless a field's type can change based on the representation, this should be ok right?
> 
> I'm not really too concerned about that (it's fully possible with `custom` and nested objects). The real thing is I'm wondering if the types themselves should force the implementer to verify that the fields are there

I don't like having that forcing everywhere, but I think we can try it and remove it later if we want (easier to go from restrictive to loose, instead of vice versa).

So how about something like this pattern? I think having every attribute being `?` is equivalent to your example.
```ts
// this already exists:
export interface OpenmrsResource {
  uuid: string;
  display?: string;
  [anythingElse: string]: any;
}

// we add this as the base type for almost everything
export interface OpenmrsResourceStrict {
  uuid: string;
  display?: string;
  links?: Array<Link>;
}

interface Person extends OpenmrsResourceStrict  {
  // everything with a '?'
  givenName?: string;
  familyName?: string;
  ...
}
```

ibacher:
> I don't like having that forcing everywhere

I'm not hugely enamored with my own proposal, but I think that having the type system setup like that will help some of our community developers do the right thing.

And, yeah, I'm happy with the pattern you suggest.

packages/framework/esm-api/src/types/concept-resource.ts
```diff
@@ -0,0 +1,5 @@
+import { type OpenmrsResource } from './openmrs-resource';
+
+export interface Concept extends OpenmrsResource {
```
mseaton, line 3: Does this need to be `OpenmrsResourceStrict` too?  Should it have more properties?

chibongho: Not every type needed by the Patient type (directly or transitively) is completely typed yet. Specifically this `Concept` type and the `User` type below are just using `OpenmrsResource` as a flexible placeholder for now, while still being more correct than `any`. A completely typed type should extend `OpenmrsResourceStrict` instead. This convention should allow us to add more typings piecemeal.

ibacher (approved):

ibacher:
Thanks @chibongho! I think this actually helps solve a long-standing problem we've had with types returned from APIs.

PR #1007, opened 2024-05-22
(feat) Add additional properties for OpenmrsDatePicker in Carbon

kajambiya:
# Requirements

## For changes to apps

## If applicable

## Summary
There's a requirement to adopt the OpenmrsDatepicker on the REACT form engine. As such, it was imperative that the OpenmrsDatepicker is extended to support all the features that the RFE Datepicker currently supports.

### DIFF
--- a/packages/framework/esm-styleguide/src/datepicker/openmrs/openmrs-date-picker.component.tsx
+++ b/packages/framework/esm-styleguide/src/datepicker/openmrs/openmrs-date-picker.component.tsx
@@ -3,6 +3,7 @@ import { supportedLocales as reactSpectrumSupportedLocales } from '../react-spec
 import { getLocale } from '@openmrs/esm-utils';
 import ReactSpectrumDatePickerWrapper from '../react-spectrum/adobe-react-spectrum-date-wrapper.component';
 import { DatePicker, DatePickerInput } from '@carbon/react';
+import dayjs from 'dayjs';
 
 // TODO: should be locale sensitive
 // see: https://issues.openmrs.org/browse/O3-998
@@ -11,14 +12,14 @@ const DEFAULT_PLACEHOLDER = 'dd/mm/yyyy';
 
 export interface OpenmrsDatePickerProps {
   id: string;
-  labelText: string;
+  labelText: string | any;
   onChange: (value: Date) => void;
   value?: Date | string;
   /* Not supported by Carbon's picker */
   defaultValue?: Date | string;
   minDate?: Date | string;
   maxDate?: Date | string;
-  readonly?: boolean;
+  readonly?: string | boolean;
   dateFormat?: string;
   invalid?: boolean;
   invalidText?: string;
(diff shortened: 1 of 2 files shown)


denniskigen (approved):
Thanks

packages/framework/esm-styleguide/src/datepicker/openmrs/openmrs-date-picker.component.tsx
```diff
@@ -11,14 +12,14 @@ const DEFAULT_PLACEHOLDER = 'dd/mm/yyyy';
...
-  labelText: string;
+  labelText: string | any;
   onChange: (value: Date) => void;
   value?: Date | string;
   /* Not supported by Carbon's picker */
   defaultValue?: Date | string;
   minDate?: Date | string;
   maxDate?: Date | string;
-  readonly?: boolean;
+  readonly?: string | boolean;
```
ibacher, line 22: [Carbon only supports booleans here](https://github.com/carbon-design-system/carbon/blob/5078eb5204b21763244768af3b0b283551561d47/packages/react/src/components/DatePickerInput/DatePickerInput.tsx#L178). The RFE must convert these to booleans before passing the prop to the component. Otherwise, `"false"` will be treated as true.

```suggestion
  readonly?: boolean;
```

packages/framework/esm-styleguide/src/datepicker/openmrs/openmrs-date-picker.component.tsx
```diff
@@ -11,14 +12,14 @@ const DEFAULT_PLACEHOLDER = 'dd/mm/yyyy';
 
 export interface OpenmrsDatePickerProps {
   id: string;
-  labelText: string;
+  labelText: string | any;
```
ibacher, line 15: Technically, what's supported here is [this](https://github.com/carbon-design-system/carbon/blob/5078eb5204b21763244768af3b0b283551561d47/packages/react/src/components/DatePickerInput/DatePickerInput.tsx#L19). No explicit `any`s should be added to the framework unless its unavoidable.

```suggestion
  labelText: string | ReactElement;
```

packages/framework/esm-styleguide/src/datepicker/openmrs/openmrs-date-picker.component.tsx
```diff
@@ -93,6 +94,14 @@ export const OpenmrsDatePicker: React.FC<OpenmrsDatePickerProps> = ({
         invalidText={invalidText}
         placeholder={carbonOptions?.placeholder || DEFAULT_PLACEHOLDER}
         style={carbonOptions?.pickerInputStyle}
+        // Added for testing purposes.
+        // Notes:
+        // Something strange is happening with the way events are propagated and handled by Carbon.
+        // When we manually trigger an onchange event using the 'fireEvent' lib, the handler below will
+        // be triggered as opposed to the former handler that only gets triggered at runtime.
+        onChange={(e) => onChange(dayjs(e.target.value, DEFAULT_PLACEHOLDER.toUpperCase()).toDate())}
```
ibacher, line 102: `onChange` in the `DatePickerInput` is not a React pseudo-event and cannot be correctly triggered using the React Testing Library. That's because the Carbon DatePicker is a wrapper around the `flatpickr` library. I'd prefer we didn't add a do-nothing handler here. The right thing to do would be to use a `ref` to get the `div` the `DatePicker` creates, get the actual `input` element underneith it and fire it's change event.

ibacher (changes requested):
See my remarks for changes that should be made here. Thanks!

kajambiya:
> See my remarks for changes that should be made here. Thanks!

Thanks @ibacher 
I've updated the code accordingly

PR #1024, opened 2024-05-31
(feat) Add CarbonMRS SVGs

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This adds all of the current SVGs available on Zero Height. This means that all placeholder SVGs are replaced with their CarbonMRS counterparts, except for Save and Reset.

## Other

Although CarbonMRS icons are similar to Carbon icons, they are all re-drawn with different size and line-weights, etc. from the Carbon set. This means they may looks slightly off next to Carbon icons. The solution here is to replace all the app icons with CarbonMRS equivalents.

Although this PR includes all the SVGs, I have not actually implemented every icon. This PR only adds 6 or so new icons. I've also included all of the pictograms, but none of the pictograms are actually used at this point.


denniskigen (approved):
LGTM! Thanks, @ibacher.

denniskigen:
The only thing that looks off to me so far is the `X` button in the workspace: 

CleanShot 2024-05-31 at 11  20 31@2x

ibacher:
@denniskigen I think I just fixed that? I was having some issues with the primary nav icons and I probably set the default too large.

denniskigen:
Looks good! Thanks!

PR #1032, opened 2024-06-05
(feat) O3-3370: In development mode, import maps should not use absolute URLs

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

When running the `openmrs develop` command, the full URL for both the import map and the scripts referenced in the map. this works if you are accessing the development server locally, but it fails with tools like GitPod, which host the development server through their own reverse proxies. This PR changes things so that we just use relative paths instead of absolute URLs, which should solve the issue without affecting local devs.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3370


denniskigen (approved):
LGTM. Thanks, @ibacher.

PR #1036, opened 2024-06-10
(fix) Login page tweaks

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR updates the Login component with the following changes:

- Removes the `value` Prop from the [PasswordInput](https://react.carbondesignsystem.com/?path=/docs/components-passwordinput--overview) component. This prevents passwords from appearing in the browser elements panel as they are typed in. This change has no impact on the login functionality. 
- Updates some CSS class names to follow the camel-case naming convention.
- Tweaks the login flow error logic so the user only gets navigated back to the `/login` route when logging in fails and the `showPasswordOnSeparateScreen` configuration property is set to `true`. When the `showPasswordOnSeparate` screen is set to `false`, the username and password are shown on the same screen and the same route (`/login`). It doesn't make sense to navigate the user to the route they're already on. Additionally, I've removed the code that clears the username and password inputs when login fails. 

- Adds translations for the logo title and alt text.

### Password revealed in the browser elements panel

https://github.com/openmrs/openmrs-esm-core/assets/8509731/b42bfa85-4596-4bac-b9b7-a102d5839bef

### DIFF
--- a/packages/apps/esm-login-app/src/login/login.component.tsx
+++ b/packages/apps/esm-login-app/src/login/login.component.tsx
@@ -1,17 +1,17 @@
 import React, { useState, useRef, useEffect, useCallback } from 'react';
 import { type To, useLocation, useNavigate } from 'react-router-dom';
-import { useTranslation } from 'react-i18next';
+import { type TFunction, useTranslation } from 'react-i18next';
 import { Button, InlineLoading, InlineNotification, PasswordInput, TextInput, Tile } from '@carbon/react';
 import {
-  useConfig,
-  interpolateUrl,
-  useSession,
-  refetchCurrentUser,
-  useConnectivity,
-  navigate as openmrsNavigate,
-  getCoreTranslation,
   ArrowLeftIcon,
   ArrowRightIcon,
+  getCoreTranslation,
+  interpolateUrl,
+  navigate as openmrsNavigate,
+  refetchCurrentUser,
+  useConfig,
+  useConnectivity,
+  useSession,
 } from '@openmrs/esm-framework';
 import { type ConfigSchema } from '../config-schema';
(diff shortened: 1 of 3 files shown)


ibacher (approved):

brandones:
@denniskigen This seems to have made it so that now filling a stored password from the browser doesn't work from the username entry page.

https://github.com/openmrs/openmrs-esm-core/assets/1031876/9949d593-4425-4846-99ea-586e59b40d0e

I think to fix this we probably want to refactor the login flow so that it all actually takes place in one component, nothing gets mounted or unmounted in the transition between the username "page" and the password "page".

ibacher:
I think it's just [this line](https://github.com/openmrs/openmrs-esm-core/pull/1036/files#diff-a91a2347b0df20dbcee119ced765bfb975088ee720a045c3bc842080987c71a8L228) that needs to be reverted...

brandones:
Well but that's the line that fixes the security issue that @denniskigen identified, where the password shows up in cleartext in the HTML. Though I'm not sure to what extent that constitutes a real problem, since
- It's unlikely that the value will get written to disk since it's not actually in HTML being sent over the wire
- At least in Chrome you don't need to authenticate to view stored passwords. Though I think for Apple products you do. So it would constitute a leak for Apple users, since it provides a way for an unauthenticated user with device access to obtain the stored cleartext password.

ibacher:
I see... I can see how when screen-sharing / recording, it might become something of an issue, since you might end up accidentally recording a password you didn't mean to.

denniskigen:
That's not ideal! Thanks for catching it, @brandones. I think it makes sense to fix that first and then worry about the security issue afterwards. Let me file a PR.

PR #1050, opened 2024-06-24
(fix) Replace deprecated substr() with slice() for string operations

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR replaces usages of the deprecated [String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) method with [String.prototype.slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice).

### DIFF
--- a/packages/tooling/openmrs/src/utils/helpers.ts
+++ b/packages/tooling/openmrs/src/utils/helpers.ts
@@ -1,19 +1,19 @@
-import { createHash } from 'crypto';
+import { util } from 'webpack';
 
-export function trimEnd(text: string, chr: string): string {
+export function trimEnd(text: string, chr: string) {
   while (text.endsWith(chr)) {
     text = text.slice(0, text.length - chr.length);
   }
   return text;
 }
 
-export function removeTrailingSlash(path: string): string {
+export function removeTrailingSlash(path: string) {
   const i = path.length - 1;
   return path[i] === '/' ? removeTrailingSlash(path.slice(0, i)) : path;
 }
 
-export function contentHash(obj: object): string {
-  const hash = createHash('xxhash64');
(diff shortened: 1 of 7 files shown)


packages/tooling/openmrs/src/utils/helpers.ts
```diff
@@ -1,20 +1,19 @@
-import { util } from 'webpack';
+import { createHash } from 'crypto';
```
denniskigen, line 1: This now uses Node.js's native [crypto](https://nodejs.org/api/crypto.html) module instead of Webpack's utility.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -36,23 +36,23 @@
...
+import { inc } from 'semver';
 import {
-  DefinePlugin,
   container,
+  DefinePlugin,
   type ModuleOptions,
-  type WebpackOptionsNormalized as WebpackConfiguration,
   type RuleSetRule,
+  type WebpackOptionsNormalized as WebpackConfiguration,
 } from 'webpack';
```
denniskigen, line 53: I've just reorganized the imports here.

packages/tooling/openmrs/src/utils/helpers.ts
ibacher: I think we need to continue using the Webpack utility here, at least to retain the `xxhash64` hash; it isn't implemented by OpenSSL, so it isn't available through Node's crypto module.

ibacher (commented):
LGTM apart from the hashing change.

vasharma05 (approved):

PR #1053, opened 2024-06-24
(feat) O3-1991: Enhancements and fixes for OMRS date picker

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This addresses a number of issues with the current version of the datepicker:

* Allows additional props, specifically, most of the DOM-related properties, etc.
* Basic support for validation styles
* Now can click anywhere in the field to launch the pop-over
* Label can support arbitrary React elements
* Years are now correct for non-Gregorian calendars (to the best of my ability to test)

## Other

Outstanding: `minDate` and `maxDate` are currently values that only control which dates display as available. They are not (currently) part of the validation, so manually entering a date outside this range is allowed. That'll be for the next PR along with whatever else comes up.

### DIFF
--- a/packages/framework/esm-styleguide/src/datepicker/index.tsx
+++ b/packages/framework/esm-styleguide/src/datepicker/index.tsx
@@ -127,6 +127,19 @@ function dateToInternationalizedDate(date: DateInputValue): DateValue | undefine
   }
 }
 
+function getYearAsNumber(date: Date, intlLocale: Intl.Locale) {
+  return parseInt(
+    formatDate(date, {
+      calendar: intlLocale.calendar,
+      locale: intlLocale.baseName,
+      day: false,
+      month: false,
+      noToday: true,
+      numberingSystem: 'latn',
+    }),
+  );
+}
+
 const MonthYear = forwardRef<Element, PropsWithChildren<HTMLAttributes<HTMLSpanElement>>>(
   function MonthYear(props, ref) {
     const { className } = props;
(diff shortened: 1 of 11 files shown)


packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -75,10 +91,6 @@ export interface OpenmrsDatePickerProps {
    * The earliest date it is possible to select
    */
   minDate?: DateInputValue;
-  /**
-   * Handler that is called when the value changes
-   */
-  onChange?: (date: DateValue) => void;
```
ibacher, line 81: NB on this and `isDateUnavailable`: the props still exist, they are just part of the `DatePickerProps` that we now inherit from.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -108,14 +120,26 @@ function dateToInternationalizedDate(date: DateInputValue): DateValue | undefine
...
 
+function getYearAsNumber(date: Date, intlLocale: Intl.Locale) {
+  return parseInt(
+    formatDate(date, {
+      calendar: intlLocale.calendar,
+      locale: intlLocale.baseName,
+      day: false,
+      month: false,
+      noToday: true,
+      numberingSystem: 'latn',
```
ibacher, line 138: numberingSystem needs to be specified or else for languages with alternative numbering systems (e.g., Arabic, Persian, etc.), the returned value will be in those number systems; Javascript's `number` is not compatible with non-Latin digits.

denniskigen (commented):
LGTM. Thanks for the quick turnaround, @ibacher.

denniskigen (approved):

PR #1071, opened 2024-07-03
(feat) Attempt to shift performance

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

Saving this for after the release. This is an experimental set of changes aiming to improve performance. Does the following:

* Add a shared dep for `@carbon/react`
* Mark most shared dependencies as not being eager (basically, the framework is now on a separate chunk)
* Re-writes a few more lodash modules to lodash-es

## Other

Goal:

`@carbon/react` as shared dep adds a 500kB chunk to each module (the full `@carbon/react`), but, on the other hand, we only need to load this once. This is the most likely change to be reversed.

Marking shared deps as not eager should drastically reduce the overall size of openmrs.js, at the expense of requiring more network calls. This is because almost every shared dep is now loaded asynchronously. On my machine, the whole app-shell bundle is substantially smaller, but it may be that delaying the loading of various libraries turns out to be bad performance-wise.

The lodash rewrites are trivial and will likely remain in. Carbon uses the versions of Lodash that bundle individual functions (and their dependencies), but this results in duplicative code at runtime. Re-writing them to use lodash-es like this should get a similar bundle, but with all duplicative code... deduplicated.


samuelmale:
Finally we spike at having Carbon as a transitive dependency

ibacher:
@samuelmale I spent a long time evaluating various Webpack stats for various packages and came to the conclusion that this might actually make sense. The biggest issue is that Carbon, and especially Carbon's icons, are not published in a way that makes it possible for them to be tree-shaken, so most apps end up bundling 200-300kB (minified) of code. That means that the overhead of loading a 500kB chunk is probably worthwhile. I had been hoping (before I examined it in depth) that we were tree-shaking things so that the Carbon size would be closer to 30-50kB, in which case, the large chunk may not be worthwhile... or would at least make the app appear slow while that large chunk was loading over a poor network connection.

PR #1078, opened 2024-07-10
(fix) O3-3563 OpenmrsDatepicker should use calendar to calculate date values

kajambiya:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR fixes an issue whereby if the locale was amharic, the datepicker was defaulting to a wrong day and month. This issue would happen when the date picker was called with the `default`, `min` or `max` values. This was because these values were not being converted to the corresponding locale calendar

## Screenshots
Before fix:
Screenshot 2024-07-10 at 08 16 49

After fix:

Screenshot 2024-07-10 at 08 15 50

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -365,7 +365,7 @@ ___
 
 #### Defined in
 
-[packages/framework/esm-utils/src/omrs-dates.ts:20](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/omrs-dates.ts#L20)
+[packages/framework/esm-utils/src/omrs-dates.ts:26](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/omrs-dates.ts#L26)
 
 ___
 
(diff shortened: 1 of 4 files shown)


NethmiRodrigo:
The issue was that we relied on `@internationalized` library's `getLocalTimezone` function which returns the the user's literal time zone while `getLocale` relies on the i18n language.

kajambiya:
> @kajambiya could you also update the today's date to an internationalized date like: `const today_ = dateToInternationalizedDate(today(getLocalTimeZone()), locale);`

Nice catch @NethmiRodrigo. I've however called the `convertToLocaleCalendar` function directly because the `dateToInternationalizedDate` function can return `undefined` which would cause [this](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/datepicker/index.tsx#L338) line to fail.

kajambiya:
Hi @ibacher 
Thanks for the re-write.
We've however noticed some unusual behaviour. The date picker seems not to correctly calculate the `maxDate` thus allowing a user to select a future date on `DOB`. Also when a date is selected, the selected input is in the Gregorian format and not ethiopic.

PR #1085, opened 2024-07-18
(feat) O3-3260 Add ability to register feature flags via routes.json

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR adds three relatively simple features:

1. Support for feature flags being defined in `routes.json`
2. Support for hiding pages via declaring that they need a specific feature flag
3. There is now a framework function to an initialize an application whether or not it's been started; this hasn't been exposed via `window`, but it may be useful for testing or digging ourselves out of the holes that not calling `startupApp()` sometimes leave.

The rest of the changes are just refactorings for the following things:

* The registration logic for modals and workspaces are moved into esm-extensions so they can be used in esm-routes
* The logic for handling the contents of routes.json has been moved to esm-routes
* The app.ts code is the app shell is removed, since this code is now all in esm-routes

## Other

NB When updating to this version of the framework, the tsconfig of each repo will likely need to be updated to include "es2022" as a valid option. This is because the code I moved from the app shell into the framework relied on `Object.hasOwn()`. If there are objections to this, it can be trivially rewritten, but hey.


brandones (approved):
LGTM. Looks like mostly moving things around and a few very minor changes. I'm not sure what problem `initializeApp` will solve though.

ibacher:
@brandones The main use case is for apps that depend on another app's configuration (which is rare, but not unheard of). I'm also thinking about integrating it into the dev tools so if a similar bug arises we have an escape hatch.

denniskigen:
Does this require any changes to the routes.json [standard schema](https://github.com/openmrs/openmrs-contrib-json-schemas/blob/main/routes.schema.json)?

PR #1090, opened 2024-07-22
(fix) O3-3505 Allow OpenmrsDatePicker enable user to type date.

suubi-joshua:
# Requirements

## For changes to apps

## If applicable

## Summary
Adjusted the input component to allow input text which is then checked to validate by the handleInputChange Function to validate if it is a date formate.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3505


suubi-joshua:
Hello @denniskigen requesting for a review. Also not quiet sure why other commits are part of this branch. They seem to be unrelated to the issue am working on.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -106,6 +106,10 @@ const defaultProps: OpenmrsDatePickerProps = {
   size: 'md',
 };
 
+const locale = useLocale();
+const intlLocale = new Intl.Locale(locale.locale);
+const tz = Intl.DateTimeFormat(intlLocale.toString()).resolvedOptions().timeZone;
+
```
NethmiRodrigo, line 112: `useLocale` is a hook that can't be used outside of a React component. Check the other places where we've used this and you'd have to do something similarly.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -214,6 +214,16 @@ const DatePickerInput = forwardRef<HTMLDivElement, DateInputProps>(function Date
   const inputRef = useRef<HTMLInputElement>(null);
   const { fieldProps, inputProps } = useDateField({ ...dateFieldProps, inputRef }, state, fieldRef);
 
+  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const typedDate = event.target.value;
+    const parsedDate = dayjs(typedDate, 'MM/DD/YYYY', true);
+    if (parsedDate.isValid()) {
+      const dateValue = new CalendarDate(parsedDate.year(), parsedDate.month() + 1, parsedDate.date());
+      state.setValue(dateValue);
```
NethmiRodrigo, line 222: Why is it adding +1 to the parsed month? Also please refer to elsewhere of how the date is internationalised, I figure it would have to be done here too.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -232,7 +242,14 @@ const DatePickerInput = forwardRef<HTMLDivElement, DateInputProps>(function Date
       >
         {state.segments.map((segment, i) => cloneElement(props.children(segment), { key: i }))}
       </Group>
-      <Input />
+      <Input
+          type="text"
+          value={state.value ? dayjs(state.value.toDate(tz)).format('MM/DD/YYYY') : ''}
+          onChange={handleInputChange}
+          placeholder="MM/DD/YYYY"
+          className="date-input"
```
NethmiRodrigo, line 250: Since this is a style from the scss file,
```suggestion
          className={styles.dateInput}
```

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -232,7 +242,14 @@ const DatePickerInput = forwardRef<HTMLDivElement, DateInputProps>(function Date
...
+          {...inputProps}
+        />
```
NethmiRodrigo, line 252: I believe the reason we were holding this off for a while was to be able to not stick to a specific format and base it off the locale, among other things. But @ibacher is probably the best to advise on this here.

packages/framework/esm-styleguide/src/datepicker/datepicker.module.scss
```diff
@@ -127,6 +127,9 @@
       color: theme.$text-disabled;
     }
   }
+  .date-input:focus {
```
NethmiRodrigo, line 130:
```suggestion
  .dateInput:focus {
```

NethmiRodrigo (changes requested):

NethmiRodrigo:
> Also not quiet sure why other commits are part of this branch. They seem to be unrelated to the issue am working on. 

After making sure that your main is up to date, please try `git rebase -i main` and when it opens up a vim editor, make sure your commits come on top of the latest commit thats on main and if so exit vim(:wq to confirm and write). `git log --oneline` and make sure the git history looks clean, and then `git push --force-with-lease`. 
But when you do try to push, the tests might fail because of the hook being used outside in the date picker @suubi-joshua

NethmiRodrigo:
Thanks for taking a hack at this! But now that I've had a bit more time to think about this, I have concerns/reservations (aside from the requested changes) about using an `Input` field:
The `Input` component, since its just a text field, unlike the `DateField` component, would allow the user to type in the date in any format (and not stick to `mm/dd/yyyy` or any other), which would likely lead to unexpected behaviour if the user does deviate from the expected format. E.g. if you type in the year as 01999 in the Carbon date picker, it auto corrects to 1999. As for the React Aria's DateField component, I think it doesn't let you type in starting with leading zeros. 
Thinking of everything that DateField does for out of the box and trying to custom implement for the `Input` field, maybe `Input` isn't the best choice to use? @ibacher @denniskigen

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -232,7 +242,14 @@ const DatePickerInput = forwardRef<HTMLDivElement, DateInputProps>(function Date
       >
         {state.segments.map((segment, i) => cloneElement(props.children(segment), { key: i }))}
       </Group>
-      <Input />
+      <Input
+          type="text"
+          value={state.value ? dayjs(state.value.toDate(tz)).format('MM/DD/YYYY') : ''}
+          onChange={handleInputChange}
+          placeholder="MM/DD/YYYY"
```
NethmiRodrigo, line 249: I'm not sure how this would work. The current input field is rendered inside the `DatePickerInput`, which is still there.

ibacher:
Fundamentally, the reason for having a DatePicker component is to be able to enter dates in non-Gregorian calendars. Any change to the DatePicker has to be able to accommodate that requirement and this PR does not.

I'm not necessarily opposed to an approach that would take text entered into an input field and convert it, but a) the date format cannot be hard-coded programmatically and b) it must support dates in formats appropriate to the user's calendar system.

The reason React Aria adopts the div-date component approach of inputs is because this allows it to accommodate formats for arbitrary calendar systems.

suubi-joshua:
> Before you proceed with the changes @suubi-joshua could you please update the PR with a video of how it works? For some reason I can't seem to override the import map to be able to test this out.

Actually this blocking me from doing a test as well. When I run a local server, the server runs but no UI loads in my browser. @NethmiRodrigo. Something related to import map.

NethmiRodrigo:
This doesn't seem like an import map error, its more likely to be caused by something that you've changed. Also, please refer to the README in this repo on how to correctly run the core (specifically the [linking the framework](https://promptlayer.com/workspace/7428/prompt/27050/version/11) section). Right now, the OpenmrsDatePicker is used in patient-management and form-builder, so that is the two frontend modules that you could test your changes on. You should link the `esm-styleguide` into one of those modules. I usually use the [method of using import map overrides](https://promptlayer.com/workspace/7428/prompt/27050/version/11) to run the modules.

ibacher:
This is fixed in #1099.

PR #1094, opened 2024-07-25
(feat) O3-3244 & O3-3250: Allow data sharing between workspaces with same sidebar family name

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR focuses on the workspaces with an action menu (sidebar) and can navigate between workspaces within the workspace itself.

1. Workspace with no action menu in itself, instead having the action menu outside of the workspace.
image

2. Workspace with action menu inside the workspace itself, where other workspace can be accessed via the action menu buttons only.
image
image

In the 1st scenario, the workspaces (order workspace, notes workspace) are independent of each other.

In the 2nd scenario, the workspaces are interlinked with some common data (as per the example screenshots, `patientUuid` and `bedId` is common data between the workspaces, and the new workspaces in the action menu are meant to be opened in the same context as of original workspace)

Taking an example of ward, user can click on the patient assigned to a bed and then the patient chart workspace will open on the right side. The other action menu buttons in the same workspaces are (for example) Transfers, in-patient notes, etc. Now the Transfers and In patient Notes workspaces share common data with the patient chart workspace, such as `patientUuid` and `bedNumber` assigned to the patient. 

To open the patient chart workspace, we can call: `launchPatientWorkspace('ward-patient-workspace', {patientUuid: <patientUuid>, bedNumber: <bedNumber>}').

Let's say, now we want to open the **Transfers** workspace, the action menu button will open it as `launchPatientWorkspace('transfer-patient-workspace')`. But the issue is that this function also needs the `patientUuid` and `bedNumber`. 

One way for this to happen is to pass the `additionalProps`(second arg of the `launchPatientWorkspace`) from the [WorkspaceRenderer](https://github.com/openmrs/openmrs-esm-core/blob/e1c8bd36795a8286576547c319c565680737b0db/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx#L127) to the [ActionMenu](https://github.com/openmrs/openmrs-esm-core/blob/e1c8bd36795a8286576547c319c565680737b0db/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx#L235), which can then be passed into the [ExtensionSlot](https://github.com/openmrs/openmrs-esm-core/blob/e1c8bd36795a8286576547c319c565680737b0db/packages/framework/esm-styleguide/src/workspaces/container/action-menu.component.tsx#L40) of the action menu, which then can be accessed by rendered Action Menu Buttons inside the above mentioned ExtensionSlot and then the Action Menu Button can take the same data from it's props and pass into the `launchPatientWorkspace` when launching, say, Transfers or In Patient Workspace.

In the above mechanism, the passing of the props from Action Menu to its respective `launchPatientWorkspace` is to be done by developer, which might be missed in some cases and it's a lot of props passing. Hence to improve the developer experience, we have come up with a workspace family store.

### Solution
The workspaces which have action menu in themselves has a registration property [hasOwnSidebar](https://github.com/openmrs/openmrs-esm-core/blob/e1c8bd36795a8286576547c319c565680737b0db/packages/framework/esm-extensions/src/workspaces.ts#L40) and [sidebarFamily](https://github.com/openmrs/openmrs-esm-core/blob/e1c8bd36795a8286576547c319c565680737b0db/packages/framework/esm-extensions/src/workspaces.ts#L41). If the workspaces have the same sidebarFamilyName, they need to share some common data, as explained above.

For these workspaces, a new property `workspaceFamilyStore` is added to the [OpenWorkspace](https://github.com/openmrs/openmrs-esm-core/blob/e1c8bd36795a8286576547c319c565680737b0db/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L96), which will point to the store of the workspaces sharing the same `sidebarFamilyName`. The workspaces can access the store using the `useWorkspaceFamilyStore` hook, passing the `sidebarFamilyName` into the hook and can access the data directly, without taking care of passing props from one component to another.

In case a workspace doesn't have a sidebar, the `workspaceFamilyStore` will be undefined for such workspaces, and the `useWorkspaceFamilyStore` hook will return empty data and dummy functions for such workspaces, since they are standalone workspaces.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-3244
https://issues.openmrs.org/browse/O3-3250

## Other
Greatly thankful to @brandones and @usamaidrsk for coming with the above solution.

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -182,7 +182,6 @@
 
 - [WorkspaceContainer](API.md#workspacecontainer)
 - [isOnline](API.md#isonline)
-- [useWorkspaceFamilyStore](API.md#useworkspacefamilystore)
 
 ### Store Functions
 
@@ -5611,31 +5610,6 @@ ___
 #### Defined in
 
 [packages/framework/esm-utils/src/is-online.ts:3](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/is-online.ts#L3)
-
-___
-
-### useWorkspaceFamilyStore
-
-▸ **useWorkspaceFamilyStore**(`sidebarFamilyName`): `void`
-
-This hook is used to interact with the store of a workspace family.
-A workspace family is defined as a group of workspaces that share the same sidebarFamilyName.
-
-In case a workspace doesn't have a sidebarFamilyName, it will be considered as a standalone workspace, and hence this hook will return an empty object and updateFunction as an empty function.
-
-#### Parameters
-
-| Name | Type | Description |
-| :------ | :------ | :------ |
-| `sidebarFamilyName` | `string` | The sidebarFamilyName of the workspace used when registering the workspace in the module's routes.json file. |
-
-#### Returns
-
-`void`
-
-#### Defined in
-
-[packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceFamilyStore.ts:13](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceFamilyStore.ts#L13)
 
 ___
 
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -182,6 +182,7 @@
 
 - [WorkspaceContainer](API.md#workspacecontainer)
 - [isOnline](API.md#isonline)
+- [useWorkspaceFamilyStore](API.md#useworkspacefamilystore)
 
 ### Store Functions
 
@@ -5613,6 +5614,31 @@ ___
 
 ___
 
+### useWorkspaceFamilyStore
+
+▸ **useWorkspaceFamilyStore**(`sidebarFamilyName`): `void`
+
+This hook is used to interact with the store of a workspace family.
+A workspace family is defined as a group of workspaces that share the same sidebarFamilyName.
+
+In case a workspace doesn't have a sidebarFamilyName, it will be considered as a standalone workspace, and hence this hook will return an empty object and updateFunction as an empty function.
+
+#### Parameters
+
+| Name | Type | Description |
+| :------ | :------ | :------ |
+| `sidebarFamilyName` | `string` | The sidebarFamilyName of the workspace used when registering the workspace in the module's routes.json file. |
+
+#### Returns
+
+`void`
+
+#### Defined in
+
+[packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceFamilyStore.ts:13](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceFamilyStore.ts#L13)
+
+___
+
 ## Store Functions
 
 ### createGlobalStore
@@ -7031,7 +7057,7 @@ Function to close an opened workspace
 
 #### Defined in
 
-[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:282](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L282)
+[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:295](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L295)
 
 ___
 
@@ -7082,7 +7108,7 @@ prop named `workspaceTitle` will override the title of the workspace.
 
 #### Defined in
 
-[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:165](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L165)
+[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:177](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L177)
 
 ___
 
@@ -7108,7 +7134,7 @@ Use this function to navigate to a new page and launch a workspace on that page.
 
 #### Defined in
 
-[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:245](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L245)
+[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:258](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L258)
 
 ___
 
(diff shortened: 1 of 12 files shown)


packages/framework/esm-framework/docs/API.md
```diff
@@ -5613,6 +5614,31 @@ ___
 
 ___
 
+### useWorkspaceFamilyStore
```
brandones, line 5617: Hi @vasharma05 , I would suggest doing this a little differently. I think we should not expose anything new in the API. Rather, new workspaces being opened in a sidebar family should just receive the correct additional props like magic. The workspace system itself will use `useWorkspaceFamilyStore` (or something like it) to pass the correct props to new workspaces. So everything happens behind the scenes. Does that make sense?

brandones (changes requested):

packages/framework/esm-styleguide/src/workspaces/container/workspace-renderer.component.tsx
```diff
@@ -35,6 +37,7 @@ export function WorkspaceRenderer({ workspace, additionalPropsFromPage }: Worksp
         setTitle: workspace.setTitle,
         ...additionalPropsFromPage,
         ...workspace.additionalProps,
+        ...workspaceFamilyState,
```
brandones, line 40:
```suggestion
        ...workspaceFamilyState,
        ...workspace.additionalProps,
```

The `additionalProps` that come directly from `launchWorkspace` are more specific, so I'd have those override the family state.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -95,6 +96,17 @@ export interface WorkspaceStoreState {
...
+   * The workspace family store is a store that is specific to the workspace sidebar family.
+   * If the workspace has its own sidebar, the store will be created and passed to the workspace.
+   * This store can be used to store data that is specific to the workspace sidebar family.
+   * The store will be same for all the workspaces with same sidebar family name.
+   *
+   * For workspaces with no sidebarFamilyName or sidebarFamilyName as 'default', the store will be undefined.
+   *
+   * The store will be cleared when all the workspaces with the store's sidebarFamilyName are closed.
+   */
+  workspaceFamilyStore?: StoreApi<object> | undefined;
```
brandones, line 109:
```suggestion
  workspaceFamilyStore?: StoreApi<object>;
```

Seems redundant with the `?`, right?

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -491,3 +516,20 @@ function getWorkspaceTitle(workspace: WorkspaceRegistration, additionalProps?: o
...
+): StoreApi<object> | undefined {
+  if (!sidebarFamilyName || sidebarFamilyName === 'default') {
+    return undefined;
+  }
+  const store = getGlobalStore<object>(sidebarFamilyName, {});
+  if (additionalProps) {
+    store.setState((prev) => ({
+      ...prev,
+      ...additionalProps,
+    }));
```
brandones, line 532: Are you sure that the store clearing on line 315 is working? In my experience that wouldn't work, because setState does a shallow merge with the existing state object, and merging `{ foo: 0 }` with `{}` yields `{ foo: 0 }`, not `{}`. This was at least true of our old state manager (what was it called?); I'm not sure about zustand.

The way I worked around that in the past was to put everything inside a top-level element. So I suggest having everything contained in a top-level property called `props`. Then when you clear it by doing `setState({ props: {} })` it will actually set the props element to `{}`, because it is a shallow merge.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -167,13 +179,14 @@ export function launchWorkspace<
...
   const workspace = getWorkspaceRegistration(name);
-  const newWorkspace = {
+  const newWorkspace: OpenWorkspace = {
     ...workspace,
     title: getWorkspaceTitle(workspace, additionalProps),
     closeWorkspace: (options: CloseWorkspaceOptions = {}) => closeWorkspace(name, options),
     closeWorkspaceWithSavedChanges: (options: CloseWorkspaceOptions) =>
       closeWorkspace(name, { ignoreChanges: true, ...options }),
     promptBeforeClosing: (testFcn) => promptBeforeClosing(name, testFcn),
+    workspaceFamilyStore: getWorkspaceFamilyStore(workspace.sidebarFamily, additionalProps),
```
brandones, line 189: Why are we adding this element to `OpenWorkspace`? It looks like we aren't using it.

brandones (changes requested):
A few questions and suggestions

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
vasharma05: Hi @brandones!
Yes, I was facing the store clean-up issue and realized that we have a second arg `replace` in the `setState`. This overrides the previous state and makes it an empty object.

I have also written a test for the same here:
image

vasharma05:
Hi @brandones!
I have pushed the required changes.
Thanks!

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -188,6 +189,20 @@ export function launchWorkspace<
     additionalProps: additionalProps ?? {},
   };
 
+  if (newWorkspace.sidebarFamily) {
+    /**
```
brandones, line 193: I'd put this docstring with either the `useWorkspaceFamilyStore` function or `getWorkspaceFamilyStore`?

brandones: Though I might add a comment here to the effect of

```
// initialize workspace family store
```

if that is indeed what is happening

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -429,4 +437,146 @@ describe('workspace system', () => {
...
+    });
+
+    it('should return store if workspace sidebar name is passed', () => {
+      const workspaceFamilyStore = getWorkspaceFamilyStore('ward-patient-sidebar', {
+        foo: true,
+      });
+      expect(workspaceFamilyStore).toBeTruthy();
+      expect(workspaceFamilyStore?.getState()?.['foo']).toBe(true);
+    });
+    it('should update the store state with new additionalProps if workspaces with same sidebarFamily name calls the function', () => {
```
brandones, line 456:
```suggestion
    });

    it('should update the store state with new additionalProps if workspaces with same sidebarFamily name calls the function', () => {
```

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -429,4 +437,146 @@ describe('workspace system', () => {
...
+      workspaceFamilyStore = getWorkspaceFamilyStore('ward-patient-sidebar', {
+        bar: true,
+      });
+      expect(workspaceFamilyStore).toBeTruthy();
+      expect(workspaceFamilyStore?.getState()?.['foo']).toBe(true);
+      expect(workspaceFamilyStore?.getState()?.['bar']).toBe(true);
+    });
+  });
+
+  describe('Testing workspace family store', () => {
```
brandones, line 471: Hmm, it's maybe a shame we don't have tests for WorkspaceRenderer. These tests feel like they're really heavy on implementation details—pretty fragile and we aren't even able to test the main idea of the change, which is "when I open a workspace with sidebar family 'foo' and args XYZ, and then I open another workspace with the same sidebar family, it should also receive args XYZ."

But that's just an aside; don't worry about it for this PR. Thanks for these tests.

brandones (approved):
A couple minor suggestions but LGTM

brandones:
Oh and please do file a ticket as @ibacher requested, and link it with the existing tickets that this supports.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
vasharma05: Yes @brandones , that is what I am doing!

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
vasharma05: Yes, we have the test that store has args from both workspace, but aren't checking if the real workspaces are getting it or not. We should add that.

brandones (approved):
Great work, thanks @vasharma05 !

PR #1100, opened 2024-07-30
(feat) O3-3684: Improve duration representation of `age` function

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

This commit improves on the existing `age()` function to display an age (or time duration) that follows [this](https://webarchive.nationalarchives.gov.uk/ukgwa/20160921162509mp_/http://systems.digital.nhs.uk/data/cui/uig/patben.pdf) guideline for NHS.

image

### DIFF
--- a/packages/framework/esm-utils/src/age-helpers.ts
+++ b/packages/framework/esm-utils/src/age-helpers.ts
@@ -1,101 +1,115 @@
 /** @module @category Utility */
+import dayjs from 'dayjs';
 import { getLocale } from './omrs-dates';
 
 /**
- * Gets the number of days in the year of the given date.
- * @param date The date to compute the days within the year.
- * @returns The number of days.
- */
-export function daysIntoYear(date: Date) {
-  return (
-    (Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate()) - Date.UTC(date.getUTCFullYear(), 0, 0)) /
-    24 /
-    60 /
-    60 /
-    1000
-  );
-}
-
-/**
- * Checks if two dates are representing the same day.
- * @param firstDate The first date.
- * @param secondDate The second date.
- * @returns True if both are located on the same day.
- */
-export function isSameDay(firstDate: Date, secondDate: Date) {
-  const firstISO = firstDate.toISOString();
-  const secondISO = secondDate.toISOString();
-  return firstISO.slice(0, firstISO.indexOf('T')) === secondISO.slice(0, secondISO.indexOf('T'));
-}
-
-/**
- * Gets a human readable and locale supported age represention of the provided date string.
- * @param dateString The stringified date.
+ * Gets a human readable and locale supported representation of a duration between 2 dates,
+ * fromDate and toDate. toDate is Optional, defaulting to the current date.
+ * Note that `age(birthDate)` gets the age representation of a person with the specified `birthDate`.
+ * The representation logic follows the guideline here:
+ * https://webarchive.nationalarchives.gov.uk/ukgwa/20160921162509mp_/http://systems.digital.nhs.uk/data/cui/uig/patben.pdf
+ * (See Tables 7 and 8)
+ *
+ * @param fromDate The left bound of the duration.
+ * @param toDate Optional. The right bound of the duration. Defaults to the current date
  * @returns A human-readable string version of the age.
  */
-export function age(dateString: string): string {
-  // Different from npm packages such as https://www.npmjs.com/package/timeago
+export function age(fromDate: dayjs.ConfigType, toDate: dayjs.ConfigType = dayjs()): string {
+  const to = dayjs(toDate);
+  const from = dayjs(fromDate);
 
-  // First calculate the age in years
-  const today = new Date();
-  const birthDate = new Date(dateString);
-  const monthDifference = today.getUTCMonth() - birthDate.getUTCMonth();
-  const dateDifference = today.getUTCDate() - birthDate.getUTCDate();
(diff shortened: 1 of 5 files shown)


packages/framework/esm-utils/src/age-helpers.ts
```diff
@@ -1,101 +1,115 @@
 /** @module @category Utility */
+import dayjs from 'dayjs';
 import { getLocale } from './omrs-dates';
 
 /**
- * Gets the number of days in the year of the given date.
- * @param date The date to compute the days within the year.
- * @returns The number of days.
- */
-export function daysIntoYear(date: Date) {
```
chibongho, line 9: I did a search on github and these exported functions don't seem used anywhere else.

packages/framework/esm-utils/src/age-helpers.ts
```diff
@@ -1,101 +1,115 @@
...
- * @param dateString The stringified date.
+ * Gets a human readable and locale supported representation of a duration between 2 dates,
+ * fromDate and toDate. toDate is Optional, defaulting to the current date.
+ * Note that `age(birthDate)` gets the age representation of a person with the specified `birthDate`.
+ * The representation logic follows the guideline here:
+ * https://webarchive.nationalarchives.gov.uk/ukgwa/20160921162509mp_/http://systems.digital.nhs.uk/data/cui/uig/patben.pdf
+ * (See Tables 7 and 8)
+ *
+ * @param fromDate The left bound of the duration.
+ * @param toDate Optional. The right bound of the duration. Defaults to the current date
```
ibacher, line 14: I don't find `fromDate` or `toDate` very intuitive here. I _think_ the variables are something like `birthDate` and `ageAtDate` or something like that.

packages/framework/esm-utils/src/age-helpers.ts
```diff
@@ -1,101 +1,115 @@
...
-  // First calculate the age in years
-  const today = new Date();
-  const birthDate = new Date(dateString);
-  const monthDifference = today.getUTCMonth() - birthDate.getUTCMonth();
-  const dateDifference = today.getUTCDate() - birthDate.getUTCDate();
-  let age = today.getUTCFullYear() - birthDate.getUTCFullYear();
-  if (monthDifference < 0 || (monthDifference === 0 && dateDifference < 0)) {
-    age--;
-  }
+export function age(fromDate: dayjs.ConfigType, toDate: dayjs.ConfigType = dayjs()): string {
```
chibongho, line 17: I changed the function to take in the `toDate` as well since I needed it for testing.

packages/framework/esm-utils/src/age-helpers.ts
```diff
@@ -1,101 +1,115 @@
...
+    if (remainderDayDiff > 0) {
+      const day = new Intl.NumberFormat(locale, {
         style: 'unit',
         unit: 'day',
         unitDisplay: 'short',
-      }).format(totalDaysAgo);
+      }).format(remainderDayDiff);
 
-      return totalDaysAgoStr;
+      return isLtr ? week + ' ' + day : day + ' ' + week;
```
ibacher, line 64: It might be worth using the [formatjs `Intl.DurationFormat` polyfill](https://formatjs.io/docs/polyfills/intl-durationformat) for this instead of rolling our own...

Ditto for other places we're doing similar things.

packages/framework/esm-utils/src/age-helpers.ts
```diff
@@ -1,101 +1,115 @@
...
- * Checks if two dates are representing the same day.
- * @param firstDate The first date.
- * @param secondDate The second date.
- * @returns True if both are located on the same day.
- */
-export function isSameDay(firstDate: Date, secondDate: Date) {
-  const firstISO = firstDate.toISOString();
-  const secondISO = secondDate.toISOString();
-  return firstISO.slice(0, firstISO.indexOf('T')) === secondISO.slice(0, secondISO.indexOf('T'));
-}
```
ibacher, line 29: I don't quite get why we're removing these? Since they're exports, isn't there the chance  that we're relyiing on them elsewhere?

ibacher (changes requested):
Code seems fine. I think we need better parameter names, though, before mergin this in.

packages/framework/esm-utils/src/age-helpers.ts
chibongho: > It might be worth using the [formatjs `Intl.DurationFormat` polyfill](https://formatjs.io/docs/polyfills/intl-durationformat) for this instead of rolling our own...
> 
> Ditto for other places we're doing similar things.

Thanks. Much simpler with it.

packages/framework/esm-utils/src/age-helpers.ts
chibongho: Changed to `birthDate` and `currentDate` (optional) instead.

packages/framework/esm-utils/src/age-helpers.ts
chibongho: They looked really specific to the prior `age()` implementation, and was likely marked as export by mistake. I did a search on frontend code in `site:openmrs` on github and we don't seem to be using these elsewhere.

ibacher (approved):

ibacher:
Thanks @chibongho! This is great!

packages/framework/esm-utils/src/age-helpers.ts
ibacher: Excellent! All I wanted to know was that we at least checked!

PR #1101, opened 2024-07-31
(fix) Config system and Implementer Tools both crash when 'null' passed for Array value

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

Passing 'null' for a config value which is supposed to be an Array causes both the Implementer Tools and the config system to crash. This fixes the problem, making them issue the 'wrong type' warning as expected.

### DIFF
--- a/packages/apps/esm-implementer-tools-app/src/implementer-tools.component.tsx
+++ b/packages/apps/esm-implementer-tools-app/src/implementer-tools.component.tsx
@@ -57,9 +57,5 @@ function PopupHandler() {
 }
 
 export default function ImplementerTools() {
-  return (
-    <UserHasAccess privilege="coreapps.systemAdministration">
-      <PopupHandler />
-    </UserHasAccess>
-  );
+  return <PopupHandler />;
 }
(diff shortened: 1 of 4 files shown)


packages/apps/esm-implementer-tools-app/src/implementer-tools.component.tsx
```diff
@@ -57,9 +57,5 @@ function PopupHandler() {
 }
 
 export default function ImplementerTools() {
-  return (
-    <UserHasAccess privilege="coreapps.systemAdministration">
-      <PopupHandler />
-    </UserHasAccess>
-  );
+  return <PopupHandler />;
```
denniskigen, line 60: Do we not want to require this permission for accessing the implementer tools?

denniskigen (approved):
LGTM! Thanks, @brandones.

packages/apps/esm-implementer-tools-app/src/implementer-tools.component.tsx
brandones: OMG that was not intended to be included in the change. That would be bad. Great catch, thank you @denniskigen

PR #1103, opened 2024-07-31
(fix) O3-3691: The workspace family store shouldn't reset if workspace of same family is launched

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
The issue comes from [this line](https://github.com/openmrs/openmrs-esm-core/blob/98118452809d1d8217059e24d4dc37096231d500/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L305) which does check that the workspaces remaining do have the same sidebar family name and doesn’t check the fact that the new workspace to be opened is of the same family, hence it shouldn’t clear the workspace.

The reason behind this is that when we launch 2nd workspace, it will first close the 1st workspace and as per the condition linked above, it will reset the store, which is not ideal. We need a condition that if the new workspace is opened of the same sidebar family and the previous workspace is closing, the store shouldn't clear itself.

Addition: Added tests for workspace renderer to test if the valid props are being passed into the workspace or not.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-3691

## Other
Fun fact:
This issue would've been caught earlier by my unit tests if I hadn't added [this line](https://github.com/openmrs/openmrs-esm-core/blob/98118452809d1d8217059e24d4dc37096231d500/packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts#L512) in a test case in my previous PR #1094

### DIFF
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-renderer.test.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-renderer.test.tsx
@@ -27,7 +27,7 @@ describe('WorkspaceRenderer', () => {
     });
     render(
       <WorkspaceRenderer
-        // @ts-ignore
+        // @ts-ignore The workspace is of type OpenWorkspace and not all properties are required
         workspace={{
           closeWorkspace: mockedCloseWorkspace,
           name: 'workspace-name',
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-renderer.test.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-renderer.test.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { WorkspaceRenderer } from './workspace-renderer.component';
+import { getWorkspaceFamilyStore, OpenWorkspace } from '../workspaces';
+import Parcel from 'single-spa-react/parcel';
+
+const mockFn = jest.fn();
+
+jest.mock('single-spa-react/parcel', () =>
+  jest.fn((props) => {
+    mockFn(props);
+    return <div data-testid="mocked-parcel" />;
+  }),
+);
+
+describe('WorkspaceRenderer', () => {
+  it('should render workspace', async () => {
+    const mockedCloseWorkspace = jest.fn();
+    const mockedCloseWorkspaceWithSavedChanges = jest.fn();
+    const mockedPromptBeforeClosing = jest.fn();
+    const mockedSetTitle = jest.fn();
+    const mockedLoadFn = jest.fn().mockImplementation(() => Promise.resolve({ default: 'file-content' }));
+    getWorkspaceFamilyStore('test-sidebar-family')?.setState({
+      // Testing that the workspace family state should be overrided by additionalProps
+      foo: false,
+      workspaceFamilyState: {},
+    });
+    render(
+      <WorkspaceRenderer
+        // @ts-ignore
+        workspace={{
+          closeWorkspace: mockedCloseWorkspace,
+          name: 'workspace-name',
+          load: mockedLoadFn,
+          title: 'Workspace title',
+          closeWorkspaceWithSavedChanges: mockedCloseWorkspaceWithSavedChanges,
+          promptBeforeClosing: mockedPromptBeforeClosing,
+          setTitle: mockedSetTitle,
+          additionalProps: {
+            foo: 'true',
+          },
+          sidebarFamily: 'test-sidebar-family',
+        }}
+        additionalPropsFromPage={{ bar: 'true' }}
+      />,
+    );
+
+    expect(screen.getByText('Loading ...')).toBeInTheDocument();
+    expect(mockedLoadFn).toHaveBeenCalled();
+    await screen.findByTestId('mocked-parcel');
+
+    expect(mockFn).toHaveBeenCalledWith({
+      config: 'file-content',
+      mountParcel: undefined,
+      closeWorkspace: mockedCloseWorkspace,
+      closeWorkspaceWithSavedChanges: mockedCloseWorkspaceWithSavedChanges,
+      promptBeforeClosing: mockedPromptBeforeClosing,
+      setTitle: mockedSetTitle,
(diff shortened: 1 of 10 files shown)


packages/framework/esm-styleguide/src/workspaces/container/workspace-renderer.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    const mockedSetTitle = jest.fn();
+    const mockedLoadFn = jest.fn().mockImplementation(() => Promise.resolve({ default: 'file-content' }));
+    getWorkspaceFamilyStore('test-sidebar-family')?.setState({
+      // Testing that the workspace family state should be overrided by additionalProps
+      foo: false,
+      workspaceFamilyState: {},
+    });
+    render(
+      <WorkspaceRenderer
+        // @ts-ignore
```
ibacher, line 30: Why are we using @ts-ignore here? (`// @ts-ignore` should generally have a comment explaining its use).

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -24,6 +24,14 @@ export interface CloseWorkspaceOptions {
    * @returns void
    */
   onWorkspaceClose?: () => void;
+  /**
+   * If set to true, the workspace family store will be cleared when the workspace is closed. Defaults to true.
+   *
+   * If set to false, the workspace family store will not be cleared when the workspace is closed. This happens when the new workspace is of the same sidebar family as the current workspace.
+   *
+   * @default true
```
ibacher, line 32:
```suggestion
  /**
   * Controls whether the workspace family store will be cleared when this workspace is closed. Defaults to true except when opening a new workspace of the same family.
   *
   * @default true
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -117,13 +125,17 @@ function promptBeforeLaunchingWorkspace(
...
   const { name, additionalProps } = newWorkspaceDetails;
+  const newWorkspaceRegistration = getWorkspaceRegistration(name);
 
   const proceed = () => {
     workspace.closeWorkspace({
       ignoreChanges: true,
       // Calling the launchWorkspace again, since one of the `if` case
       // might resolve, but we need to check all the cases before launching the form.
       onWorkspaceClose: () => launchWorkspace(name, additionalProps),
+      // This condition is added here to check if the new workspace is of the same sidebar family as the current workspace.
```
ibacher, line 136:
```suggestion
```

ibacher (changes requested):
Thanks @vasharma05! A couple of minor things. Overall question though: doesn't this remove the check for existing open workspaces while fixing a bug when opening a new workspace?

brandones:
I would rather the family workspace store stayed out of the API entirely. There's no reason that workspace client code should know about family workspace stores, and if you expose that in the API, you have to explain the whole thing. We can avoid that by creating a new function, called something like `closeWorkspaceInternal`, which takes the `clearFamilyStore` option, and is called by both `closeWorkspace` and `promptBeforeLunchingWorkspace`. `closeWorkspace` would basically become a thin wrapper around `closeWorkspaceInternal`.

vasharma05:
Hi @brandones @ibacher, requesting your review here.
Thanks!

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -280,18 +292,26 @@ export function cancelPrompt() {
...
  */
-export function closeWorkspace(
-  name: string,
-  options: CloseWorkspaceOptions = {
-    ignoreChanges: false,
-    onWorkspaceClose: () => {},
-  },
-): boolean {
+export function closeWorkspace(name: string, options: CloseWorkspaceOptions = {}) {
+  closeWorkspaceInternal(name, { ...options, clearWorkspaceFamilyStore: true });
```
brandones, line 306:
```suggestion
  return closeWorkspaceInternal(name, { ...options, clearWorkspaceFamilyStore: true });
```

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -499,7 +500,50 @@ describe('workspace system', () => {
...
+      });
+      const workspaceFamilyStore = getWorkspaceFamilyStore('ward-patient-sidebar');
+      expect(workspaceFamilyStore).toBeTruthy();
+      expect(workspaceFamilyStore?.getState()?.['foo']).toBe(true);
+      closeWorkspace('ward-patient-workspace');
+      expect(workspaceFamilyStore?.getState()?.['foo']).toBeUndefined();
+      expect(workspaceFamilyStore?.getState()).toStrictEqual({});
+    });
+
+    it('should not clear workspace family store if the workspace is closed and `clearWorkspaceFamilyStore` option is passed false', async () => {
```
brandones, line 524: I would rather we just test that the props are maintained if a workspace of the same family is launched. That's what we care about; the `clearWorkspaceFamilyStore` stuff is implementation details. And then we don't even have to export the internal function.

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -559,25 +595,134 @@ describe('workspace system', () => {
...
+       *
+       * function closeWorkspace(
+       *  workspaceName: string,
+       *  options?: CloseWorkspaceOptions={}
+       * ) {
+       *  options = {...defaultOptions, ...options}
+       * }
+       *
+       */
+
```
brandones, line 722:
```suggestion
      // test that default options are interpolated when providing options to `closeWorkspace`
```

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
vasharma05: True @brandones!
I verified that the scenario is being handled by the other test for validating that the workspace is not cleared when opening new workspace of same type.
Thanks!

ibacher (approved):

PR #1108, opened 2024-08-02
(feat) O3-3708: Add a safer expression runner

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

There are many places in the application where it would be helpful to allow implementers, form designers, etc. to use Javascript expressions to control various behaviour, e.g., hiding or showing widgets and extensions based on logic that we don't want to hard-code in the application. Currently, the supported way of doing this is something like this (from the form engine):

```ts
Function(...Object.keys(expressionContext), `"use strict"; return (${expression})`).call(
  undefined,
  ...Object.values(expressionContext),
 );
```

Essentially creating a single-use function that evaluates an arbitrary JS expression. This technique is substantially safer than just calling `eval()`; however it still suffers from some risk of security issues, specifically in that any CSP for OpenMRS must allow these type of dynamic function executions, which opens a door that third-party scripts could exploit to define an execute arbitrary code. Additionally, it allows expressions to be written and executed which may have harmful side-effects, e.g., by accessing the `window` object or other globals.

This PR introduces a new approach to the problem by implementing a simple Javascript expression evaluator based on the [jsep](https://ericsmekens.github.io/jsep/) parser with several plugins. It supports the bulk of Javascript that we should need for such use-cases via a (hopefully) simple API:

```ts
evaluate('myValue * 2', { myValue: 10 }); // produces 20
```

Although, it also supports more realistic scenarios like (modified from the form engine tests):

```ts
evaluateAsBoolean(
  "includes(referredToPreventionServices, '88cdde2b-753b-48ac-a51a-ae5e1ab24846') && !includes(referredToPreventionServices, '1691AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA')",
  {
      referredToPreventionServices: ['88cdde2b-753b-48ac-a51a-ae5e1ab24846'],
      includes: <T>(collection: Array<T>, value: T) => Array.prototype.includes.call(collection, value),
  },
),
```

The test suite should give an idea of the features supported.

Finally for use-cases where an expression needs to be evaluated multiple times (e.g., on each render), there is a `compile()` function which takes an expression, runs it through jsep and returns the resulting AST. This AST can be re-used for multiple function calls and should eliminate whatever overhead there is to parsing (though its pretty minimal). E.g.,

```ts
const expr = compile('myValue * 2');

evaluate(expr, { myValue: 10 }); // produces 20
evaluate(expr, { myValue:5 }); // produces 10
```

In addition to adding the engine, this PR also adds an `expression` property to the `Display conditions` configuration for extensions. If there is an expression defined, it will be evaluated with access to a parameter called `session` that holds the current `Session` object. This can be used to, e.g., hide extensions based on the user's currently logged-in location or their user name, whether or not they have a current provider account, etc.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3708

## Other

**Known limitations:**

* jsep was initially written in 2013 and does not have full support for more modern versions of ECMAScript. So far, the only limitation I've come across is an inability to support BigInt literals.
* The expression evaluator only evaluates Javascript expressions, so it does not support assignment operators or blocks of code.
* The evaluator intentionally does not implement all possible features including:
  * Only `Date` and `RegExp` objects can be constructed with `new` in an expression
  * None of the bit-shifting or bit-wise operations are implemented
  * The available `Object` global only implements a subset of functions (`assign`, `fromEntries`, `hasOwn`, `keys`, `values` and `is`).
  * The `Function` global is not available. Arrow functions can be used, though.
  * Object expressions are not supported (e.g. `{ a: 1, b:2 }[b]`; you can access properties of objects supplied as variables however)
  * Expression results can only be strings, numbers, booleans, Dates, and null or undefined by default. It is possible to limit this further, and theoretically to support additional types, but not all types can be created due to the above restrictiions.
* Providing access to libraries, e.g., dayjs, likely requires importing the whole object, i.e., `import * as dayjs from 'dayjs';` then passing `dayjs` into the variables object.


ibacher:
@samuelmale If you have some time, could you take a look at this PR and let me know if you think it's sufficient to be plugged into the form engine.

ibacher:
When trying this with the expression `session?.user?.name?.contains('Jared')`, I get an error "Error while evaluating expression session?.user?.name?.contains('Jared') No function named [object Object] is defined in this context" instead of "Error while evaluating expression session?.user?.name?.contains('Jared') No function named 'contains' is defined in this context".

samuelmale:
I like the idea of having a global expression runner. 

>it will be evaluated with access to a parameter called session that holds the current Session object. This can be used to, e.g., hide extensions based on the user's currently logged-in location or their user name, whether or not they have a current provider account, etc.

Do we plan on having the ability to extend the session object or context to include objects like the current "patient" or "visit" in relevant routes? A good use-case would be mounting an extension for a patient based on gender, age etc.

>could you take a look at this PR and let me know if you think it's sufficient to be plugged into the form engine.

For the most part, this seems pluggable into the form-engine. The only concerns are around supporting async expressions. 

#### How we currently approach async expressions in RFE
We decompose the expression and extract async fragments, we resolve the fragments in a granular manner and then reconstruct the expression with resolved refs ([see](https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/utils/expression-runner.ts#L149-L174)). This makes it possible to reference promised values within the same expression:

```js
"resolve(api.fetchSomeAsyncValue(someField)).value === someOtherField ? doSomeOtherOperationAndReturn() : runBlahBlahBlahAndReturn()"

```

Assuming the above isn't overkill and something worth supporting, we should consider exposing a parser fn that accepts an expression and returns fragments for preprocessing.

packages/framework/esm-expression-evaluator/src/evaluator.ts
```diff
@@ -0,0 +1,755 @@
...
+  }
+}
+
+function visitBinaryExpression(expression: jsep.BinaryExpression, context: EvaluationContext) {
+  let left = visitExpression(expression.left, context);
+  let right = visitExpression(expression.right, context);
+
+  switch (expression.operator) {
+    case '+':
+      return left + right;
```
brandones, line 416: What is JSEP doing if we're doing all this?

ibacher: Its just parsing the code to an AST, which is what allows this be written as a series of visitors; it also handles things like the order of operations and parsing ambiguities (e.g., proper handling for `a + -b`). E.g., `1 + 1` gets turned into a series of objects like:

```
{
  type: 'BinaryExpression',
  operator: '+',
  left: {
    type: 'Literal'
    value: 1
    raw: '1'
  },
  right: {
    type: 'Literal'
    value: 1
    raw: '1'
  }
}
```

While `obj.b() + arr[4](c)` becomes something like:


```
{
  type: 'BinaryExpression'
  operator: '+',
  left: {
    type: 'CallExpression'
    arguments: []
    callee: {
      type: 'MemberExpression'
      computed: false
      object:
          type: 'Identifier'
          name: 'obj'
      property:
          type: 'Identifier'
          name: 'b'
    }
  },
  right: {
    type: 'CallExpression',
    arguments: [
      {
        type: 'Identifier'
        name: 'c'
      }
    ],
    callee: {
      type: 'MemberExpression',
      computed: true,
      object: {
          type: 'Identifier'
          name: 'arr'
      },
      property: {
        type: 'Literal',
        value: 4,
        raw: '4'
      }
    }
  }
}
```

This makes evaluation of an expression into a matter of walking the AST and converting the innermost expression into a concrete value, then working outwards, etc., until the outermost expression has a concrete value. It probably saves us around 1000 LOC we don't have to write and test.

I picked JSEP because:

1. Its small
2. The AST it generates is reasonably close to [ESTree's](https://github.com/estree/estree) output for es5 (estree being basically the standard AST when working with JS), which means it should be replaceable with minimal effort
3. It's limited to just parsing the expression subpart of JavaScript
4. It's extensible via plugins if we need to support something not currently supported.

packages/framework/esm-expression-evaluator/src/evaluator.ts
```diff
@@ -0,0 +1,755 @@
...
+  if (typeof expression !== 'string' && (typeof expression !== 'object' || !expression || !('type' in expression))) {
+    throw `Unknown expression type ${expression}. Expressions must either be a string or pre-compiled string.`;
+  }
+
+  if (typeof variables === 'undefined' || variables === null) {
+    variables = {};
+  }
+
+  const context = createSynchronousContext(variables);
+  const result = visitExpression(typeof expression === 'string' ? jsep(expression) : expression, context);
```
brandones, line 246: So `visit` in this file is used to mean `evaluate`, basically?

brandones (approved):
LGTM. Also the test suite looks excellent, which gives me confidence.

packages/framework/esm-expression-evaluator/src/evaluator.ts
ibacher: Yes; the naming convention here is basically because this is (roughly) the visitor pattern.

PR #1113, opened 2024-08-14
(feat) O3-3773 Support Summary Tiles component in esm-styleguide commons-lib 

kajambiya:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR introduces the summary tile/card summary component to esm-styleguide.

## Screenshots
Screenshot 2024-08-14 at 14 26 52

### DIFF
--- a/packages/framework/esm-styleguide/src/tile/summary-card-component.tsx
+++ b/packages/framework/esm-styleguide/src/tile/summary-card-component.tsx
@@ -1,0 +1,63 @@
+import React, { useMemo } from 'react';
+import { Tile, Column } from '@carbon/react';
+import styles from './tile.scss';
+
+export interface SummaryCardProps {
+  columns: Array<SummaryCardColumn>;
+  headerTitle: string;
+  maxRowItems?: number;
+}
+
+export interface SummaryCardColumn {
+  header: string;
+  value: string;
+  summary?: string;
+}
+
+export const SummaryCard: React.FC<SummaryCardProps> = ({ columns, headerTitle, maxRowItems }: SummaryCardProps) => {
+  const groupedColumns = useMemo(() => {
+    if (maxRowItems && columns.length > maxRowItems) {
+      let groups: SummaryCardColumn[][] = [];
+      for (let i = 0; i < columns.length; i += maxRowItems) {
+        groups.push(columns.slice(i, i + maxRowItems));
+      }
+      return groups;
+    }
+    return [columns];
+  }, [columns, maxRowItems]);
+
+  return (
+    <Tile className={styles.tile}>
+      <div className={styles.cardTitle}>
+        <h4 className={styles.title}> {headerTitle} </h4>
+      </div>
+      {maxRowItems ? (
+        groupedColumns.map((group) => (
+          <Column className={styles.columnContainer}>
+            {group.map((column) => (
+              <SummaryItem column={column} />
+            ))}
+          </Column>
+        ))
+      ) : (
+        <Column className={styles.columnContainer}>
+          {columns.map((column) => (
+            <SummaryItem column={column} />
+          ))}
+        </Column>
+      )}
+    </Tile>
+  );
+};
+
+function SummaryItem({ column }: { column: SummaryCardColumn }) {
+  return (
+    <div className={styles.tileBox}>
+      <div className={styles.tileBoxColumn}>
+        <span className={styles.tileTitle}> {column.header} </span>
+        <span className={styles.tileValue}>{column.value}</span>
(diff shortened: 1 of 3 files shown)


packages/framework/esm-styleguide/src/tile/tile.scss
```diff
@@ -0,0 +1,112 @@
+@use '@carbon/styles/scss/colors';
+@use '@carbon/styles/scss/spacing';
+@use '@carbon/styles/scss/type';
+@import '../root.scss';
```
ibacher, line 4: Don't do this. Follow the guidelines Dennis suggested yesterday on the call (or look through the developer docs)

packages/framework/esm-styleguide/src/tile/summary-card-component.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return [columns];
+  }, [columns, maxRowItems]);
+
+  return (
+    <Tile className={styles.tile}>
+      <div className={styles.cardTitle}>
+        <h4 className={styles.title}> {headerTitle} </h4>
+      </div>
+      {maxRowItems ? (
+        groupedColumns.map((group) => (
```
ibacher, line 35: Any time you're using a `map` you should supply a `key` to the component, either derived from the data or at least using the index of the item.

packages/framework/esm-styleguide/src/tile/tile.scss
```diff
@@ -0,0 +1,112 @@
...
+  @extend .productiveHeading03;
+}
+
+.cardTitle {
+  display: flex;
+  justify-content: space-between;
+  padding: spacing.$spacing-04 0 spacing.$spacing-04 spacing.$spacing-05;
+}
+
+.cardTitle > h4:after {
```
ibacher, line 16: The `h4` element has a class, so it seems like this could be better written using `.title:after`

packages/framework/esm-styleguide/src/tile/tile.scss
```diff
@@ -0,0 +1,112 @@
...
+  content: '';
+  display: block;
+  width: 2rem;
```
ibacher, line 19: Please use the Carbon tokens for defining spacing wherever possible. This helps ensure that the UI remains consistent with Carbon guidelines.

packages/framework/esm-styleguide/src/tile/tile.scss
```diff
@@ -0,0 +1,112 @@
...
+  padding-top: 0.188rem;
+  border-bottom: 0.375rem solid $brand-teal-01;
+}
+
+.tile {
+  height: 100%;
+  padding: 1px 0 16px 16px;
```
ibacher, line 26: Nothing over 1px should be specified in px.

packages/framework/esm-styleguide/src/tile/tile.scss
```diff
@@ -0,0 +1,112 @@
...
+  flex-direction: row;
+}
+
+.tileBoxColumn {
+  width: 100%;
+  display: 'flex';
+  flex-direction: 'row';
+}
+
+.widgetContainer {
```
ibacher, line 59: A number of these rules seem to not be used by this component, so they probably don't belong here.

ibacher (changes requested):
* Please look through the checklist on the PR. There are items there that absolutely need to be done.
* Were you able to generate that screen shot using _this version_ of the component?
* I don't really understand the need for this component per se. Where is this intended to be used? How many times do we need it?

gracepotma:
@ibacher to answer your question: 
> I don't really understand the need for this component per se. Where is this intended to be used? How many times do we need it?

It's a useful feature to have a clinical area-specific summary within different clinical views. This was actually a make-or-break requirement for Ampath at the very beginning of O3 :) The user story is, "As a healthcare staff member focused on X condition (eg HIV clinic, MCH clinic, etc), I need to immediately see some very specific information summarized for me, so I can rapidly get the key information relevant to my specialty." Examples here from OHRI:

image

image

ibacher:
It seems then like this belongs more readily in the patient chart then, rather than the styleguide.

ebambo:
> > It seems then like this belongs more readily in the patient chart then, rather than the styleguide, more exactly, in the patient-common-lib.
> 
> My thoughts too. If there's no objection to this, I'm going to proceed and and transfer this code to patient-chart. I guess the same applies to [encounter-list](https://openmrs.atlassian.net/browse/O3-3772) as well. CC: @ebambo

No objections from my side, as long as this still works fine with the work @hadijahkyampeire is doing with the config schema.

brandones:
Ok closing this as it seems we've decided to do it in patient-chart.

PR #1114, opened 2024-08-15
(feat) - O3-3189 - add hook to handle server-side pagination

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

Most REST endpoints are server-side paginated, which means the client might need to make multiple fetch requests to get all items in a data set from an endpoint. This PR adds 2 new hooks for handling this:
- `useServerPagination` is similar to existing `usePagination` hook and is intended to be used for UIs that offer paginated view of data (like a table), but it lazy-loads the data of each page as needed
- `userServerInfinite` is intended to be used for UIs that offer infinite scrolling / loading.

## Screenshots

Here's a video of a test table that loads test data from `getTestData()` (included in this PR), intentionally delayed with 1 second of lag per fetch. The table uses the `useServerPagination` hook to handle loading and navigation of different pages in the table. The table is clearly missing CSS polish, but it demonstrates handling of pagination. 
https://github.com/user-attachments/assets/44fd5abe-a004-4086-9e40-4bfd15bf2bd4

(For some reason, the video didn't capture rendering of browser dropdown menus. Those definitely rendered when I was recording the video)

## Related Issue

[link to ticket](https://openmrs.atlassian.net/issues/O3-3189?jql=text%20~%20%22pagination%2A%22%20AND%20project%20%3D%20O3)

## Other

The code for the table included in the video is not included in this PR. I do think it'd be nice to check it in somewhere (when it's more polished) so people can easily play around with it. Perhaps `esm-styleguide`?

### DIFF
--- a/packages/framework/esm-react-utils/src/useServerInfinite.ts
+++ b/packages/framework/esm-react-utils/src/useServerInfinite.ts
@@ -3,6 +3,7 @@ import { type FetchResponse, openmrsFetch } from '@openmrs/esm-api';
 import useSWRInfinite from 'swr/infinite';
 import { type OpenMRSPaginatedResponse } from './useServerPagination';
 import { type KeyedMutator } from 'swr';
+import { useCallback } from 'react';
 
 export interface UseServerInfiniteReturnObject<T> {
   /**
@@ -51,7 +52,7 @@ export interface UseServerInfiniteReturnObject<T> {
 /**
  * Most REST endpoints that return a list of objects, such as getAll or search, are server-side paginated.
  * The server limits the max number of results being returned, and multiple requests are needed to get the full data set
- * if its size exceed this limit.
+ * if its size exceeds this limit.
  * The max number of results per request is configurable server-side
  * with the key "webservices.rest.maxResultsDefault". See: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25469882/REST+Module
  *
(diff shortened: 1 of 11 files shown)


packages/framework/esm-react-utils/src/useServerPagination.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  const totalPages = Math.ceil(totalCount.current / pageSize);
+
+  const goTo = useCallback(
+    (page: number) => {
+      if (0 < page && page <= totalPages) {
+        setCurrentPage(page);
+      }
+    },
```
ibacher, line 73: Should report some kind of error is the user calls this with a value out of bounds?

packages/framework/esm-react-utils/src/useServerPagination.ts
```diff
@@ -0,0 +1,98 @@
...
+    [url, pageSize, currentPage, totalPages],
```
ibacher, line 74: I kind of get having `url` here, but is `pageSize` here useful?

ibacher: Also the next couple of places...

packages/framework/esm-react-utils/src/useServerInfinite.ts
```diff
@@ -0,0 +1,109 @@
...
+    return data?.data?.links?.find((link) => link.rel == 'next');
+  };
+
+  const getKey = (pageIndex: number, previousPageData: FetchResponse<OpenMRSPaginatedResponse<T>>) => {
+    if (pageIndex == 0) {
+      return url;
+    } else {
+      return getNextUri(previousPageData)?.uri ?? null;
+    }
+  };
```
ibacher, line 91: Might be a good idea to wrap these in `useCallback()`s?

ibacher (commented):
Basically LGTM. A couple of questions. I also find something slightly off about the "useServer..." naming, but I don't really have a better suggestion.

mseaton (commented):
I fully support this, and also anything we can do to ensure that our consumers default to getting the right (full) set of results.  Defer to others on the nuances of the PR.  Thanks @chibongho !

packages/framework/esm-react-utils/src/useServerInfinite.ts
chibongho: Yeah, I thought about it. The `useSWRInfinite` hook seems to function fine without needing the `getKey()` function memoized, but that's not true for all hooks that take in other callbacks. I'll add it in.

chibongho:
Added new changes per PR comments.

I'm honestly not a big fan of the names either, especially `useServerInfinite`, but I don't have anything better.

packages/framework/esm-react-utils/src/useServerInfinite.ts
ibacher: I figured it might be. Really, I'm more concerned about stuff that gets returned to the client because our code usually isn't written in a way that would be helpful.

ibacher (approved):

chibongho:
> as all OpenMRS REST responses are potentially paginated, we will likely want to replace a lot of the existing calls with this new hook?

Yeah. At least a few usages of `usePagination` are not doing this properly, and probably other places too that fetch array of items server-side. Filed https://openmrs.atlassian.net/browse/O3-3806.

chibongho:
Merging this in, feel free to comment post-hoc.

PR #1115, opened 2024-08-15
(refactor) Refactor PatientBannerContactDetails component

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR refactors the PatientBannerContactDetails component as follows:

- Removes the `usePatient`, `usePatientAttributes` and `usePatientContactAttributes` hooks. We don't need the `usePatient` hook to resolve the patient object because it's already getting passed down as a prop to `PatientBannerContactDetails`. We don't need `usePatientAttributes` and `usePatientContactAttributes` to fetch person attributes for populating the `Contact Details` section of the Patient Banner because we can already extract these from the FHIR patient object's `telecom` property.
- Cleans up the Address UI by removing the [use](https://hl7.org/fhir/R4/valueset-contact-point-use.html) property from the listed address. We're only using it to fetch the patient's `home` address, so it probably shouldn't be listed as part of the address. If we want to clarify to the user that the address shown is the patient's home address, we could consider changing the section label to `Home address` instead of `Address` (as is the case presently).

This PR is related to https://github.com/openmrs/openmrs-esm-patient-chart/pull/1961, which tried to fix the issue of a patient's telephone number not showing up in the Contact Details UI. Per [@ibacher's suggestion](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1961#pullrequestreview-2238472131), we shouldn't need to fetch contact attributes from the backend. The FHIR patient object already includes the phone number in its `telecom` field.

## Screenshots

https://github.com/user-attachments/assets/b2779fb3-9f8c-412f-8705-961b9fc6df0b

## Other

[All references](https://github.com/search?q=org%3Aopenmrs%20%3CPatientBannerContactDetails%20&type=code) to `PatientBannerContactDetails` will need to switch over from using the `patientId` prop to the full `patient` object.


ibacher (approved):

denniskigen:
The patient search app has the only occurrence of `PatientBannerContactDetails` that passes in a REST patient object https://github.com/openmrs/openmrs-esm-patient-management/blob/859a44979436703a8fa0d2a8ef70d25184bd43ee/packages/esm-patient-search-app/src/patient-search-page/patient-banner/banner/patient-banner.component.tsx#L125

ibacher:
Ah... right. That makes sense. IIRC, I added a search URL for patients to FHIR2 that duplicates the OpenMRS search logic, but I'm not sure if its a good idea to switch over (it would give you FHIR patients though).

denniskigen:
Yeah, that nullifies this refactor to some degree.

ibacher:
Hmmm... Maybe we go back to the version before I said anything?

denniskigen:
Closing owing to https://github.com/openmrs/openmrs-esm-core/pull/1115#issuecomment-2291268342.

PR #1116, opened 2024-08-16
(feat) O3-3774: Add implementation for the Nav Group

hadijahkyampeire:
# Requirements

## For changes to apps

## If applicable

## Summary
- Moved the implementation of Nav Groups to the style guide.

## Screenshots
I couldn't link the framework to the patient chart so I tested with the primary navigation app instead and below is a demo.

https://github.com/user-attachments/assets/76f9e647-7963-409e-b261-a205ee0663aa

Screenshot 2024-08-22 at 02 13 57

## Related Issue
- https://openmrs.atlassian.net/browse/O3-3774


packages/framework/esm-styleguide/src/left-nav/CreateDashboardGroup.tsx
ibacher, line 1: These components shouldn't be in `left-nav`. The `left-nav` component is it's own thing.

packages/framework/esm-styleguide/src/left-nav/nav-group.ts
```diff
@@ -0,0 +1,18 @@
+/** @module @category UI */
+import { createGlobalStore } from '@openmrs/esm-state';
+import { createUseStore } from '@openmrs/esm-react-utils';
+
+export * from './generic-nav-group.component';
```
ibacher, line 5: Why are we exporting stuff from here?

packages/framework/esm-styleguide/src/left-nav/nav-group.ts
```diff
@@ -0,0 +1,18 @@
...
+
+interface NavGroupStoreState {
+  navGroups: string[];
+}
+
+const navGroupStore = createGlobalStore<NavGroupStoreState>('nav-groups', { navGroups: [] });
+
+export function registerNavGroup(slotName: string) {
```
ibacher, line 13: Why do we need this store at all?

packages/framework/esm-styleguide/src/left-nav/generic-nav-group.component.tsx
```diff
@@ -0,0 +1,32 @@
+/** @module @category UI */
+import React from 'react';
+import { useConfig } from '@openmrs/esm-react-utils';
+import { Type } from '@openmrs/esm-config';
+import { DashboardGroupExtension } from './DashboardGroupExtension';
+
+export const genericNavGroupConfigSchema = {
```
ibacher, line 7: This needs to be registered somewhere to have any kind of effect.

ibacher (changes requested):
I think this PR both extracts too much and too little. The only things we should need to move (AFAICT) are the generic-nav-group and maybe the generic-dashboard. I don't really see the value of or how anyone would even make use of the DashboardGroup stuff generically (it works in the chart, let's keep it there and simplify what this PR needs to do).

packages/framework/esm-styleguide/src/left-nav/generic-nav-group.component.tsx
ibacher: To use this as an extension, it also needs to actually be registered with the extension system in some way.

hadijahkyampeire: Yes, I have registered it in the index of the style guide, thanks @ibacher

packages/framework/esm-styleguide/src/left-nav/nav-group.ts
hadijahkyampeire: I wanted to export all from one file, but I have fixed that.

packages/framework/esm-styleguide/src/left-nav/index.tsx
```diff
@@ -2,7 +2,6 @@
 import React from 'react';
 import { ExtensionSlot, useStore } from '@openmrs/esm-react-utils';
 import { createGlobalStore } from '@openmrs/esm-state';
-import type { SideNavProps } from '@carbon/react';
```
ibacher, line 5: ?

packages/framework/esm-styleguide/src/nav-group/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import React from 'react';
+import { ExtensionSlot, useConfig } from '@openmrs/esm-react-utils';
+import { Type } from '@openmrs/esm-config';
+import { Accordion } from '@carbon/react';
+import { AccordionItem } from '@carbon/react';
+
+export const navGroupFeatureName = 'Nav Group';
+
+export const navGroupConfigSchema = {
+  title: {
```
ibacher, line 11: Instead of `title` we should probably specify this as a translation key...

packages/framework/esm-styleguide/src/nav-group/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+};
+
+export interface NavGroupConfig {
+  title: string;
+  slotName: string;
+  isExpanded?: boolean;
+}
+
+interface NavGroupProps {
```
ibacher, line 34:
```suggestion
export interface NavGroupProps {
```

packages/framework/esm-styleguide/src/nav-group/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    _default: 'my-group-nav-slot',
+  },
+  isExpanded: {
+    _type: Type.Boolean,
+    _description: 'The boolean to determine whether the nav group is expanded or not.',
+    _default: true,
+  },
+};
+
+export interface NavGroupConfig {
```
ibacher, line 28:
```suggestion 
interface NavGroupConfig {
```

packages/framework/esm-styleguide/src/left-nav/index.tsx
hadijahkyampeire: This was some unused import I came across and just cleaned it up

packages/framework/esm-styleguide/src/nav-group/index.tsx
hadijahkyampeire: @ibacher I don't understand the above review 🤔

hadijahkyampeire: I have tried to check for some examples but I can't get any, but I have done it in this way, I don't know if this is what you meant.
```js
 title: {
    _type: Type.Object,
    _description: 'The title of the nav group.',
    _default: {
      key: 'myGroupKey',
      value: 'My Group',
    },
  },
```

ibacher: So what I said was that the `title` should specify a translation key passed through the `t()` function. What I mean is we leave this configuration as-is except change the default value to something like `myGroup` and then in the component, we have:

```jsx
const { t } = useTranslations();
//...
// t('myGroup', 'My Group')
<AccordianItem title={t(title)} />
```

An object isn't terribly useful here because the goal is to not hard-code the string in the configuration, but to allow it to be translated.

ibacher:
So two other things need to happen for this PR to work:

1. We need to actually register this extension with the extension system (try renaming the extension to global-nav-group and defining one in configuration).
2. I still think we need to move the generic dashboard stuff into the styleguide.

packages/framework/esm-styleguide/src/nav-group/index.tsx
hadijahkyampeire: ok, thanks

hadijahkyampeire:
@ibacher Most of what is in the generic-nav component is what is in the nav-group component unless if i am missing something

hadijahkyampeire:
By registering, do you mean creating a store and registering it in the store as they did in the common-lib? like this;
```js
const navGroupStore = createGlobalStore('global-nav-groups', { navGroups: [] });

export function registerNavGroup(slotName: string) {
  const store = navGroupStore.getState();
  navGroupStore.setState({ navGroups: [slotName, ...store.navGroups] });
}
```

ibacher:
> By registering, do you mean creating a store and registering it in the store as they did in the common-lib?

No. I mean actually registering the extension with the extension system, via the `registerExtension()` call at some point. For most apps, this is handled by having an appropriate entry in `routes.json`, so patient-chart does this with it's `nav-group` extension [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/627fc0e3fba80f58157a79c3a098a1a96b933469/packages/esm-patient-chart-app/src/routes.json#L138); without registering this extension with the extension system, there's no way to actually _use_ it (your video above uses the `nav-group` defined in patient chart).

This is why I commented suggesting to rename the extension `global-nav-group` for testing.

hadijahkyampeire:
@ibacher anything we haven't covered in the PR, just want to know if there is more work or just the final review.

packages/apps/esm-primary-navigation-app/src/index.ts
```diff
@@ -53,3 +54,8 @@ export const changeLanguageModal = getAsyncLifecycle(
   () => import('./components/change-language/change-language.modal'),
   options,
 );
+
+export const globalNavGroup = getSyncLifecycle(GlobalNavGroup, {
```
ibacher, line 58: Use `getAsyncLifecycle()` here please.

```suggestion
export const globalNavGroup = getSyncLifecycle(GlobalNavGroup, {
```

packages/apps/esm-primary-navigation-app/src/routes.json
```diff
@@ -28,6 +28,12 @@
       "offline": true,
       "order": 0
     },
+    {
+      "name": "global-nav-group",
```
ibacher, line 32: It's fine to call this `nav-group`. I was just using `global-nav-group` as an example to show you that there was no extension registration.

packages/framework/esm-styleguide/src/index.ts
```diff
@@ -24,6 +26,7 @@ export * from './icons';
 export * from './pictograms';
 
 defineConfigSchema('@openmrs/esm-styleguide', esmStyleGuideSchema);
+defineExtensionConfigSchema('global-nav-group-slot', globalNavGroupConfigSchema);
```
ibacher, line 29:
```suggestion
defineExtensionConfigSchema('nav-group', globalNavGroupConfigSchema);
```

ibacher: Slots are only used for _locating_ extensions. They are a space on a page. They have no _functional_ meaning in this context. Note, this needs to match the _name the extension is registered as_.

packages/framework/esm-styleguide/src/nav-group/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const globalNavGroupConfigSchema = {
+  title: {
+    _type: Type.String,
+    _description: 'The title of the nav group.',
+    _default: 'myGroup',
+  },
+  slotName: {
+    _type: Type.String,
+    _description: 'The name of the slot to create, which links can be added to.',
+    _default: 'global-nav-group-slot',
```
ibacher, line 20:
```suggestion
    _default: 'nav-group-slot',
```

hadijahkyampeire:
@ibacher do we still want to move this into core or we close it

ibacher (approved):

PR #1117, opened 2024-08-17
(fix) - O3-3684 - improve performance of age() function

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

The `age()` function refactored in this [PR](https://github.com/openmrs/openmrs-esm-core/pull/1100) uses a [`DurationFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat) polyfill that is a really expensive object to create (takes 66 - 100ms on my laptop depending on whether it's plugged in). We rely on it in the `esm-ward-app` to render the age, time since admission and time on ward for each patient card. That's about 200ms per patient on those 3 elements. On a ward with ~5 patients, the rendering is cripplingly slow.

This PR just adds caching to the `DurationFormat` object so we only get the performance penalty on the first `age()` call.

## Screenshots

Profiling before this change. Those 3 elements take 200ms to render:
image

Profiling after this change. PatientAge element takes 100ms on first render, but is reasonable faster afterwards. The other 2 time elements aren't even showing in the waterfall (probably because their render times are insignificant).
withformatjs

## Other

I initially tried a solution that does away with the polyfill altogether. It eliminates the first call penalty, but is a bit iffy with the i18n stuff. I checked it in [here](https://github.com/openmrs/openmrs-esm-core/commit/2b531bde02bd556bb45cd3bd7c4986be80e4c6e5) in case we want to not use the polyfill.

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -6644,7 +6644,7 @@ ___
 
 Most REST endpoints that return a list of objects, such as getAll or search, are server-side paginated.
 The server limits the max number of results being returned, and multiple requests are needed to get the full data set
-if its size exceed this limit.
+if its size exceeds this limit.
 The max number of results per request is configurable server-side
 with the key "webservices.rest.maxResultsDefault". See: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25469882/REST+Module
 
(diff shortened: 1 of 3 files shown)


ibacher:
Two comments: 

* We should probably cache by locale to support locale switching (specifically in the case where a user has a locale preference that’s different from the default locale for an implementation).
* Reading through some of the issues, it seems like the default locale matching algorithm (“best fit”—which is unspecified, but formatjs tries to use a the Unicode LDML algorithm which is comprehensive and aims to be correct, but slow). It may help substantially cut down the time by switching to the “lookup” algorithm via the `localeMatcher` option to the constructor. While less comprehensive, this should be sufficient for our purposes. This would likely remove much of the remaining overhead. See [the discussion here](https://github.com/formatjs/formatjs/issues/4463).

chibongho:
`localeMatcher: "lookup"` worked beautifully. < 1ms of runtime. No caching needed.

mseaton (approved):
Not sure what I'm reviewing here, but LGTM?

mogoodrich (approved):
I probably spend too long trying to understand this, but LGTM, thanks @chibongho !

PR #1119, opened 2024-08-22
(feat) O3-3857: Support extension overrides in per-slot extension config

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

The "Translation overrides" configuration stuff only works when defined at the app level because the translation namespace does not communicate any extension-specific information.

This PR fixes that so that you can provide translation overrides at the level of specific extensions-in-slots, which is useful for some of the implementation of "Clinical views" and other scenarios using "generic extension components".

## Screenshots

With this configuration:

Example Configuration

When showing the extension in the slot, we get:

Specific slot override

When showing it in a different slot we get:

General override

## Related Issue
https://issues.openmrs.org/browse/O3- 3857

## Other

This uses `___` as a field demarcator since other characters I thought of either have special handling in module names or in i18next itself.

### DIFF
--- a/packages/framework/esm-config/package.json
+++ b/packages/framework/esm-config/package.json
@@ -43,11 +43,13 @@
   "peerDependencies": {
     "@openmrs/esm-globals": "5.x",
     "@openmrs/esm-state": "5.x",
+    "@openmrs/esm-utils": "5.x",
     "single-spa": "5.x"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


denniskigen (approved):
LGTM (I haven't been able to test it out locally because webpack OOMs when I run the build jobs - but the code looks great)

PR #1125, opened 2024-08-27
(fix) Remove uses of externalModuleName in styleguide components

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR removes uses of `externalModuleName` from two components in the styleguide. AFAIK, `externalModuleName` is used when you want to load a config schema from a separate frontend module into another. In this case, both components are colocated within the style guide and reference the style guide config schema, so there's no need to use `externalModuleName`. Unrelatedly, this exposed an issue with the [useConfig stub](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-react-utils/mock.tsx#L27) in the framework mock (which is responsible for the failing tests from [this refactor](https://github.com/openmrs/openmrs-esm-patient-management/pull/1275) in Patient Management). The stub does not correctly handle `externalModuleName` and fails silently in the test environment when the component under test leverages `externalModuleName`.

### DIFF
--- a/packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
+++ b/packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
@@ -87,15 +87,13 @@ export const PageHeader: React.FC<PageHeaderProps> = (props) => {
  * ```
  */
 export const PageHeaderContent: React.FC<PageHeaderContentProps> = ({ title, illustration, className }) => {
-  const { implementationName } = useConfig<StyleguideConfigObject>({
-    externalModuleName: '@openmrs/esm-styleguide',
-  });
+  const config = useConfig<StyleguideConfigObject>();
 
   return (
     <div className={classNames(styles.pageHeaderContent, className)}>
       {illustration}
       <div className={styles.pageLabels}>
-        <p>{implementationName}</p>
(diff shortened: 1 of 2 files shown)


denniskigen:
FYI @Twiineenock

denniskigen:
Ok. This is not as straightforward as I first thought.

brandones:
> Maintain the externalModuleName option in the useConfig hook so that the usePatientPhoto hook knows to obtain the patientPhotoConceptUuid property from the styleguide config. The benefit of this approach is that there's one source of truth for patientPhotoConceptUuid and that the concerns are nicely centralized in one place.

Yeah, this is why we did it this way, and how I suggest we keep doing it

ibacher:
The issue with doing away with the `externalModuleName` is that `useConfig()` defaults [to the context-provided](https://github.com/openmrs/openmrs-esm-core/blob/886629e665c26749aba4b0f6da2163405f33cc6e/packages/framework/esm-react-utils/src/useConfig.ts#L143) module, which will be whatever extension mounts the styleguide component. So effectively, this won't pick up the actual component (at least not without wrapping it in something like the `openmrsComponentDecorator`).

We probably should just fix the stub.

denniskigen:
Closing as discussed in favour of updating the stub. Thanks, guys.

PR #1132, opened 2024-08-30
(fix) Add better support for loading multiple namespaces at once

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This is a follow-up for #1119 that addresses a very narrow issue. Specifically, with #1119, when we loaded a namespace for an extension instance in a slot, this resulted in the namespace for the app itself not yet being populated. Basically, we would populate a namespace like `@openmrs/esm-login-app___location-picker___location-picker` correctly, but leave the `@openmrs/esm-login-app` namespace unpopulated. Since we _generally_ rely on the `useTranslation()` to actually load the namespace, this meant there were some circumstances (like using the `translateFrom()` function) where the namespace for a specific extension instance was loaded but the namespace for the module was not currently registered, and `i18next` had no idea that it should try the longer namespace name.

This addresses this issue by ensuring that when we populate the namespace for an extension instance, we also populate the namespace for the app itself.

## Other

Unsolved by this PR is the issue where `translateFrom()` cannot guarantee that a namespace is actually loaded, and so it may not correctly translate given translation keys under circumstances where the module hasn't otherwise been loaded first. While this sounds narrow, it remains a live possibility. I have added a note about this to the docs for `translateFrom()`.

### DIFF
--- a/packages/framework/esm-config/src/module-config/module-config.test.ts
+++ b/packages/framework/esm-config/src/module-config/module-config.test.ts
@@ -1259,11 +1259,13 @@ describe('translation overrides', () => {
     });
     Config.registerModuleWithConfigSystem('corge-module');
     const translationOverrides = await Config.getTranslationOverrides('corge-module');
-    expect(translationOverrides).toStrictEqual({
-      en: {
-        'foo.bar': 'baz',
+    expect(translationOverrides).toStrictEqual([
+      {
+        en: {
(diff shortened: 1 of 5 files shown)


denniskigen (approved):
Thanks, @ibacher. Lets see how this goes.

ibacher:
At least locally, this fixed the issue, so 🤞.

PR #1133, opened 2024-09-02
(refactor) Load styleguide config asynchronously in PageHeader

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR refactors the PageHeader component to load the styleguide config asynchronously via the [getConfig](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-config/src/module-config/module-config.ts#L259) function. This fixes an issue where the useConfig hook usage would result in the consuming component failing to render in the test environment when running tests. I'm not sure why this only happens in the test environment, but it's likely related to the way the styleguide config is loaded in the test environment. It's also possible that the better approach here would be to refactor the useConfig stub to make it more flexible, but this PR fixes the issue for now. Functionality should be unchanged.

FWIW, the [other use](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts#L139-L140) of the styleguide config in the Patient Management app loads the config asynchronously via the getConfig function, so this PR is consistent with that approach.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3782

### DIFF
--- a/packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
+++ b/packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
@@ -1,7 +1,7 @@
 /** @module @category UI */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import classNames from 'classnames';
-import { useConfig } from '@openmrs/esm-react-utils';
+import { getConfig } from '@openmrs/esm-config';
 import { type StyleguideConfigObject } from '../config-schema';
 import styles from './page-header.module.scss';
 
@@ -87,10 +87,13 @@ export const PageHeader: React.FC<PageHeaderProps> = (props) => {
  * ```
  */
 export const PageHeaderContent: React.FC<PageHeaderContentProps> = ({ title, illustration, className }) => {
-  const config = useConfig<StyleguideConfigObject>({
-    // Do not remove this property. See https://github.com/openmrs/openmrs-esm-core/pull/1125#issuecomment-2313230513 for more context.
-    externalModuleName: '@openmrs/esm-styleguide',
-  });
+  const [config, setConfig] = useState<StyleguideConfigObject | null>(null);
+
+  useEffect(() => {
+    getConfig('@openmrs/esm-styleguide').then((fetchedConfig: StyleguideConfigObject) => {
+      setConfig(fetchedConfig);
+    });
+  }, []);
 
   return (
     <div className={classNames(styles.pageHeaderContent, className)}>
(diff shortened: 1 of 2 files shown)


packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
```diff
@@ -87,10 +87,13 @@ export const PageHeader: React.FC<PageHeaderProps> = (props) => {
...
-    // Do not remove this property. See https://github.com/openmrs/openmrs-esm-core/pull/1125#issuecomment-2313230513 for more context.
-    externalModuleName: '@openmrs/esm-styleguide',
-  });
+  const [config, setConfig] = useState<StyleguideConfigObject | null>(null);
+
+  useEffect(() => {
+    getConfig('@openmrs/esm-styleguide').then((fetchedConfig: StyleguideConfigObject) => {
+      setConfig(fetchedConfig);
+    });
+  }, []);
```
vasharma05, line 96: This won't work for mutation, instead we should `subscribe` to the store returned by the [getConfig](https://github.com/openmrs/openmrs-esm-core/blob/c8abf51c15068625bb53c6b433b958298dd40940/packages/framework/esm-config/src/module-config/state.ts#L114).

```
 useEffect(() => {
    const unsubscribe = getConfig('@openmrs/esm-styleguide').subscribe((fetchedConfig: StyleguideConfigObject) => {
      setConfig(fetchedConfig);
    });
	return () => {
	unsubscribe();
}
  }, []);
```

packages/framework/esm-styleguide/src/page-header/page-header.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+jest.mock('@openmrs/esm-config', () => ({
+  getConfig: jest.fn(),
+}));
+
+describe('PageHeaderContent', () => {
+  const mockIllustration = <svg data-testid="mock-illustration" />;
+
+  it('renders title and illustration', async () => {
+    (getConfig as jest.Mock).mockResolvedValue({});
```
vasharma05, line 14: Why are we not using:
```
const mockedGetConfig = jest.mocked(getConfig)
```
Basically, it returns the Zustand store, hence we won't have to take care of the TS handling, and will be easier to return the mocked value.
Thanks!

vasharma05 (approved):
I believe that we should improve the test stub of the `useConfig` but shared before, it's giving issues, hence we must revert the change here when a valid fix is found for the mocked `useConfig`.
Thanks!

packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
vasharma05: Yes, my bad.

vasharma05: Please ignore this comment.
Thanks!

brandones:
Wait, was the problem actually with the `useConfig` mock? Seems not ideal to change the implementation because the mock was messing up.

PR #1134, opened 2024-09-02
(fix) O3-3891: Adjust Change Language Modal Size to Small

Samstar10:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR adjusts the size of the change-language pop-up modal to small.

## Screenshots
Find attached the screenshot with the changes.
Screenshot 2024-09-02 at 11 26 40

## Related Issue
https://openmrs.atlassian.net/browse/O3-3891

## Other
@denniskigen @vasharma05 please review.

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/change-language/change-language-link.extension.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/change-language/change-language-link.extension.tsx
@@ -1,28 +1,31 @@
-import React, { useCallback } from 'react';
+import React, { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Button, SwitcherItem } from '@carbon/react';
-import { TranslateIcon, showModal, useSession } from '@openmrs/esm-framework';
+import { TranslateIcon, useSession } from '@openmrs/esm-framework';
 import styles from './change-language-link.scss';
+import ChangeLanguageModal from './change-language.modal';
 
 /** The user menu item that shows the current language and has a button to change the language */
 export function ChangeLanguageLink() {
   const { t } = useTranslation();
   const session = useSession();
-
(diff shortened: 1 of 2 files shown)


denniskigen:
Thanks for taking a stab at this, @Samstar10. As mentioned last week, we've created an [abstraction](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/modals/index.tsx#L41) that handles creating the modal container element and injecting it it into the DOM. The idea behind this is that modals should be framework agnostic (the Angular form engine in O3 runs Angular and can leverage the same modal container we use in the React context). In https://github.com/openmrs/openmrs-esm-core/pull/1151, I tweaked the underlying implementation of the abstraction to support passing in a optional size prop to your `showModal` definition. Check out the PR description for a guide on how you could leverage that here to achieve the same goal.

Samstar10:
> Thanks for taking a stab at this, @Samstar10. As I mentioned last week, we've built an [abstraction](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/modals/index.tsx#L41) in the styleguide that handles creating the modal container and injecting it into the DOM. The goal is to make modals framework-agnostic, so both Angular (used in O3) and React can share the same modal container.
> 
> In [PR #1151](https://github.com/openmrs/openmrs-esm-core/pull/1151), I updated the abstraction to support an optional `size` prop in the `showModal` definition. You can check out the PR description for guidance on how to use it to achieve the same result.

Thank you @denniskigen. I will check it out and thank you for the feedback.

vasharma05:
@Samstar10, you can refer #1155 from @suubi-joshua, and do the same for this PR too.
Thanks!

Samstar10:
> @Samstar10, you can refer #1155 from @suubi-joshua, and do the same for this PR too. Thanks!

Thank you for the feedback @vasharma05. Let me get on it.

PR #1142, opened 2024-09-06
(DRAFT) (TEST) Removing the translation for workspace title

vasharma05:
# Requirements

## For changes to apps

## If applicable


denniskigen:
We made this intentionally to try and break the E2E tests in Patient Chart - the idea was to break the logic that determines which namespace to load translations for in the Workspace header. The E2E tests wound up breaking as we expected, so this in some way validates that making an API change that breaks stuff in Core would in fact break things in a repo that leverages APIs exported from Core.

PR #1154, opened 2024-09-18
(BREAKING) Retire concept of ConnectedExtension

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

esm-core tracks extensions in various forms (there are others; these are the relevant ones here):

* `ExtensionDefinition`: An extension as defined in an apps `routes.json`
* `ExtensionRegistration`: An extension registered with the extension system
* `AssignedExtension`: An extension instance, representing an extension in a specific slot
* `ConnectedExtension`: An "assigned" extension that is also available due to either the presence of a feature flag or the browsers online status and the extensions definition.

This PR collapses the distinction between `AssignedExtension` and `ConnectedExtension`, that is, an extension is only assigned to a slot now if:

* It is attached to the slot
* The current user has all require privileges to load the extension
* If the extension is behind a feature flag, the feature flag is active
* If the browser is online, the extension is configured to load online and if the browser is offline the extension is configured to load while offline

This is accomplished by making the "output extension store" dependent on the featureFlag store, the browser's online status, and the current session store. These dependencies are now centralized as part of the internal `getAssignedExtensionsFromSlotData()` and the relevant public APIs have been updated to supply the necessary data.

Prior to this change, assigned extensions were not recalculated globally, but instead, each instance of an `ExtensionSlot` would call `useConnectedExtensions()`, which would load the current feature flags and online status and use these to filter the `AssignedExtensions`.

There are two reasons for this change:

1. It should be more efficient to calculate the extensions that should be rendered, since the session store, current feature flags, and browser online status only needs to be loaded once rather than once-per-`useConnectedExtensions()` instance.
2. This makes the extension system easier to consume correctly outside of React, because all relevant changes are purely part of the global extension store and not layered on top of that.

This PR removes all of the `ConnectedExtension` internals, however, the `useConnectedExtensions()` hook (now an alias for `useAssignedExtensions()`) remains so as to not break too much code.

I've also taken this opportunity to completely remove the support for the `wrap` property from the React `Extension` object.


packages/framework/esm-react-utils/src/Extension.tsx
```diff
@@ -1,17 +1,12 @@
...
-import type { ExtensionData } from './ComponentContext';
 
-export type ExtensionProps = {
-  state?: Record<string, any>;
-  /** @deprecated Pass a function as the child of `ExtensionSlot` instead. */
-  wrap?(slot: React.ReactNode, extension: ExtensionData): ReactElement<any, any> | null;
-} & Omit<React.HTMLAttributes<HTMLDivElement>, 'children'> & {
-    children?: React.ReactNode | ((slot: React.ReactNode, extension?: ExtensionData) => React.ReactNode);
-  };
+export interface ExtensionProps extends React.HTMLAttributes<HTMLDivElement> {
```
brandones, line 7:
```suggestion
/** @noInheritDoc */
export interface ExtensionProps extends React.HTMLAttributes<HTMLDivElement> {
```

packages/framework/esm-framework/src/integration-tests/extension-config.test.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { act, render, screen, waitFor } from '@testing-library/react';
+import { act, render, prettyDOM, screen, waitFor } from '@testing-library/react';
```
brandones, line 2:
```suggestion
import { act, render, screen, waitFor } from '@testing-library/react';
```

Unused?

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -234,8 +234,8 @@ export function launchWorkspace<
   } else if (isWorkspaceAlreadyOpen) {
     const openWorkspace = openWorkspaces[workspaceIndexInOpenWorkspaces];
     // Only update the title if it hasn't been set by `setTitle`
-    if (openWorkspace.title == getWorkspaceTitle(openWorkspace, openWorkspace.additionalProps)) {
-      openWorkspace.title = getWorkspaceTitle(openWorkspace, newWorkspace.additionalProps);
+    if (openWorkspace.title === getWorkspaceTitle(openWorkspace, openWorkspace.additionalProps)) {
+      openWorkspace.title = getWorkspaceTitle(newWorkspace, newWorkspace.additionalProps);
```
brandones, line 238: Nice catch

brandones (approved):
Excellent change, thank you @ibacher . A couple suggestions but generally LGTM.

packages/framework/esm-framework/src/integration-tests/extension-config.test.tsx
ibacher: Yeah, I was using it for troubleshooting and forgot to remove it.

packages/framework/esm-react-utils/src/Extension.tsx
ibacher: Hmmm... I couldn't get this to do anything useful. It should work. In any case, if I turn it into a `type` instead of `interface`, it seems to have the intended effect.

denniskigen (approved):
LGTM. Thanks, @ibacher.

PR #1157, opened 2024-09-19
(feat) Disable automatic revalidations by default

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

These changes to the global SWR config disable automatic revalidations by default. The idea is to set up a conservative caching strategy that minimizes automatic data revalidations. In practice, it is very unlikely that the data in a chart will change frequently, so it makes sense to limit the number of revalidations to a bare minimum.

Note that the SWRConfig component merges the configuration from the parent context. This means that alternative configurations provided further down the line will override the parent values.


packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx
```diff
@@ -17,8 +17,13 @@ const defaultSwrConfig = {
   errorRetryCount: 3,
   // default fetcher function
   fetcher: openmrsFetch,
-  // only revalidate once every 30 seconds
-  focusThrottleInterval: 30000,
+  // only revalidate once every 10 minutes
+  focusThrottleInterval: 600000,
+  // disable automatic revalidations by default
+  revalidateIfStale: false,
```
ibacher, line 23: I'm a little more hesitant about turning on `revalidateIfStale`. Maybe we leave that out for now?

ibacher (approved):

packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx
chibongho: I agree. If `revalidateIfStale`, `revalidateOnFocus`, `revalidateOnReconnect` are all set to false, there is no longer a distinction between `useSWR` and `useSWRInfinite`. 

https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations

Side note: Should we have a convention where the SWRConfig values are override-able on a per-app basis? Maybe an edge case, but I'm thinking about multiple clinician using the queues / appointments / ward app and we might want faster re-validate internals.

denniskigen: Yeah, @chibongho. As I mentioned [here](https://github.com/openmrs/openmrs-esm-core/pull/929#issue-2136160373), `SWRConfig` [merges](https://swr.vercel.app/docs/global-configuration#nesting-configurations) the configuration from the parent context. This means that alternative configurations provided further down the line will override the parent values.

ibacher: Tbh, actually, using SWR configuration instead of `useSWRImmutable` actually has some advantages (the global `mutate()` function cannot invalidate content loaded with either `useSWRInfinite()` or `useSWRImmutable()` and we've tended to use Immutable in a few inappropriate places).

> Should we have a convention where the SWRConfig values are override-able on a per-app basis? 

I mean, those decisions are always possible to make, but I'd rather focus on doing so locally where it makes sense to explore whether some kind of app-wide configuration even makes sense (these sort of technical details definitely shouldn't be part of the standard configuration system).

chibongho (approved):

PR #1159, opened 2024-09-20
(chore) O3-4011 - refactor patient banner to have patient identifiers…

chibongho:
… component be reuseable

# Requirements

## For changes to apps

## If applicable

## Summary

This is motivated from a [discussion](https://openmrs.slack.com/archives/C06P2GVUGEP/p1726780219518729) on how the patient identifiers should show in the patient card in the ward app:

>  The identifier AND the number should appear in a tag, with the number bolded. This happens only when there are multiple patient identifiers and one identifier has been configured to the be the primary identifier

The above was said in regards to the patient banner, but should also apply to the ward patient card. This PR refactors the Patient Identifiers as a separate re-useable component, so we can reuse the same logic and configurations for patient banner and have them applied in the ward patient card (and elsewhere).

## Related Issue

https://openmrs.atlassian.net/browse/O3-4011

## Other

- Right now, we do not follow the logic quoted above for the patient banner (having the primary identifier label and number appear in a tag only when there are multiple identifiers present). We can fix it in a separate PR.
- The patient banner was moved from `patient-chart` to `esm-styleguide` a while ago, but there are some remnants of the `excludePatientIdentifierCodeTypes` config value still present in `patient-chart`, mostly used when the patient chart needs to display the patient identifiers outside of the patient banner. This change should help to with completely removing the `excludePatientIdentifierCodeTypes` from the patient-chart.

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -213,6 +213,7 @@
 - [CustomOverflowMenu](API.md#customoverflowmenu)
 - [PatientBannerActionsMenu](API.md#patientbanneractionsmenu)
 - [PatientBannerContactDetails](API.md#patientbannercontactdetails)
+- [PatientBannerPatientIdentifier](API.md#patientbannerpatientidentifier)
 - [PatientBannerPatientInfo](API.md#patientbannerpatientinfo)
 - [PatientBannerToggleContactDetailsButton](API.md#patientbannertogglecontactdetailsbutton)
 - [PatientPhoto](API.md#patientphoto)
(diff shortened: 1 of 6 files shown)


mogoodrich:
@chibongho  can we have a screenshot, preferrably in the main patient banner, as I think this functionality extends beyond the Ward App?

mogoodrich (approved):
LGTM, also fyi @denniskigen  and @ibacher

chibongho:
> @chibongho can we have a screenshot, preferrably in the main patient banner, as I think this functionality extends beyond the Ward App?

No UI changes in this PR. but there will be in upcoming ones. Specifically, the patient identifier still gets incorrectly shown with the label (and not the number) in a tag:
image

Desired behavior: the tag should enclose both the label and the identifier number, but only when multiple ones are present. The tag should be applied to the preferred identifier in that case. Mockup:
image

denniskigen:
Didn't get round to having a look at this, sorry. Thanks for getting it off the mark, @chibongho!

PR #1173, opened 2024-10-03
(feat) Add variable name extractor to expression runner

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This adds a companion function to the expression-evaluator, called `extractVariableNames()` which takes either a "pre-compiled" or raw expression and returns a list of any unbound identifiers ("variables") in the expression. This is useful for tracking expression dependencies or just understanding what variables need to be loaded to be passed into an expression.

While it uses the same parsing framework, the variable name extractor does not do any real validation of the expression; that is, an expression passing through this function does not guarantee that it will not throw an error when executed. This is because we do not make any attempt to actual evaluate the expression at this point.

See the examples and test cases for a better outline of what this actually does.

Variables are returned in roughly the order they first appear in the expression.


samuelmale (approved):
LGTM, thanks @ibacher!

PR #1183, opened 2024-10-20
(fix) O3-3627: Preserve original filename when saving an attachment

suubi-joshua:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR amends the logic that uploads a file to the server, ensuring that when a File object is provided, the original `filename` is saved. Presently, if the user manually edits the filename, the existing logic will save the original filename from the `File` object (typically the user's selection from the file dialog). This change makes it so that the filename is saved as the user edits it, rather than the original filename.

## Related Issue

[](https://openmrs.atlassian.net/browse/O3-3627)

### DIFF
--- a/packages/framework/esm-api/src/attachments.ts
+++ b/packages/framework/esm-api/src/attachments.ts
@@ -23,7 +23,7 @@ export async function createAttachment(patientUuid: string, fileToUpload: Upload
   formData.append('patient', patientUuid);
 
   if (fileToUpload.file) {
-    formData.append('file', fileToUpload.file);
+    formData.append('file', fileToUpload.file, fileToUpload.fileName);
   } else {
     formData.append('file', new File([''], fileToUpload.fileName), fileToUpload.fileName);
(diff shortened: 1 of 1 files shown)


denniskigen:
Thanks, @suubi-joshua. Could you fix the casing and minor typos in the PR title and description?

packages/framework/esm-api/src/attachments.ts
```diff
@@ -23,7 +23,7 @@ export async function createAttachment(patientUuid: string, fileToUpload: Upload
   formData.append('patient', patientUuid);
 
   if (fileToUpload.file) {
-    formData.append('file', fileToUpload.file);
+    formData.append('file', fileToUpload.file, fileToUpload.fileName);
```
denniskigen, line 26: I've tested this out and found that it works as expected @dkayiwa, @ibacher. Any qualms with the approach?

denniskigen (approved):

PR #1184, opened 2024-10-23
(refactor) Switch visit requests to use promises instead of observables

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

Ports over requests to the Visit API to use promise-based fetch requests instead of observables. Promises are generally simpler to work with than observables, and this should make the code easier to understand. This change also makes the code more consistent with other requests the frontend makes to the REST API.

I've also amended the `UpdateVisitPayload` type annotation to a partial of `NewVisitPayload`, as not all fields are required when updating a visit. We should only really need to provide the stop datetime when ending a visit, for example.

### DIFF
--- a/packages/framework/esm-api/src/shared-api-objects/visit-utils.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/visit-utils.ts
@@ -1,10 +1,26 @@
 /** @module @category API */
-import type { Observable } from 'rxjs';
 import { BehaviorSubject } from 'rxjs';
-import { take, map } from 'rxjs/operators';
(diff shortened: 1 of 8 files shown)


denniskigen:
@ibacher Is this technically a breaking change?

ibacher:
Technically, yes, but maybe we can ignore that?

ibacher (approved):

denniskigen:
@ibacher I think I've covered all the bases with my linked PRs.

ibacher:
Yeah, we'd exepct the tests to break until everything is updated

PR #1185, opened 2024-10-28
(refactor) O3-4077: Improve the workspace group workflow

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
Currently, for the related workspaces (current implementation: patient dashboard workspace in the Ward View), the sidebar family (group of workspaces is termed as a family) is defined in the workspace registration using 2 properties `hasSidebar` and `sidebarFamily`.

@brandones suggests that we can move this handling to the `WorkspaceContainer`, and handling of workspace group should be independent to the workspace implementation.

### Workspace group (rename of the workspace family)
Workspace group refers to the workspaces with a sidebar and shares some common state among different workspaces.

### Launching workspace group
Workspace group should be launched with `launchWorkspaceGroup` and we can pass state, as well as a callback when workspace is launched and a cleanup function when the workspace group closes.

### Supported group for a workspace
A new property `workspaceGroups` is defined in the workspace registration, to validate if a workspace launched is a part of the currently opened workspace group and if not, the current workspace group will be closed and then the workspace will be launched.

### Example
```
<button
        className={styles.wardPatientCardButton}
        onClick={() => {
          launchWorkspaceGroup('ward-patient', {
            state: {
              wardPatient,
              WardPatientHeader,
            },
            onWorkspaceGroupLaunch: () => {
              const store = getPatientChartStore();
              store.setState({
                patientUuid: patient.uuid,
              });
              launchWorkspace<PatientWorkspaceAdditionalProps>('patient-transfer-swap-workspace');
            },
            workspaceGroupCleanup: () => {
              const store = getPatientChartStore();
              store.setState({
                patientUuid: undefined,
              });
            },
          });
        }}>
        {getPatientName(patient.person)}
      </button>
```

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-4077

### DIFF
--- a/packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
+++ b/packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
@@ -21,7 +21,7 @@
 
 Controls whether the workspace group should be closed and store to be
 cleared when this workspace is closed.
-Defaults to true except when opening a new workspace of the same family.
+Defaults to true except when opening a new workspace of the same group.
 
 **`default`** true
 
--- a/packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
+++ b/packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
@@ -6,6 +6,7 @@
 
 ### Workspace Properties
 
+- [closeWorkspaceGroup](CloseWorkspaceOptions.md#closeworkspacegroup)
 - [ignoreChanges](CloseWorkspaceOptions.md#ignorechanges)
 
 ### Workspace Methods
@@ -13,6 +14,20 @@
 - [onWorkspaceClose](CloseWorkspaceOptions.md#onworkspaceclose)
 
 ## Workspace Properties
+
+### closeWorkspaceGroup
+
+• `Optional` **closeWorkspaceGroup**: `boolean`
+
+Controls whether the workspace family store will be cleared when this workspace is closed. Defaults to true except when opening a new workspace of the same family.
+
+**`default`** true
+
+#### Defined in
+
+[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:31](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L31)
+
+___
 
 ### ignoreChanges
 
--- a/packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
+++ b/packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
@@ -19,13 +19,15 @@
 
 • `Optional` **closeWorkspaceGroup**: `boolean`
 
-Controls whether the workspace family store will be cleared when this workspace is closed. Defaults to true except when opening a new workspace of the same family.
+Controls whether the workspace group should be closed and store to be
+cleared when this workspace is closed.
+Defaults to true except when opening a new workspace of the same family.
 
 **`default`** true
 
 #### Defined in
 
-[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:31](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L31)
+[packages/framework/esm-styleguide/src/workspaces/workspaces.ts:33](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L33)
 
 ___
 
--- a/packages/framework/esm-globals/src/types.ts
+++ b/packages/framework/esm-globals/src/types.ts
@@ -302,13 +302,30 @@ export type WorkspaceDefinition = {
    * The width "extra-wide" is for workspaces that contain their own sidebar.
    */
   width?: 'narrow' | 'wider' | 'extra-wide';
-  preferredWindowSize?: WorkspaceWindowState;
   /**
-   * Workspaces can open either individually or in a group of workspaces. The workspace groups
-   * will define the groups in which a workspace can be opened.
+   * Launches the workspace in the preferred size, it defaults to the 'narrow' width
+   */
+  preferredWindowSize?: WorkspaceWindowState;
+
+  /**
+   * Workspaces can open either independently or as part of a "workspace group". A
+   * "workspace group" groups related workspaces together, so that only one is visible
+   * at a time. For example,
    *
-   * In case the currently opened workspace is not present in the groups of a workspace,
-   * the current group will close and then the workspace will be launched.
+   * @example
+   *
+   * {
+   *  name: 'order-basket',
+   *  type: 'order',
+   *  groups: ['ward-patient']
+   * }
+   *
+   * This means that the 'order-basket' workspace can be opened independently, or only
+   * in the 'ward-patient'.
+   * If a workspace group is already open and a new workspace is launched, and the
+   * groups in the newly launched workspace do not include the currently open group’s
+   * name, the entire workspace group will close, and the new workspace will launch independently.
+   *
    */
   groups: Array<string>;
 } & (
--- a/packages/framework/esm-globals/src/types.ts
+++ b/packages/framework/esm-globals/src/types.ts
@@ -302,19 +302,14 @@ export type WorkspaceDefinition = {
    * The width "extra-wide" is for workspaces that contain their own sidebar.
    */
   width?: 'narrow' | 'wider' | 'extra-wide';
-  /**
-   * Controls whether the workspace has its own sidebar. If true, the sidebar will be displayed when the workspace is open.
-   */
-  hasOwnSidebar?: boolean;
-  /**
-   * Sidebars have icons that representing workspaces. The sidebar family is the name of the
-   * sidebar that should contain the icon for this workspace. This is generally only needed if
-   * `hasOwnSidebar` is true, in which case this is the name of that sidebar. If multiple
-   * workspaces have `hasOwnSidebar` set to true and the same family name, then the sidebar
-   * within the workspace area will have icons for each of those workspaces.
-   */
-  sidebarFamily?: string;
   preferredWindowSize?: WorkspaceWindowState;
+  /**
+   * Workspaces can open either individually or in a group of workspaces. The workspace groups
+   * will define the groups in which a workspace can be opened.
+   *
+   * In case the currently opened workspace is not present in the workspaceGroups of a workspace,
+   * the current group will close and then the workspace will be launched.
+   */
   workspaceGroups: Array<string>;
 } & (
   | {
--- a/packages/framework/esm-globals/src/types.ts
+++ b/packages/framework/esm-globals/src/types.ts
@@ -315,6 +315,7 @@ export type WorkspaceDefinition = {
    */
   sidebarFamily?: string;
   preferredWindowSize?: WorkspaceWindowState;
+  workspaceGroups: Array<string>;
 } & (
   | {
       /**
--- a/packages/framework/esm-globals/src/types.ts
+++ b/packages/framework/esm-globals/src/types.ts
@@ -307,10 +307,10 @@ export type WorkspaceDefinition = {
    * Workspaces can open either individually or in a group of workspaces. The workspace groups
    * will define the groups in which a workspace can be opened.
    *
-   * In case the currently opened workspace is not present in the workspaceGroups of a workspace,
+   * In case the currently opened workspace is not present in the groups of a workspace,
    * the current group will close and then the workspace will be launched.
    */
-  workspaceGroups: Array<string>;
+  groups: Array<string>;
 } & (
   | {
       /**
--- a/packages/framework/esm-routes/src/loaders/components.ts
+++ b/packages/framework/esm-routes/src/loaders/components.ts
@@ -194,6 +194,7 @@ supported, so the workspace will not be loaded.`,
       canMaximize: workspace.canMaximize,
       width: workspace.width,
       preferredWindowSize: workspace.preferredWindowSize,
+      workspaceGroups: workspace.workspaceGroups ?? [],
     });
   }
 }
--- a/packages/framework/esm-routes/src/loaders/components.ts
+++ b/packages/framework/esm-routes/src/loaders/components.ts
@@ -193,8 +193,6 @@ supported, so the workspace will not be loaded.`,
       canHide: workspace.canHide,
       canMaximize: workspace.canMaximize,
       width: workspace.width,
-      hasOwnSidebar: workspace.hasOwnSidebar,
-      sidebarFamily: workspace.sidebarFamily,
       preferredWindowSize: workspace.preferredWindowSize,
     });
   }
--- a/packages/framework/esm-routes/src/loaders/components.ts
+++ b/packages/framework/esm-routes/src/loaders/components.ts
@@ -194,7 +194,7 @@ supported, so the workspace will not be loaded.`,
       canMaximize: workspace.canMaximize,
       width: workspace.width,
       preferredWindowSize: workspace.preferredWindowSize,
-      workspaceGroups: workspace.workspaceGroups ?? [],
+      groups: workspace.groups ?? [],
     });
   }
 }
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
@@ -74,7 +74,8 @@ export function WorkspaceContainer({
   actionMenuProps,
 }: WorkspaceContainerProps) {
   const layout = useLayoutType();
-  const { workspaces, workspaceWindowState, currentGroupName } = useWorkspaces();
+  const { workspaces, workspaceWindowState, workspaceGroup } = useWorkspaces();
+  const currentGroupName = workspaceGroup?.name;
   const activeWorkspace = workspaces[0];
   const isHidden = workspaceWindowState === 'hidden' || activeWorkspace == null;
   const isMaximized = workspaceWindowState === 'maximized';
@@ -144,7 +145,8 @@ interface WorkspaceProps {
 function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspaceProps) {
   const { t } = useTranslation();
   const layout = useLayoutType();
-  const { workspaceWindowState, currentGroupName } = useWorkspaces();
+  const { workspaceWindowState, workspaceGroup } = useWorkspaces();
+  const currentGroupName = workspaceGroup?.name;
   const isMaximized = workspaceWindowState === 'maximized';
 
   // We use the feature name of the app containing the workspace in order to set the extension
@@ -159,7 +161,7 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
   const {
     canHide = false,
     canMaximize = false,
-    workspaceContainerName = '',
+    currentWorkspaceGroup = '',
     closeWorkspace,
   } = useMemo(() => workspaceInstance ?? ({} as OpenWorkspace), [workspaceInstance]);
 
@@ -182,7 +184,7 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
           <div className={styles.overlayHeaderSpacer} />
           <HeaderGlobalBar className={styles.headerButtons}>
             <ExtensionSlot
-              name={`workspace-header-family-${workspaceInstance.workspaceContainerName}-slot`}
+              name={`workspace-header-family-${workspaceInstance.currentWorkspaceGroup}-slot`}
               state={workspaceProps}
             />
             <ExtensionSlot name={`workspace-header-type-${workspaceInstance.type}-slot`} state={workspaceProps} />
@@ -245,7 +247,7 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
         </Header>
         <div
           className={classNames(styles.workspaceContent, {
-            [styles.marginWorkspaceContent]: Boolean(workspaceContainerName),
+            [styles.marginWorkspaceContent]: Boolean(currentWorkspaceGroup),
           })}
         >
           <WorkspaceRenderer
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
@@ -14,7 +14,6 @@ import styles from './workspace.module.scss';
 
 export interface WorkspaceContainerProps {
   contextKey: string;
-  containerName?: string;
   overlay?: boolean;
   showSiderailAndBottomNav?: boolean;
   additionalWorkspaceProps?: object;
@@ -256,7 +255,6 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
             additionalPropsFromPage={additionalWorkspaceProps}
           />
         </div>
-        {/* {hasOwnSidebar && <ActionMenu isWithinWorkspace name={workspaceInstance.sidebarFamily} />} */}
       </>
     )
   );
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
@@ -74,14 +74,14 @@ export function WorkspaceContainer({
   actionMenuProps,
 }: WorkspaceContainerProps) {
   const layout = useLayoutType();
-  const { workspaces, workspaceWindowState, currentContainerName } = useWorkspaces();
+  const { workspaces, workspaceWindowState, currentGroupName } = useWorkspaces();
   const activeWorkspace = workspaces[0];
   const isHidden = workspaceWindowState === 'hidden' || activeWorkspace == null;
   const isMaximized = workspaceWindowState === 'maximized';
   const width = activeWorkspace?.width ?? (overlay ? 'wider' : 'narrow');
   const showActionMenu = useMemo(
-    () => showSiderailAndBottomNav || (currentContainerName && !isHidden),
-    [currentContainerName, isHidden, showSiderailAndBottomNav],
+    () => showSiderailAndBottomNav || (currentGroupName && !isHidden),
+    [currentGroupName, isHidden, showSiderailAndBottomNav],
   );
 
   useBodyScrollLock(!isHidden && !isDesktop(layout));
@@ -127,7 +127,7 @@ export function WorkspaceContainer({
       </div>
       {showActionMenu && (
         <ActionMenu
-          name={currentContainerName}
+          name={currentGroupName}
           isWithinWorkspace={!showSiderailAndBottomNav}
           actionMenuProps={actionMenuProps}
         />
@@ -144,7 +144,7 @@ interface WorkspaceProps {
 function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspaceProps) {
   const { t } = useTranslation();
   const layout = useLayoutType();
-  const { workspaceWindowState } = useWorkspaces();
+  const { workspaceWindowState, currentGroupName } = useWorkspaces();
   const isMaximized = workspaceWindowState === 'maximized';
 
   // We use the feature name of the app containing the workspace in order to set the extension
@@ -208,7 +208,7 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
                     {isMaximized ? <Minimize /> : <Maximize />}
                   </HeaderGlobalAction>
                 )}
-                {canHide ? (
+                {canHide && !currentGroupName ? (
                   <HeaderGlobalAction
                     align="bottom-right"
                     aria-label={getCoreTranslation('hide', 'Hide')}
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
@@ -68,21 +68,20 @@ export interface WorkspaceContainerProps {
  */
 export function WorkspaceContainer({
   contextKey,
-  containerName,
   overlay,
   showSiderailAndBottomNav,
   additionalWorkspaceProps,
   actionMenuProps,
 }: WorkspaceContainerProps) {
   const layout = useLayoutType();
-  const { workspaces, workspaceWindowState } = useWorkspaces(containerName);
+  const { workspaces, workspaceWindowState, currentContainerName } = useWorkspaces();
   const activeWorkspace = workspaces[0];
   const isHidden = workspaceWindowState === 'hidden' || activeWorkspace == null;
   const isMaximized = workspaceWindowState === 'maximized';
   const width = activeWorkspace?.width ?? (overlay ? 'wider' : 'narrow');
   const showActionMenu = useMemo(
-    () => showSiderailAndBottomNav || (containerName && !isHidden),
-    [containerName, isHidden, showSiderailAndBottomNav],
+    () => showSiderailAndBottomNav || (currentContainerName && !isHidden),
+    [currentContainerName, isHidden, showSiderailAndBottomNav],
   );
 
   useBodyScrollLock(!isHidden && !isDesktop(layout));
@@ -128,7 +127,7 @@ export function WorkspaceContainer({
       </div>
       {showActionMenu && (
         <ActionMenu
-          name={containerName}
+          name={currentContainerName}
           isWithinWorkspace={!showSiderailAndBottomNav}
           actionMenuProps={actionMenuProps}
         />
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
@@ -183,7 +183,7 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
           <div className={styles.overlayHeaderSpacer} />
           <HeaderGlobalBar className={styles.headerButtons}>
             <ExtensionSlot
-              name={`workspace-header-family-${workspaceInstance.currentWorkspaceGroup}-slot`}
+              name={`workspace-header-group-${workspaceInstance.currentWorkspaceGroup}-slot`}
               state={workspaceProps}
             />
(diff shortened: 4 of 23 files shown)


chibongho:
With this change, does it mean that the `workspaceFamily` of a workspace is only defined when we call `launchWorkspace(workspaceName, props, workspaceFamily)`? For example, let's say we have a patient search workspace defined in `esm-patient-search-app`. Can we have the same workspace be launched in different workspace families (such as `service-queue-family` and `ward-family`, depending on which app it's launched from)? 

Also, is there a way to close a all workspaces in a workspace family?

If yes to the above questions, then yeah, I'm generally on board. 

This change is a bit hard for me to review without seeing it in action. If people are on board with this, maybe we can also have a draft for corresponding changes in patient chart or patient management?

vasharma05:
> let's say we have a patient search workspace defined in esm-patient-search-app. Can we have the same workspace be launched in different workspace families (such as "service-queues-app-family" and "ward-app-family", depending on which app it's launched from)?

Exactly, we won't have to duplicate workspaces to configure it and to open it in the workspace family, now we use the same workspace either individually or in the workspace family, with minimal to no changes. I am using the same to add Order Basket to the Ward Workspace.

> Also, is there a way to close a all workspaces in a workspace family?

Yes, we have a function to close all workspaces, I need to tweak it to allow closing only the workspaces in a family.

> This change is a bit hard for me to review without seeing it in action. If people are on board with this, maybe we can also have a draft for corresponding changes in patient chart or patient management?

I understand, hence I am implementing this with the order basket addition to the ward workspaces, and will share the results here, currently keeping this PR in draft and will make changes as I find more observations.

Thanks @chibongho for supporting the work!

packages/framework/esm-globals/src/types.ts
```diff
@@ -315,6 +315,7 @@ export type WorkspaceDefinition = {
    */
   sidebarFamily?: string;
   preferredWindowSize?: WorkspaceWindowState;
+  workspaceGroups: Array<string>;
```
vasharma05, line 318: Write the JS Doc for `workspaceGroup`

packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
```diff
@@ -14,6 +14,7 @@ import styles from './workspace.module.scss';
 
 export interface WorkspaceContainerProps {
   contextKey: string;
+  containerName?: string;
```
vasharma05, line 17: Remove this

packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
```diff
@@ -243,7 +256,7 @@ function Workspace({ workspaceInstance, additionalWorkspaceProps }: WorkspacePro
             additionalPropsFromPage={additionalWorkspaceProps}
           />
         </div>
-        {hasOwnSidebar && <ActionMenu isWithinWorkspace name={workspaceInstance.sidebarFamily} />}
+        {/* {hasOwnSidebar && <ActionMenu isWithinWorkspace name={workspaceInstance.sidebarFamily} />} */}
```
vasharma05, line 259: Remove this

packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceFamilyStore.ts
```diff
@@ -12,7 +12,7 @@ import type { StoreApi } from 'zustand/vanilla';
  *
  * @param {string} sidebarFamilyName The sidebarFamilyName of the workspace used when registering the workspace in the module's routes.json file.
  */
-export function useWorkspaceFamilyStore(sidebarFamilyName: string) {
+export function useWorkspaceFamilyStore(sidebarFamilyName?: string) {
```
vasharma05, line 15: Replace sidebarFamilyName with workspaceGroupName

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -23,17 +23,16 @@ export interface CloseWorkspaceOptions {
...
-}
-
-interface CloseWorkspaceInternalOptions extends CloseWorkspaceOptions {
   /**
    * Controls whether the workspace family store will be cleared when this workspace is closed. Defaults to true except when opening a new workspace of the same family.
    *
    * @default true
    */
-  clearWorkspaceFamilyStore?: boolean;
+  closeWorkspaceGroup?: boolean;
```
vasharma05, line 31: Update the JSDoc

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,21 +124,88 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+  const currentGroupName = store.getState()?.workspaceGroup?.name;
+  if (!currentGroupName || groupName !== currentGroupName) {
+    return;
+  }
+  const filter: (workspace: OpenWorkspace) => boolean = currentGroupName
+    ? (workspace) => workspace.currentWorkspaceGroup === currentGroupName
+    : () => true;
+
+  closeAllWorkspaces(() => {
+    // Clearing the workspace family store if there are no more workspaces with the same sidebar family name and the new workspace is not of the same sidebar family, which is handled in the `launchWorkspace` function.
```
vasharma05, line 139: Update this

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,21 +124,88 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+}
+
+interface LaunchWorkspaceGroupArg {
+  groupName: string;
+  state: object;
+  onWorkspaceGroupLaunch?: Function;
+  workspaceGroupCleanup?: Function;
+}
+
+export function launchWorkspaceGroup(args: LaunchWorkspaceGroupArg) {
```
vasharma05, line 167: Add JSDoc

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,21 +124,88 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
   return !promptBeforeFn || !promptBeforeFn();
 }
 
+function closeWorkspaceGroup(groupName: string, onWorkspaceCloseup?: Function) {
```
vasharma05, line 127: Add JSDoc

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,21 +124,88 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+    }));
+    getWorkspaceFamilyStore(groupName, state);
+    onWorkspaceGroupLaunch?.();
+  }
+}
+
 function promptBeforeLaunchingWorkspace(
   workspace: OpenWorkspace,
-  newWorkspaceDetails: { name: string; additionalProps?: object },
+  newWorkspaceDetails: { name: string; additionalProps?: object; workspaceContainerName?: string },
```
vasharma05, line 196: Remove `workspaceContainerName`

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -294,46 +373,41 @@ export function cancelPrompt() {
...
+
+    const workspaceGroupName = store.getState().workspaceGroup?.name;
+
     if (
-      options.clearWorkspaceFamilyStore &&
-      workspaceFamilyStore &&
-      !newOpenWorkspaces.some((w) => w.sidebarFamily === workspaceSidebarFamilyName)
+      workspaceGroupName &&
+      options.closeWorkspaceGroup
+      // !newOpenWorkspaces.some((w) => w.workspaceContainerName === workspaceSidebarFamilyName)
```
vasharma05, line 398: Remove this comment

vasharma05 (commented):
Some notes to myself

brandones:
As a total aside: looking at this code

```
            onWorkspaceGroupLaunch: () => {
              const store = getPatientChartStore();
              store.setState({
                patientUuid: patient.uuid,
              });
              launchWorkspace<PatientWorkspaceAdditionalProps>('patient-transfer-swap-workspace');
            },
```

I think we should wrap the patient chart state in a higher-level API. Something like `setCurrentPatient` and `getCurrentPatient`.

packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
```diff
@@ -14,6 +15,22 @@
 
 ## Workspace Properties
 
+### closeWorkspaceGroup
+
+• `Optional` **closeWorkspaceGroup**: `boolean`
+
+Controls whether the workspace group should be closed and store to be
+cleared when this workspace is closed.
+Defaults to true except when opening a new workspace of the same family.
```
brandones, line 24: I thought `family` is no more? How is a workspace's family set?

packages/framework/esm-routes/src/loaders/components.ts
```diff
@@ -193,9 +193,8 @@ supported, so the workspace will not be loaded.`,
       canHide: workspace.canHide,
       canMaximize: workspace.canMaximize,
       width: workspace.width,
-      hasOwnSidebar: workspace.hasOwnSidebar,
-      sidebarFamily: workspace.sidebarFamily,
       preferredWindowSize: workspace.preferredWindowSize,
+      workspaceGroups: workspace.workspaceGroups ?? [],
```
brandones, line 197: "Workspace" is redundant with the context, this can just be `groups`

packages/framework/esm-styleguide/src/workspaces/container/workspace-container.component.tsx
```diff
@@ -73,11 +73,16 @@ export function WorkspaceContainer({
   actionMenuProps,
 }: WorkspaceContainerProps) {
   const layout = useLayoutType();
-  const { workspaces, workspaceWindowState } = useWorkspaces();
+  const { workspaces, workspaceWindowState, workspaceGroup } = useWorkspaces();
```
brandones, line 76: With reference to my comment above on `workspaceGroups` vs `groups`, it's ok/good if the API for `useWorkspaces` remains `workspaceGroup` though

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -501,60 +506,69 @@ describe('workspace system', () => {
...
-      const workspaceStore = getWorkspaceStore();
-      launchWorkspace('ward-patient-workspace', {
-        foo: true,
+      launchWorkspaceGroup(workspaceGroup, {
+        state: {
+          foo: true,
+        },
+        onWorkspaceGroupLaunch: () => {
+          launchWorkspace('ward-patient-workspace');
+        },
```
brandones, line 558: We should consider a simpler API for this, if this is the pattern. Like, why would we ever launch a group without launching a workspace at the same time? Perhaps `launchWorkspaceGroup` should take a `workspaceToLaunch` parameter.

brandones:
I've provided some feedback. I did some pairing with Vineet and Usama on this so it would be great if someone else could review after the next revision. Generally I think the approach is good. @mogoodrich @chibongho @denniskigen @ibacher

packages/framework/esm-framework/docs/interfaces/CloseWorkspaceOptions.md
vasharma05: Sorry, family is renamed to group, let me correct this.

chibongho (commented):
Thanks. I think this will simplify the workspace sidebar

Questions: 
- Should a workspace instance have access to its `currentWorkspaceGroup` (since it is required to close its workspace group)?
- When `launchWorkspace` is called from within a workspace or from an action menu item, does the new workspace get launched within the same workspace group?
- Is it still true that a `WorkspaceContainer` is not tied to any particular workspace or workspace group? (so we don't need to create multiple WorkspaceContainer for multiple workspace groups?)
- Tangent. I'm not that comfortable with the idea of `contextKey` for `WorkspaceContainer`, which defines the URL subpath to match on to determine whether any workspaces in the `WorkspaceContainer` should remain open. It feels like we should just close any open workspaces when the `<WorkspaceContainer>` dismounts. Thoughts?

brandones:
@chibongho thanks for the thoughtful review, much appreciated. Just a note:

> I'm not that comfortable with the idea of contextKey for WorkspaceContainer, which defines the URL subpath to match on to determine whether any workspaces in the WorkspaceContainer should remain open. It feels like we should just close any open workspaces when the <WorkspaceContainer> dismounts. Thoughts?

The reason we designed it this way is because it doesn't always line up well with mounts/unmounts. In particular, when transitioning between different patients in the patient chart, the WorkspaceContainer will not be unmounted; however, any open workspaces should certainly be closed. On the home page, I don't remember whether things like Service Queues are dashboards or independent pages that just happen to be served under the `/home` path, but in the latter case, this would be a thing where the workspace should stay open even though the component would need to unmount.

packages/framework/esm-globals/src/types.ts
```diff
@@ -302,19 +302,15 @@ export type WorkspaceDefinition = {
...
-   * within the workspace area will have icons for each of those workspaces.
+   * Workspaces can open either individually or in a group of workspaces. The workspace groups
+   * will define the groups in which a workspace can be opened.
+   *
+   * In case the currently opened workspace is not present in the groups of a workspace,
+   * the current group will close and then the workspace will be launched.
    */
-  sidebarFamily?: string;
-  preferredWindowSize?: WorkspaceWindowState;
+  groups: Array<string>;
```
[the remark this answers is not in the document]
vasharma05: No, the registration of the workspace is the right place to define the groups, else, the developer will have to handle different scenarios of whether to open it in the group or individual, which right now is handled by the workspace system.

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -440,29 +441,29 @@ describe('workspace system', () => {
     expect(store.getState().openWorkspaces.length).toBe(0);
   });
 
-  describe('Testing `getWorkspaceFamilyStore` function', () => {
+  describe('Testing `getWorkspaceGroupStore` function', () => {
```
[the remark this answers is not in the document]
vasharma05: Sure, I'm on it

vasharma05:
> * Should a workspace instance have access to its currentWorkspaceGroup (since it is required to close its workspace group)?

Yes, when the `launchWorkspace` is called, the `openWorkspace` stores the group name.

> * When launchWorkspace is called from within a workspace or from an action menu item, does the new workspace get launched within the same workspace group?

If the workspace registration (inside the `routes.json`) defines the current open workspace group inside the workspace's `groups` properties, it will open inside the container, else it will close the workspace group and then open the new workspace.

> * Is it still true that a WorkspaceContainer is not tied to any particular workspace or workspace group? (so we don't need to create multiple WorkspaceContainer for multiple workspace groups?)

Yes, we just need 1 workspace container

vasharma05:
Hi @brandones @chibongho ,
Requesting your review here and if all good, may we merge it, given it's needed for some other tasks too.
Thanks!

packages/framework/esm-globals/src/types.ts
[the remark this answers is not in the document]
chibongho: Ok, I think I misunderstood our earlier conversation. Does that mean that a workspace must have a workspace group defined in `routes.json` for it to be launched in that workspace group? (in other words, it's impossible to create a workspace that can be launched in any arbitrary workspace group.) 

I'm thinking of something like the patient search workspace in `esm-patient-serach-app`, where it's really meant for other apps (ex: `esm-service-queues-app`) to launch it. There is no way to launch the patient search workspace in the "service-queues-app" group, correct? (Admittedly, if we really want this, we can work around it by having the patient search workspace be an extension instead, and slot it into a workspace in the "service-service-app" group defined within `esm-service-queues-app`)

chibongho:
> Yes, when the `launchWorkspace` is called, the `openWorkspace` stores the group name.

Shouldn't the group name also be accessible in `DefaultWorkspaceProps`? If not, how does a workspace access the group name?

> If the workspace registration (inside the routes.json) defines the current open workspace group inside the workspace's groups properties, it will open inside the container, else it will close the workspace group and then open the new workspace.

Ok, that makes sense.

packages/framework/esm-globals/src/types.ts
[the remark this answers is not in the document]
vasharma05: > Admittedly, if we really want this, we can work around it by having the patient search workspace be an extension instead, and slot it into a workspace in the "service-service-app" group defined within esm-service-queues-app

I didn't understand this

chibongho:
> > Shouldn't the group name also be accessible in DefaultWorkspaceProps? If not, how does a workspace access the group name?
> 
> What's the use case for this? As far as the common data of a group is concerned, all the common data is shared in the workspace props, as previously implemented. Thanks!

Ok, never mind. Your answer to the other question makes this moot. I was thinking something like this:

```
function MyWorkspace(props: DefaultWorkspaceProps) {
  const onSubmit = () => {
     submitData().then(() => closeWorkspaceGroup('myWorkspaceGroup')); 
  }
  return (<button onClick={onSubmit} />);
}
```

But since `MyWorkspace` must be defined with a workspace group (`myWroskpaceGroup`), we should always know what it is without needing to access it programmatically.

packages/framework/esm-globals/src/types.ts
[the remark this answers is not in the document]
ibacher: Or just a set of React components, so it can be added to the styleguide? That seems the cleanest to me...

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+    if (currentWorkspaceGroup && typeof currentWorkspaceGroup?.cleanup === 'function') {
+      currentWorkspaceGroup?.cleanup?.();
+    }
+
+    store.setState((prev) => ({
+      ...prev,
+      workspaceGroup: undefined,
+    }));
+    if (onWorkspaceCloseup && typeof onWorkspaceCloseup === 'function') {
+      onWorkspaceCloseup?.();
```
ibacher, line 171: We already did the checking that this is defined, so...

```suggestion
      onWorkspaceCloseup();
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+  closeAllWorkspaces(() => {
+    // Clearing the workspace group and respective store if the new workspace is not part of the current group, which is handled in the `launchWorkspace` function.
+    const workspaceGroupStore = getWorkspaceGroupStore(groupName);
+    if (workspaceGroupStore) {
+      workspaceGroupStore.setState({}, true);
+      const unsubscribe = workspaceGroupStore.subscribe(() => {});
+      unsubscribe?.();
+    }
+    if (currentWorkspaceGroup && typeof currentWorkspaceGroup?.cleanup === 'function') {
+      currentWorkspaceGroup?.cleanup?.();
```
ibacher, line 163:
```suggestion
      currentWorkspaceGroup.cleanup();
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+
+  closeAllWorkspaces(() => {
+    // Clearing the workspace group and respective store if the new workspace is not part of the current group, which is handled in the `launchWorkspace` function.
+    const workspaceGroupStore = getWorkspaceGroupStore(groupName);
+    if (workspaceGroupStore) {
+      workspaceGroupStore.setState({}, true);
+      const unsubscribe = workspaceGroupStore.subscribe(() => {});
+      unsubscribe?.();
+    }
+    if (currentWorkspaceGroup && typeof currentWorkspaceGroup?.cleanup === 'function') {
```
ibacher, line 162: You really don't need more than this:

```suggestion
    if (typeof currentWorkspaceGroup?.cleanup === 'function') {
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+      currentWorkspaceGroup?.cleanup?.();
+    }
+
+    store.setState((prev) => ({
+      ...prev,
+      workspaceGroup: undefined,
+    }));
+    if (onWorkspaceCloseup && typeof onWorkspaceCloseup === 'function') {
```
ibacher, line 170: I think there should be a newline here.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
ibacher: There should also probably be a newline here...

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+    ? (workspace) => workspace.currentWorkspaceGroup === currentGroupName
+    : () => true;
+
+  closeAllWorkspaces(() => {
+    // Clearing the workspace group and respective store if the new workspace is not part of the current group, which is handled in the `launchWorkspace` function.
+    const workspaceGroupStore = getWorkspaceGroupStore(groupName);
+    if (workspaceGroupStore) {
+      workspaceGroupStore.setState({}, true);
+      const unsubscribe = workspaceGroupStore.subscribe(() => {});
+      unsubscribe?.();
```
ibacher, line 160: I think it's safe to assume that `unsubscribe()` is always defined here, no?

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+ * - Closes all workspaces associated with the group
+ * - Clears the workspace group store state
+ * - Executes cleanup function if defined in the workspace group
+ * - Updates the main workspace store to remove the workspace group
+ * - Calls the optional closeup callback if provided
+ */
+function closeWorkspaceGroup(groupName: string, onWorkspaceCloseup?: Function) {
+  const store = getWorkspaceStore();
+  const currentWorkspaceGroup = store.getState()?.workspaceGroup;
+  const currentGroupName = store.getState()?.workspaceGroup?.name;
```
ibacher, line 146:
```suggestion
  const currentGroupName = currentWorkspaceGroup?.name;
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+      ...prev,
+      workspaceGroup: {
+        name: groupName,
+        cleanup: workspaceGroupCleanup,
+      },
+    }));
+    getWorkspaceGroupStore(groupName, state);
+    onWorkspaceGroupLaunch?.();
+    if (workspaceToLaunch) {
+      launchWorkspace(workspaceToLaunch?.name, workspaceToLaunch?.additionalProps ?? {});
```
ibacher, line 228:
```suggestion
      launchWorkspace(workspaceToLaunch.name, workspaceToLaunch.additionalProps ?? {});
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -179,6 +289,15 @@ export function launchWorkspace<
 >(name: string, additionalProps?: Omit<T, keyof DefaultWorkspaceProps> & { workspaceTitle?: string }) {
   const store = getWorkspaceStore();
   const workspace = getWorkspaceRegistration(name);
+  const currentWorkspaceGroup = store.getState().workspaceGroup;
+
+  if (currentWorkspaceGroup && !workspace.groups?.includes(currentWorkspaceGroup?.name)) {
+    closeWorkspaceGroup(currentWorkspaceGroup?.name, () => {
```
ibacher, line 295:
```suggestion
    closeWorkspaceGroup(currentWorkspaceGroup.name, () => {
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -179,6 +289,15 @@ export function launchWorkspace<
...
+      launchWorkspace(name, additionalProps);
+    });
+    return;
+  }
+  const currentGroupName = store.getState().workspaceGroup?.name;
```
ibacher, line 300:
```suggestion
  const currentGroupName = currentWorkspaceGroup?.name;
```

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -388,13 +498,19 @@ export function updateWorkspaceWindowState(value: WorkspaceWindowState) {
...
-    const canCloseWorkspace = canCloseWorkspaceWithoutPrompting(name);
-    return canCloseWorkspace;
-  });
+  const canCloseAllWorkspaces = store
+    .getState()
+    .openWorkspaces.filter(filter)
+    .every(({ name }) => {
+      const canCloseWorkspace = canCloseWorkspaceWithoutPrompting(name);
+      return canCloseWorkspace;
+    });
```
ibacher, line 513:
```suggestion
    .every(({ name }) => canCloseWorkspaceWithoutPrompting(name));
```

packages/framework/esm-globals/src/types.ts
```diff
@@ -302,19 +302,15 @@ export type WorkspaceDefinition = {
    * The width "extra-wide" is for workspaces that contain their own sidebar.
    */
   width?: 'narrow' | 'wider' | 'extra-wide';
+  preferredWindowSize?: WorkspaceWindowState;
```
ibacher, line 305: This needs docs...

packages/framework/esm-globals/src/types.ts
```diff
@@ -302,19 +302,15 @@ export type WorkspaceDefinition = {
...
-   * Sidebars have icons that representing workspaces. The sidebar family is the name of the
-   * sidebar that should contain the icon for this workspace. This is generally only needed if
-   * `hasOwnSidebar` is true, in which case this is the name of that sidebar. If multiple
-   * workspaces have `hasOwnSidebar` set to true and the same family name, then the sidebar
-   * within the workspace area will have icons for each of those workspaces.
+   * Workspaces can open either individually or in a group of workspaces. The workspace groups
+   * will define the groups in which a workspace can be opened.
+   *
+   * In case the currently opened workspace is not present in the groups of a workspace,
+   * the current group will close and then the workspace will be launched.
```
ibacher, line 311: I feel like the description here is unclear. Maybe:

```suggestion
   * Workspaces can open either independently or as part of a "workspace group". A
   * "workspace group" groups related workspaces together, so that only one is visible
   * at a time. For example, <USE CASE DESCRIPTION HERE>.
   *
   * In case the currently opened workspace is not present in the groups of a workspace,
   * the current group will close and then the workspace will be launched.
```

ibacher, line 311: These last two lines are really ambiguous. Can you think of a way to rephrase it? I'm not entirely sure what this is trying to communicate, so I'm not sure what to suggest.

ibacher (changes requested):
Great overall work here, aside from the nit-picking comments I left. 

There's some documentation work to be improved and, since this affects the routes.json file, we should also have a a PR [to update the schema](https://github.com/openmrs/openmrs-contrib-json-schemas/blob/main/routes.schema.json).

packages/framework/esm-globals/src/types.ts
ibacher: What's unclear to me: What does "currently opened workspace" here mean? Is the workspace being opened, the workspace currently opened, any workspace in the workspace stack? What does this have to do with "launching"?

packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceGroupStore.ts
```diff
@@ -12,18 +11,18 @@ import type { StoreApi } from 'zustand/vanilla';
  *
  * @param {string} sidebarFamilyName The sidebarFamilyName of the workspace used when registering the workspace in the module's routes.json file.
  */
-export function useWorkspaceFamilyStore(sidebarFamilyName: string) {
+export function useWorkspaceGroupStore(workspaceGroupName?: string) {
```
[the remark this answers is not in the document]
vasharma05: Since this function was initially made to be called for workspaces launched with and without a group name, the argument is hence optional. If the argument is `undefined`, this function will return `undefined`, else will return a global store.

I have already added tests for this function.

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
```diff
@@ -501,60 +506,69 @@ describe('workspace system', () => {
...
       });
-      const sidebarFamilyStore = getWorkspaceFamilyStore(sidebarFamily);
+      const workspaceStore = getWorkspaceStore();
+      const workspaceGroupStore = getWorkspaceGroupStore(workspaceGroup);
       expect(workspaceStore.getState().openWorkspaces.length).toBe(1);
-      expect(sidebarFamilyStore).toBeTruthy();
-      expect(sidebarFamilyStore?.getState()?.['foo']).toBe(true);
-      launchWorkspace('transfer-patient-workspace', { bar: false });
+      expect(workspaceGroupStore).toBeTruthy();
+      expect(workspaceGroupStore?.getState()?.['foo']).toBe(true);
```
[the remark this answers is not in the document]
vasharma05: It has become a habit of mine to use null checks, so that I can check things validity, and also it doesn't break the code at any point. I'll try to improve on this.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+ * @remarks
+ * This function performs the following operations:
+ * - Validates if the provided group name matches the current workspace group
+ * - Closes all workspaces associated with the group
+ * - Clears the workspace group store state
+ * - Executes cleanup function if defined in the workspace group
+ * - Updates the main workspace store to remove the workspace group
+ * - Calls the optional closeup callback if provided
+ */
+function closeWorkspaceGroup(groupName: string, onWorkspaceCloseup?: Function) {
```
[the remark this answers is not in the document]
vasharma05: Done

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
vasharma05: After closing the workspace group, the updated workspace group might have been updated, hence I am fetching the workspace group name from the state.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -120,6 +124,112 @@ export function canCloseWorkspaceWithoutPrompting(name: string, ignoreChanges: b
...
+ * @param args.workspaceGroupCleanup - Optional cleanup function to be executed when the workspace group is closed
+ *
+ * @example
+ * launchWorkspaceGroup("myGroup", {
+ *   state: initialState,
+ *   onWorkspaceGroupLaunch: () => console.log("Workspace group launched"),
+ *   workspaceGroupCleanup: () => console.log("Cleaning up workspace group")
+ * });
+ */
+export function launchWorkspaceGroup(groupName: string, args: LaunchWorkspaceGroupArg) {
```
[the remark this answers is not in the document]
vasharma05: Yes, it makes sense to have a separate `launchWorkspaceGroup` function for the following reasons:

1. Making clear distinction between launching workspace and launching workspace group. Providing different function orderloads will in the end confuse the developer too.
2. `launchWorkspace` function is used for every workspace launch, and on the other hand, `launchWorkspaceGroup` function is used only at the initial launch.

packages/framework/esm-styleguide/src/workspaces/workspace-sidebar-store/useWorkspaceGroupStore.ts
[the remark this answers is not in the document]
ibacher: Ok. Can we update the docs to match the code?

packages/framework/esm-styleguide/src/workspaces/workspaces.test.ts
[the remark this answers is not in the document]
ibacher: The only reason I bring this up is because of how it gets compiled. Specifically, since the optional operator isn't supported across browsers, e.g.,

```ts
store?.getState()?.['option']
```

Gets can get compiled into this gnarly mess:

```js
var _store_getState, _store;
(_store = store) === null || _store === void 0 ? void 0 : (_store_getState = _store.getState()) === null || _store_getState === void 0 ? void 0 : _store_getState['option'];
```

Which is semantically the same, but which involves 3 branching statements and 2 comparisons per branch ("native" support for `?`, this is pushed into the interpreter itself, which is usually faster).

packages/framework/esm-globals/src/types.ts
vasharma05: Updated the doc here

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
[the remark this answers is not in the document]
ibacher: Are clients expected to call both of these functions? When would I, as a code consumer, call `launchWorkspaceGroup()` instead of `launchWorkspace()`?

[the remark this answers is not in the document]
vasharma05: No, to launch the workspace group, the client only needs to call the `launchWorkspaceGroup` function and they can pass the new workspace's details into the argument. You can refer to the argument structure [here](https://github.com/openmrs/openmrs-esm-core/blob/deb79bde7e9b17a9eabe186bd23ea607f04525ad/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L182).
Thanks!

vasharma05:
Hi @ibacher!
Please note that this PR is not a breaking change at all. It doesn't change any existing workspace API/ functionality, just add improvements for workspace groups.
Thanks!

ibacher (approved):

PR #1191, opened 2024-10-30
(feat) O3-4143: useLocations should support searching by tag or query…

mogoodrich:
… string

# Requirements

## For changes to apps

## If applicable

## Summary

see: https://openmrs.atlassian.net/browse/O3-4143

## Related Issue

https://openmrs.atlassian.net/browse/O3-4143

### DIFF
--- a/packages/framework/esm-api/src/shared-api-objects/location.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/location.ts
@@ -11,8 +11,16 @@ export function toLocationObject(openmrsRestForm: any): Location {
   };
 }
 
-export function getLocations(tagUuidOrName: string | null = null): Observable<Array<Location>> {
-  const url = `${restBaseUrl}/location` + (tagUuidOrName ? '?tag=' + tagUuidOrName : '');
+export function getLocations(
+  tagUuidOrName: string | null = null,
+  query: string | null = null,
+): Observable<Array<Location>> {
+  const url =
+    `${restBaseUrl}/location` +
+    (tagUuidOrName || query ? '?' : '') +
+    (tagUuidOrName ? 'tag=' + tagUuidOrName : '') +
+    (tagUuidOrName && query ? '&' : '') +
+    (query ? 'q=' + query : '');
   return openmrsObservableFetch<any>(url)
     .pipe(
       map((results) => {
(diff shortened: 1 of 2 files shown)


mogoodrich:
I wasn't seeing any existing tests around these functions to update?  What is our policy for tests around stuff like this?

mogoodrich:
@ibacher  @denniskigen  just a ping... thoughts on this one?  I need this to be able to add the custom behavior to filter visit locations by the "Visit Location" tag while still maintaining the ability to search locations by string.

packages/framework/esm-api/src/shared-api-objects/location.ts
```diff
@@ -11,8 +11,16 @@ export function toLocationObject(openmrsRestForm: any): Location {
...
+  tagUuidOrName: string | null = null,
+  query: string | null = null,
+): Observable<Array<Location>> {
+  const url =
+    `${restBaseUrl}/location` +
+    (tagUuidOrName || query ? '?' : '') +
+    (tagUuidOrName ? 'tag=' + tagUuidOrName : '') +
+    (tagUuidOrName && query ? '&' : '') +
+    (query ? 'q=' + query : '');
   return openmrsObservableFetch<any>(url)
```
denniskigen, line 24:
```suggestion
  const params = new URLSearchParams();
  if (tagUuidOrName) {
    params.set('tag', tagUuidOrName);
  }
  if (query) {
    params.set('q', query);
  }
  const queryString = params.toString();
  const url = `${restBaseUrl}/location${queryString ? '?' + queryString : ''}`;

  return openmrsObservableFetch<{ results: Array<Location> }>(url)
```

Would something like this work as well? The current URL construction could produce an invalid URL when `tagUuidOrName` is null but query is provided. In this case, the URL would include an empty string where the tag= parameter should be, potentially resulting in malformed URLs like:

`/location?&q=searchterm`

Also, this annotates the return type of the locations response correctly.

denniskigen (approved):
Thanks, @mogoodrich. Mostly LGTM save for one concern about the URL construction.

packages/framework/esm-api/src/shared-api-objects/location.ts
mogoodrich: Perfect!  Thanks so much @denniskigen ... I *think* I handled the null case, but this is *much* cleaner... I will commit your change, test, and then merge if all is well...

PR #1193, opened 2024-10-31
(chore) export openmrsComponentDecorator

chibongho:
# Requirements

## For changes to apps

## If applicable

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -192,6 +192,7 @@
 - [evaluateAsync](API.md#evaluateasync)
 - [extractVariableNames](API.md#extractvariablenames)
 - [isOnline](API.md#isonline)
+- [openmrsComponentDecorator](API.md#openmrscomponentdecorator)
 - [useFhirFetchAll](API.md#usefhirfetchall)
 - [useFhirInfinite](API.md#usefhirinfinite)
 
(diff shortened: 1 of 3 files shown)


mogoodrich (approved):
Defer to others, but looks straightforward to expose this!

ibacher:
Can you say more about the motivation here? I'm not terribly keen on having the component decorator become part of the public API, as it limits our ability to make changes to what is mostly an internal implementation detail.

chibongho:
Yeah, this can be a somewhat dangerous change.

In this ([PR](https://github.com/openmrs/openmrs-esm-core/pull/919)), different sub-components of the patient banner is moved to in style guide. I think the idea is that we can mix and match these components into different looking patient cards in the search results workspace if desired, and eliminate the need of `patient-banner-slot`. We have been mostly using extensions for this kind of thing (and thus avoiding the need for `openmrsComponnetDecorator`), but I think there is limitation to that in scenarios where we only want a custom patient card for a specific app.

Not implemented in that PR, but I think the next step would be to have the patient search workspace accept a custom `PatientCardComponent` as a prop. So something like this:

```
launchWorkspace('patient-search', {
  PatientCardComponent={MyPatientCard} // optional prop
})

function MyPatientCard(prop: DefaultPatientCardProps) {
  return (
    <div>
      <PatientInfo />
      <PatientIdentifier />
      <PatientActions />
    </div>
  )
}
```

For this to work, any sub-components included in `MyPatientCard` needs to be wrapped in a `openrmsComponentDecorator` in order to have proper translations and configurations.

(This approach also avoids the pattern of rendering N extensions for N patient cards in a loop, something we had difficulty with in the ward app.)

Besides search results, I also have another use case of needing to show a patient header at the top of the clinical form workspace. In the patient chart, there is no ambiguity on which patient we opened the workspace for. But in the ward app (where we can click on any ward patient to open their workspace), we need to display a patient header. (The ward app also has its own app-specific patient header different from the regular one)

chibongho:
@ibacher and others, interested in your thoughts / concerns about this, especially on:
- the proposed way to allow for custom patient cards in my previous comment's example
- whether `openrmsComponentDecorator` is the right API to expose to enable this.

ibacher:
If we're just exporting components from the styleguide, there's no reason to make this a public export. We can just use the component decorator directly in the styleguide (things inside the framework are allowed to import from the framework internals / non-public APIs; other code can, but we don't support it long-term). That is, I think, my preferred solution here.

chibongho:
I think we also want to use components defined outside of styleguide though, and they require the decorator or something similar to wrap the elements. In the `MyPatientCard` example above, it can include components that are defined in an app outside of styleguide.

ibacher:
Sorry... I think I dropped the ball on this. Anyways, if we _do_ need to export the openmrsComponentDecorator, then we need to add doc strings to it explaining how it is meant to be used. Nothing should exist in the public API with documentation like that.

chibongho:
Sure. I originally meant to use the function as a proof of concept. If there is no objection to the proposed use cases (to pass configurable / translatable components across different ESMs), I can improve on it.

TBH, while the `openmrsComponentDecorator()` function gets the job done, it doesn't feel intuitive to use. We really just need it to wrap a component in react contexts for SWR and translation. I think exposing the wrapper component would be simpler. So we can use it like this:

```
<OpenmrsReactComponent> {# this provides the context for SWRConfig and I18nextProvider #}
  {children}
</OpenmrsReactComponent>
```

chibongho:
Yeah... I was looking into it just now and realized the higher order component is harder to refactor that I thought. I'll let it be for now.

PR #1209, opened 2024-11-19
(refactor) Refactor reusable patient banner components

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR refactors the reusable patient banner components as follows:

### PatientBannerPatientIdentifier

- Renames the component export from `PatientBannerPatientIdentifier` to `PatientBannerPatientInfoIdentifiers` to better describe its purpose. The pluralized name is more representative of the fact that it renders a list of identifiers.
- Renames the `identifiers` array prop from `identifier` to `identifiers` to better describe its purpose.
- Fixes incorrect usage of the `showIdentifierLabel` prop. Per its description, it should toggle the display of the identifier label, not the entire identifier.
- Removes the `title` prop from the Carbon `Tag` component which has since been deprecated.
- Adds a test for the component.

### PatientBannerPatientInfo

- Adds a default export.
- Adds a test for the component.

### Misc

- Adds a stub for `usePrimaryIdentifierCode` to the `esm-react-utils` mock. I'm not sure why we have a stub for the resource file, instead of just one for the function it exports. I'll look into this more closely.

## Screenshots

No UI changes

## Related Issue
https://openmrs.atlassian.net/browse/O3-4160

## Other

This would require an update to the framework and tooling versions of Core in the places where we're using these components.


chibongho (approved):
Thanks!

PR #1210, opened 2024-11-20
(fix) Fix change password modal

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR fixes the change password modal as follows:

- Adds default values to the change password form schema - this is to prevent the form from being empty when the user clicks on the change password button.
- Adds min validation to the old password, new password and password confirmation fields to ensure that the user has entered a value.
- Adds the appropriate dependencies to the submit handler useCallback hook.
- Surfaces the error message returned from the change password API.
- Showing an inline error notification in the modal instead of a snackbar.
- Fixes a translation key/value pair.

### Surfacing error messages from the Password resource

CleanShot 2024-11-20 at 4  10 10@2x

### DIFF
--- a/packages/apps/esm-login-app/src/change-password/change-password.component.tsx
+++ b/packages/apps/esm-login-app/src/change-password/change-password.component.tsx
@@ -44,29 +44,32 @@ const ChangePassword: React.FC = () => {
     resolver: zodResolver(changePasswordFormSchema),
   });
 
-  const onSubmit: SubmitHandler<z.infer<typeof changePasswordFormSchema>> = useCallback((data) => {
-    setIsChangingPassword(true);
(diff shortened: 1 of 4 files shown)


denniskigen:
Would it make more sense to inline the error notification in the modal? Probably does...

chibongho (approved):
LGTM!

chibongho:
> Would it make more sense to inline the error notification within the modal? Probably does... I've amended it to use the inline notification instead of a snackbar.

I'm not the best person to ask, but looks fine to me.

ojwanganto:
> Would it make more sense to inline the error notification within the modal? Probably does... I've amended it to use the inline notification instead of a snackbar.

Considering this is on a modal, I would go with the inline notification within the modal

PR #1213, opened 2024-11-23
(fix) Correct Typo in Openmrs Component Decorator For Strict Mode

kb019:
# Requirements

## For changes to apps

## If applicable

## Summary
 I am not sure if this is a typo, Please feel free to close the PR if its not . The default value for the strictMode is true. But in the if condition we don't render the component with the strictMode when its true.


brandones:
Your change looks correct to me; it looks like there was indeed a typo. It may or may not uncover a lot of console warnings on the dev server or production environments. I suggest approving but think @ibacher or @denniskigen should take a look, in case they have a better idea whether/where we're using unsafe components in practice.

Thanks @kb019

ibacher:
There are no unsafe components I'm aware of. This is also a very clear and obvious explanation for why Strict Mode never seemed to work. Thanks!

PR #1228, opened 2024-12-06
(fix) Notify config system when app is fully loaded

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This is a companion commit for openmrs/openmrs-esm-patient-chart#2139. The underlying issue in that PR is that a module is only registered as loaded when `defineConfigSchema()` is called, causing the promises used to load the configuration to never resolve when `defineConfigSchema()` isn't called. (It is technically an error to call `useConfig()` without having called `defineConfigSchema()`, but its unclear when to log this error since we can't always know when the config is _read_).

This PR adds a call at the end of the app loading process to ensure that the module is marked as "loaded" in the configuration system.

Note that openmrs/openmrs-esm-patient-chart#2139 should still be committed even with this change.

### DIFF
--- a/packages/framework/esm-config/src/module-config/module-config.test.ts
+++ b/packages/framework/esm-config/src/module-config/module-config.test.ts
@@ -51,6 +51,8 @@ describe('defineConfigSchema', () => {
       bar: true,
     };
     Config.defineConfigSchema('foo-module', schema);
+    // load to trigger validation
+    Config.registerModuleLoad('foo-module');
     expect(console.error).toHaveBeenCalledWith(expect.stringMatching(/foo-module.*bar/));
   });
 
@@ -59,6 +61,8 @@ describe('defineConfigSchema', () => {
       bar: { baz: 'bad bad bad' },
     };
     Config.defineConfigSchema('foo-module', schema);
+    // load to trigger validation
+    Config.registerModuleLoad('foo-module');
     expect(console.error).toHaveBeenCalledWith(expect.stringMatching(/foo-module.*bar\.baz/));
   });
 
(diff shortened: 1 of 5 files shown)


brandones (approved):
This looks like a much better way to handle updating `moduleLoaded`.

Should we remove [the line in `defineConfigSchema`](https://github.com/openmrs/openmrs-esm-core/blob/7787ef8c369335452b2af6cc3c34e4509bb9afa0/packages/framework/esm-config/src/module-config/module-config.ts#L173)?

I wonder if defining extension config works at all, and whether this fix will allow it to work. The fact that [extension-config.test.tsx](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/src/integration-tests/extension-config.test.tsx) does not have any tests using `defineExtensionConfigSchema` gives me some pause.

ibacher:
> Should we remove [the line in defineConfigSchema](https://github.com/openmrs/openmrs-esm-core/blob/7787ef8c369335452b2af6cc3c34e4509bb9afa0/packages/framework/esm-config/src/module-config/module-config.ts#L173)?

Honestly, that's initially what I did and I got lazy fixing test, but I think you're right.

> I wonder if defining extension config works at all

It actually does:

Screenshot adding obs-by-encounter-widget to slot

This is just the obs-by-encounter widget added to a random slot with no additional config. Since it's showing something, the schema defaults are taking, so something is parsing the schema and correctly associating it. That said, it doesn't _look_ like it should work to me...

ibacher:
So, not marking the module as loaded at the end of `defineConfigSchema()` causes a lot of failures. I'm not sure why... in my mental model of how things work, this should be the equivalent of just having the module marked as loaded as soon as `startupApp()`, if any, is run.

brandones:
This fixed https://openmrs.atlassian.net/browse/O3-4254 if I'm not mistaken

ibacher:
@brandones Yes, though not as fully as I'd like.

PR #1230, opened 2024-12-09
(chore) Add FetchError interface for more granular error typing

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR adds a `FetchError` interface to the framework that can be used to annotate errors that occur during fetch requests. It also amends the `OpenmrsFetchError` class to implement the `FetchError` interface. Presently, the `openmrsFetch` hook throws a custom error that includes the response status code and status text if a request fails. This is not always the most helpful message, and it's not always clear what the underlying error was.

The REST API returns a more useful message in the responseBody, and it would be preferable to show this message instead of the status code and status text. With this change, we can then use the FetchError interface further down the line in SWR hooks to annotate errors.

### DIFF
--- a/packages/framework/esm-api/src/openmrs-fetch.ts
+++ b/packages/framework/esm-api/src/openmrs-fetch.ts
@@ -292,7 +292,7 @@ export function openmrsObservableFetch<T>(url: string, fetchInit: FetchConfig = 
   });
 }
 
-export class OpenmrsFetchError extends Error {
+export class OpenmrsFetchError extends Error implements FetchError {
   constructor(url: string, response: Response, responseBody: ResponseBody | null, requestStacktrace: Error) {
     super();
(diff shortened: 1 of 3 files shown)


ibacher (approved):
I'm fine with this, but is there any superiority to using `FetchError` rather than `OpenmrsFetchError` in the SWR typings?

denniskigen:
Not at all. Is `OpenmrsFetchError` exported as an interface?

ibacher:
Well, the class itself is exported, so the types should be there by default.

denniskigen:
For anyone wondering what the resolution here was - `OpenmrsFetchError` did not provide the types for FetchError - which is why I added the interface explicitly. Now that this PR is merged in, consumers of SWR hooks can optionally annotate their error objects using this interface.

PR #1231, opened 2024-12-09
(feat) OA-41: openmrsFetch() should handle redirects

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This implements a simple protocol for the backend to be able to send redirect responses to the frontend.

Normally, redirect responses are handled by a HTTP response in the 300-399 range, with a `Location` header indicating where the page should redirect to. The problem is how this is handled with `fetch()` requests. Specifically, while using the `fetch()` API, 3xx responses cause the fetch request itself to redirect to the new URL, meaning that the user's UI doesn't change at all. For SPAs, this is usually the desired behaviour, since SPAs will usually do the work on the frontend.

Instead, this hijacks the meaning of a 204 ("No Content") response from the backend. If the response is a 204 and there's a `Location` header _and_ this feature hasn't been turned off via config, then `openmrsFetch()` will redirect to the URL specified in the header. This gives us a small escape hatch so the backend can redirect the browser to a specific page.

### DIFF
--- a/packages/framework/esm-api/src/openmrs-fetch.ts
+++ b/packages/framework/esm-api/src/openmrs-fetch.ts
@@ -152,6 +152,14 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
     const response = r as FetchResponse<T>;
     if (response.ok) {
       if (response.status === 204) {
+        const { followRedirects } = await getConfig('@openmrs/esm-api');
+        if (followRedirects && response.headers.has('location')) {
+          const location = response.headers.get('location');
+          if (location) {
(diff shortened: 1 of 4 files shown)


wluyima:
This is a great addition @ibacher because it will also automatically address one of the requirements for https://openmrs.atlassian.net/browse/OA-37, i.e. redirect the user to the IDP upon logout when the oauth2 module sets the location header in the response, when do you intend to move this PR out of the draft state?

ibacher:
Basically, once I'm able to test that it works as expected.

denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1253, opened 2024-12-30
(feat) O3-4277: Enable app-specific SWRConfig

Twiineenock:
# Requirements

## For changes to apps

## If applicable

### **Description**

Work in this PR allows app-specific SWR configurations for example the **service queues** and **ward apps**, ensuring their data refresh rates align with their real-time requirements.

### **Background**

Both the service queues and ward apps are used to display near real-time dashboards for **OPD/IPD workflows**. A recent global configuration change in the SWR library increased the default data cache time to 30 minutes. While this change optimizes performance for many apps, it is not suitable for apps requiring more frequent updates.

## Related Issue
[https://openmrs.atlassian.net/browse/O3-4277](https://openmrs.atlassian.net/browse/O3-4277)

### DIFF
--- a/packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx
+++ b/packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx
@@ -1,6 +1,6 @@
 import React, { type ComponentType, type ErrorInfo, Suspense } from 'react';
 import { I18nextProvider } from 'react-i18next';
-import { SWRConfig } from 'swr';
+import { SWRConfig, type SWRConfiguration } from 'swr';
 import type {} from '@openmrs/esm-globals';
 import { openmrsFetch } from '@openmrs/esm-api';
 import { ComponentContext, type ComponentConfig, type ExtensionData } from './ComponentContext';
@@ -31,6 +31,7 @@ export interface ComponentDecoratorOptions {
   featureName: string;
   disableTranslations?: boolean;
   strictMode?: boolean;
+  swrConfig?: Partial<Omit<SWRConfiguration, 'fetcher'>>;
 }
 
 export interface OpenmrsReactComponentProps {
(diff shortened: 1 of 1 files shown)


Twiineenock:
CC
@chibongho , @brandones, @denniskigen

chibongho (dismissed):
Thanks @Twiineenock . I think we should be able to have app-specific SWRConfig without changes to core. I'm thinking of just wrapping the root component in `root.comonent.tsx` (in the ward app and the service queues app) with `<SWRConfig>`.

brandones (approved):
I support this change. Doing it this way has two advantages:
1. We can merge custom SWR config options with our application default options
2. We avoid creating extra nested SWRConfig contexts

Maybe @ibacher or @denniskigen can weigh in, since @chibongho and I are at odds here?

packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx
```diff
@@ -31,6 +31,7 @@ export interface ComponentDecoratorOptions {
   featureName: string;
   disableTranslations?: boolean;
   strictMode?: boolean;
+  swrConfig?: Partial<typeof defaultSwrConfig>;
```
ibacher, line 34: This should really be `Partial<Omit<SWRConfig, 'fetcher'>>`, I think, or something similar.

packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx
```diff
@@ -52,7 +53,7 @@ export function openmrsComponentDecorator<T>(userOpts: ComponentDecoratorOptions
     throw new Error('Invalid options');
   }
 
-  const opts = Object.assign({}, defaultOpts, userOpts);
+  const opts = { ...defaultOpts, ...userOpts };
```
ibacher, line 56: Can we remove this non-functional change?

ibacher (approved):

ibacher:
My own vote here is that we merge this or something like it in. I don't see any downsides to it and it does make it possible for app-specific default configurations to be applied in a relatively low-effort way. I don't think there's any reason that we can't just support both this _and_ adding a nested SWRConfig, if someone feels like doing so.

ibacher:
@Twiineenock Can you remove the updates to the JSON files from this PR? They should _not_ be part of this and are causing conflicts.

ibacher:
So, the problem is this now _deletes_ a lot of files. Maybe it would be easier to start from a new branch? We can't merge a PR that will remove those files either.

Twiineenock:
> So, the problem is this now _deletes_ a lot of files. Maybe it would be easier to start from a new branch? We can't merge a PR that will remove those files either.

@ibacher, I have `hard-reset` the branch with the upstream and re-pushed the changes

ibacher:
Thanks @Twiineenock. Please close out the ticket.

PR #1256, opened 2025-01-06
(feat) Allow registering workspace groups via routes.json

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR allows workspace groups to be registered via routes.json, which is similar to how workspaces are registered. In the previous implementation of `launchWorkspaceGroup` in #1185, there was a design flaw, wherein, for a workspace to be part of a workspace group, it should define the respective group under its definition in the `routes.json`. Now, the workspace group registration will define the individual workspaces that will be the `members` of the respective workspace group, hence the individual workspace will not need to know what all workspace groups it is a part of.

## Screenshots
Yet to test out changes, since dev3 is down.

## Related Issue
https://issues.openmrs.org/browse/O3-4077

### DIFF
--- a/packages/framework/esm-extensions/src/workspaces.ts
+++ b/packages/framework/esm-extensions/src/workspaces.ts
@@ -20,7 +20,9 @@ export interface WorkspaceRegistration {
   groups: Array<string>;
 }
 
-export type WorkspaceGroupRegistration = WorkspaceGroupDefinition;
+export type WorkspaceGroupRegistration = WorkspaceGroupDefinition & {
+  members: Array<string>;
+};
 
 interface WorkspaceRegistrationStore {
   workspaces: Record<string, WorkspaceRegistration>;
@@ -74,20 +76,18 @@ export function registerWorkspace(workspace: RegisterWorkspaceOptions) {
   }));
 }
 
-export type RegisterWorkspaceGroupOptions = WorkspaceGroupRegistration;
-
 /**
  * Tells the workspace system about a workspace group. This is used by the app shell
  * to register workspace groups defined in the `routes.json` file.
  * @internal
  */
-export function registerWorkspaceGroup(workspaceGroup: RegisterWorkspaceGroupOptions) {
+export function registerWorkspaceGroup(workspaceGroup: WorkspaceGroupRegistration) {
   workspaceGroupStore.setState((state) => ({
     workspaceGroups: {
       ...state.workspaceGroups,
       [workspaceGroup.name]: {
         name: workspaceGroup.name,
-        members: workspaceGroup.members ?? [],
+        members: workspaceGroup.members,
       },
     },
   }));
@@ -123,7 +123,7 @@ export function getWorkspaceRegistration(name: string): WorkspaceRegistration {
         canHide: workspaceExtension.meta?.canHide ?? false,
         canMaximize: workspaceExtension.meta?.canMaximize ?? false,
         width: workspaceExtension.meta?.width ?? 'narrow',
-        groups: workspaceExtension.meta?.groups,
+        groups: workspaceExtension.meta?.groups ?? [],
       };
     } else {
       throw new Error(`No workspace named '${name}' has been registered.`);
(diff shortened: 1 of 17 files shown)


packages/framework/esm-extensions/src/workspaces.ts
```diff
@@ -17,9 +17,10 @@ export interface WorkspaceRegistration {
   preferredWindowSize: WorkspaceWindowState;
   load: () => Promise<{ default?: LifeCycles } & LifeCycles>;
   moduleName: string;
-  groups?: Array<string>;
 }
 
+export type WorkspaceGroupRegistration = WorkspaceGroupDefinition;
```
ibacher, line 22: This seems silly to me? Why do we need two different public aliases for the same type?

ibacher: I know we have, e.g., `ExtensionRegistration` and `ExtensionDefinition`, but those are, in fact, different types with differrent (though overlapping) properties.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -1,6 +1,11 @@
 /** @module @category Workspace */
```
ibacher, line 1: Future work: We should probably refactor things to extract the workspace system from the styleguide. The styleguide should really only have components.

packages/framework/esm-extensions/src/workspaces.ts
```diff
@@ -28,6 +29,14 @@ const workspaceRegistrationStore = createGlobalStore<WorkspaceRegistrationStore>
   workspaces: {},
 });
 
+interface WorkspaceGroupRegistrationStore {
+  workspaceGroups: Record<string, { name: string; members: Array<string> }>;
```
ibacher, line 33: Why do we have both a key and a name property?

packages/framework/esm-routes/src/loaders/components.ts
```diff
@@ -194,11 +196,51 @@ supported, so the workspace will not be loaded.`,
...
+  if (!name) {
+    console.error(
+      `A workspace group definition in ${appName} is missing a name and thus cannot be registered.
+To fix this, ensure that you define the "name" field inside the workspace definition.`,
+      workspaceGroup,
+    );
+    return;
+  }
+
+  if (!workspaceGroup.members) {
```
ibacher, line 220: Is it strictly required that a workspace group have `members`? It seems to me that we would want to allow workspace group registrations without members.

packages/framework/esm-routes/src/loaders/pages.ts
```diff
@@ -153,6 +161,22 @@ export function registerApp(appName: string, routes: OpenmrsAppRoutes) {
       }
     });
 
+    availableWorkspaceGroups.forEach((workspaceGroup) => {
+      if (
+        workspaceGroup &&
+        typeof workspaceGroup === 'object' &&
+        Object.hasOwn(workspaceGroup, 'name') &&
+        Object.hasOwn(workspaceGroup, 'members')
```
ibacher, line 169: Ditto above.

ibacher (commented):
Basically LGTM, but a couple of questions / comments.

packages/framework/esm-extensions/src/workspaces.ts
vasharma05: Yeah, I was following the same convention, but yes those are 2 different types.

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
vasharma05: True that!

packages/framework/esm-extensions/src/workspaces.ts
vasharma05: Yes, the key and the name will be the same, but name is kept to get the whole workspace group definition whenever intended, instead of defining the name as the key again.

vasharma05:
Hi @ibacher,
Requesting your re-review here.
Thanks!

usamaidrsk (approved):
LGTM
Thanks @vasharma05

PR #1259, opened 2025-01-07
(feat) Add icons for various order types

vasharma05:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR adds icons for drug, imaging, procedure, material and referral order types.

## Screenshots
None

## Related Issue
None


ibacher:
So, this needs a few changes.

1. Pictograms are these things , i.e., SVGs rendered at the top of the app in 92x92 format. These are much closer to icons.
2. The icons we use are SVGs with a 20x20 viewport. These SVGs are 24x24. Normally, I've just asked Sonder to convert things into the correct format.
3. Both icons and pictograms should be uncolored, which means we need to somewhat modify those SVGs to remove the fills, etc.

vasharma05:
Hi @ibacher!

> The icons we use are SVGs with a 16x16 viewport. These SVGs are 24x24. Normally, I've just asked Sonder to convert things into the correct format.

Yeah, I tried removing the size factor from these SVGs, but wasn't being able to do the same.

> Both icons and pictograms should be uncolored, which means we need to somewhat modify those SVGs to remove the fills, etc.

Yes, they should be uncolored, but as per designs for the order basket, these pictograms are colored, I have asked for these icons from this slack thread: https://openmrs.slack.com/archives/C06SSBQ7FPU/p1733426639388409

Thanks!

ibacher:
> Yeah, I tried removing the size factor from these SVGs, but wasn't being able to do the same.

Basically, the problem here is that neither the pictogram nor icon setups are really that adaptable. The SVGs need to be in the expected viewport size or else they won't render correctly. So either we need versions of these upscaled to a 80x80 view box (sorry, that's the actual view port size of the pictograms) or downscaled to a 16x16 view box or we need a new custom component for images with a viewbox sized to the size of these. These won't render correctly without one of those three things being done.

> Yes, they should be uncolored, but as per designs for the order basket, these pictograms are colored

Hmmm... maybe we should treat them more similarly to pictograms then. They'll still need to be either resized or will need a custom component, I think to make things work correctly.

ibacher:
@vasharma05 The size that they are meant to displayed at here is irrelevant. We can support that. It's an SVG. The problem is that the `Icon` component is designed to work with SVGs that have a 16 x 16 view box. That view box can be scaled to whatever size we need, but if the SVGs added don't conform to the right view box size, then then won't be rendered correctly.

What's going to happen is that the `Icon` thing will pick out a 16x16 view box from the 24x24 image, dropping everything that doesn't fix, so the image will not look right.

vasharma05:
Hi @ibacher, I have updated the icons and have also added icon for Drug order

packages/framework/esm-styleguide/src/icons/svgs/imaging-order.svg
ibacher, line 1: Isn't this just a colorized version of `user-xray`? I'm not sure we need this.

ibacher: This actually makes me think we should go back to stripping the fills from these icons and set them at the point we're rendering them and need the colors (none of these icons have more than two colors, so we can handle it as `fill` and `background-color`.

vasharma05: Hi @ibacher, I accept that we can fill the colors using the `background-color` and `fill`. But, I have 2 points to consider, 1. These icons are specific to the order basket, and secondly, we are able to add the order types in the order basket via configuration, hence defining the `background-color` and `fill` will be something that will have to handled by configuration. Something similar being faced with the order type colors, for which I created a PR here: https://github.com/openmrs/openmrs-esm-patient-chart/pull/2188

Either we don't put any colors in the SVGs, or we should keep the icons as is. I am hoping to pass SVG registration name in the configuration as is implemented [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/284c69281b4f418bdf06f8ce8537e49192abaa8a/packages/esm-patient-tests-app/src/config-schema.ts#L75) and [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/284c69281b4f418bdf06f8ce8537e49192abaa8a/packages/esm-patient-tests-app/src/test-orders/lab-order-basket-panel/lab-order-basket-panel.extension.tsx#L147). Similar are implemented for General Order types.
Thanks!

vasharma05: Also to add to the above points, the icons will have to modified, because the icon size is smaller than the viewbox to fill in the background colors, hence these icons will appear smaller than the rest of the icons, something which will have to be taken care by the developer implementing these icons, instead of using them out of styleguide directly.
Thanks!

vasharma05: Ian, your final thoughts on this?

ibacher: > the icon size is smaller than the viewbox to fill in the background colors

Every icon has a certain amount of padding. All we've done is remove colors, not change the drawing. Note that every icon we get is always colored, they're just usually black.

> These icons are specific to the order basket

In what way? These icons are being _used_ in the order basket in the designs, but the actual images are not necessarily unique to the order basket, thus, for example, this`imaging-order` icon is the exact image that we have as the `user-xray` icon, except that one is colored green and the other isn't and I think that applies to all of these—even if we don't use them yet, there's no reason that, e.g., the Rx icon is _only_ going to be used here.

I also want to point out that if the argument here is that we _need_ to support users adding custom order types, then we want to make it as flexible as possible for users to add icons for those new order types, i.e., it makes it ridiculous if I can add an order type via config, but get the icon for it, I need to make a PR to update the styleguide. So that seems to suggest that we really _should_ support specifying the background and fill colors via configuration, because then I can define an order type and reuse an existing icon for it.

I'm not trying to be difficult here; I'm trying to make sure that we're not committing stuff that's going to require lots of churn. O3 is meant to be becoming stable.

packages/framework/esm-styleguide/mock.tsx
```diff
@@ -84,6 +84,11 @@ export const UserXrayIcon = () => <span>UserXrayIcon</span>;
 export const ViewIcon = () => <span>ViewIcon</span>;
 export const ViewOffIcon = () => <span>ViewOffIcon</span>;
 export const WarningIcon = () => <span>WarningIcon</span>;
+export const ImagingOrderIcon = () => <span>ImagingOrderIcon</span>;
+export const MaterialOrderIcon = () => <span>MaterialOrderIcon</span>;
+export const ProcedureOrderIcon = () => <span>ProcedureOrderIcon</span>;
+export const ReferralOrderIcon = () => <span>ReferralOrderIcon</span>;
+export const DrugOrderIcon = () => <span>DrugOrderIcon</span>;
```
ibacher, line 91: Can we keep these alphabetized? It makes it easier for us to ensure that all the icons are created.

packages/framework/esm-styleguide/src/icons/icons.tsx
```diff
@@ -226,6 +226,14 @@ export const DownloadIcon = memo(
   }),
 );
 
+/**
+ */
+export const DrugOrderIcon = memo(
+  forwardRef<SVGSVGElement, IconProps>(function DrugOrderIcon(props, ref) {
+    return <Icon ref={ref} icon="omrs-icon-drug-order" iconProps={props} />;
```
ibacher, line 233: If we're adding colorized icons, I can see having a new naming convention for the colored versions. However, we should be consistent...

ibacher:
To be clear about the status here, I'll approve this if you can address the two small technical comments.

vasharma05:
Yes @ibacher , I'll make the changes. Thanks a lot for the above explanations, I align with them.
Thanks again!

vasharma05:
Hi @ibacher, 
The following changes are made as per your suggestions:
1. The background and path fills are removed from all the icons
2. Imaging order icon is removed and user-xray will be used for imaging orders
3. Updated the mock

vasharma05:
@ibacher, requesting your final review here

ibacher (approved):
Thanks @vasharma05!

denniskigen (approved):

PR #1264, opened 2025-01-14
(fix) O3-4339: Frontend should properly support locales with region specifiers

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This attempts to fix a couple of different issues bound up with how specifiers for locales with regions work. Per BCP 47, locales are expressed in the form `<ISO 639 Language Tag>[-<ISO 639 Region Tag>]` (the full spec is slightly more complicated; this is the part we care about), so, e.g., `en` represents "English" and `en-GB` represents "British English".

However, some tooling supports something like BCP 47 with underscores instead of hyphens as separators, representing "English" as `en` and "British English" as `en_GB`. Because both formats agree on the representation of "English" (and the like) we hadn't really noticed until we actually got some translations that were sensitive to regional changes. 

It turns out that the stuff we have that deals with locales wasn't adapted for this change. Specifically:

* Transifex uses the underscore variant, so files created from Transifex are named like `en.json` and `en_GB.json`, but when loading the translations, we use the "correct" BCP 47 tags, and so try to load `en-GB.json`. Since this doesn't resolve correctly, the frontend could never find region-specific translations.
* The backend _correctly_ translates the `locale` property into the correct BCP 47 tag, but this doesn't work with the "defaultLocale" user property, which appears to get set to the the incorrect BCP 47 tag. Since the code to `setUserLanguage()` didn't correctly handle locales with underscores instead of hyphens, the user's locale would not be updated when configured to use a "defaultLocale" with a region specifier

This PR addresses both of those issues.

### DIFF
--- a/packages/framework/esm-api/src/shared-api-objects/current-user.ts
+++ b/packages/framework/esm-api/src/shared-api-objects/current-user.ts
@@ -114,14 +114,22 @@ function isValidLocale(locale: unknown): locale is string {
 }
 
 export function setUserLanguage(data: Session) {
-  const locale = data?.user?.userProperties?.defaultLocale ?? data.locale;
-  const htmlLang = document.documentElement.getAttribute('lang');
+  let locale = data.user?.userProperties?.defaultLocale ?? data.locale;
 
-  if (isValidLocale(locale) && locale !== htmlLang) {
+  if (locale && locale.includes('_')) {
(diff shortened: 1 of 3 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher.

PR #1266, opened 2025-01-14
(fix) - O3-4349 - open workspace in declared workspace group

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

We want to support both:
1. having a workspace group be able to declare member workspaces
2. having a workspace be able to declare which groups it belongs to

I think there was a regression with [this commit](https://github.com/openmrs/openmrs-esm-core/commit/a61deb8b#diff-1c03ea6c69f77e98ed8c9eb205bbb1b02b37aa6c3fe6dc0ddf45c1389b024aceL296) that dropped support for 2. This PR adds it back.

## Screenshots

"Visit Summary" is a workspace that we define in a PIH custom ESM. It declares that it belongs to the `ward-patient` workspace group.

```
  "workspaces": [
    {
      "name": "o2-visit-summary-workspace",
      "component": "o2VisitSummaryWorkspace",
      "title": "visitSummary",
      "type": "o2-visit-summary",
      "sidebarFamily": "ward-patient",
      "hasOwnSidebar": true,
      "width": "extra-wide",
      "groups": ["ward-patient"],
      "canMaximize": true
    },
  }
```

Before:
(notice the workspace action menu icons fail to load)
image

After:
image

## Other

I don't know why, but as seen in the screenshot, the content of the workspace is blank when I run `yarn run:shell` from `esm-core`. It seems not related to this change. It also doesn't seem to be a real issue in dev3, or when running `yarn start --sources esm-ward-app` in patient-management locally

### DIFF
--- a/packages/framework/esm-extensions/src/workspaces.ts
+++ b/packages/framework/esm-extensions/src/workspaces.ts
@@ -82,15 +82,33 @@ export function registerWorkspace(workspace: RegisterWorkspaceOptions) {
  * @internal
  */
 export function registerWorkspaceGroup(workspaceGroup: WorkspaceGroupRegistration) {
-  workspaceGroupStore.setState((state) => ({
-    workspaceGroups: {
-      ...state.workspaceGroups,
-      [workspaceGroup.name]: {
-        name: workspaceGroup.name,
-        members: workspaceGroup.members,
-      },
-    },
-  }));
+  workspaceGroupStore.setState((state) => {
+    const group = state.workspaceGroups[workspaceGroup.name];
+    if (group) {
+      // This condition arises when a workspace with `groups` property registers
+      // before the workspace group is registered, hence the workspace group is
+      // registered by the `attachWorkspaceToGroup` function.
+      return {
+        workspaceGroups: {
+          ...state.workspaceGroups,
+          [workspaceGroup.name]: {
+            ...group,
+            members: Array.from(new Set([...group.members, ...workspaceGroup.members])),
+          },
+        },
+      };
+    } else {
+      return {
+        workspaceGroups: {
+          ...state.workspaceGroups,
+          [workspaceGroup.name]: {
+            name: workspaceGroup.name,
+            members: workspaceGroup.members,
+          },
+        },
+      };
+    }
+  });
 }
 
 const workspaceExtensionWarningsIssued = new Set();
--- a/packages/framework/esm-extensions/src/workspaces.ts
+++ b/packages/framework/esm-extensions/src/workspaces.ts
@@ -85,9 +85,9 @@ export function registerWorkspaceGroup(workspaceGroup: WorkspaceGroupRegistratio
   workspaceGroupStore.setState((state) => {
     const group = state.workspaceGroups[workspaceGroup.name];
     if (group) {
-      // This condition arises when a workspace with `groups` property registers
-      // before the workspace group is registered, hence the workspace group is
-      // registered by the `attachWorkspaceToGroup` function.
+      // This condition occurs when a workspace with a `groups` property is registered before
+      // the corresponding workspace group. In such cases, the workspace group is registered
+      // by the `attachWorkspaceToGroup` function.
       return {
         workspaceGroups: {
           ...state.workspaceGroups,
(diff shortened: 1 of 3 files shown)


mseaton (approved):
Thank you!

packages/framework/esm-styleguide/src/workspaces/workspaces.ts
```diff
@@ -301,7 +301,7 @@ export function launchWorkspace<
   const workspace = getWorkspaceRegistration(name);
   const currentWorkspaceGroup = store.getState().workspaceGroup;
 
-  if (currentWorkspaceGroup && !currentWorkspaceGroup.members?.includes(name)) {
+  if (currentWorkspaceGroup && !(currentWorkspaceGroup.members?.includes(name) || workspace?.groups.includes(currentWorkspaceGroup.name))) {
```
vasharma05, line 304: Hi @chibongho, instead of adding conditions here, we have all the members of a workspaceGroup in the `workspaceGroup` store and the `groups` adds the given workspace to the members of a workspaceGroup at [this place](https://github.com/openmrs/openmrs-esm-core/blob/aef6778cd27c3d072993d43aa56008fbdaa85fa7/packages/framework/esm-extensions/src/workspaces.ts#L166). Hence all the workspaces are in the end added in the to the workspaceGroup's members.

One thing that might be happening is that there might be a race condition going on in the above function, hence the group's definition might be overrided. This is something which must be looked into.

mseaton: @vasharma05 - is this something you are able to fix?  Do you know where the proper fix is?  @ibacher ?  This is trying to fix a regression in our system that has been broken for the last few days and arose (we think) following this commit:  https://github.com/openmrs/openmrs-esm-core/commit/a61deb8b00d527f6d7386d096385bf5d1737a87f .  @ibacher any guidance appreciated.

vasharma05: Yes @mseaton, I made the change. Let me have a look at it.
Thanks!

vasharma05: @chibongho, we connected on a meeting, and it was working well at that time, right?

mseaton: @chibongho is not 100% available today, but he and I met about this and were able to reproduce it locally as well as on our testing server.  I'm not sure why it may have appeared working at the time, but it is an issue for both of us now @vasharma05

vasharma05: Looking into it @mseaton, will there be a way that I can use your implementation to reproduce the issue?

mseaton: Yes, I'll DM you on slack

vasharma05: @mseaton @chibongho, it's the race condition between the `workspaceGroup` registration by the ward-app and the custom workspace's `group` property.

image

mseaton: @vasharma05 - shouldn't your PR remove this change made by @chibongho , if your follow-up change fixes things?

vasharma05: Yes @mseaton, pushed the commit.

Correction: It's not a race condition, actually the `registerWorkspaceGroup` overrides the already registered workspace group and it's members.

vasharma05:
## Update
The workspace registration adds a new entry in the `workspaceGroupStore`, but in case the entry already exists, it actually overrides the existing members of the `workspaceGroup` in consideration.

### Workspace registering before the workspace group is registered
image
image


https://github.com/user-attachments/assets/63a4e3ad-b655-443b-8e67-0b5dfd004a31

vasharma05:
@ibacher, requesting your review here.
Thanks!

packages/framework/esm-extensions/src/workspaces.ts
```diff
@@ -164,28 +182,8 @@ function createNewWorkspaceGroupInfo(groupName: string): WorkspaceGroupRegistrat
 }
```
mseaton, line 183: The function above this is no longer used or needed, right @vasharma05 -  `createNewWorkspaceGroupInfo` ?

packages/framework/esm-extensions/src/workspaces.ts
```diff
@@ -164,28 +182,8 @@ function createNewWorkspaceGroupInfo(groupName: string): WorkspaceGroupRegistrat
...
 
 export function attachWorkspaceToGroup(workspaceName: string, groupName: string) {
```
mseaton, line 184: Do we even need this new function now?  All it does is delegate to another existing function with basically the same arguments, right?

Couldn't any code that calls `attachWorkspaceToGroup(workspaceName, groupName)` instead just call `registerWorkspaceGroup({name: groupName, members: [workspaceName]});` ?

packages/framework/esm-extensions/src/workspaces.ts
vasharma05: Yes, forgot to clean it up

mseaton (approved):
Thanks!

ibacher (approved):

PR #1276, opened 2025-01-23
(chore) O3-4372: Remove non-existent feature flags from local storage

Bharath-K-Shetty:
# Requirements

## For changes to apps

## If applicable

## Summary

This pull request addresses the issue of removing feature flags that no longer exist from local storage. The following changes have been made:

1. Added cleanupObsoleteFeatureFlags function: This function removes feature flags from local storage that no longer exist in the current state.

2. Called cleanupObsoleteFeatureFlags within featureFlagsStore.subscribe: Ensures that obsolete feature flags are cleaned up whenever the state is updated.

3. Called cleanupObsoleteFeatureFlags within registerFeatureFlag: Ensures that obsolete feature flags are cleaned up whenever a new feature flag is registered.

**Testing**
To verify the functionality, the following steps were performed:

1. Registered a new feature flag:

```
registerFeatureFlag('testFlag', 'Test Flag', 'This is a test flag');
```

2. Checked local storage:

- Opened the browser's developer tools.

- Navigated to the "Application" tab

- Verified that the key openmrs:feature-flag:testFlag was present in local storage.

3. Removed the feature flag from the application's state:

```
window.featureFlagsStore.setState((state) => {
  const { testFlag, ...remainingFlags } = state.flags;
  return { flags: remainingFlags };
});
```

4. Checked local storage again:

- Verified that the key openmrs:feature-flag:testFlag was removed from local storage.

## Screenshots

https://github.com/user-attachments/assets/eb0c60aa-b5ab-4c2b-9418-1a7e897f56c6

## Related Issue

https://openmrs.atlassian.net/browse/O3-4372


ibacher (changes requested):
Thanks @Bharath-K-Shetty! So, the deletion code looks correct, but I don't think calling it at the two points that this PR does will delete them at the right point. This is because `registerFeatureFlag()` is called _as each feature flag is registered_, which means, on each run, this would—presumably—delete all existing flags from local storage, which is not what we want.

We need to find a point in the app start-up cycle when the feature flags will all be registered to call this, so likely at some point in the app shell itself.

ibacher:
> I moved the cleanupObsoleteFeatureFlags function to core-config.ts

I don't think that's the right point. `setupCoreConfig()` is called way too early in the cycle. I'd add a new `then()` call after the last one in the chain [here](https://github.com/openmrs/openmrs-esm-core/blob/d2f3b6de53fc0aa7210e373eb7e83dd549380198/packages/shell/esm-app-shell/src/run.ts#L423C11-L423C44). All feature flags should be registered by then.

> I noticed that the billing module is not registering its feature flags

The app needs to be updated to register the feature flags via it's routes.json, rather than [this](https://github.com/openmrs/openmrs-esm-billing-app/blob/3f6ba6da3922f1f07565c8aeaf9fb031475d32d3/src/index.ts#L23), which is only called at an unpredictable point in time.

Bharath-K-Shetty:
@ibacher While I called cleanup in run.js, it runs after registering the patient flags. However, the other two flags, like bedmanagement-module and emrapi-module, are registered after the cleanup.
Should these modules be updated to register their feature flags via routes.json?

packages/framework/esm-feature-flags/src/feature-flags.ts
```diff
@@ -62,6 +63,22 @@ export function registerFeatureFlag(flagName: string, label: string, description
...
+ */
+function cleanupObsoleteFeatureFlags(currentFlags?: FeatureFlagsStore['flags']) {
+  const flags = currentFlags || featureFlagsStore.getState().flags;
+  const storedFlags = Object.keys(localStorage).filter((key) => key.startsWith('openmrs:feature-flag:'));
+  storedFlags.forEach((key) => {
+    const flagName = key.replace('openmrs:feature-flag:', '');
+    if (!flags[flagName]) {
+      localStorage.removeItem(key);
+      localStorage.removeItem(`openmrs:feature-flag-meta:${flagName}`);
+    }
```
Pratibha0280, line 80: Hey @Bharath-K-Shetty  one small suggestion Should we add error handling around localStorage operations? In cases like private browsing (especially Safari), accessing or modifying localStorage might throw an error.

denniskigen:
Ok, this PR is ready to progress following the merging of these dependent PRs:

- https://github.com/openmrs/openmrs-esm-billing-app/pull/77
- https://github.com/openmrs/openmrs-esm-patient-chart/pull/2365

denniskigen:
@ibacher I think we can move this along

ibacher:
@Bharath-K-Shetty We've reworked things so that all the feature flags are registered in the routes.registry.json. Could we switch this back to the implementation that fires off after that step is done? Thanks!

Bharath-K-Shetty:
Yes @denniskigen i will update this asap..!

denniskigen:
@ibacher @Bharath-K-Shetty mentioned on the coffee break call that he had an a problem with removing the feature flags after trying the suggested approach. Do you mind pushing a commit implementing the change so we can close this one out?

ibacher (approved):

PR #1279, opened 2025-02-03
(feat) O3-4425 - add ability to define collapsed left nav

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR adds the ability for apps to define a left nav that's always collapsed into the top nav hamburger menu.

This PR also includes refactoring to remove the need to render `<LeftNavMenu>` within the home or patient-chart app. Currently, the home / patient-chart app needs to render `<LeftNavMenu>` in its root component in order for the left nav to show in desktop mode. The top nav, however, also has a `<SideMenuPanel>`, which renders the menu when in tablet mode and the hamburger button is toggled on. (Incidentally, `<SideMenuPanel>` actually uses `<LeftNavMenu>` as child component)

Screenshot: `<LeftNavMenu>` rendered by home app:
image

Screenshot: `<SideMenuPanel>` rendered by primary-navigation-app:
image

The `<LeftNavMenu>` is a carbon's `<SideNav>` component, which is a `position: fixed` element. This means we shouldn't really need the home or patient chart app to render it as part of its component tree. The refactoring makes it so the that the `<SideMenuPanel>` renders both:
-  the pop-up menu when the top nav hamburger is pressed, in tablet mode
- the actual left nav, in desktop mode

## Screenshots

Video of the home app with this change. No noticeable difference from before. (Not part of this PR, but in the video capture, `<LeftNavMenu>` is removed from the home app.)
https://github.com/user-attachments/assets/f89bae32-5b20-4270-aa44-8874fff62cf4

Video of the dispensing app with this change, The dispensing app added the following to define the left nav to be always collapsed:

    useEffect(() => {
      const basePath = window.spaBase + "/dispensing"; 
      setLeftNav({ name: 'homepage-dashboard-slot', basePath, mode: 'collapsed'});
      return () => unsetLeftNav('homepage-dashboard-slot');
    }, []);

https://github.com/user-attachments/assets/12f8b46f-2776-4676-a287-054db27ded27

## Related Issue

https://openmrs.atlassian.net/browse/O3-4425

## Other

I attempted to refactor `<LeftNavMenu>`  hoping this will give us a cleaner way to define a collapsed left nav. While this change removes the need to render `<LeftNavMenu>` in home / patient-chart, it fails to completely abstract the left nav logic away from those apps. In particular, the home dashboard actually needs to maintain a `margin-left` in order to be rendered to the right of the left nav in desktop mode. (That margin is set to 0 in tablet mode). ~~I'm not sure if there is a way around it as the left nav is `position: fixed`.~~ This is probably doable with a more invasive change by overriding the `position: fixed` in the carbon `<SideNav>` and adding a `top-nav-container` div in `index.ejs` (so we can render the top nav there via `createPortal())`.

image

### DIFF
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
@@ -25,9 +25,10 @@ const SideMenuPanel: React.FC<SideMenuPanelProps> = ({ expanded, hidePanel }) =>
   const leftNavContainer = window.document.getElementById('omrs-left-nav-container-root');
   return (
     <>
-      {(!isDesktop(layout) || mode == 'collapsed') && expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />}
+      {(!isDesktop(layout) || mode === 'collapsed') && expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />}
       {isDesktop(layout) &&
-        mode != 'collapsed' &&
+        mode !== 'collapsed' &&
+				leftNavContainer &&
         createPortal(<LeftNavMenu ref={menuRef} isChildOfHeader />, leftNavContainer)}
     </>
   );
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
@@ -1,11 +1,17 @@
 import React, { useEffect } from 'react';
-import { LeftNavMenu, useOnClickOutside } from '@openmrs/esm-framework';
+import { isDesktop, LeftNavMenu, useLayoutType, useOnClickOutside } from '@openmrs/esm-framework';
+import { createPortal } from 'react-dom';
 
 interface SideMenuPanelProps {
   expanded: boolean;
   hidePanel: Parameters<typeof useOnClickOutside>[0];
 }
 
+/**
+ * This is the menu that pops up when clicking on the hamburger button
+ * on the top nav. It's also responsible for rendering the left nav
+ * in desktop mode (via a react portal).
+ */
 const SideMenuPanel: React.FC<SideMenuPanelProps> = ({ expanded, hidePanel }) => {
   const menuRef = useOnClickOutside(hidePanel, expanded);
 
@@ -13,8 +19,15 @@ const SideMenuPanel: React.FC<SideMenuPanelProps> = ({ expanded, hidePanel }) =>
     window.addEventListener('popstate', hidePanel);
     return () => window.removeEventListener('popstate', hidePanel);
   }, [hidePanel]);
+  const layout = useLayoutType();
 
-  return expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />;
+  const leftNavContainer = window.document.getElementById('omrs-left-nav-container-root');
+  return (
+    <>
+      {!isDesktop(layout) && expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />}
+      {isDesktop(layout) && createPortal(<LeftNavMenu ref={menuRef} isChildOfHeader />, leftNavContainer)}
+    </>
+  );
 };
 
 export default SideMenuPanel;
--- a/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
+++ b/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react';
-import { isDesktop, LeftNavMenu, useLayoutType, useOnClickOutside } from '@openmrs/esm-framework';
+import { isDesktop, LeftNavMenu, useLayoutType, useLeftNavStore, useOnClickOutside } from '@openmrs/esm-framework';
 import { createPortal } from 'react-dom';
 
 interface SideMenuPanelProps {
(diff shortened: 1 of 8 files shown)


packages/framework/esm-styleguide/src/left-nav/index.tsx
```diff
@@ -25,16 +25,30 @@ export function unsetLeftNav(name) {
...
+  return useStore(leftNavStore);
+}
 type LeftNavMenuProps = SideNavProps;
 
+/**
+ * This component renders the left nav in desktop mode. It's also used to render the same
+ * nav when the hamburger menu is clicked on in tablet mode. See side-menu-panel.component.tsx
+ *
+ * Use of this component by anything other than <SideMenuPanel> (where isChildOfHeader == false)
+ * is deprecated; it simply renders nothing.
```
chibongho, line 38: This is a bit hacky. If we feel comfortable with this, we can opt for more aggressive breaking change where we stop exporting `<LeftNavMenu>` in `@openmrs/esm-framework`, and force other ESMs to remove it.

chibongho:
@ibacher @denniskigen @samuelmale @vasharma05 This touches core code, so I'm hoping to get feedback from outside PIH. Please review when you have a chance. Thanks!

chibongho:
Bump. Can I get a review for this? @ibacher @denniskigen @vasharma05 @samuelmale 

I just investigated the E2E failure. It's caused by the patient chart having a custom VisitHeader in the top nav. It should be fixed when we remove the VisitHeader.  (There is a [separate PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2262) for that.)

Is there a way to either: bypass the E2E test failure in patient chart, or run the E2E test again a different branch?

packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx
```diff
@@ -44,7 +46,10 @@ const HeaderItems: React.FC = () => {
     [],
   );
 
-  const showHamburger = useMemo(() => !isDesktop(layout) && navMenuItems.length > 0, [navMenuItems.length, layout]);
+  const showHamburger = useMemo(
+    () => (!isDesktop(layout) || mode == 'collapsed') && navMenuItems.length > 0,
```
denniskigen, line 50:
```suggestion
    () => (!isDesktop(layout) || mode === 'collapsed') && navMenuItems.length > 0,
```

nit

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -1,20 +1,36 @@
...
     return () => window.removeEventListener('popstate', hidePanel);
   }, [hidePanel]);
+  const layout = useLayoutType();
+  const { mode } = useLeftNavStore();
 
-  return expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />;
+  const leftNavContainer = window.document.getElementById('omrs-left-nav-container-root');
+  return (
+    <>
+      {(!isDesktop(layout) || mode == 'collapsed') && expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />}
```
denniskigen, line 28:
```suggestion
      {(!isDesktop(layout) || mode === 'collapsed') && expanded && <LeftNavMenu ref={menuRef} isChildOfHeader />}
```

denniskigen: nit

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -1,20 +1,36 @@
...
+      {isDesktop(layout) &&
+        mode != 'collapsed' &&
```
denniskigen, line 30:
```suggestion
        mode !== 'collapsed' &&
```

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
```diff
@@ -1,20 +1,36 @@
...
+        createPortal(<LeftNavMenu ref={menuRef} isChildOfHeader />, leftNavContainer)}
```
denniskigen, line 31:
```suggestion
				leftNavContainer &&
        createPortal(<LeftNavMenu ref={menuRef} isChildOfHeader />, leftNavContainer)}
```

nit: Would it be useful to assert the existence of the `omrs-left-nav-container-root` DOM element here, or is that guaranteed?

chibongho:
> Is this artifacting to be expected?
> artifacting.mp4

Yeah, it's caused by the patient chart having a custom VisitHeader in the top nav. It should be fixed when we remove the VisitHeader. (There is a https://github.com/openmrs/openmrs-esm-patient-chart/pull/2262 for that.) That PR is pretty big, so maybe I can make a smaller one in patient chart just to unblock this.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/side-menu-panel.component.tsx
chibongho: The `omrs-left-nav-container-root` div is hard-coded into `index.ejs`, so it should be guaranteed.

packages/framework/esm-styleguide/src/left-nav/index.tsx
```diff
@@ -25,16 +38,30 @@ export function unsetLeftNav(name) {
...
-  );
+  if (props.isChildOfHeader) {
+    return (
+      <SideNav ref={ref} expanded aria-label="Left navigation" className={styles.leftNav} {...props}>
+        <ExtensionSlot name="global-nav-menu-slot" />
+        {slotName ? <ExtensionSlot name={slotName} state={{ basePath, currentPath }} /> : null}
+      </SideNav>
+    );
+  } else {
+    return <></>;
```
denniskigen, line 65: nit: Is it preferable to return `null` in an early return if this is `!isChildOfHeader`?

```tsx
  // Early return if not a child of header
  if (!props.isChildOfHeader) {
    return null;
  }
```

denniskigen:
> That PR is pretty big, so maybe I can make a smaller one in patient chart just to unblock this.

We probably should so that merging this doesn't break things and we end up in a situation where the fix PR isn't ready to check in yet.

denniskigen (approved):
LGTM

PR #1280, opened 2025-02-12
(feat) O3-4013: Add support for partial (estimated) dates

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This should resolve O3-4013. In essence, for _estimated_ birthdates, the FHIR API returns only a partial date. DayJS interpolates this with (essentially) the "zero value" for each field, so, e.g., the FHIR string "2014" (representing "born sometime in 2014") gets interpreted as "2014-01-01", etc. which leads to screens that looks incorrect.

Because I was lazy, this uses the `any-date-parser` even though we don't actually need most of it's features. However, it was the only date parser I could find that would return a partial object instead of trying to construct a date (at least without hacking into non-public APIs). Eventually, we may want to replace this with something lighter-weight.

## Screenshots

Screenshot showing how this renders partial dates

## Other

I'm not entirely sure the changes to `age()` are necessary.

Note that there are some serious edge cases with estimated birth dates less than a day old, but hopefully that's not something we need to deal with.

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced date formatting capabilities for displaying partial dates, improving localization and clarity in user interfaces.
- **Documentation**
  - Updated API documents to reflect the new date formatting features and simplified parameter types.
- **Tests**
  - Refactored test cases for age calculation to enhance clarity and scalability.
- **Chores**
  - Revised configuration and dependency settings to improve module compatibility and overall system stability.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -144,7 +144,6 @@ jobs:
       - name: 📦 Install dependencies
         if: steps.cache-dependencies.outputs.cache-hit != 'true'
         run: yarn install --immutable
-        working-directory: e2e_repo
 
       - name: 🚀 Setup local cache server for Turborepo
         uses: felixmosh/turborepo-gh-artifacts@v3
@@ -206,7 +205,7 @@ jobs:
           name: report-${{ matrix.repo }}
           path: e2e_repo/playwright-report/
           retention-days: 30
-      
+
       - name: 📝 Capture Server Logs
         if: always()
         uses: jwalton/gh-docker-logs@v2
@@ -221,4 +220,3 @@ jobs:
           path: "./logs"
           retention-days: 2
           overwrite: true
-
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -33,6 +33,7 @@
 - [toVisitTypeObject](API.md#tovisittypeobject)
 - [updateVisit](API.md#updatevisit)
 - [useAttachments](API.md#useattachments)
+- [useEmrConfiguration](API.md#useemrconfiguration)
 - [useLocations](API.md#uselocations)
 - [usePatient](API.md#usepatient)
 - [usePrimaryIdentifierCode](API.md#useprimaryidentifiercode)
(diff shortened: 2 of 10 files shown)


denniskigen:
FYI @veronicamuthee @paulsonder @ciaranduffy

NethmiRodrigo:
So one of the failing e2e tests are expected "e2e/specs/register-new-patient.spec.ts:87:5 › Register an unknown patient › And I should see the newly registered patient's details displayed in the patient banner" where its expecting to see a full birthdate by searching for "01-Jan-2000" which this PR changes to just "2000". So we'll need to update the e2e test itself on patient-management once this goes in.

NethmiRodrigo:
The other failing test (is for one very badly written), which tests managing patient lists fails because it tries to test the deletion of a patient from a list (expect it does so by directly calling the API in the test - https://github.com/openmrs/openmrs-esm-patient-management/blob/9201ed389bd8f811ff2c1f265c4ccbf60d5c3079/e2e/specs/patient-list.spec.ts#L84 instead of interacting the UI (as an e2e test should). Anyways the API call apparently fails weirdly which is why the test fails. CC @denniskigen

gracepotma:
Two questions.
1) @ibacher what is the use case that would be a problem, when you mentioned "serious edge cases with estimated birth dates less than a day old"? Since PIH is planning to use O3 with freshly just-born babies at the MCOE this could come up after they go live in October so would be good to know what might happen.

2) The screenshot shows "11 years". What would be shown if the estimated DOB was entered as "11 years 8 months"? Or "6 years 5 months"? Or "2 years 2 months"? Would it be "11" and "6" and "2", or 11y 8mo, and 6y5mo, and 2y2mo?

ibacher:
> "serious edge cases with estimated birth dates less than a day old"?

Effectively, they'd show up as "0 yrs" or something. But that only applies if they have an estimated birthdate and there's currently no way to estimate a birth date less than 1 month.

> The screenshot shows "11 years". What would be shown if the estimated DOB was entered as "11 years 8 months"?

This is a limitation of the data model for estimated birthdates. We do not retain any information on what was actually entered. Instead it's converted into a date and time, so 11 years 8 months would show up as eleven years. But for anyone under 5, we also report the month as stored in the database, so you could get "2 years 2 months". Again, this only applies to _estimated_ birth dates. Anyone with a non-estimated birthdate will show the full date and accurate age as before.

denniskigen:
Have you looked into the test failure, @ibacher?

ibacher:
@denniskigen Yes. The failing test needs to be adjusted because it's testing for exactly the condition this PR aims to fix (an estimated birthdate being interpreted as Jan 1).

denniskigen (approved):
Thanks, @ibacher!

PR #1283, opened 2025-02-13
(fix) Address some minor issues with OpenmrsDatePicker overlay

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

The goal of this PR is to fix a couple of small issues with the OpenmrsDatePicker. Specifically:

1. In the chart, opening the Popover causes some shifts in how things render in general. This is because, by default react-aria assumes that the Popover is rendered as a modal and aggressively hides, but in this case, it is not a modal.
2. While clicking on the button correctly works with the Popover, clicking on the field doesn't correctly toggle it. I _think_ this is due to some funkiness with how react-aria button work internally. In any case, I've moved to a simpler (to me) setup where there's just a state in the root component a toggle function that's passed to the points we want to enable toggling the popover.

See the little GIFs below for what this is fixing.

## Screenshots

**Before**

datepicker-broken

**After**

datepicker-working

## Other

I've also updated to the latest version of react-aria-components and adopted a small set of changes from the updates to the code that is modified from there.

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -546,7 +546,7 @@ A type for any of the acceptable date formats
 
 #### Defined in
 
-[packages/framework/esm-styleguide/src/datepicker/index.tsx:73](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/datepicker/index.tsx#L73)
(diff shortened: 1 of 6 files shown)


ibacher:
This change isn't nearly as big as the Git commit says. Almost all of that seems to be in the Yarn lockfile.

vasharma05:
@ibacher, there's also one more issue I found, which is that the width of the date picker if fixed, and when placing the date picker in compact places, it doesn't squeeze, and hence it makes the container horizontally scrollable. One example of this is in the start visit form on the patient chart.

NethmiRodrigo (approved):

ibacher:
@vasharma05 Could you add that as a ticket?

PR #1285, opened 2025-02-14
(fix) O3-2076 & O3-4436: Navigation occasionally breaks & fix logout with OAuth2

wluyima:
[O3-4436](https://openmrs.atlassian.net/browse/O3-4436): Navigation occasionally breaks

# Requirements

## For changes to apps

## If applicable

## Other

[O3-4436]: https://openmrs.atlassian.net/browse/O3-4436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
+++ b/packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
@@ -12,7 +12,9 @@ const RedirectLogout: React.FC = () => {
   useEffect(() => {
     clearHistory();
     if (!session.authenticated || !isLoginEnabled) {
-      navigate({ to: '${openmrsSpaBase}/login' });
+      if (config.provider.type !== 'oauth2') {
+        navigate({ to: '${openmrsSpaBase}/login' });
+      }
     } else {
       performLogout()
         .then(() => {
@@ -24,9 +26,7 @@ const RedirectLogout: React.FC = () => {
             sessionId: '',
           });
 
-          if (config.provider.type === 'oauth2') {
-            navigate({ to: config.provider.logoutUrl });
-          } else {
+          if (config.provider.type !== 'oauth2') {
             navigate({ to: '${openmrsSpaBase}/login' });
           }
         })
--- a/packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
+++ b/packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
@@ -72,11 +72,10 @@ describe('RedirectLogout', () => {
     expect(mockNavigate).toHaveBeenCalledWith({ to: '${openmrsSpaBase}/login' });
   });
 
-  it('should redirect to `provider.logoutUrl` if the configured provider is `oauth2`', async () => {
+  it('should not redirect if the configured provider is `oauth2`', async () => {
     mockUseConfig.mockReturnValue({
       provider: {
         type: 'oauth2',
-        logoutUrl: '/oauth/logout',
       },
     });
 
@@ -95,7 +94,7 @@ describe('RedirectLogout', () => {
       authenticated: false,
       sessionId: '',
     });
-    expect(mockNavigate).toHaveBeenCalledWith({ to: '/oauth/logout' });
+    expect(mockNavigate).toHaveBeenCalledTimes(0);
   });
 
   it('should redirect to login if the session is already unauthenticated', async () => {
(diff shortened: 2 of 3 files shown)


packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
```diff
@@ -24,9 +26,7 @@ const RedirectLogout: React.FC = () => {
             sessionId: '',
           });
 
-          if (config.provider.type === 'oauth2') {
-            navigate({ to: config.provider.logoutUrl });
```
ibacher, line 28: We should also remove the corresponding configuration element from the configuration schema.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
```diff
@@ -72,20 +72,15 @@ describe('RedirectLogout', () => {
...
-        logoutUrl: '/oauth/logout',
       },
     });
 
     render(<RedirectLogout />);
 
-    expect(mockOpenmrsFetch).toHaveBeenCalledWith(`${restBaseUrl}/session`, {
-      method: 'DELETE',
-    });
-
```
ibacher, line 88: We still should assert that this was called.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
```diff
@@ -153,15 +148,29 @@ describe('RedirectLogout', () => {
 
     mockUseConfig.mockReturnValue({
       provider: {
-        type: 'oauth2',
-        logoutUrl: '/new/logout/url',
+        type: 'testProvider',
```
ibacher, line 151: Just remove this test. The whole point was to test the `oauth2` value with `logoutUrl`. If we're not using that, there's no point in retaining this test.

ibacher (changes requested):
Thanks @wluyima! A few small things. Also please note that we have slightly different conventions in the frontend than in the backend, specifically, we ask that you:

> Ensure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing#contributing-guidelines) label (such as feat, fix, or chore, among others). See existing PR titles for inspiration.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
wluyima: In my opinion we might just have to remove the entire oauth2 provider but I suggest let's do that in a separate ticket and after a proper inclusive discussion.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
wluyima: I wanted to remove it but my assumption was that there could be other provider implementation that would still need this behavior.

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
```diff
@@ -153,15 +148,29 @@ describe('RedirectLogout', () => {
...
 
     rerender(<RedirectLogout />);
 
     await waitFor(() => {
-      expect(mockNavigate).toHaveBeenCalledWith({ to: '/new/logout/url' });
+      expect(mockNavigate).toHaveBeenCalledWith({ to: '${openmrsSpaBase}/login' });
+    });
+  });
+
+  it('should not redirect to login if user is not authenticated the provider is oauth2', async () => {
```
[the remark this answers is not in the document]
wluyima: Yeah, am missing an **and**

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.test.tsx
ibacher: There aren't any. The "provider" concept was really only so we could redirect to /oauth2logout and /oauth2login as appropriate.

ibacher: Basically, if we don't need it to work with OAuth2, we can gut it with no additional risk. If we really needed a login provider system, we'd likely need to do a major refactoring of everything anyway.

wluyima: I updated the config to remove the logout URL field

dkayiwa (approved):

PR #1296, opened 2025-02-27
Fix: O3-4478: Redirect logged-in users from login page to home

Lemeri123:
# Requirements

## For changes to apps

## If applicable

## Summary
fixes [O3-4478](https://openmrs.atlassian.net/browse/O3-4478)

### DIFF
--- a/packages/apps/esm-login-app/src/login/login.component.tsx
+++ b/packages/apps/esm-login-app/src/login/login.component.tsx
@@ -39,6 +39,12 @@ const Login: React.FC = () => {
   const usernameInputRef = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
+    if (user) {
+      navigate('/home');
+    }
+  }, [user, navigate]);
+
+  useEffect(() => {
     if (!user) {
       if (loginProvider.type === 'oauth2') {
         openmrsNavigate({ to: loginProvider.loginUrl });


Muppasanipraneeth (commented):
@Lemeri123  Thanks for picking this 😊 but I tested It locally but its not working as expected  
- Login to the application it means already  login and 
- Now go back to /openmrs/spa/login
- Since you are already logged in, it should’ve redirected you to /openmrs/spa, but it doesn’t
# screen recording 
https://github.com/user-attachments/assets/86275049-c4f2-4033-9969-62acb775b80b

Muppasanipraneeth:
> @Lemeri123 Thanks for picking this 😊 but I tested It locally but its not working as expected
> 
> * Login to the application it means already  login and
> * Now go back to /openmrs/spa/login
> * Since you are already logged in, it should’ve redirected you to /openmrs/spa, but it doesn’t
> 
> # screen recording
>  Screen.Recording.2025-02-27.at.5.41.43.PM.mov

if user is authenticated should be redirected

Lemeri123:
[Screencast from 2025-02-27 19-00-38.webm](https://github.com/user-attachments/assets/c1447d8a-c78d-4df6-8c5d-964a00643eea)
Here is a screen recording. I just tested it and it works fine

Muppasanipraneeth:
I will check it once again


> [Screencast from 2025-02-27 19-00-38.webm](https://github.com/user-attachments/assets/c1447d8a-c78d-4df6-8c5d-964a00643eea) Here is a screen recording. I just tested it and it works fine

I will check it once again

Muppasanipraneeth:
https://github.com/user-attachments/assets/df118d81-c1a5-42ac-b4c7-673cf7ca045e
ya its working

Muppasanipraneeth:
@denniskigen @vasharma05 Can you check this? It looks fine to me

Lemeri123:
@vasharma05 @denniskigen

shivarm (approved):
LGTM 👌

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -38,6 +38,12 @@ const Login: React.FC = () => {
   const passwordInputRef = useRef<HTMLInputElement>(null);
   const usernameInputRef = useRef<HTMLInputElement>(null);
 
+  useEffect(() => {
+    if (user) {
+      navigate('/home');
```
ipiyumii, line 43: To discuss: Shouldn't the redirect be using the ConfigSchema -> loginSuccess  and use the /home hardcoded path if it doesn't exist. I'm new to this. so I may be wrong suggestion is based on login.component.tsx: line 97

Screenshot 2025-03-24 at 23 16 13
Screenshot 2025-03-24 at 23 19 18

gracepotma:
@ibacher looks like this may have fixed the problem (concern being if someone had bookmarked the /login URL). Pls kindly review :)

ibacher:
This doesn't really solve the issue.

1. From the component we're on, if the user is logged in, then the correct workflow is to redirect to the location picker. This kind of works by redirecting to home which will then call the session endpoint and redirect to the location picker. Much better to just go there.
2. As pointed out in the comment above, redirecting to `/home` is not the right behaviour. However implementing 1 above correctly would also ensure we hit the code path that does this redirection correctly.

PR #1298, opened 2025-02-28
(fix) Allow user to login by selecting a location if default one is invalid

icrc-psousa:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR fixes an issue where the location selection screen would get stuck in a loading state after login if the user had an invalid location configured in their properties. The fix ensures that the default location exists before posting it to server.

## Screenshots

Before fix:
image

### DIFF
--- a/packages/apps/esm-login-app/src/login.resource.ts
+++ b/packages/apps/esm-login-app/src/login.resource.ts
@@ -135,7 +135,7 @@ export function useValidateLocationUuid(userPreferredLocationUuid: string) {
 
   const results = useMemo(
     () => ({
-      isLocationValid: data?.ok,
+      isLocationValid: data?.ok && data?.data.total > 0,
       defaultLocation: data?.data?.entry ?? [],
       error,
       isLoading,


NethmiRodrigo (approved):
Thanks @icrc-psousa! I tested this out, and it fixes the bug.
If anyone wants to recreate this issue, just create a test location, login and set that test location as the preferred location. Then, delete the location and logout, and login again, and you'll be able to see the error.
@vasharma05 @denniskigen @ibacher Two things:
1. The [default login location is set to `data?.data?.entry`](https://github.com/openmrs/openmrs-esm-core/blob/cbd6dc4ae66b6778ff6de57538e9225625e36863/packages/apps/esm-login-app/src/login.resource.ts#L139), so for `isLocationValid` shouldn't we also ensure that this is present?
2. If a login location is invalid, shouldn't we remove it from the session or [userProperties](https://github.com/openmrs/openmrs-esm-core/blob/cbd6dc4ae66b6778ff6de57538e9225625e36863/packages/apps/esm-login-app/src/location-picker/location-picker.resource.ts#L20)? Currently, this fix works without explicitly doing that because the user is forced to select a new location anyway so I guess that updates everything.

denniskigen (approved):
LGTM. Thanks, @icrc-psousa.

samuelmale (approved):

PR #1303, opened 2025-03-06
(feat) O3-4509: Should be able to configure preferred datepicker locale for each language locale

brandones:
# Requirements

## For changes to apps

## If applicable

## Summary

This allows implementers to override the datepicker date format for any locale using a styleguide config element.

## Screenshots

https://github.com/user-attachments/assets/56a7b777-5beb-40fb-808e-718863c2084c

I don't know what the single-spa error is, but it's not relevant to this PR.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4509

## Other

This PR contains some refactoring that is not strictly necessary. This is because I thought I was also going to implement this for `formatDate`. I was like, "You could configure it to show Chinese dates to English users! How fun!" But here's the thing. We have a lot of wacky date format overriding specific to British-style dates. The wide format should always be `DD — MMM — YYYY` in English. Also the `en-u-ca-ethiopic` locale is supposed to display like `30-Hedar-2014` for standard mode, which turns out to be _very touchy_. So I gave up on adding `formatDate` support for this config option. No one has asked for it anyway.

I am also finally deleting the deprecated date/time format functions. I have checked that they are not in use in any public repository (aside from archived ones and personal forks).


packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -3,7 +3,6 @@ import { act, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {
   openmrsFetch,
-  useConfig,
```
brandones, line 6: misc. cleanup

packages/framework/esm-styleguide/src/config-schema.ts
```diff
@@ -48,37 +48,15 @@ export const esmStyleGuideSchema = {
     _description:
       "Used to look up the patient photo, which is stored as an attachment obs. Set to `null` in order to disable the feature and use only generated avatars. To remove the avatars entirely, use extension configuration's `remove` feature.",
   },
-  preferredCalendar: {
```
brandones, line 51: This was cruft, no longer used.

packages/framework/esm-utils/src/get-locale.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Returns the current locale of the application.
+ * @returns string
+ */
+export function getLocale() {
+  let language = window.i18next.language;
+  language = language.replace('_', '-'); // just in case
+  // Hack for `ht` until all browsers update their unicode support with ht to fr mapping.
+  // See https://unicode-org.atlassian.net/browse/CLDR-14956
```
brandones, line 9: Unicode has updated the standard. It's probably in most browsers at this point, but Jest still doesn't follow it.

brandones:
I'm not sure the E2E failure is related?

packages/framework/esm-styleguide/tsconfig.json
```diff
@@ -21,5 +21,5 @@
       "esnext"
     ]
   },
-  "include": ["src/**/*"]
+  "include": ["src/**/*", "../esm-utils/src/dates"]
```
ibacher, line 24: This seems off to me?

packages/framework/esm-styleguide/src/config-schema.ts
ibacher: Uh... the `preferredCalendar` is not cruft and really should be left in! That's how we control which calendar we use for cases where we _don't_ want to use the Gregorian calendar, as is the case in Ethiopia!

We could maybe support the same thing together by having, e.g., `am` map to `am-u-ca-ethiopic` but we cannot remove this feature.

packages/framework/esm-styleguide/src/datepicker/index.tsx
```diff
@@ -447,14 +448,16 @@ export const OpenmrsDatePicker = /*#__PURE__*/ forwardRef<HTMLDivElement, Openmr
       ...datePickerProps
     } = Object.assign({}, defaultProps, props);
 
+    const config = useConfig<StyleguideConfigObject>({ externalModuleName: '@openmrs/esm-styleguide' });
+    const preferredDateLocaleMap = config.preferredDateLocale;
```
ibacher, line 452: So what happens if the user's current locale is not in the `preferredDateLocaleMap`?

ibacher (changes requested):
Thanks @brandones! Unfortunately, as-is, this breaks support for other calendar systems, so this cannot be merged, but otherwise great work.

packages/framework/esm-styleguide/tsconfig.json
brandones: Whoops, IDE did that...

packages/framework/esm-styleguide/src/datepicker/index.tsx
brandones: Then the locale is not changed, everything works the same as it does presently

brandones:
@ibacher , there are  [no references to preferredCalendar in esm-core](https://github.com/search?q=repo%3Aopenmrs%2Fopenmrs-esm-core%20preferredCalendar&type=code). Are you sure this code is still used?

ibacher:
@brandones Huh... I need to dig a bit deeper then.

ibacher (approved):

PR #1311, opened 2025-03-12
(feat) O3-4535 Add widget to display diagnosis tags in styleguide

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR adds a configurable component to display diagnoses as Carbon tags, with the tag color configurable based on the diagnoses' ranks (1 = Primary, 2 = Secondary). While the component will only be used in patient chart visit's table / visits summary cards, I think it's a component that is universal enough for other ESMs to use.

Not in this PR, but I intend to also add a similar component for displaying coded obs value as Carbon tags. The configuration of that component will be a bit more complex, to allow for some form of "configuration registration" where we can define multiple configs for the same component. [here](https://talk.openmrs.org/t/proposal-to-add-more-configurable-components-in-o3-esm-core/45168/3). 

However, I didn't make the `<DiagnosisTag>` component as complex; instead, it's just one config that defines how all diagnoses should be displayed. (I think it makes more sense to have one universal config for diagnoses tag colors. For obs, OTOH, it makes sense to have multiple configuration for different concepts.)

## Screenshots

image

## Other

I had trouble testing this component locally in patient-chart, where I intend to use it. I followed the instructions in the README to to `yarn link`, run `yarn run:shell` and override the module URL, but it was not able to get the component. I "tested" it by including this component in the left nav. I'm not sure if I exported it incorrectly.


packages/framework/esm-styleguide/src/config-schema.ts
```diff
@@ -1,5 +1,47 @@
...
+  'gray',
+  'orange',
+  'green',
+  'warm-gray',
+  'cool-gray',
+  'high-contrast',
+  'outline',
+] as const;
+
+export const diagnosisTagConfigSchema: ConfigSchema = {
```
ibacher, line 20: Could we make the config schema here just:

```json
{
  "primary": "red",
  "secondary": "blue"
}
```

I realize that's not as flexible as the underlying data model, but we've never really used anything other than primary and secondary diagnoses and I'm not sure the "ranks" part is actually well understood.

packages/framework/esm-styleguide/src/config-schema.ts
```diff
@@ -1,5 +1,47 @@
-import { Type, validators } from '@openmrs/esm-config';
+import { type ConfigSchema, Type, validators } from '@openmrs/esm-config';
+import { type DiagnosisTagsConfig } from './diagnoses-tags/diagnosis-tags.resource';
```
ibacher, line 2: I think it's probably preferable to have this type defined here.

packages/framework/esm-styleguide/src/diagnoses-tags/diagnosis-tags.resource.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  return rankToColorMap;
+}
+
+export interface DiagnosisTagsConfig {
+  tagColor: string;
+  appliedToRanks: Array<number>;
+}
+
+export const diagnosisTagConfigSchema: ConfigSchema = {
```
ibacher, line 34: I assume this will go before we merge it in?

packages/framework/esm-styleguide/src/diagnoses-tags/diagnosis-tags.resource.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Return a map of diagnosis rank to the configured color to use to display
+ * the diagnosis tag
+ */
+export function useDiagnosisRankToColorMap() {
+  const { diagnosisTags: config } = useConfig<StyleguideConfigObject>({
+    externalModuleName: '@openmrs/esm-styleguide',
+  });
+  const key = '@openmrs/esm-styleguide/diagnosisTags';
+  const { data: rankToColorMap } = useSWRImmutable(config ? key : null, () => {
```
ibacher, line 16: I strongly dislike this use of SWR as a generic cache. I feel like we can more easily just cache the results in an object or map stored outside of the hook and get a better benefit here.

packages/framework/esm-styleguide/src/utils.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Array of valid values for Carbon tag's "type" attribute.
+ * Options from https://react.carbondesignsystem.com/?path=/docs/components-tag--overview
+ */
+export const carbonTagColors = [
```
ibacher, line 5: This also seems duplicative of what we have in the config schema, but in this case, it makes some sense to keep this version and remove it from the config schema.

ibacher (changes requested):
Thanks @chibongho! I think this is something that makes for a very reasonable component, but there are a few things we should clean-up in this PR before it's merged and I'm hoping we can do something similar for the configuration, since I don't think what's here is going to be very easy for anyone to use.

ibacher:
> Have you done it recently, @ibacher?

`yarn link` doesn't appropriately handle `workspace:*`-style version declarations, but removing those declarations makes the linking process _within the repo_ more challenging and that's what we've optimized for. What I've done where I need to do this is manually shift those versions, i.e., replace `workspace:*` with whatever the version in package.json is.

packages/framework/esm-styleguide/src/index.ts
```diff
@@ -1,27 +1,28 @@
...
 export * from './patient-photo';
-export * from './custom-overflow-menu';
-export * from './icons';
 export * from './pictograms';
+export * from './responsive-wrapper';
+export * from './snackbars';
+export * from './spinner';
+export * from './toasts';
+export * from './workspaces';
+export * from './diagnoses-tags';
```
ibacher, line 25: Shouldn't this be after `./datepicker`?

packages/framework/esm-styleguide/src/diagnoses-tags/diagnosis-tags.component.tsx
```diff
@@ -0,0 +1,30 @@
+import { Tag } from '@carbon/react';
+import { type Diagnosis } from '@openmrs/esm-api';
+import React from 'react';
+import { useDiagnosisRankToColorMap } from './diagnosis-tags.resource';
+
+interface DiagnosisTagsProps {
+  diagnoses: Diagnosis[];
```
ibacher, line 7: Style-wise we've preferred this syntax. They're equivalent, but consistency is important.

```suggestion
  diagnoses: Array<Diagnosis>;
```

packages/framework/esm-styleguide/src/diagnoses-tags/diagnosis-tags.component.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/**
+ * This component takes a list of diagnoses and displays them as
+ * Carbon tags, with colors configured base on the rank (e.g. primary or
+ * secondary) of the diagnoses.
+ */
+export const DiagnosisTags: React.FC<DiagnosisTagsProps> = ({ diagnoses }) => {
```
ibacher, line 15: I think `DiagnosisTags` is the appropriate name for the component. Should we rename the directories to reflect taht?

chibongho:
> Thanks @chibongho! I think this is something that makes for a very reasonable component, but there are a few things we should clean-up in this PR before it's merged and I'm hoping we can do something similar for the configuration, since I don't think what's here is going to be very easy for anyone to use.

Addressed the feedback. This definitely simplifies the config a bit.

My next component is for displaying coded obs values as tags. Config will be more complex, but I'll try to model it similarly.

ibacher (approved):

PR #1313, opened 2025-03-12
(feat) O3-4511 & O3-4512: Fix several date picker issues

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

The main goal of this PR is to solve an issue that's been reported a few times where the OpenmrsDatePicker takes up more horizontal space than the Carbon DatePicker did (style docs be damned!). I'm in no way sure _how_ that worked, but I think this is a slightly better solution (though it may need to be adapted a bit to grow a bit more, we'll see).

Additionally, this partially reverts #1283, adopting a slightly different solution to controlling the behaviour of the Popover, which necessitated extracting another element from react-aria-components so, more code, but now selecting a date once again closes the Popover.

I've also re-added a mock for the OpenmrsDatePicker, since not mocking it seems to result in issues. We'll need to verify this.

Finally, this also turns the `preferredCalendar` configuration option added a while ago into something useful.

## Screenshots

**Before**

datepicker-broken

**After**

datepicker-fixed

## Related Issue

https://issues.openmrs.org/browse/O3-4511
https://issues.openmrs.org/browse/O3-4512
https://issues.openmrs.org/browse/O3-4525

## Other

I realize that eventually I probably need to split the massive datepicker index file into several components, etc.

And yes, this is too many things for one PR. Apologies.

### DIFF
--- a/packages/apps/esm-primary-navigation-app/__mocks__/openmrs-esm-framework.mock.tsx
+++ b/packages/apps/esm-primary-navigation-app/__mocks__/openmrs-esm-framework.mock.tsx
@@ -7,10 +7,6 @@ export function openmrsFetch() {
   return new Promise(() => {});
 }
 
-export function useOnClickOutside() {
-  return React.useRef();
-}
-
 export function getCurrentUser() {
   return of({ authenticated: false });
 }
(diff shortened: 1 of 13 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1316, opened 2025-03-14
(test) Add mock for OpenmrsDatePicker

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

Adds a mock implementation of the `OpenmrsDatePicker` component for testing purposes. This mock implementation:

- Takes `id`, `labelText`, `value`, and `onChange` props
- Renders a simple date input with a labelText
- Formats dates using dayjs
- Handles changes by converting the input value back to a date object and passing it to the `onChange` prop

There's probably more that could be done here, but this seems like a good baseline per previous attempts at stubbing the component.

### DIFF
--- a/packages/framework/esm-framework/mock.tsx
+++ b/packages/framework/esm-framework/mock.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { NEVER } from 'rxjs';
 import type {} from '@openmrs/esm-globals';
 import * as utils from '@openmrs/esm-utils';
-export { OpenmrsDatePicker } from './src';
+import dayjs from 'dayjs';
 
 window.i18next = { ...window.i18next, language: 'en' };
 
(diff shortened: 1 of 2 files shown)


packages/framework/esm-framework/mock.tsx
```diff
@@ -96,6 +96,18 @@ export const launchWorkspaceGroup = jest.fn();
...
+    <label htmlFor={id}>{labelText}</label>
+    <input
+      id={id}
+      type="date"
+      value={value ? dayjs(value).format('DD/MM/YYYY') : ''}
+      onChange={(evt) => onChange?.(dayjs(evt.target.value).toDate())}
+    />
+  </>
+));
+
```
NethmiRodrigo, line 110: Looking at form-engine's mock -https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/b1c70b14a27d10070daa1d8688b611fc3bcd2933/src/components/inputs/date/date.test.tsx#L18-L32, would it make sense to include the `isInvalid` and `invalidText` props too?

NethmiRodrigo (approved):
Thanks @denniskigen!

denniskigen:
Yeah, I thought about that but I wasn't entirely sure whether we have any tests that actually test the invalid or invalid text behavior of the day picker component. If we do I'd be happy to modify the stub in future to accommodate that. I wasn't entirely sure about the DOM structure as well, there's no error related classes being applied to the error span, for example. 

CleanShot 2025-03-14 at 12  28 37@2x

denniskigen:
@ibacher Looks like this needs a bit more tweaking to accommodate the spinbuttons...

NethmiRodrigo:
> I wasn't entirely sure about the DOM structure either - for example, there's no carbon specific error classes getting applied to the error span:
> 
> CleanShot 2025-03-14 at 12  28 37@2x

Sorry, I forgot to clarify this before. I added a custom style for the datepicker's error text because adding the carbon error message name didn't work for some reason - https://github.com/openmrs/openmrs-esm-core/blob/b90f200f06046218bc0dc69a601acdda57b888fb/packages/framework/esm-styleguide/src/datepicker/datepicker.module.scss#L321

PR #1328, opened 2025-03-28
(tests): Add invalid text to datepicker mock and change datepicker mock's role

NethmiRodrigo:
# Requirements

## For changes to apps

## If applicable

## Summary

This PR changes the role of the datepicker mock from `date` to `textbox` because when set to the type `date` browsers automatically think its a literal date input and renders it with a calendar and in tests you won't be able to do `await user.type(dateInput, '2020-05-05');`. We would have to use `fireEvent` instead and do `fireEvent.change(onsetDateInput, { target: { value: '2020-05-05' } });`.
Since this is a mock anyway, it would be simpler to change it into a textbox that we can type into for tests.
This PR also adds the invalid text so that we could (possibly) test out validation errors (for example as mentioned in Chi Bong's PR additional refactoring last point [here](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2240))

## Summary by CodeRabbit

- **New Features**
	- Enhanced the date selection component to improve validation feedback. The component now accepts text input and displays an error message when an invalid value is entered.

### DIFF
--- a/packages/framework/esm-framework/mock.tsx
+++ b/packages/framework/esm-framework/mock.tsx
@@ -96,15 +96,16 @@ export const launchWorkspaceGroup = jest.fn();
 export const navigateAndLaunchWorkspace = jest.fn();
 export const useWorkspaces = jest.fn();
 
-export const OpenmrsDatePicker = jest.fn(({ id, labelText, value, onChange }) => (
+export const OpenmrsDatePicker = jest.fn(({ id, labelText, value, onChange, isInvalid, invalidText }) => (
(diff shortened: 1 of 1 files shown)


chibongho (commented):
Thanks! Seems reasonable, but I'll defer to others.

ibacher (approved):

ibacher:
From experimentation, `type=date` makes it harder to set the value, at least in Jest DOM and since it's a mock anyways, this seemed good enough.

PR #1329, opened 2025-03-31
(fix) Setup redirect for static assets served from wrong location

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

Basically, on some pages (e.g., the chart) its possible to do a hard refresh that results in trying to load the Carbon's Plex fonts relative to the current route rather than the base SPA URL. This adds a rewrite so that these cases serve the right asset.

I couldn't reproduce the issue either on dev3 or using the `develop` command, so I haven't made any additional changes.

## Other

@denniskigen  Could you let me know if you think I've missed a case. This just fixes things with `run:shell` which was the only case I could reproduce.

## Summary by CodeRabbit

- **New Features**
  - Introduced dynamic URL rewriting logic in the build configuration to streamline asset request routing.

- **Refactor**
  - Enhanced internal utility methods with explicit return types for improved consistency and reliability.

### DIFF
--- a/packages/shell/esm-app-shell/webpack.config.js
+++ b/packages/shell/esm-app-shell/webpack.config.js
@@ -93,11 +93,11 @@ function escapeRegExp(string) {
 }
 
 /**
- * @param {*} env
- * @param {*} argv
+ * @param {Record<string, string>} env
+ * @param {Array<string>} argv
  * @returns {import("webpack").Configuration}
  */
-module.exports = (env, argv = {}) => {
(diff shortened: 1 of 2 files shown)


denniskigen (approved):
The `run:shell` case is the only one I'm both aware of and consistently able to reproduce. Thanks for the fix!

ibacher:
Admittedly, I don't understand why the same issue doesn't show up in other cases, but, hey.

PR #1330, opened 2025-03-31
(fix)O3-4516: Design standards: use the appropriate workspace size for workspaces

EDSONZ-WASSWA:
I have made sure that the “Create new appointment“ workspace, the “Add patient to queue“ workspace and the “Admission requests“ workspaces now use the  “normal“ workspace size.
 These widgets now match the correct size as the ticket describes

## Screenshots
before:
Appointment-before
Admission-request-before
Add patient-before
After:
Appointment-after
Admission-request-after
Add-patient-after

Ticket-Link:
[ticket](https://openmrs.atlassian.net/browse/O3-4516l)

@ibacher  please review this

## Summary by CodeRabbit

- **Style**
  - Adjusted the workspace layout in wide mode for a more balanced display, resulting in a noticeably refined appearance.

### DIFF
--- a/packages/framework/esm-styleguide/src/workspaces/container/workspace.module.scss
+++ b/packages/framework/esm-styleguide/src/workspaces/container/workspace.module.scss
@@ -126,7 +126,7 @@ $extraWideWorkspaceWidth: 48.25rem;
     width: $widerWorkspaceWidth;
 
     .workspaceFixedContainer {
-      width: $widerWorkspaceWidth;
+      width: $narrowWorkspaceWidth;
     }
 
     &.hiddenRelative {


denniskigen (changes requested):
Thanks, @EDSONZ-WASSWA for taking a stab at this. So, this is not the correct approach. Each workspace should have it's own workspace registration logic that lives in its source frontend module, typically in a `routes.json` file under the `workspaces` key. For example, the `CreateQueueEntry` workspace is registered [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/routes.json#L131). 

You can see that the registration logic includes a [width](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/routes.json#L135C7-L135C23) property. That `width` is what dynamically determines the width of the workspace based on the classes in this file in Core. So, to change the workspace size there to the `narrow` variant, you'd just need to change that value in the workspace registration. The same idea applies to the other workspaces you've mentioned.

PR #1332, opened 2025-04-04
(test) Fix age check in patient banner test

chibongho:
# Requirements

## For changes to apps

## If applicable

### DIFF
--- a/packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx
+++ b/packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx
@@ -3,6 +3,7 @@ import { type i18n } from 'i18next';
 import { screen, render } from '@testing-library/react';
 import { usePrimaryIdentifierCode } from '@openmrs/esm-react-utils';
 import { PatientBannerPatientInfo } from './patient-banner-patient-info.component';
+import { age } from '@openmrs/esm-framework';
 
 window.i18next = { language: 'en' } as i18n;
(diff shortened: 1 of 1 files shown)


packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx
```diff
@@ -67,7 +68,7 @@ describe('PatientBannerPatientInfo', () => {
 
     expect(screen.getByText(/john wilson/i)).toBeInTheDocument();
     expect(screen.getByText(/male/i)).toBeInTheDocument();
-    expect(screen.getByText(/52 yrs/i)).toBeInTheDocument();
+    expect(screen.getByText(new RegExp(age(mockPatient.birthDate)!, 'i'))).toBeInTheDocument();
```
SarangVT, line 71: Yeah, everything looks correct, as now it calculates age based on current date, not hardcoded, and as age is used in [patient banner](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.component.tsx) from esm-utils, so same [function](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/age-helpers.ts) using here is even safe, with even wrapped with RegExp.
And, I don't think that even handling null is required here as registering with null BirthDate is not allowed in OpenMRS, **but still if there is an entry with null birthDate should we add something for it.**
Thanks!

denniskigen (approved):
Thanks, @chibongho!

PR #1195, opened 2024-11-04
(feat) O3-4160: Patient header redesign

denniskigen:
# Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. Ensure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing#contributing-guidelines) label (such as `feat`, `fix`, or `chore`, among others). See existing PR titles for inspiration.

## For changes to apps

- [x] My work conforms to the [**O3 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://om.rs/o3ui).

## If applicable

- [x] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

## Summary

This PR introduces the first iteration of changes to align the Patient Header with the [new designs](https://zpl.io/5EdQzDW). Specifically, this PR does the following:

- It fixes the avatar to match the [designs](https://zeroheight.com/23a080e38/p/6663f3-patient-header) by: 
	- Adding a loading state	
	- Adding a empty state placeholder for cases where a valid image source exists, but the image can't load for some reason (like when the CSP blocks loading of the patient avatar from a different origin).
	- Customizing the loaded state to match the designs
- Fixes up the gender display to show a corresponding gender icon next to the gender value.
- Fixes up the identifiers display to match the design.
- Changing the `Show / hide details` buttons to read `Show more` and `Show less` to match the designs.

## Screenshots

CleanShot 2024-10-29 at 2  05 36@2x

https://github.com/user-attachments/assets/a85bad25-b649-42ae-8479-8b58579347fb

PR #1334, opened 2025-04-06
(fix) O3-4606 Enable display of Spanish format names

xprl-gjf:
# Requirements

- [X] This PR has a title that briefly describes the work done including the ticket number. Ensure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing#contributing-guidelines) label (such as `feat`, `fix`, or `chore`, among others). See existing PR titles for inspiration.

## For changes to apps

## If applicable

- [X] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.  (**not applicable**)

## Summary

1. When displaying the patient name in the patient banner, patient search bar etc.,  use the formatted display name rather than simply a concatenation of givenName(s) and familyName. In particular, this affects locales with a different name layout, e.g. `layout.name.format=latinamerica`.

2. Allow up to 4 initials in the patient avatar tile. Again, in Spanish-speaking countries it is customary to have 4 names.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4606

### DIFF
--- a/packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx
+++ b/packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx
@@ -65,9 +65,9 @@ describe('PatientBannerPatientInfo', () => {
   it("renders the patient's name, demographics, and identifier details in the banner", () => {
     render(<PatientBannerPatientInfo patient={mockPatient} />);
 
-    expect(screen.getByText(/john wilson/i)).toBeInTheDocument();
+    expect(screen.getByText(/wilson, john/i)).toBeInTheDocument();
     expect(screen.getByText(/male/i)).toBeInTheDocument();
-    expect(screen.getByText(/52 yrs/i)).toBeInTheDocument();
+    expect(screen.getByText(/53 yrs/i)).toBeInTheDocument();
     expect(screen.getByText(/04-Apr-1972/i)).toBeInTheDocument();
     expect(screen.getByText(/openmrs id/i)).toBeInTheDocument();
     expect(screen.getByText(/100gej/i)).toBeInTheDocument();
--- a/packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx
+++ b/packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx
@@ -47,7 +47,7 @@ export function PatientPhoto({ patientUuid, patientName, size }: PatientPhotoPro
     <Avatar
       alt={altText}
       color="rgba(0,0,0,0)"
-      maxInitials={3}
+      maxInitials={4}
       name={patientName}
       size="56"
       src={validImageSrc ?? undefined}
--- a/packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx
+++ b/packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx
@@ -47,7 +47,7 @@ export function PatientPhoto({ patientUuid, patientName, size }: PatientPhotoPro
     <Avatar
       alt={altText}
       color="rgba(0,0,0,0)"
-      maxInitials={4}
+      maxInitials={3}
       name={patientName}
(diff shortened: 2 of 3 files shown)


xprl-gjf:
@denniskigen would you be available to review this?

packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx
```diff
@@ -47,7 +47,7 @@ export function PatientPhoto({ patientUuid, patientName, size }: PatientPhotoPro
     <Avatar
       alt={altText}
       color="rgba(0,0,0,0)"
-      maxInitials={3}
+      maxInitials={4}
```
ibacher, line 50: This feels a little detached from this change otherwise and perhaps something we should make into a configuration setting rather than something hard-coded?

packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx
```diff
@@ -66,7 +66,7 @@ describe('PatientBannerPatientInfo', () => {
   it("renders the patient's name, demographics, and identifier details in the banner", () => {
     render(<PatientBannerPatientInfo patient={mockPatient} />);
 
-    expect(screen.getByText(/john wilson/i)).toBeInTheDocument();
+    expect(screen.getByText(/wilson, john/i)).toBeInTheDocument();
```
denniskigen, line 69: Is this display locale-specific or would all names display with the comma following the surname in the patient header?

ibacher: By default the `text` value returned from FHIR is whatever is returned from `getFullName()`, which used the default name layout template if one exists. So, it's not locale-specific, but it is implementation-specific and controllable. In this case this just reflects [the mock data](https://github.com/openmrs/openmrs-esm-core/blob/b962c360882aade5837eb6a91fc7d0f37e5f0d3f/packages/framework/esm-styleguide/src/patient-banner/patient-info/patient-banner-patient-info.test.tsx#L15). On dev3, with this change I think the display would be "John Wilson".

packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx
xprl-gjf: Yes, but neither did we provide a value for this until [packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx](https://github.com/openmrs/openmrs-esm-core/pull/1195). Were there problems with the avatar display prior to this?

denniskigen: Ah, I'd forgotten about that. Thanks for providing that context, @xprl-gjf. So the decision was made [here](https://openmrs.slack.com/archives/CKS32D55G/p1730111656091649?thread_ts=1730111565.689469&cid=CKS32D55G). Here's a snippet of the conversation from the #ux-design-advisory channel on Slack:

> – What should the maximum number of initials to render in the GeoPattern be? There’s no maximum set at the moment, so if a patient is named Test Kurosaki Kun Ichigo, it will render TKKI. Should we limit it to a certain number (it looks like 2 is a reasonable default). (edited)

And the response from the design team:

> I think having a 1 character minimum and a 3 character limit would suffice? The full name is written next to the avatar, so it’s only intended to support making things ‘glanceable’ in the absence of a photo.

I think the problem was that if you assigned a patient say 6 names, you'd see 6 initials. So we needed to cap it to some reasonable number. And per the design choices, 3 seemed like a sensible compromise at the time.

xprl-gjf: I'll drop this change for now and look into making this configurable in a separate PR.

PR #1336, opened 2025-04-07
(test) Add explicit login step to navbar e2e test

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

Ensure navbar e2e tests have proper authentication before accessing the home page. This change adds the necessary login flow that was previously missing, which should resolve authentication-related test failures.

### DIFF
--- a/e2e/specs/navbar.spec.ts
+++ b/e2e/specs/navbar.spec.ts
@@ -1,17 +1,28 @@
 import { test } from '../core';
 import { expect } from '@playwright/test';
-import { HomePage } from '../pages';
+import { HomePage, LoginPage } from '../pages';
 
-test('action buttons in top nav', async ({ page }) => {
+test('View action buttons in the navbar', async ({ page }) => {
+  const loginPage = new LoginPage(page);
   const homePage = new HomePage(page);
   const topNav = page.getByRole('banner', { name: 'OpenMRS' });
+
+  await test.step('When I log in as admin', async () => {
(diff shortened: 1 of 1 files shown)


denniskigen:
I'm not sure whether this is an actual fix, or whether the test is just flaky and fails intermittently. I was able to replicate the test failure locally, so I think I'm on the right track with this solution. Thoughts, @jayasanka-sack, @chibongho?

denniskigen:
The CI failures are owing to the check-documentation workflow failing. That's to be expected because we're facing some infra issues.

chibongho:
The [`global-setup.ts`](https://github.com/openmrs/openmrs-esm-core/blob/97b0b50836985d2f25d01265b211458d1dc56a28/e2e/core/global-setup.ts) file should have logged in the test user already. The `logout.spec.ts` file however, explicitly logs out the user as part of its test. The test ordering matters too, but I also tried to make the logout test log back it in its [`afterEach` function](https://github.com/openmrs/openmrs-esm-core/blob/e474822a3b7d50187ba16b1f32b341cc5c44dd49/e2e/specs/logout.spec.ts), but maybe that's brittle? 

Maybe it would be more robust to just explicitly call `globalSetup()` in a `beforeEach()` in each `*.spec.ts` file? That said, doing it via UI actions work too, if slightly slower.

denniskigen:
I'm going ahead with merging this to fix the current stituation.

PR #1344, opened 2025-04-19
(feat) add useOnVisible() hook to call function when component is visible

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

Refactored the code for triggering infinite load from the `LocationPicker` into its own hook.


packages/framework/esm-react-utils/src/useOnVisible.ts
```diff
@@ -0,0 +1,39 @@
...
+          threshold: 1,
+        },
+      );
+      if (node) observer.current.observe(node);
+    },
+    [callBack],
+  );
+
+  return ref;
+}
```
ibacher, line 39: Does this clear the IntersectionObserver when the containing component unmounts? Normally, I see `IntersectionObserver`s created in a `useEffect()` hook.

chibongho: Good catch. This was copied from the LocationPicker, but I think the warp app has an implementation that does `unobserve` on unmount.

ibacher (approved):

PR #1345, opened 2025-04-21
(feat) O3-4517: Implement a range variant of the OpenmrsDatePicker

jwnasambu:
# Requirements

## For changes to apps

## If applicable

## Summary
 Implemented a DateRangePicker component that allows users to select a range of dates in a single widget using react-aria's DateRangePicker as the base component to mimic the look and feel of Carbon’s “range” variant as closely as possible. With this implementation, it does not need to be split into two separate components like Carbon’s implementation.

## Screenshots

https://github.com/user-attachments/assets/f721b0aa-7c29-420d-992b-553774f93dbb

## Related Issue
https://openmrs.atlassian.net/browse/O3-4517


packages/framework/esm-styleguide/src/datepicker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,73 @@
...
+  label = 'Trip dates',
+  className = '',
+  minValue,
+  maxValue,
+  value,
+  onChange,
+}) => {
+  return (
+    <DateRangePicker
+      className={`react-aria-DateRangePicker ${className}`}
```
ibacher, line 38: It's not important, I think, to retain the react-aria classes. It is, however, important to maintain the Carbon styling. As much as possible, that means we should include relevant Carbon classes for the component.

ibacher: Also, use the `classnames` library to combine multiple classes. It's much easier to work with.

packages/framework/esm-styleguide/src/datepicker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,73 @@
...
+  RangeCalendar,
+} from 'react-aria-components';
+
+interface DateRangePickerFieldProps {
+  label?: string;
+  className?: string;
+  minValue?: CalendarDate;
+  maxValue?: CalendarDate;
+  value?: DateRange;
+  onChange?: (range: DateRange) => void;
```
ibacher, line 25: Like with the DatePicker itself, we want the standard `onChange()` hook and all the other types to not depend on the `@internationalized/date` types like `CalendarDate` and `DateRange`. Ideally, we can just return an array of `[start, stop]` points similar to what Carbon does.

ibacher: We may need to refactor the existing `datepicker` to extract out common components and functions, since presumably we want to reuse as much of that as possible.

ibacher: We'd also want to support a fuller range of properties here. Ideally, as many of the properties the Carbon DatePicker has as we possibly can.

packages/framework/esm-styleguide/src/datepicker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,73 @@
...
+      className={`react-aria-DateRangePicker ${className}`}
+      value={value}
+      onChange={onChange}
+      minValue={minValue}
+      maxValue={maxValue}
+    >
+      <Label>{label}</Label>
+
+      <Group className="react-aria-Group">
+        <DateInput slot="start">{(segment) => <DateSegment segment={segment} />}</DateInput>
```
ibacher, line 47: These probably need to use some of the internal components from the DatePicker in order to appear correctly with the correct stylings.

packages/framework/esm-styleguide/src/datepicker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+        <span aria-hidden="true" className="mx-1">
```
ibacher, line 49: I'm not sure the `mx-1` class means anything in this context?

ibacher (changes requested):
Remember, as much as possible, the goal here is to wind up with something that looks like it was rendered by Carbon, even though it's using react-aria in the background.

I'm also confused why we have two different component files here?

packages/framework/esm-styleguide/src/datepicker/date-range-picker.component.tsx
jwnasambu: @ibacher, thanks for the review. Kindly feel free to review my changes, please!

jwnasambu:
@ibacher, @samuelmale thanks for the guidance. Kindly feel free to review my PR at your convinient time please!

gitcliff:
helo @jwnasambu kindly look at the CI build failure here ..
Thanks

PR #1346, opened 2025-04-22
(feat) Re-work Framework packages for lighter-weight builds

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This is a mostly non-breaking reworking of the framework with the goal of making builds of esm-core faster and more reliable. To that end:

1. All of the esm-framework _internal_ modules are converted to pure ESM libraries, except esm-styleguide and esm-framework itself.
2. Everything except esm-framework and esm-styleguide in the framework has been migrated to using just SWC to transcompile from TS -> JS and TSC to generate typings.
3. All parts of the framework now share a standardized Typescript configuration
4. All tests in the framework have been migrated from Jest to Vitest, a substantially faster, but more or less API-compatible library. As part of this migration, most tests in the framework use happy-dom instead of jsdom, which is lighter-weight.
5. I've updated all the framework components to use the `exports` setup from Node 12+. This is mostly useful for ensuring that we have a well-defined public API. However, this is _potentially_ breaking if I've missed something we need to import. In that case, we should just fix esm-core.
6. Over time, we created a circular dependency between esm-api and esm-offline since for some parts of the API like getting visits and patients, we need to check the offline database, but esm-offline also relies on `openmrsFetch` and other primitives in esm-api. In this PR, I split out the "higher-level" stuff (most types, visits, patients, locations) into esm-emr-api and confining esm-api to just `openmrsFetch` and friends and the login concepts since we need those too. Unfortunately, this split is a little rougher than I'd like as the types have some circular dependencies, but it does work.
7. This does actually bump Carbon to 3.8.1, as that has some fixes I contributed to make things work with Rspack.

On my machine, this takes the time it takes to run `yarn turbo build` with no cache down to just under 1 minute. In the future, I'll aim to move Rspack into more places, which should improve that.

## Other

This began as an attempt to migrate _apps_ to Rspack, which is still a goal, ~but which is not included in this PR as its large enough already. Note, though, that as a result, this _does_ include an `@openmrs/rspack-config` module which does appear to successfully work, but currently does not integrate with the CLI tooling (so `yarn start` will not work in packages that depend on it).~ I've updated the tooling to work with Rspack. This PR currently migrates the login-app to Rspack as a demo

For the most part, migrating from Jest to Vitest is pretty trivial: just add `vitest` and `happy-dom` as a dependency, add a vitest.config.ts file like this:

```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
  test: {
    environment: 'happy-dom',
    mockReset: true,
    setupFiles: ['./setup-tests.ts'],
  },
});
```

In the `setup-tests.ts` file ensure you have this:

```ts
import { afterEach, vi } from 'vitest';
import { cleanup } from '@testing-library/react';
import '@testing-library/jest-dom/vitest';

afterEach(cleanup);
afterEach(vi.resetAllMocks);
```

And in the test files add:

```ts
import { describe, expect, it, vi } from 'vitest';
import '@testing-library/jest-dom/vitest'; // this replaces import '@testing-library/jest-dom`
```

I'm not certain about this, but these sets of changes have appeared to resolve the issue where rebuilding esm-core repeatedly continually slowed down until it became completely broken. I never worked out what caused or might have caused that bug, so it's hard to be confident it's resolved, but I have run `yarn turbo build --force` close to 100 times while working on this with no noticeable degradation of performance so 🤞.

PS I realize this is probably not a PR anyone can really review in any level of detail, but I'd appreciate a quick check that I haven't screwed up something obvious.


packages/apps/esm-devtools-app/package.json
```diff
@@ -10,8 +10,8 @@
     "start": "openmrs develop",
     "serve": "webpack serve --mode=development",
     "debug": "npm run serve",
-    "test": "cross-env TZ=UTC jest --config jest.config.js --verbose false --passWithNoTests --color",
-    "test:watch": "cross-env TZ=UTC jest --watch --config jest.config.js --color",
+    "test": "cross-env TZ=UTC vitest run --passWithNoTests",
+    "test:watch": "cross-env TZ=UTC vitest watch",
```
ibacher, line 14: Oh... I guess I moved devtools and the help menu to vitest. The biggest impetus for migrating to Vitest is just that it natively supports ESMs and Jest still does not.

packages/apps/esm-implementer-tools-app/jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   clearMocks: true,
   transform: {
-    '^.+\\.tsx?$': ['@swc/jest'],
+    '^.+\\.m?[jt]sx?$': ['@swc/jest'],
```
ibacher, line 4: This is a change other apps may need to implement. Since we now rely on JS files from the framework rather than TS file. Hopefully not needing to transpile between them speeds things up a bit.

packages/apps/esm-implementer-tools-app/jest.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     '^lodash-es$': 'lodash',
     '^lodash-es/(.*)$': 'lodash/$1',
     '\\.(s?css)$': 'identity-obj-proxy',
-    '@openmrs/esm-framework': '@openmrs/esm-framework/mock.tsx',
+    '@openmrs/esm-framework': '@openmrs/esm-framework/mock',
```
ibacher, line 14: This change may also be needed other places as the export is now `@openmrs/esm-framework/mock`.

packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/user-menu-panel.component.tsx
```diff
@@ -14,22 +14,23 @@ interface UserMenuPanelProps extends HeaderPanelProps {
...
-      expanded={expanded}
-      aria-label={t('userMenu', 'User menu')}
-    >
-      <Switcher className={styles.userPanelSwitcher} aria-label={t('userMenuOptions', 'User menu options')}>
-        <ExtensionSlot className={styles.fullWidth} name="user-panel-slot" />
-        <SwitcherDivider className={styles.divider} aria-hidden="true" />
-        <ExtensionSlot className={styles.fullWidth} name="user-panel-bottom-slot" />
-      </Switcher>
-    </HeaderPanel>
+    <div ref={userMenuRef} style={{ display: 'inline' }}>
```
ibacher, line 21: This is a result of now using typings from Carbon. `HeaderPanel` doesn't actually take a `ref`, so I just wrapped it in a `div`. I don't think this affects the display.

packages/framework/esm-api/mock-jest.ts
ibacher, line 1: This type of file is one of the more brittle things here. Internally in the framework, the mocks need to be imported by Vitest, but for most apps, they need to work via Jest. This is accomplished by having `@openmrs/esm-api/mock` resolve to `mock.ts` when called as an ESM and `mock-jest.ts` when called using `require()`.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -216,7 +216,11 @@ export default (env: Record<string, string>, argv: Record<string, string> = {})
               cssLoader,
               {
                 loader: require.resolve('sass-loader'),
-                options: { sassOptions: { quietDeps: true } },
+                options: {
+                  api: 'modern-compiler',
+                  implementation: require.resolve('sass-embedded'),
+                  sassOptions: { quietDeps: true },
+                },
```
ibacher, line 223: The one change to our Webpack configurations here is what the Sass team currently recommends. This may mean that sass-embedded needs to be added to package dependencies, but it should be imported with the CLI.

ibacher:
Alright the e2e failures are due to the Patient Identifier now having a tooltip with the same text, so they fail because the matcher matches two elements.

I currently can't reproduce the failure in the `build`, but I'm looking into it.

ibacher:
> Should we make the default concurrency less aggressive?

That's probably sensible if it causes issues. Does it work better if you use a percentage value like '100%' or '50%'?

> This PR is really big. Are there any particular files we should pay attention to? Anything else we should try locally?

I think I highlighted most of the things I'm a little concerned about in comments on the PR. 

The big thing to try locally is:

1. `yarn turbo build` works (need to run `yarn install` before)
2. `yarn run:shell` works and correctly brings up the app shell
3. `yarn run:omrs --sources packages/apps/esm-login-app` (or hopefully any other) still works

> Do we plan to have this as part of the 3.4 release?

Dennis has convinced me that's a bad idea.

brandones:
> On my machine, this takes the time it takes to run yarn turbo build with no cache down to just under 1 minute

.github/workflows/ci.yml
```diff
@@ -46,12 +46,12 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           server-token: ${{ env.TURBO_TOKEN }}
 
-      - name: 🔍 Run lint, tests and typecheck code
-        run: yarn run verify
-
       - name: 🏗️ Run build
         run: yarn turbo run build --color --concurrency=5
 
+      - name: 🔍 Run lint, tests and typecheck code
```
NethmiRodrigo, line 52: Shouldn't this step run before the build step?

ibacher: Sorry, I meant to reply to this as soon as you wrote it, but if fell off my radar. The short answer is: this is very intentional, but we probably won't be adopting it anywhere.

The longer answer is that our existing setup forces Typescript to parse the TS source files for the entire framework every time it needs to check types. With this PR, we now rely on generated `.d.ts` files instead, but that means we need the build step (where we generate those files) to come before we do any type-checking.

The goal of switching to generating the `.d.ts` files and using that is to allow TSC to get away with doing less work, since the `.d.ts` _only_ have type definitions with none of that pesky executable code.

We probably won't be adopting it elsewhere because usually doing the type-checks before the builds makes sense and other repos won't have the same need to generate types (there are fewer libraries outside of esm-core, so the performance there is less of an issue).

samuelmale (commented):
So I've tested this out and I'm impressed with how fast the build takes! However, the behaviour seems to vary across Node versions. 

**Node v22.11.0 (LTS):**

- yarn turbo build --force --concurrency 3; -> Takes ~ 1m3.189s
- yarn turbo run test --force; -> Tests pass consistently; Takes ~ 23.534s

**Node v18.18.2** (Presently, we [use Node 18](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/main/frontend/Dockerfile#L5C1-L6C1) within the distro):

- yarn turbo build --force --concurrency 3; -> Takes ~ 2m54.3s (much slower)
- yarn turbo run test --force -> Fails ([log](https://pastebin.com/NuH6sTME))

--- 
@ibacher did you test this out on Node 18.x?

ibacher:
> did you test this out on Node 18.x?

Not specifically, because 18 is EOL. However, that just seems to be a failure on a test that's known to be flaky. Try re-running it.

packages/framework/esm-emr-api/mock.ts
```diff
@@ -0,0 +1,33 @@
+import { of } from 'rxjs';
+import { type SessionStore } from '@openmrs/esm-api';
+import { createGlobalStore } from '@openmrs/esm-state/mock';
+
+export const setSessionLocation = jest.fn(() => Promise.resolve());
```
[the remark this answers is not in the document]
ibacher: Good catch!

PR #1350, opened 2025-04-30
(feat) Update TypeDoc and TypeScript

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This work updates TypeDoc and Typescript to their latest versions, including a new TypeDoc plugin to replace the previous custom plugin we were using.

Note that output between these two versions are not identical due to changes to TypeDoc. Hopefuly, we can figure out how to adjust anything that isn't convenient.

## Other

I'm a little concerned that certain `@internal` methods might be appearing in the documentation.

This is built on top of #1346 and relies on that being merged first.


denniskigen:
Can we fix this up, @ibacher?

ibacher:
Note: Despite the huge number of line changes, most of those are just due to how Typedoc now generates documentation, as most functions, etc. are now split into separate files. We can always revisit how API docs are generated.

ibacher:
Also, this doesn't appear to be ready yet. Sorry.

PR #1356, opened 2025-05-08
(fix) Add extra guard-rails around `displayedValidationMessages`

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This fixes an obscure and what should be semantically impossible corner case.

The previous implementation of this code was this:

```ts
const displayedValidationMessages = new Set<string>();

function logError(keyPath: string, message: string) {
  const key = `${keyPath}:::${message}`;
  if (!displayedValidationMessages.has(key)) {
    console.error(message);
    displayedValidationMessages.add(key);
  }
}
```

Which transpiles into this (beautified):

```js
var Ie = new Set;
function Re(e, t) {
    var n = "".concat(e, ":::").concat(t);
    Ie.has(n) || (console.error(t),
    Ie.add(n))
}
```

These two snippets are _almost_ semantically identical except that `var` declarations are treated slightly differently than `const` declarations (see [this MDN section](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#variable_hoisting)). Specifically, `var`s are able to be referenced within their scope before the line defining them has been defined.

What seems to be happening here is that the function `logError()` gets called [in](https://github.com/openmrs/openmrs-esm-core/blob/c0100b7d658d363c5be475f683952bf299971254/packages/framework/esm-config/src/module-config/module-config.ts#L458) [a](https://github.com/openmrs/openmrs-esm-core/blob/c0100b7d658d363c5be475f683952bf299971254/packages/framework/esm-config/src/module-config/module-config.ts#L615) [few](https://github.com/openmrs/openmrs-esm-core/blob/c0100b7d658d363c5be475f683952bf299971254/packages/framework/esm-config/src/module-config/module-config.ts#L735) places _before_ the line assigning a value to the `Ie` variable, resulting in a `Cannot find property "has" of type undefined` style error message, which isn't caught or handled resulting in an app crash.

This PR does two things to attempt to make this impossible:

1. The definition of `displayedValidationMessages` is moved to the top of the file, so it should be executed before any functions in this scope are defined, meaning that it should always have a value when the `logError()` function is called, which is how the code-as-written would behave.
2. Adds a couple of guards to try to provide a value for `displayedValidationMessages` if `logError()` somehow gets called before `displayedValidationMessages` is defined.

### DIFF
--- a/packages/framework/esm-config/src/module-config/module-config.ts
+++ b/packages/framework/esm-config/src/module-config/module-config.ts
@@ -74,6 +74,9 @@ temporaryConfigStore.subscribe((tempConfigState) => {
   computeExtensionConfigs(configInternalStore.getState(), configExtensionStore.getState(), tempConfigState);
 });
 
+/** Keep track of which validation errors we have displayed. Each one should only be displayed once. */
+let displayedValidationMessages = new Set<string>();
+
 function computeModuleConfig(state: ConfigInternalStore, tempState: TemporaryConfigStore) {
   for (let moduleName of Object.keys(state.schemas)) {
     // At this point the schema could be either just the implicit schema or the actually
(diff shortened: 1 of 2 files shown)


NethmiRodrigo (approved):
Thanks @ibacher!

PR #1360, opened 2025-05-16
(fix) O3-4707 Update visit context when active visit is created or ended

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

Currently, the visit context does not get updated when we create an active visit. This is bad because:
- When a user creates a new active visit and tries to place an order, the order should be associated with the newly created visit, but is associated with whichever visit is in context instead.
- When the visit context is set to the active visit, and the user ends the visit. The user should not be able to place an order, but is able to do so with the visit in context.

The above behavior is especially confusing with RDE feature flag off, because the order basket’s visit context is hidden from the user.

To fix this:
- When an active visit is created, if there is no visit in visit context, we set the visit context to the active visit
- When the visit in context was an active visit, but it’s just been made inactive, we clear the visit context (make it null)

## Screenshots

Video. Note that:
- the "start visit" modal is prompted when there is no active visit
- The placed order is correctly associated with the active visit

https://github.com/user-attachments/assets/eec04b84-4279-4657-8d8e-5769104fa5eb

## Related Issue

https://openmrs.atlassian.net/browse/O3-4707

### DIFF
--- a/packages/framework/esm-framework/docs/API.md
+++ b/packages/framework/esm-framework/docs/API.md
@@ -6597,7 +6597,7 @@ ___
 
 ### useVisitContextStore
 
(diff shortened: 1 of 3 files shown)


donaldkibet (approved):

makombe (approved):
Thanks @chibongho for the fix, we encountered the same when testing RDE feature. :+1:

PR #1380, opened 2025-06-03
(feat) Allow left nav to preserve correct component context

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

There's a regression, accidentally introduced by [this commit](https://github.com/openmrs/openmrs-esm-core/commit/47e2a1def83eb3ece030156e27c009166c67f019), which tries to switch to using a single left nav. The issue here is that the LeftNav is now always rendered by the primary navigation app, which removes the ability to actually configure the left nav slot.

This PR fixes this by storing the underlying component context for the app containing the LeftNav in the `leftNavStore`, so that that context can be used when rendering the LeftNav.

Because actually fetching the component context is not generally something we require app-authors to do, this PR also introduces a React Hook to make things easy, called `useLeftNav()`. The component that defines a page rendered by the framework can call `useLeftNav()` in place of `setLeftNav()`. This hook takes care of two things:

1. Calling `setLeftNav()` when mounted and calling `unsetLeftNav()` on unmount
2. It will grab the module name from the ComponentContext.

This is a non-breaking change as we do not remove any APIs, but it is highly recommended that all apps switch to using this new hook or it's equivalent.

## Other

In line with the principle that components in the styleguide should be stateless, the internal implementation of the LeftNav and store, etc. are moved in esm-extensions. I'm not sure that this is the ideal place for them, but it seemed better than creating another component library for ~50 lines of code.

### DIFF
--- a/packages/framework/esm-extensions/src/extensions.ts
+++ b/packages/framework/esm-extensions/src/extensions.ts
@@ -24,12 +24,15 @@ import { type FeatureFlagsStore, featureFlagsStore } from '@openmrs/esm-feature-
 import { subscribeConnectivityChanged } from '@openmrs/esm-globals';
 import { isOnline as isOnlineFn } from '@openmrs/esm-utils';
 import { isEqual } from 'lodash-es';
-import { type AssignedExtension, checkStatusFor, type ExtensionSlotState, getExtensionInternalStore } from '.';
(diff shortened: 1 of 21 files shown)


NethmiRodrigo (approved):
Thanks @ibacher!

PR #1393, opened 2025-06-11
(feat) Output real ESM modules from Webpack config

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

One goal for O3 has been to support "real" ES modules—after all, we've named most of our projects `esm-*` as if that's the case. This PR is a small step towards realizing that goal.

This changes the Webpack and Rspack configurations to output using the "module" format instead of the "var" format, meaning that package entry points, at least, are ESMs. I've also patched the dynamic loading code to first try loading a MFE as a ES module and only fall back to the old process for code that cannot be imported this way.

## Other

The medium-term plan is to drop support for module federation and switch to using shared libraries on the import map. Those steps will necessarily be breaking changes.


ibacher:
Actually, no. I called the branch breaking because I thought it would be, but this should actually be backwards-compatible (once I fix whatever is causing the failures there).

packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -198,9 +198,15 @@ module.exports = (env, argv = []) => {
...
            * @param {Response} proxyRes
+           * @param {Request} req
+           * @param {Response} res
            */
-          onProxyRes(proxyRes) {
+          onProxyRes(proxyRes, req) {
             proxyRes.headers && delete proxyRes.headers['content-security-policy'];
+
+            if (req.url.endsWith('importmap.json')) {
+              proxyRes.headers['content-type'] = 'application/importmap+json';
```
ibacher, line 208: Mostly a note to myself, but we'll need to make a similar change in the CLI, the RefApp, and the SPA module

packages/framework/esm-globals/src/types.ts
```diff
@@ -46,6 +46,12 @@ declare global {
      * The build number of the app shell. Set when the app shell is built by webpack.
      */
     spaVersion: string;
+    /**
+     * Globals added by the @single-spa/import-map-injector package.
+     */
+    importMapInjector: {
```
jolyndenning, line 52: This is defined in the types published to npm. Is it possible to reuse those types?

https://app.unpkg.com/@single-spa/import-map-injector@2.0.2/files/types/import-map-injector.d.ts#L11

packages/shell/esm-app-shell/src/index.ejs
```diff
@@ -10,35 +10,35 @@
...
       href="<%= openmrsImportmapUrl %>"
       as="fetch"
       crossorigin="anonymous"
     >
-    <script type="systemjs-importmap" src="<%= openmrsImportmapUrl %>"></script>
+    <script type="injector-importmap" src="<%= openmrsImportmapUrl %>"></script>
 <% } %>
 <% if (openmrsCoreImportmap) { %>
-  <script type="systemjs-importmap"><%= openmrsCoreImportmap %></script>
+    <script type="injector-importmap"><%= openmrsCoreImportmap %></script>
```
jolyndenning, line 27: why spaces added? For indentation because of ejs? Output html will have indentation bc of this?

packages/shell/esm-app-shell/src/index.ts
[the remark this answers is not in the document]
jolyndenning: I think the following would fix the `initializeSpa` problem:

```html
<script>
  window.importMapInjector.initPromise
    .then(() => import('@openmrs/whichever-module-defines-initializeSpa'))
    .then(module => {
      module.initializeSpa({...});
    })
    .catch(err => {
      // some kind of error handling?
    })
</script>
```

packages/shell/esm-app-shell/src/index.ts
```diff
@@ -67,13 +68,16 @@ function initializeSpa(config: SpaConfig) {
...
   setupPaths(config);
   wireSpaPaths();
-  return Promise.resolve(__webpack_init_sharing__('default')).then(async () => {
-    const { configUrls = [], offline = false } = config;
-    window.offlineEnabled = offline;
 
-    const { run } = await import(/* webpackPreload: true */ './run');
-    return run(configUrls);
-  });
+  return window.importMapInjector.initPromise.then(() =>
```
jolyndenning, line 72: How is esm-app-shell built and loaded into the html file? What is the module format of the built version of it?

Is it part of htmlWebpackPlugin's body tags? `<%= htmlWebpackPlugin.tags.bodyTags %>`?

What I would expect is to see a script tag only loading import-map-injector within index.ejs, and then it loads esm-app-shell after `initPromise` resolves. The reason is that native `import()` or `<script type=module>` cannot be used until the import map is installed.

In index.ejs, it looks like sometimes there is an external import map and sometimes not. When there is not an external import map, waiting for `initPromise` is not required, since import-map-injector inserts the native import map synchronously. The optimal solution might account for that, with two code paths for external vs inline import maps.

packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -225,7 +231,7 @@ module.exports = (env, argv = []) => {
       ignored: ['.git', 'test-results'],
     },
     mode,
-    devtool: isProd ? 'hidden-nosources-source-map' : 'eval-source-map',
+    devtool: isProd ? 'hidden-nosources-source-map' : 'source-map',
```
jolyndenning, line 234: Why this change? Since eval-source-map is for dev?

jolyndenning (commented):
Will the `feat()` PR title result in this being applied as a minor rather than major change?

The branch name is `breaking/use-real-esms`

packages/shell/esm-app-shell/src/index.ts
ibacher: > Is it part of htmlWebpackPlugin's body tags?

Yes.

> What I would expect is to see a script tag only loading import-map-injector within index.ejs

For now because I didn't move the app shell entry point to an ESM, it wasn't necessary, but yes.

> In index.ejs, it looks like sometimes there is an external import map and sometimes not.

We could take that into account, but presently for production builds, it's always using an external import map; the embedded import map is currently only used in development mode.

ibacher:
> Will the feat() PR title result in this being applied as a minor rather than major change?

Well, we version things manually, but as-is, this is currently backwards-compatible with existing code.

ibacher:
>  Is there other code outside of preloadImport that relies upon window[slugifiedPackageName]?

That and `importDynamic`. This falls into the "gory details" we try to confine to core. The definition of breaking change I'm working from is whether or not it will require code changes for code downstream of the esm-core, but you are right that this is probably a breaking change anyway.

jwnasambu:
@ibacher Thanks for working on this issue. Could you kindly fix the merge conflicts, please?

PR #1396, opened 2025-06-12
(chore) Downgrade swr to 2.2.5 to try and fix revalidation issues

denniskigen:
# Requirements

## For changes to apps

## If applicable

## Summary

Presently, when running the latest version of the core tooling and framework in Patient Chart, we're running into regressions in the automatic revalidation of data. More specifically, it appears as though mutations for the order endpoint are not firing off the revalidation of the data. To replicate this, you can run the following steps:

1. Check out [this branch](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2492) locally
2. Run `yarn` to install the dependencies
3. Run `yarn start --sources packages/esm-patient-medications-app packages/esm-patient-orders-app` to run Medications and Orders apps simultaneously
4. Navigate to a Patient Chart with existing drug orders (like 100008E)
5. Modify an existing order by changing the value of the `Indication` field
6. Sign and close the order basket
7. Note how despite a success notification, the order is not updated in the active medications table
8. Refresh the page and note how the updated indication finally appears in the active medications table

This is a regression from the previous version of the core tooling and framework, where the automatic revalidation of data was working as expected. In https://github.com/openmrs/openmrs-esm-core/pull/1395, we rolled back to the original approach of federating SWR in both the Webpack and Rspack configurations. That didn't seem to do the trick, however. So, in this PR, I'm downgrading to the original version of SWR (2.2.5) to see if that fixes the issue.

You can follow along with the steps above to see the issue in action. You can also track the e2e workflow runs [here](https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/15594649539?pr=2492).


NethmiRodrigo (approved):

denniskigen:
So, I've tried testing this out locally using import map overrides by running the app shell and swapping in local importmap URLs for both the Medications and Orders apps (which are both running a local copy of `@openmrs/esm-common-lib`). No luck. Despite the `order` request URL being available in the SWR cache:

CleanShot 2025-06-12 at 15 08 16@2x

No revalidation request is getting sent after mutating the drug order:

CleanShot 2025-06-12 at 15 10 58@2x

PR #1407, opened 2025-06-26
(fix) Fix the Dashboard and NavGroup components to work as expected

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

When using the `dashboard` and `nav-group` extensions, we want these extensions to appear to be defined in the app rendering the slot containing either the dashboard or the nav-group containing the dashboard. This is done by capturing the `ComponentContext` of either the `dashboard` or the `nav-group` and, if it's a `nav-group`, passing the moduleName as a prop to the slot.

Secondarily, when mounting components in this way, the per-extension-in-slot configuration was not being properly passed into the rendered extension. I've fixed this by adjusting the internal function `getAssignedExtensionsFromSlotData()` to take into account any configuration for the extension in the extension slot configuration store. This should hopefully be purely additive and not negatively affect anything.

## Other

Note that `nav-group`s are not currently designed to be nested.


packages/framework/esm-config/src/types.ts
```diff
@@ -56,15 +59,6 @@ export interface ExtensionSlotConfigureValueObject {
   [key: string]: object;
 }
 
-export interface ExtensionSlotConfigObject {
```
ibacher, line 59: N.B. ExtensionSlotConfigObject is the same as ExtensionSlotConfig with the `configure` property removed.

tools/i18next-parser.config.js
```diff
@@ -30,11 +30,31 @@ module.exports = {
...
-    mjs: ['JavascriptLexer'],
-    js: ['JavascriptLexer'], // if you're writing jsx inside .js files, change this to JsxLexer
-    ts: ['JavascriptLexer'],
-    jsx: ['JsxLexer'],
-    tsx: ['JsxLexer'],
+    mjs: [{
+      lexer: 'JavascriptLexer',
+      functions: ['t'],
+      namespaceFunctions: [],
+    }],
```
ibacher, line 37: By default, `namespaceFunctions` is set to `['useTranslations', 'withTranslations']`. However, for the sake of the parser (though not at runtime), we always use the namespace "translations". At runtime, we use the name of the module as the namespace. This leads to some situations where we need to change the namespace in the runtime code, but at parse-time we don't want to change the namespace, so dropping these allows that. 

(Note though, that even with this, changing the namespace in the `t` function itself will still break things).

We may eventually want to further customize the parser to deal with this...

denniskigen: Maybe newer versions of the package offer more granular ways for dealing with this.

denniskigen (approved):
Thanks, @ibacher! Let's get folks unblocked.

tools/i18next-parser.config.js
ibacher: Yeah, we're definitely due for an i18next upgrade. That said, in this particular case, I actually want a much blunter tool...

PR #1409, opened 2025-06-26
(fix) O3-4844: Update visit store only when useVisit is called for current patient

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

With ongoing work on RDE, we changed patient chart actions that require a visit (like adding a visit note) to check for the `currentVisit` instead of the `activeVisit`. This means that it's really important that `currentVisit` is updated accordingly on any user action. In particular, we want to preserve this following existing behavior:

- When the user starts a visit V on the patient chart patient, clicking on the visit notes action item should not prompt for a visit.
  - This means that when the user starts an active visit V, if the patient has no `currentVisit,` we need to set it to the newly created active visit
- When the user ends visit V on the patient chart patient, clicking on the visit notes action item should prompt for a visit.
  - This means that when the user ends active visit V, if `currentVisit == V`, we need to set `currentVisit` to null.

Note that there are multiple ways to end / un-end visits in the UI (for example, via the visit history table and via the action menu in the patient banner), so setting the `currentVisit` is implemented reactively in a `useEffect()` hook in `usevisit()`.

The current bug arises because the `useVisit()` function improperly sets the `currentVisit`.

This PR fixes this by **only** modifying the visit store when `useVisit()` is called on the patient currently in the store. This is still a bit iffy, as `useVisit()` should probably not have any side effect to modify the store at all. (`useEffect()` is called in places like patient search, on every single patient in the search result.) Once we consolidate the `useVisit()` calls in patient chart, we should be able to move the `useEffect()` hook to the store instead of here.

## Related Issue
https://openmrs.atlassian.net/browse/O3-4844


vasharma05 (approved):
Thanks @chibongho for the help!

packages/framework/esm-react-utils/src/useVisit.ts
```diff
@@ -77,7 +77,7 @@ export function useVisit(patientUuid: string, representation = defaultVisitCusto
 
   useEffect(() => {
     // if an active visit is created and there is no visit in context, set the context to the active visit
```
denniskigen, line 79:
```suggestion
    // if an active visit is created and either there is no visit in context or we've switched to a different patient, set the context to the active visit
```

denniskigen: Or something along those lines

PR #1414, opened 2025-07-02
(fix): infer currentVisitIsRetrospective from the absence of cu

amosmachora:
…rrentvisit.stopDateTime

# Requirements

## For changes to apps

## If applicable

## Summary
This PR is an attempt to fix a bug with the `currentVisitIsRetrospective` variable. Previously, the variable would have a value of true regardless of whether the current visit was an active visit. In this PR, I check for the presence of a `stopDateTime` value on the `currentVisit` object and infer whether the visit is retrospective.

### DIFF
--- a/packages/framework/esm-react-utils/src/useVisit.ts
+++ b/packages/framework/esm-react-utils/src/useVisit.ts
@@ -109,7 +109,7 @@ export function useVisit(patientUuid: string, representation = defaultVisitCusto
     isValidating: activeIsValidating || retroIsValidating,
     activeVisit,
     currentVisit,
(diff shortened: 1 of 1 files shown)


amosmachora:
@chibongho and anyone else of interest. I have not tested this new implementation on the `esm-core` because I tried running the patient chart against a local copy of the `react-utils` package in the `esm-core`. Everything got too complicated and too hard fast therefore I was not able to really test it.

I, however, have tested this approach on [this PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2531) and it looks fine for the purposes I have in mind at least.

amosmachora:
Closing this as the hook will be rewritten and this logic will eventually be redefined.

PR #1415, opened 2025-07-03
(fix) Ensure that i18next namespace matches state rather than props

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

For some reason, in the `nav-group` extension, the `I18nextProvider` wasn't "seeing" the correct extension context, but the wrapping `ComponentContext` was. This switches the `I18nextProvider` to extract the extension context from the component state like the `ComponentContext` does.

I also made a small update to the `dashboard` component so that if falls back more gracefully to the correct module name.

## Other

I think this is the last bit in the nav-groups / dashboard thing.


NethmiRodrigo (approved):
Thanks @ibacher!

denniskigen (approved):

PR #1416, opened 2025-07-07
(feat) O3-4864 Support including CSS and JS assets in generated HTML …

chibongho:
…file

# Requirements

## For changes to apps

## If applicable

## Summary

This PR provides a way for the `openmrs build` command to copy static CSS or JS assets in the build output and include those files as `<link>` and `<script>` tags in the generated `index.html` file.

Testing done:

- go to openmrs-esm-core/packages/tooling/openmrs, create 2 files: `assets/a.css` and `assets/b.js`, fill them with dummy content.
- run command: `npx openmrs build --asset assets/a.css --asset assets/b.js`
- verify that the output `dist/` directory contains the `assets/a.css` and `assets/b.js` files,
- verify that the `dist/index.html` file includes those assets as `<link>` and `<script>` tags.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4864

### DIFF
--- a/packages/shell/esm-app-shell/webpack.config.js
+++ b/packages/shell/esm-app-shell/webpack.config.js
@@ -4,6 +4,7 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 const WebpackPwaManifest = require('webpack-pwa-manifest');
+const HtmlWebpackTagsPlugin = require('html-webpack-tags-plugin');
 const { InjectManifest } = require('workbox-webpack-plugin');
 const { DefinePlugin, container } = require('webpack');
 const { basename, dirname, resolve } = require('path');
@@ -39,6 +40,10 @@ const openmrsConfigUrls = (process.env.OMRS_CONFIG_URLS || '')
   .filter((url) => url.length > 0)
   .map((url) => JSON.stringify(url))
   .join(', ');
+const openmrsHtmlCssAssets = (process.env.OMRS_HTML_CSS_ASSETS || '')
+  .split(';')
+  .filter((filePath) => filePath.length > 0);
+
 const openmrsCleanBeforeBuild =
   (() => {
     try {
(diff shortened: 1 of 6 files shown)


packages/tooling/openmrs/src/cli.ts
```diff
@@ -300,12 +300,19 @@ yargs.command(
         default: 'production',
         describe: 'The environment to build for. e.g., development, production.',
         type: 'string',
+      })
+      .option('asset', {
+        default: [],
+        describe:
+          'The path to CSS or HTML assets to copy into the build directory and include in <link> and <script> tags in generated HTML. Can be used multiple times.',
```
ibacher, line 307: This description doesn't seem to accurately describe how this works, right?

packages/tooling/openmrs/src/cli.ts
```diff
@@ -300,12 +300,19 @@ yargs.command(
...
+        type: 'array',
```
ibacher, line 308: I think it's probably safer to include the `coerce` function we use for `build-config` and other properties?

packages/shell/esm-app-shell/webpack.config.js
```diff
@@ -136,6 +141,10 @@ module.exports = (env, argv = []) => {
     });
   }
 
+  openmrsHtmlCssAssets.forEach(asset => {
+    appPatterns.push({from: asset, to: dirname(asset) });
```
ibacher, line 145: I don't love the re-use of `appPatterns` here and I'm not sure why we're doing it?

ibacher: Specifically, this kind of throws-off whether or not we use the `coreImportMap`.

packages/tooling/openmrs/src/cli.ts
chibongho: I can't do `coerce: (arg) => resolve(process.cwd(), arg)` like other args that take in files because I need the relative path of each asset file so I can put them in the same corresponding place inside `dist`. I checked in Ubuntu (bash) and Windows (git bash, powershell) and it seems to do the right thing without it.

chibongho: Yeah, I guess that's fine as long as the assets have unique file names.

ibacher (approved):

PR #1417, opened 2025-07-09
(feat) O3-5206: Migrate to using RSpack everywhere

ibacher:
# Requirements

## For changes to apps

## If applicable

## Summary

This is basically a follow-on to #1346 with the following changes:

* Everything in esm-core (including the app shell) is built with either SWC or RSpack
* Updated SWC version
* Update RSpack version

This PR does have things that are technically breaking changes, but all related to removing functionality that was not widely used, specifically:

* The `debug` command is removed from the CLI. Please use either `openmrs develop` or `webpack/rspack serve`
* The service worker is completely removed in this PR. (The workbox-webpack-plugin is incompatible with RSpack and while there is an [alternative](https://github.com/Clarkkkk/workbox-rspack-plugin), we currently have all the offline tools disabled.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5206

## Other

We are now using ModuleFederation 1.5 everywhere. The issue turned out to be that RSpack is "cleverer" than Webpack's MF setup and de-duplicated `@openmrs/esm-framework` when it was aliased to `@openmrs/esm-framework/src/internal` in such a way that lead to a deadlock. [This](https://github.com/openmrs/openmrs-esm-core/pull/1417/changes#diff-12432dd49111f479d0ca42c726e6bdf8a883645ba1a6680010e332833c46e930R72) is how we work around the issue.


packages/shell/esm-app-shell/rspack.config.js
```diff
@@ -407,23 +404,6 @@ module.exports = (env, argv = []) => {
...
-              { url: openmrsRoutesUrl, revision: null },
-            ],
-          })
-        : new InjectManifest({
-            swSrc: resolve(__dirname, './src/service-worker/noop.ts'),
-            swDest: 'service-worker.js',
-            // this is a no-op service worker, so we don't want to cache anything
-            maximumFileSizeToCacheInBytes: 0,
-            exclude: [/.*/],
-          }),
```
denniskigen, line 426: Should we remove the service worker generation in https://github.com/openmrs/openmrs-esm-core/blob/main/packages/shell/esm-app-shell/src/run.ts#L318 now that we're no longer including the `InjectManifest` plugin and `workbox` dependencies in the Rspack config?

denniskigen (approved):
LGTM. I've posed a couple of questions.

packages/shell/esm-app-shell/rspack.config.js
ibacher: Yes!

denniskigen (approved):

denniskigen:
Set to draft because the e2e test failures are owing to changes in this diff.

PR #1422, opened 2025-07-14
O3-4931: Add ability to evaluate expressions in extensions

CynthiaKamau:
# Requirements

## For changes to apps

## If applicable

## Summary

Add ability to evaluate expressions in extensions

## Screenshots

https://github.com/user-attachments/assets/a7c6dc2d-bac2-4acb-a360-939c872ce387

### DIFF
--- a/packages/framework/esm-framework/src/integration-tests/extension-config.test.tsx
+++ b/packages/framework/esm-framework/src/integration-tests/extension-config.test.tsx
@@ -383,9 +383,11 @@ describe('Interaction between configuration and extension systems', () => {
 
     render(<App />);
 
-    await screen.findByTestId(/slot/);
-    expect(screen.getByTestId('slot').firstChild).toHaveAttribute('data-extension-id', 'Wilma');
-    expect(screen.queryAllByText(/\bSchmoo\b/)).toHaveLength(0);
+    await waitFor(() => {
+      const slot = screen.getByTestId('slot');
+      expect(slot.firstChild).toHaveAttribute('data-extension-id', 'Wilma');
+      expect(screen.queryAllByText(/\bSchmoo\b/)).toHaveLength(0);
+    });
   });
 
   it('should show extension when user has configured privilege', async () => {
@@ -445,28 +447,29 @@ describe('Interaction between configuration and extension systems', () => {
     expect(screen.getByTestId('slot').firstChild).toHaveAttribute('data-extension-id', 'Schmoo');
   });
 
-  // TODO This test fails on CI but not locally
   it('should only show extensions users have default privilege for', async () => {
-    const promise = Promise.resolve();
-    mockSessionStore.setState({
-      loaded: true,
-      session: {
-        authenticated: true,
-        sessionId: '1',
-        user: {
-          uuid: '1',
-          display: 'Non-Admin',
-          username: 'nonadmin',
-          systemId: 'nonadmin',
-          userProperties: {},
-          person: {} as Person,
-          privileges: [{ uuid: '1', display: 'YOWTCH!' }],
-          roles: [],
-          retired: false,
-          locale: 'en',
-          allowedLocales: ['en'],
+    // Set up initial session state before registering extensions
+    await act(async () => {
+      mockSessionStore.setState({
+        loaded: true,
+        session: {
+          authenticated: true,
+          sessionId: '1',
+          user: {
+            uuid: '1',
+            display: 'Non-Admin',
+            username: 'nonadmin',
+            systemId: 'nonadmin',
+            userProperties: {},
+            person: {} as Person,
+            privileges: [{ uuid: '1', display: 'YOWTCH!' }],
+            roles: [],
+            retired: false,
+            locale: 'en',
+            allowedLocales: ['en'],
+          },
         },
-      },
+      });
     });
 
     registerSimpleExtension('Schmoo', 'esm-bedrock', true, 'Yabadabadoo!');
@@ -494,13 +497,13 @@ describe('Interaction between configuration and extension systems', () => {
       disableTranslations: true,
     })(RootComponent);
 
-    await act(async () => await promise);
-
     render(<App />);
 
-    screen.findByTestId(/slot/);
-    expect(screen.getByTestId('slot').firstChild).toHaveAttribute('data-extension-id', 'Wilma');
-    expect(screen.queryAllByText(/\bSchmoo\b/)).toHaveLength(0);
+    await waitFor(() => {
+      const slot = screen.getByTestId('slot');
+      expect(slot.firstChild).toHaveAttribute('data-extension-id', 'Wilma');
+      expect(screen.queryAllByText(/\bSchmoo\b/)).toHaveLength(0);
+    });
   });
 });
 


packages/framework/esm-react-utils/src/useAssignExtensionExpressionContext.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    return () => {
+      // Remove the context keys when component unmounts
+      if (context) {
+        const currentState = extensionExpressionContextStore.getState();
+        const newState = { ...currentState };
+        Object.keys(context).forEach((key) => {
+          delete newState[key];
+        });
+        extensionExpressionContextStore.setState(newState);
```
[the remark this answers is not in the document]
ibacher: And elsewhere...

packages/framework/esm-react-utils/src/useAssignExtensionExpressionContext.ts
ibacher, line 1: So, while I realize this was in my little specification, now that I look at it, I think there's actually a better way to implement extension expression context that will work much better and that's to be enable ExtensionSlot's to pass certain variables to the context via the `state`. So, when we render an `extensionSlot` it has a `state` property that is generally passed down to the extension.

Perhaps what we do is to push this `state` into the extension system itself, e.g., so when we call `useExtensionSlot()`, we also pass in the slot state, which can be associated with the slot [initially in the call to `registerExtensionSlot()`](https://github.com/openmrs/openmrs-esm-core/blob/c001c5169f9b12df5e7c583cb18c46574b55b5d0/packages/framework/esm-react-utils/src/useExtensionSlot.ts#L15) and then, with a second `useEffect()` hook and a new function in `esm-extensions` that updates the `state` property of the slot in `extensionInternalStore` (which doesn't currently exist). That way, we can use the `ExtensionSlot` state directly in expression evaluation. This would also mean that since most extension slots in the patient chart already pass the patient as part of their state, we'd probably have to do very little work to make this mechanism work for the patient-chart (other than ensuring that the `nav-group` correctly passes any state to any `dashboard`s underneath it).

It also means that each extension could evaluate rules against slot-specific data, which should help make this more generally useful.

ibacher (commented):
Thanks @CynthiaKamau! I think this PR as-written does much more than it should.  What we should be adding here is a core building block for functionalities that can be properly leveraged by supplying data at sensible points in the application. Additionally, I've had a re-think about how this should be implemented. See my rather long comment on `useAssignExtensionExpressionContext()`.

ibacher (changes requested):
See my previous comments.

CynthiaKamau:
> Thank you **so much** @CynthiaKamau for working on this!! Sorry you have been waiting for weeks for a review - Ian has been super swamped with some very urgent complex projects. Thank you for your patience and again, thank you SO MUCH for contributing this in follow up to Ian's suggestions [here](https://openmrs.slack.com/archives/C02UNMKFH8V/p1751992722458659?thread_ts=1751548491.921259&cid=C02UNMKFH8V). I made a ticket at https://openmrs.atlassian.net/browse/O3-4931 to reflect/track this work plan so it's not just hanging out in the heap of slack :P
> 
> While we await a review, I have a few questions, please forgive me as I think they're basic, but this will help me understand and communicate this exciting work more widely.
> 
> 1. Is the goal of this to be able to configure things like the Clinical View pages to show or not show depending on patient variables? If I understand the demo screenrecording, you show some extensions (the allergy app in the example) not showing up if the patient is Female. So does this mean that this is how people would configure "If patient = Female, do not show / show the MCH Dashboard", or "If patient = enrolled in Program X, then show the X Program Page in the Left Nav". Do I have that correct?
> 2. Can you point me to the specific place where an implementer would adjust such config/settings? (e.g. where did you adjust things to say "if female, don't show allergy app"?)
> 3. Can this work already be used to apply to navigation and pages, or is it so far only applicable to individual apps/widgets? (e.g. in the screenrecording, the Allergy Page and Left Nav still shows for the female patient, even though the esm doesn't show up.)
> 
> Thank you again!!!! I'm super impressed by your turnaround time in response to Ian's suggestions on Slack 🙏
> 
> CCing @denniskigen because I think he'll want to be aware of this important work, and @NethmiRodrigo because this directly relates to Clinical Views (IIUC) which she's been working on / thinking about a lot lately, and we'll want to use this in our HIV Package support work for DRC.

1. Yes, in OHRI, we hide nav links in the patient chart based on conditions like age and gender for specific programs, eg MCH and PMTCT, you can see the difference in the menus for male and female patients on the patient chart here : https://ethiohri-dev.globalhealthapp.net/openmrs/spa 

2. https://ethiohri-dev.globalhealthapp.net/openmrs/spa - Ethiopia 

3. The goal is do it it in nav links for starters, but it applies to extensions generally

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -357,15 +467,34 @@ function getAssignedExtensionsFromSlotData(
         }
       }
 
-      const displayConditionExpression = extensionConfig?.['Display conditions']?.expression ?? null;
+      const displayConditionExpression =
+        extensionConfig?.['Display expression']?.expression ??
+        (typeof extension.displayExpression === 'string'
+          ? extension.displayExpression
+          : (extension.displayExpression as any)?.expression) ??
```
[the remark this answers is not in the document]
samuelmale: Ideally, `extension.displayExpression` should be a `string` property. We won't have any issues if the [schema](https://github.com/openmrs/openmrs-contrib-json-schemas/blob/36ea64ac4c18a94719ddc380679cca9a6e3cf910/routes.schema.json#L133) is updated.

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -20,7 +20,7 @@ import {
   getExtensionSlotConfigFromStore,
   getExtensionSlotsConfigStore,
 } from '@openmrs/esm-config';
-import { evaluateAsBoolean } from '@openmrs/esm-expression-evaluator';
+import { evaluateAsBoolean, type VariablesMap } from '@openmrs/esm-expression-evaluator';
```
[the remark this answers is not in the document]
CynthiaKamau: removing it

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -357,15 +363,18 @@ function getAssignedExtensionsFromSlotData(
...
-      if (displayConditionExpression !== null) {
+      const displayConditionExpression =
+        extensionConfig?.['Display conditions']?.expression ?? extension.expression ?? null;
+      const slotState = internalState.slots[slotName]?.state ?? null;
+      if (displayConditionExpression && slotState) {
         try {
-          if (!evaluateAsBoolean(displayConditionExpression, { session })) {
+          const context: VariablesMap =
+            slotState && typeof slotState === 'object' ? { session, ...slotState } : { session, slotState };
+          if (!evaluateAsBooleanAsync(displayConditionExpression, context)) {
```
samuelmale, line 373: Hhhmmm, using`evaluateAsBooleanAsync(...)` will always return a Promise, hence this condition will always be `false` unless you resolve the Promise. Can you use `evaluateAsBoolean(...)` instead?

gracepotma:
Thanks for your answers above @CynthiaKamau!
Sorry I didn't follow your answer to this question:
> Can you point me to the specific place where an implementer would adjust such config/settings? (e.g. where did you adjust things to say "if female, don't show allergy app"?)

When I go to https://ethiohri-dev.globalhealthapp.net/openmrs/spa (logged in), I only see the home page. Where in the code or config or admin does someone configure the _logic_ for what is shown/hidden depending on certain patient variables?

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -357,15 +363,18 @@ function getAssignedExtensionsFromSlotData(
         }
       }
 
-      const displayConditionExpression = extensionConfig?.['Display conditions']?.expression ?? null;
-      if (displayConditionExpression !== null) {
+      const displayConditionExpression =
+        extensionConfig?.['Display conditions']?.expression ?? extension.expression ?? null;
```
samuelmale, line 367: @CynthiaKamau So, because you provide an implicit config value of `""` for the `expression` under `"Display conditions"`, this statement will never resolve expressions defined via extension def in the routes file (`extension.expression`). This is because `extensionConfig?.['Display conditions']?.expression` will never be `null` or `undefined`. You should instead:

```suggestion
        extensionConfig?.['Display conditions']?.expression || extension.expression || null;
```

samuelmale: @CynthiaKamau have you tested this out at runtime to confirm that it works extension config? If so, can you show me a sample config?

ibacher (approved):

packages/framework/esm-config/src/module-config/module-config.ts
```diff
@@ -37,42 +37,37 @@ import { type TemporaryConfigStore } from '..';
...
-  computeExtensionConfigs(configInternalStore.getState(), configExtensionStore.getState(), tempConfigState);
-});
+// Store unsubscribe functions to allow cleanup (e.g., in tests or hot module reloading)
+const configSubscriptions: Array<() => void> = [];
+
+/**
+ * Recomputes all configuration derived stores based on current state of input stores.
+ * Called whenever any input store (configInternalStore, temporaryConfigStore, configExtensionStore) changes.
+ */
+function recomputeAllConfigs() {
```
ibacher: This should be roughly equivalent to what was happening before, albeit with less chaining.

PR #1425, opened 2025-07-15
(feat) O3-4888: Use common LifeCycles loader for extensions / modals / workspaces

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

Extensions, workspaces and modals all rely on `getLoader()` function to load their single-spa lifecycles. This PR does the following refactoring:
- There is existing code that allows extensions (and co.) to specify a custom loader, in lieu of specifying `component` in `routes.json`. However, it is very unlikely to be used, as the loader is a function type, which cannot be defined in a JSON file. That code is removed.
- `getLoader()` is changed from a higher-order function to just a regular function to load a `Lifecycles`, given the `moduleName` and `componentName`. It is renamed to `loadParcel`, and is moved to `esm-dynamic-loading` (instead of `esm-routes`) so it can be easily called in other packages. (Previously, we have a create a `load()` function for each extension and save it in the registration store.) 

These changes should make it easier to:
- create a generic react component to render a Parcel by `moduleName` and `componentName`.
- separate workspace code from `esm-extensions` to `esm-workspaces`. 

See further discussion of those 2 points [here](https://github.com/openmrs/openmrs-esm-core/pull/1424#issuecomment-3071332317).

## Related Issue

https://openmrs.atlassian.net/browse/O3-4888


brandones:
I like this, but it will immediately break anything using `load` in the routes file. What's the migration plan?

brandones:
Whoops! You're right, sorry I got mixed up.

chibongho:
Yeah, should have mentioned. I ran `yarn run:shell` and opened a few extensions, modals and workspaces. They all work fine.

chibongho:
I ended up adding back the loader function in for extensions / modals / workspaces registrations. This makes it possible for `esm-styleguide` to directly call `registerModal()` to register a modal. (The "normal" way of declaring a modal in `routes.json` does not work for `esm-styleguide` because it does not have that file.) Note that the code that parses for the `load()` function in `routes.json` remains removed in this PR.

packages/framework/esm-extensions/src/modals.ts
```diff
@@ -1,11 +1,11 @@
 import { createGlobalStore } from '@openmrs/esm-state';
-import type { LifeCycles } from 'single-spa';
 import { getExtensionRegistration } from '.';
+import { type LifeCycles } from 'single-spa';
```
ibacher, line 3: I actually think, per our standards, this should probably be the first import.

packages/framework/esm-extensions/src/render.ts
```diff
@@ -3,6 +3,7 @@ import { mountRootParcel, type Parcel, type ParcelConfig } from 'single-spa';
 import { getExtensionNameFromId, getExtensionRegistration } from './extensions';
 import { checkStatus } from './helpers';
 import { updateInternalExtensionStore } from './store';
+import { loadLifeCycles } from '@openmrs/esm-dynamic-loading';
```
ibacher, line 6: This should probably be immediately after `'single-spa'`.

packages/framework/esm-dynamic-loading/src/load-lifecycles.ts
ibacher, line 1: I'm somewhat hesitant about adding more code to dynamic-loading; I'm hopping that when we shift over to ESMs, we can get rid of that whole setup.

packages/framework/esm-extensions/src/render.ts
ibacher: Also, we don't actually seem to be using this?

packages/framework/esm-extensions/src/workspaces.ts
```diff
@@ -65,6 +66,7 @@ export function registerWorkspace(workspace: RegisterWorkspaceOptions) {
       ...state.workspaces,
       [workspace.name]: {
         ...workspace,
+        load: () => loadLifeCycles(workspace.moduleName, workspace.component),
```
ibacher, line 69: Is there a reason we can't handle this in esm-routes?

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -3,6 +3,7 @@ import { mountRootParcel, type Parcel } from 'single-spa';
 import { createGlobalStore } from '@openmrs/esm-state';
 import { getModalRegistration } from '@openmrs/esm-extensions';
 import { reportError } from '@openmrs/esm-error-handling';
+import { loadLifeCycles } from '@openmrs/esm-dynamic-loading';
```
ibacher, line 6: Also not actually used here right?

ibacher (commented):
High-level, I'm ok with dropping the `load` property. The goal of that property was to allow users to define custom single-spa loader functions. But, realistically, I don't think this is a thing that anyone needs to do. Even if someone were to want to build support for Vue or Svelte or whatever, it's still preferable, I think, to keep to the same standard we're using with React: each MFE ("esm" in our parlance) has a single main entry point that defines named exports that return a single-spa application. Similar to how, for React, we have the `openmrsComponentDecorator`, custom lifecycles could be defined there.

The transformation of `getLoader()` to a standard function is probably fine, although we _may_ want to adopt single-spa's support for multiple loaders at some point (single-spa lifecycle functions are technically processed as an array of functions).

chibongho:
Updated PR. I was under the mistaken impression that I couldn't import `loadLifeCycles` in some packages when it's defined in `esm-routes`. Turns out I can, so I moved it back there.

PR #1429, opened 2025-07-22
(fix) O3-4478: Redirect logged-in users from login page to home page

Willie-theBeastMutua:
# Requirements

## For changes to apps

## If applicable

## Summary
This PR addresses the regression identified in O3-4478, where users who are already authenticated are not redirected to the home page when navigating back to /openmrs/spa/login.
✅ Changes Implemented:
Added a redirect check inside the login route handler to verify if the user is already authenticated.

If the session is valid, the user is automatically redirected to /openmrs/spa.
## Screenshots

https://github.com/user-attachments/assets/b3244cad-c8b2-4224-a304-55a38f317504

## Related Issue
https://openmrs.atlassian.net/browse/O3-4478


adarshiiit0117:
Thanks for looking into this regression. The redirect flow looks solid.

I have a quick question: do we need to consider a cleanup or guard in case the component unmounts while the async redirect logic is running, or is this safe in this route?

Also, would it be useful to add a small test for the “already authenticated user visits /login” scenario, or is this behavior already tested elsewhere?

adarshiiit0117 (commented):
Thanks for looking into this regression. The redirect flow looks solid.

I have a quick question: do we need to consider a cleanup or guard in case the component unmounts while the async redirect logic is running, or is this safe in this route?

Also, would it be useful to add a small test for the “already authenticated user visits /login” scenario, or is this behavior already tested elsewhere?

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -11,6 +11,7 @@ import {
   useConnectivity,
   useSession,
 } from '@openmrs/esm-framework';
+import { getSessionStore } from '@openmrs/esm-api';
```
ibacher: 1. Never import from anything except `@openmrs/esm-framework` in apps, please.
2. `useSession` is already a React-focused wrapper for this.

ibacher (changes requested):

PR #1432, opened 2025-07-28
(fix) Prevent calendar from closing on internal clicks

samuelmale:
# Requirements

## For changes to apps

## If applicable

## Summary

**Issue**: The calendar was closing both when clicking outside **and inside**, due to the `useOnClickOutside` hook being bound to a relative DOM element.

**Solution**: Introduce the `AutoCloseDialog` component that uses the calendar’s **modal container** as the boundary reference for the `useOnClickOutside` hook. This ensures that only true outside clicks will trigger the `state.close()` call.

This issue was making it impossible to capture dates while in the tablet viewport as reported [here](https://openmrs.slack.com/archives/C02UNMKFH8V/p1753436467861419). 

## Screenshots

Fix:

https://github.com/user-attachments/assets/4d5ca689-075e-49f2-9506-e9f26089253f


denniskigen (approved):
I've left some pernickety comments, but otherwise this is a very useful change! Thanks, @samuelmale!

PR #1434, opened 2025-07-29
(BREAKING) O3-4927: Add single spa app root container divs in index.ejs

chibongho:
# Requirements

## For changes to apps

## If applicable

## Summary

Currently, each single-spa page is mounted in a `<div>` directly under `<body>`. This PR changes it so that:
- the primary navigation app (top nav) has a hard-coded app root `<div>` directly under `<body>`.
- all other apps now has app roots under `body > #omrs-single-spa-app-roots`.

Motivation:

The workspace system supports "overlay" mode. If true, workspaces are render on top on the main app page content; if false, workspaces take up space on the right side, and reduces the total width of the main app page content. Currently, this is implemented by having `<WorkspaceContainer>` be rendered as part of the app, so that its can take up space inside the app as needed. However, this implementation is prone to mishandling workspace placement for RtL languages.

This change paves the way for us to have the new workspace system use a workspace root hard-coded inside `index.ejs` instead, and have us handle workspace and left nav placements at the esm-core level instead of inside apps. We can have the layout of our O3 application defined like this:

```
body {
  grid-template-areas:
    'topNav    topNav    topNav'
    'leftNav   appRoots  workspace';
}
```

and have it change accordingly in tablet mode or when switched to RtL languages.

Note:
- We are moving towards hard-coding the top nav to be at the top of the view port. This reduces some flexbility afforded by the system, but I think that is acceptable.
- Right now, this is technically a non-breaking change right now. That said, Implementations with their custom primary navigation app (and named it something different) will see breakage as we implementation the new workspace system and its layout.

## Screenshots

Recording with the changes. No visible difference from before:
https://github.com/user-attachments/assets/4ea9d1ac-1030-46a0-beaf-5610e679a5ce

## Related Issue

https://openmrs.atlassian.net/browse/O3-4927

### DIFF
--- a/packages/framework/esm-globals/src/types.ts
+++ b/packages/framework/esm-globals/src/types.ts
@@ -136,7 +136,7 @@ export type PageDefinition = {
    */
   offline?: boolean;
   /**
-   * If supplied, the page will be rendered within the DOM element with he specified ID. Defaults to "omrs-apps-container" if not supplied.
+   * If supplied, the page will be rendered within the DOM element with the specified ID. Defaults to "omrs-apps-container" if not supplied.
    */
   containerDomId?: string;
 } & (


ibacher:
I kind of worry about the semi-hard-coded nature of this; specifically, it works as long as there's a single thing called `@openmrs/esm-primary-navigation-app` _and_ the first page in its `routes.json` is the wrapper page. Initially, this was the thing that the `order` property for pages was supposed to take care of (so the pages in primary-navigation-app are just set to a really low precendence).

Maybe instead of the hard-coding approach, we add special handling where order is zero where those divs get attached to an, e.g., `omrs-wrapper-page` container, and other pages get wrapped in the `omrs-single-spa-app-roots`. That also gives us some additional flexibility if we have other extensions that need to be mounted "outside" all of the other divs?

chibongho:
Is the "order" property used by anything besides the primary-navigation app / top nav? If not, maybe we can deprecate that, and instead create a field for specifying the DOM element id to insert the app root into? So the primary-nagivation-app's `routes.json` would look like:

```
{
  "pages": [
    {
      "component": "root",
      "routeRegex": "^(?!(?:login|change-password)/?)",
      "online": true,
      "offline": true,
      "mountDomId": "omrs-top-nav-container" // defaults to 'omrs-single-spa-app-roots' if unspecified
    },
  ],
}
```

chibongho:
Update the PR, with an implementation where we hard-code an extra DOM element called `omrs-top-nav-app-container` instead of hard-coding the actual app's root DOM id. I feel better about this.

I also removed the "order" attribute under `pages` in `routes.json`, assuming primary navigation app is the only thing that uses it. Technically a breaking change. I can add it back if that's not the case.

brandones:
Seems reasonable enough to me. @ibacher ?

ibacher:
Yeah, I'm fine with that.

packages/framework/esm-routes/src/loaders/pages.ts
```diff
@@ -195,14 +194,13 @@ export function registerApp(appName: string, routes: OpenmrsAppRoutes) {
  */
 export function finishRegisteringAllApps() {
   pages.sort((a, b) => {
-    let sort = a.order - b.order;
-    if (sort != 0) {
-      return sort;
-    }
```
samuelmale, line 201: Can you update the JS doc comment of this function to communicate the new approach? Specifically, this part "pages are rendered in the DOM according to their order definition, especially because certain pages _must_ be first in the DOM"

brandones:
Ah sorry for the rebase, I should have merged.

brandones (approved):
LGTM

PR #1153, opened 2024-09-18
O3-3866-2

mogoodrich:
# Requirements

- [ ] This PR has a title that briefly describes the work done including the ticket number. Ensure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing#contributing-guidelines) label (such as `feat`, `fix`, or `chore`, among others). See existing PR titles for inspiration.

## For changes to apps

- [ ] My work conforms to the [**O3 Styleguide**](https://om.rs/styleguide) and [**design documentation**](https://om.rs/o3ui).

## If applicable

- [ ] My work includes tests or is validated by existing tests.
- [ ] I have updated the [esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) to reflect any API changes I have made.

PR #1465, opened 2025-09-30
(fix) Add missing workspace translation and fix extraction

denniskigen:
# Requirements

## If applicable

## Summary

This PR makes the following changes:

- Fix `unsavedChangesInOpenedWorkspace` not being extracted by using single quotes instead of backticks. The value uses backticks (`), but the [custom lexer regex](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-translations/custom-i18next-parser-lexer.mjs#L11) only matches single quotes. Keys with backticks or double quotes are skipped during extraction.
- Add missing `unsavedChangesInOpenedWorkspace` key to `en.json`.
- Update `closingAllWorkspacesPromptBody` to match current source text.
- Tangential: Remove `scripts-prepend-node-path=true` from .npmrc to fix a deprecation warning.

## Screenshots

With french translation added by hand for testing:

CleanShot 2025-10-01 at 01 15 45@2x

This screenshot reveals the other bug reported by Bradley where both the discard and cancel buttons incorrectly use the same french translation ("Annuler"). That should be fixed upstream in Transifex.

### DIFF
--- a/packages/framework/esm-translations/src/translations.ts
+++ b/packages/framework/esm-translations/src/translations.ts
@@ -27,7 +27,8 @@ const workspaceTranslations = {
   maximize: 'Maximize',
   minimize: 'Minimize',
   openAnyway: 'Open anyway',
-  unsavedChangesInOpenedWorkspace: `You may have unsaved changes in the opened workspace. Do you want to discard these changes?`,
+  unsavedChangesInOpenedWorkspace:
+    'You may have unsaved changes in the opened workspace. Do you want to discard these changes?',
   unsavedChangesInWorkspace:
     'There may be unsaved changes in "{{workspaceName}}". Please save them before opening another workspace.',
   unsavedChangesTitleText: 'Unsaved changes',
--- a/packages/framework/esm-translations/translations/en.json
+++ b/packages/framework/esm-translations/translations/en.json
@@ -15,7 +15,7 @@
   "Clinic": "Clinic",
   "close": "Close",
   "closeAllOpenedWorkspaces": "Discard changes in {{count}} workspaces",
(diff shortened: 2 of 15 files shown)


packages/framework/esm-translations/translations/en.json
```diff
@@ -15,7 +15,7 @@
   "Clinic": "Clinic",
   "close": "Close",
   "closeAllOpenedWorkspaces": "Discard changes in {{count}} workspaces",
-  "closingAllWorkspacesPromptBody": "There are unsaved changes in the following workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}",
+  "closingAllWorkspacesPromptBody": "There may be unsaved changes in the following workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}",
```
denniskigen, line 18: This effects a translation change that was effected in https://github.com/openmrs/openmrs-esm-core/pull/1153/files#diff-1c03ea6c69f77e98ed8c9eb205bbb1b02b37aa6c3fe6dc0ddf45c1389b024ace but not actually propagated to the translation file because the i18next parser doesn't change default values in place.

packages/framework/esm-translations/src/translations.ts
```diff
@@ -27,7 +27,8 @@ const workspaceTranslations = {
   maximize: 'Maximize',
   minimize: 'Minimize',
   openAnyway: 'Open anyway',
-  unsavedChangesInOpenedWorkspace: `You may have unsaved changes in the opened workspace. Do you want to discard these changes?`,
+  unsavedChangesInOpenedWorkspace:
+    'You may have unsaved changes in the opened workspace. Do you want to discard these changes?',
```
denniskigen, line 31: This is the main change.

ibacher (approved):
Should we also update the regex?

PR #1502, opened 2025-11-23
(fix) O3-4562: Fix extension configuration schema machinery

RajPrakash681:
## Requirements

 This PR has a title that briefly describes the work done including the ticket number. Ensure your PR title includes a conventional commit label (such as `feat`, `fix`, or `chore`, among others). See existing PR titles for inspiration.

## If applicable

 My work includes tests or is validated by existing tests.
 I have updated the esm-framework mock to reflect any API changes I have made.

## Summary

This PR fixes the extension configuration schema machinery so that extensions can properly define and use their own configuration schemas independent of module schemas.

**Problems Solved:**
- `defineExtensionConfigSchema()` wasn't properly registering extension schemas (missing `moduleLoaded` flag)
- `useConfig` hook was always merging module config with extension config, even when extension had its own schema

**Changes:**
- Added `extensionHasOwnConfigSchema()` function to check if an extension has its own config schema
- Fixed `defineExtensionConfigSchema()` to set the `moduleLoaded` flag
- Updated `useConfig` hook to use extension-specific schemas exclusively when they exist
- Updated mock files (`mock.ts`, `mock-jest.ts`) for testing compatibility

**Behavior:**
When an extension defines its own config schema using `defineExtensionConfigSchema()`, the `useConfig` hook now returns **only** the extension's config (not merged with module config). When no extension schema exists, it falls back to merging module + extension configs.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4562

## Other

All 95 existing tests pass. Extension config tests verify the correct exclusive schema usage. No breaking changes to existing functionality.

### DIFF
--- a/packages/framework/esm-config/mock-jest.ts
+++ b/packages/framework/esm-config/mock-jest.ts
@@ -39,10 +39,6 @@ export function defineExtensionConfigSchema(extensionName: string, schema) {
   configSchema = schema;
 }
 
-export const extensionHasOwnConfigSchema = jest.fn((extensionName: string) => {
-  return !!extensionConfigSchemas[extensionName];
-});
-
 export const clearConfigErrors = jest.fn();
 
 /**
--- a/packages/framework/esm-config/mock-jest.ts
+++ b/packages/framework/esm-config/mock-jest.ts
@@ -34,4 +34,6 @@ export function defineExtensionConfigSchema(extensionName, schema) {
   configSchema = schema;
 }
 
+export const extensionHasOwnConfigSchema = jest.fn(() => false);
+
 export const clearConfigErrors = jest.fn();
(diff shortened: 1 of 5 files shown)


ibacher (changes requested):
So thanks for the attempt here. I'm not sure exactly what this is trying to solve, but from the ticket I remember:

1. This appears to be linked to the wrong issue. I don't see how this implementation relates to O3-4562 (which is an epic and shouldn't be a PR)
2. This isn't really solving the problem which is ensuring that the _system_ sees the _schema_ for the configuration of a widget using the extension-specific schema. This needs to be solved in `esm-config`, not `esm-react-utils`.
3. `useConfig()` shouldn't have any specific knowledge of the component's configuration schema. This is entirely a concept internal to `esm-config` and the implementer tools.

PR #1551, opened 2026-01-16
(fix) O3-5350: Add error handling for JSON.parse in navigation history

sanks011:
# Requirements

## If applicable

## Summary

This PR adds defensive error handling for `JSON.parse()` operations in the navigation history module to prevent application crashes when sessionStorage contains corrupted or invalid JSON data.

### Problem
The navigation history code had three unprotected `JSON.parse()` calls in `packages/framework/esm-navigation/src/history/history.ts`. If sessionStorage data becomes corrupted (due to browser extensions, manual editing, browser crashes during write operations, or race conditions), the application would crash with an unhandled `SyntaxError`, completely breaking all navigation functionality.

### Solution
Wrapped all three `JSON.parse()` calls in try-catch blocks with:
- Graceful fallback to empty array `[]` to maintain functionality
- Console warnings to aid debugging when corruption is detected
- Explicit TypeScript type annotations for clarity

### Changes Made
Modified `packages/framework/esm-navigation/src/history/history.ts`:
- `addToHistory()` function - Added try-catch error handling
- `setupHistory()` function - Added try-catch error handling  
- `getHistory()` function - Added try-catch error handling

All three functions now safely handle corrupted sessionStorage data without crashing the application.

### Console Warning on Corrupted Data
image
*The fix logs a warning instead of crashing when sessionStorage contains invalid JSON*

### Comparison: With vs Without Fix
image
*Without fix - crashes with SyntaxError*

## Related Issue

https://issues.openmrs.org/browse/O3-5350

## Other

**Risk Assessment:** Low risk - These changes are purely defensive:
- Only adds error handling around existing code
- No changes to application logic or behavior
- Maintains existing functionality in normal operation
- Prevents catastrophic failures in edge cases
- Follows JavaScript best practices for parsing external data sources

**Testing:** Verified TypeScript compilation passes with no errors. Existing tests remain valid as behavior is unchanged during normal operation.


sanks011:
> Have you come across a circumstance where the application itself is corrupting localStorage?

@ibacher So when I was looking into JSON.parse best practices, I came across the MDN documentation that specifically shows examples of wrapping JSON.parse in try-catch when working with localStorage/sessionStorage (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse, https://app.studyraid.com/en/read/12384/399944/storing-json-in-localstorage, https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API). They mention it can throw a SyntaxError if the data isn't valid JSON.

Then I checked our codebase and noticed we have three JSON.parse calls in the navigation history code with no error handling at all. So I started thinking - when would this actually break?

**Honestly, the app itself isn't corrupting the storage.** All our writes use JSON.stringify() on simple string arrays, so under normal operation we're always writing valid JSON. That part works fine.

**But here's where it could still crash:**

1. **Browser extensions** - Some extensions mess with sessionStorage directly and could corrupt it
2. **DevTools editing** - If someone's debugging and manually edits the storage and makes a typo
3. **Browser crashes** - Super rare, but if the browser crashes mid-write, you could get partial JSON
4. **Future code changes** - If someone later adds objects with circular references or something

Right now it's more of a **defensive programming thing** than fixing an active bug. The MDN docs treat it as a best practice when reading from external storage, and since we're dealing with sessionStorage (which we don't fully control - extensions, browser state, etc.), it made sense to add the protection.

The risk is low, but if it does happen, the entire navigation system just dies. With the try-catch, it logs a warning and keeps working.

What do you think? just wanted to add the safety net since it's zero cost in normal operation.

DilshanRanabahu (commented):
I noticed that the exact same try...catch logic for parsing sessionStorage.getItem(historyKey) is repeated across three different functions (addToHistory, setupHistory, and getHistory). >
To keep the codebase DRY (Don't Repeat Yourself), would it be better to extract this parsing logic into a single reusable helper function (e.g., getParsedHistory()) and just call it within these functions? Let me know what you think!

AshThe25:
Clean approach — extracting `getParsedHistory` with a try/catch and `Array.isArray` guard is much better than the bare `JSON.parse` calls spread across three functions.

One small suggestion: when malformed data is detected, consider also clearing the corrupted entry from `sessionStorage`:

```ts
} catch (error) {
  console.warn(`Failed to parse navigation history in ${action}, returning empty array`, error);
  sessionStorage.removeItem(historyKey); // prevent repeated warnings on every navigation
  return [];
}
```

As it stands, once the storage is corrupted, every subsequent call to `addToHistory`, `setupHistory`, and `getHistory` will log the same warning until the browser session ends — which could flood the console during normal use. Clearing it on first detection means the next `addToHistory` call starts fresh with a valid empty array.

AshThe25:
A cleaner design, with a single location to call `getParsedHistory` with try/catch and `Array.isArray` guards will be better than using raw `JSON.parse` calls in three functions.

I also suggest that if malformed data is detected, you should remove the corrupted entry from `sessionStorage`:

```ts
} catch (error) {
  console.warn(`Unable to parse navigation history in ${action}, returning empty array`, error);
  sessionStorage.removeItem(historyKey); // prevent multiple warnings for each navigation
  return [];
}
```

Within the current model, once `sessionStorage` has been corrupted, every subsequent call to `addToHistory`, `setupHistory` or `getHistory` will log the same warning until the browser session ends and will likely flood the console during normal operations. Removing a corrupted object on first detection will allow for a new valid empty array to be used on the next `addToHistory` method call.

sanks011:
> A cleaner design, with a single location to call `getParsedHistory` with try/catch and `Array.isArray` guards will be better than using raw `JSON.parse` calls in three functions.
> 
> I also suggest that if malformed data is detected, you should remove the corrupted entry from `sessionStorage`:
> 
> ```ts
> } catch (error) {
>   console.warn(`Unable to parse navigation history in ${action}, returning empty array`, error);
>   sessionStorage.removeItem(historyKey); // prevent multiple warnings for each navigation
>   return [];
> }
> ```
> 
> Within the current model, once `sessionStorage` has been corrupted, every subsequent call to `addToHistory`, `setupHistory` or `getHistory` will log the same warning until the browser session ends and will likely flood the console during normal operations. Removing a corrupted object on first detection will allow for a new valid empty array to be used on the next `addToHistory` method call.

Good catch, you're right. Without clearing the corrupted entry, every subsequent call to getParsedHistory (which fires on every routing event via getHistory, addToHistory, etc.) would log the same warning repeatedly until something happens to write valid data back - flooding the console during normal navigation.

I've added sessionStorage.removeItem(historyKey) in both error paths (malformed JSON catch block and the non-array guard). This way corruption is cleared on first detection, and the next addToHistory call naturally writes a fresh valid array. Updated the tests to verify the storage is actually cleaned up too.

Addressed in c483014.

PR #1565, opened 2026-01-28
(fix) Recalculate layout on window move and focus change

keshavrathi07:
## Problem
When an OpenMRS tab is moved to a new window, the app layout does not update
because a `resize` event is not always triggered.

## Solution
Extended the layout update logic to also react to:
- `window.focus`
- `document.visibilitychange`

This ensures layout recalculation when the browser context changes,
including when tabs are moved across windows.

## How to test
1. Open OpenMRS in a wide desktop window
2. Open another smaller window
3. Drag the OpenMRS tab to the smaller window
4. Layout updates correctly without manual refresh

## Notes
This is a frontend-only change with no breaking impact.

### DIFF
--- a/packages/framework/esm-react-utils/src/useLayoutType.ts
+++ b/packages/framework/esm-react-utils/src/useLayoutType.ts
@@ -27,11 +27,24 @@ export function useLayoutType() {
   const [type, setType] = useState<LayoutType>(getLayout);
 
   useEffect(() => {
-    const handler = () => {
+    const handleLayoutChange = () => {
       setType(getLayout());
     };
-    window.addEventListener('resize', handler);
-    return () => window.removeEventListener('resize', handler);
+
+    // normal resize
+    window.addEventListener('resize', handleLayoutChange);
+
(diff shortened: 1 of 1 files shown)


suubi-joshua:
Could you the jira ticket issue your working on here. Also could you provide some videos or images showing your changes have solved the issue.

AshThe25:
Both the `focus` and `visibilitychange` events will be triggered when the user returns to the browser window or a tab becomes visible — but `getLayout()` calls a class list from `document.body.classList`, which can only be altered via the breakpoint observer when the viewport size changes. Focus and visibilitychange events cannot alter the class list; therefore, invoking `getLayout()` on these events will most often have little impact, and the state will not be updated.

Would you mind providing the context in which this was trying to address? For instance, if a window moves to another display with a distinct DPI/Resolution, the class list would only change once the breakpoint observer evaluates and modifies the class list on `document.body`. Listening to more events for `useLayoutType` won't address this problem.

Is the purpose of handling multi-screen or window snap scenarios?

PR #1582, opened 2026-02-06
(fix) O3-4552: Fix Firefox back button issue with hash URLs

RajPrakash681:
# Requirements

## Summary
Fixes Firefox back button getting stuck when navigating between pages with trailing hash fragments (#).

## Screenshots

https://github.com/user-attachments/assets/a4fed052-f9ca-49d5-803b-78260ae3c3e5

### DIFF
--- a/packages/framework/esm-navigation/src/history/history.ts
+++ b/packages/framework/esm-navigation/src/history/history.ts
@@ -4,7 +4,18 @@ import { navigate } from '../navigation/navigate';
 const historyKey = 'openmrs:history';
 
 function addToHistory(newLocation: string) {
+  // Don't add URLs ending with '#' to history as they cause issues with browser back button
+  // (they redirect to the non-# version, making back button appear broken)
+  if (newLocation.endsWith('#')) {
+    return;
+  }
+
   let history = JSON.parse(sessionStorage.getItem(historyKey) ?? '[]') || [];
+
+  if (history.length > 0 && history[history.length - 1] === newLocation) {
+    return;
+  }
+
   history.push(newLocation);
   const maxSize = 50;
   if (history.length > maxSize) {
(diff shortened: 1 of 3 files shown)


RajPrakash681:
@chibongho

chibongho:
I think this is fine? Replacing a trailing `#` in the URL seems benign, if a bit hacky. Interested in what you think @denniskigen @ibacher .

RajPrakash681:
Hey @ibacher! 

I totally get your concern about this feeling hacky. After digging deeper, I found something interesting: the trailing `#` is actually coming from **single-spa itself** due to a Firefox quirk ([issue #1063](https://github.com/single-spa/single-spa/issues/1063), been open since 2022).

When `window.location.hash` is set to an empty string in Firefox, it auto-adds the `#`. This breaks the back button loop.

**My fix normalizes URLs in our history tracking** - basically treating `/page#` and `/page` as equivalent (which they semantically are). It's defensive programming for an upstream bug we can't control.


The change only affects history tracking, preserves legitimate hashes like `#section`, and has tests. I think it's actually pretty clean for what we're dealing with.

ibacher (approved):

packages/framework/esm-navigation/src/history/history.ts
```diff
@@ -49,7 +61,8 @@ export function getHistory(): Array<string> {
 }
 
 /**
- * Rolls back the history to the specified point and navigates to that URL.
+ * Rolls back the history to the specified point. The browser has already navigated
```
ibacher: The comment added here is, I think, incorrect or only describes one workflow. The app can _also_ call `goBackInHistory(-2)` to move this.

packages/framework/esm-navigation/src/history/history.ts
```diff
@@ -60,7 +73,6 @@ export function goBackInHistory({ toUrl }: { toUrl: string }) {
   const toIndex = history.lastIndexOf(toUrl);
   if (toIndex != -1) {
     const newHistory = history.slice(0, toIndex + 1);
-    navigate({ to: history[toIndex] });
```
ibacher: Wait, this kind of breaks the core purpose of this function which is to force the browser back in history where the back-button wasn't pressed.

packages/framework/esm-react-utils/src/ConfigurableLink.tsx
```diff
@@ -63,12 +63,10 @@ export function ConfigurableLink({
       );
     }
   }, []);
+  const href = interpolateUrl(to, templateParams);
```
ibacher: I don't understand why this refactoring is part of this PR?

ibacher (changes requested):
Actually, looking a little closer this breaks `goBackInHistory()` do to, I think, thinking that function is only called from the event handler. It is, in fact, a public function to navigate in the app.

DilshanRanabahu (commented):
Hi! To build on @ibacher's concern, the current unit tests for stripTrailingHash only cover URLs with no hash or URLs ending exactly with a single #. >
We are missing a crucial test case for actual hash-based routes. Could we add a test case with a valid path (e.g., expect(stripTrailingHash('http://localhost:8080/openmrs/spa/#/login')).toBe(...))? This will explicitly verify that valid route data is NOT accidentally stripped or corrupted by this function, ensuring better test coverage.

PR #1590, opened 2026-02-10
(fix) Remove redundant dark theme CSS mixin from implementer tools

Shawn-Dong:
## Summary
- Remove the redundant `.darkTheme` CSS mixin (`@include theme.theme(themes.$g90)`) from implementer tools, keeping only Carbon's `<Theme theme="g90">` React component which is the recommended approach per Carbon guidelines
- Move the content-switcher selected state style to `popup.styles.scss` using `theme.$background-inverse` to preserve the original appearance
- Remove unused `@carbon/react/scss/themes` imports

## Why
The dark theme was being applied twice — once via a CSS mixin and once via Carbon's `<Theme>` component. This PR removes the CSS mixin approach in favor of the `<Theme>` component.

## Test plan
- [x] Open Implementer Tools panel and verify dark theme renders correctly
- [x] Verify selected tab has white/light background (`#f4f4f4`)
- [x] Check all tabs: Configuration, Frontend modules, Backend modules, Feature flags
- [x] Verify no visual regressions compared to before the change
- [x] Build passes with no errors

## Screenshots
No visual changes expected 
Screenshot 2026-02-10 at 5 42 53 PM


suubi-joshua:
Could you look into the failing e2e test

adarshiiit0117 (commented):
Nice simplification 

I noticed that the `.darkTheme` wrapper has been removed and the theme-specific styles are now applied via `:global(.cds--content-switcher--selected)`.

Just wanted to clarify , does this change rely on the parent Carbon theme already being applied higher in the tree? Previously, the wrapper ensured theme scoping at the component level.

Also, since `:global` is being used here, are we confident this won't unintentionally affect other `cds--content-switcher` instances outside this popup?

Overall this looks cleaner, just wanting to confirm the theming scope is still well contained.

PR #1629, opened 2026-02-20
(feat) O3-5423: Add unified REST error parsing utility

himanshumoral:
## Summary

Implements unified REST exception parsing utility for O3-5423 in `esm-error-handling`.

Normalizes different OpenMRS REST error shapes into a consistent `string[]` format for UI display.

Includes comprehensive unit tests.

## Related Issue

https://issues.openmrs.org/browse/O3-5423


himanshumoral:
@jayasanka-sack @denniskigen @VeronicaMuthee Could you please review this PR?

packages/framework/esm-error-handling/src/index.ts
```diff
@@ -20,60 +20,172 @@ window.onunhandledrejection = function (event: PromiseRejectionEvent) {
...
+ * @category Error Handling
  */
-export function createErrorHandler() {
-  const outgoingErr = Error();
-  return (incomingErr: unknown) => {
-    const finalErr = ensureErrorObject(incomingErr);
-    finalErr.stack += `\nAsync stacktrace:\n${outgoingErr.stack}`;
-    reportError(incomingErr);
-  };
+export function extractErrorMessagesFromResponse(error: unknown): Array<string> {
```
RajPrakash681: refactor this as per reported by sonar

himanshumoral: Refactored the function to reduce cognitive complexity and addressed the Sonar warning. Please let me know if any further changes are needed.

packages/framework/esm-error-handling/src/index.ts
```diff
@@ -76,6 +76,145 @@ export function createErrorHandler() {
...
+
+/**
+ * The error body shape returned by the OpenMRS REST API.
+ *
+ * ValidationExceptions contain {@link fieldErrors} and/or {@link globalErrors}.
+ * Other exceptions contain {@link message} (and optionally {@link translatedMessage}).
+ *
+ * @category Error Handling
+ */
+export interface RestErrorResponse {
```
suubi-joshua: Reviewing the existing logic/pattern 
`REST error → caught in ESM → reportError() → window.onerror → dispatchToastShown()`

I feel your new function sits completely outside this chain. It returns `string[]` for the caller to handle manually, meaning it doesn't integrate with `reportError`, `createErrorHandler`, or the global toast system at all. I think the issue at hand is to have a unified way to handle REST exceptions  but your changes seem to be adding a parallel system instead of extending the existing one.

sohamdhande:
hey @himanshumoral 
The getRestErrorBody function checks 'error' in body but then immediately casts to RestErrorResponse via as unknown as RestErrorResponse — this bypasses the type system entirely and could silently pass a body that has an error key of the wrong shape (e.g. a string). A proper type guard with field-level narrowing would make this safe. Also, the test for empty fieldErrors/globalErrors asserts only expect.any(String) for subtitle, which doesn't actually verify which fallback branch fired — worth tightening to the expected ensureErrorObject message. Finally, reportRestError is a new public export and needs a changeset entry, otherwise it won't be included in any package version bump.

PR #1651, opened 2026-03-12
(fix) O3-5486: Prioritize referrer redirect after authentication

Gazal-Sharma:
# Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. Ensure your PR title includes a conventional commit label.

## If applicable

- [ ] My work is based on designs.
- [ ] I have updated the esm-framework mocks to reflect any API changes.

## Summary

This PR fixes the login redirect logic to prioritize the referrer after successful authentication.

Previously, the login component could redirect users to `/home`, which may not exist in the SPA router and could cause errors such as "Cannot GET /home".

The redirect priority is now:

1. `location.state.referrer`
2. `loginLinks.loginSuccess`
3. fallback `/home`

A test has also been added to verify that the referrer is used for redirect after login.

## Screenshots

N/A

## Related Issue

N/A

## Other

All existing tests pass locally.

### DIFF
--- a/packages/apps/esm-login-app/src/login/login.component.tsx
+++ b/packages/apps/esm-login-app/src/login/login.component.tsx
@@ -115,7 +115,7 @@ const Login: React.FC = () => {
             navigate('/login/location');
           }
         } else {
-          setErrorMessage(t('invalidCredentials', 'Invalid username or password'));
+          setErrorMessage('invalidCredentials');
           setUsername('');
           setPassword('');
           if (showPasswordOnSeparateScreen) {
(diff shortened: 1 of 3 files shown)


packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -101,34 +101,33 @@ const Login: React.FC = () => {
...
+              to = location.state.referrer;
+
+              // 2️⃣ Next priority: configured login success page
+            } else if (loginLinks?.loginSuccess) {
+              to = loginLinks.loginSuccess;
+
+              // 3️⃣ Fallback: default home route
+            } else {
+              to = '/home';
             }
```
ibacher: Right, this doesn't actually change the priority. It's effectively a different (and longer) re-implementation of the existing logic.

ibacher (changes requested):
Thank you for your interest in OpenMRS. Unfortunately, I don't think this PR is a candidate for merging. AFAICT, the "core logic" here isn't any different and this breaks several things for no good reason. I think you likely have identified an issue but have mis-identified the cause of and therefore the solution to this problem.

Please try doing a more thorough investigation of why the current code isn't working.

PR #1687, opened 2026-03-20
(fix) O3-5523: Add deep equality check on ExtensionSlot state prop to prevent cascading re-renders

sohamdhande:
## Requirements
- [x] This PR has a title that briefly describes the work done including the ticket number.
- [ ] I have updated the esm-framework and storybook mocks to reflect any API changes.

## Summary

`useExtensionSlot` uses a `useEffect` with `[slotName, state]` in its dependency array. When a parent component passes `state` as an inline object:

The `useEffect` fires on every parent render because inline objects always fail JavaScript reference equality. This causes   `updateExtensionSlotState` to dispatch on every render, forcing `extensionInternalStore` to update globally and triggering   `updateExtensionOutputStore` across all extension slots — an expensive operation for what should be a no-op.

This PR introduces a **deep equality check using `lodash-es/isEqual`** to ensure that updates are only dispatched when the actual content of `state` changes, not just its reference.

Additionally:
- `prevStateRef` now stores the raw state instead of a serialized version  
- `prevStateRef` is initialized lazily (after the first effect run) to avoid unnecessary comparisons  

This approach also correctly handles non-serializable values (e.g., functions) passed via `state`, which would not work reliably with `JSON.stringify`.

## Changed files
- `packages/framework/esm-react-utils/src/useExtensionSlot.ts`
- `packages/framework/esm-react-utils/src/extensions.test.tsx`

## Screenshots

Not applicable — no UI changes.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5523

## Other

As more context variables (location, visit, encounter) are injected via `ExtensionSlot` state props, this issue would cause unnecessary cascading re-renders across extension slots. Fixing it improves performance and ensures the system scales efficiently.

### DIFF
--- a/packages/framework/esm-react-utils/src/extensions.test.tsx
+++ b/packages/framework/esm-react-utils/src/extensions.test.tsx
@@ -1,9 +1,9 @@
 /* eslint-disable */
 import React, { useReducer } from 'react';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import '@testing-library/jest-dom/vitest';
 import userEvent from '@testing-library/user-event';
-import { act, render, screen, waitFor, within } from '@testing-library/react';
+import { act, render, renderHook, screen, waitFor, within } from '@testing-library/react';
 import { registerFeatureFlag, setFeatureFlag } from '@openmrs/esm-feature-flags';
 import {
   attach,
@@ -11,6 +11,7 @@ import {
   registerExtension,
   updateInternalExtensionStore,
 } from '@openmrs/esm-extensions';
+import * as esmExtensions from '@openmrs/esm-extensions';
 import {
   getSyncLifecycle,
   Extension,
@@ -18,6 +19,7 @@ import {
   openmrsComponentDecorator,
   useExtensionSlotMeta,
   useRenderableExtensions,
+  useExtensionSlot,
 } from '.';
 
 describe('ExtensionSlot, Extension, and useExtensionSlotMeta', () => {
@@ -301,3 +303,46 @@ function registerSimpleExtension(
     featureFlag,
   });
 }
+
+describe('useExtensionSlot — state update optimization', () => {
+  let updateExtensionSlotStateSpy: ReturnType<typeof vi.spyOn>;
+  let wrapper: any;
+
+  beforeEach(() => {
+    wrapper = openmrsComponentDecorator({
+      moduleName: 'test-module',
+      featureName: 'Test',
+      disableTranslations: true,
+    })(({ children }: any) => <>{children}</>);
+
+    updateExtensionSlotStateSpy = vi.spyOn(esmExtensions, 'updateExtensionSlotState');
+  });
+
+  afterEach(() => {
+    updateExtensionSlotStateSpy.mockRestore();
+  });
+
+  it('should not dispatch when state content is unchanged but reference changes', () => {
+    const { rerender } = renderHook(({ state }) => useExtensionSlot('test-slot-dedup', state), {
+      initialProps: { state: { encounterType: 'visit-note' } },
+      wrapper,
+    });
+
+    updateExtensionSlotStateSpy.mockClear();
+    rerender({ state: { encounterType: 'visit-note' } });
+
+    expect(updateExtensionSlotStateSpy).toHaveBeenCalledTimes(0);
+  });
+
+  it('should dispatch when state content actually changes', () => {
+    const { rerender } = renderHook(({ state }) => useExtensionSlot('test-slot-changed', state), {
+      initialProps: { state: { encounterType: 'visit-note' } },
+      wrapper,
+    });
+
+    updateExtensionSlotStateSpy.mockClear();
+    rerender({ state: { encounterType: 'lab-results' } });
+
+    expect(updateExtensionSlotStateSpy).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/framework/esm-react-utils/src/useExtensionSlot.ts
+++ b/packages/framework/esm-react-utils/src/useExtensionSlot.ts
@@ -21,9 +21,15 @@ export function useExtensionSlot(slotName: string, state?: ExtensionSlotCustomSt
     isInitialRender.current = false;
   }, []);
 
+  const prevStateRef = useRef<string>(JSON.stringify(state ?? {}));
+
   useEffect(() => {
     if (!isInitialRender.current) {
-      updateExtensionSlotState(slotName, state);
+      const serialized = JSON.stringify(state ?? {});
+      if (prevStateRef.current !== serialized) {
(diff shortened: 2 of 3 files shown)


packages/framework/esm-react-utils/src/extensions.test.tsx
```diff
@@ -301,3 +303,46 @@ function registerSimpleExtension(
     featureFlag,
   });
 }
+
+describe('useExtensionSlot — state update optimization', () => {
```
ibacher, line 307: Do not name describe blocks like this. Nest them. These belong with all the other `'useExtensionSlot'` tests.

packages/framework/esm-react-utils/src/extensions.test.tsx
```diff
@@ -301,3 +303,46 @@ function registerSimpleExtension(
...
+  let updateExtensionSlotStateSpy: ReturnType<typeof vi.spyOn>;
+  let wrapper: any;
+
+  beforeEach(() => {
+    wrapper = openmrsComponentDecorator({
+      moduleName: 'test-module',
+      featureName: 'Test',
+      disableTranslations: true,
+    })(({ children }: any) => <>{children}</>);
```
ibacher, line 316: Even in tests it is desireable not to use `any` for well-known things.

```suggestion
    })(({ children }: string | JSX.Element | JSX.Element[] | () => JSX.Element) => <>{children}</>);
```

packages/framework/esm-react-utils/src/extensions.test.tsx
ibacher: Frankly, though these tests belong in `useExtensionSlot.test.ts` and not here which is for integrated extension tests.

packages/framework/esm-react-utils/src/useExtensionSlot.ts
```diff
@@ -21,9 +21,15 @@ export function useExtensionSlot(slotName: string, state?: ExtensionSlotCustomSt
     isInitialRender.current = false;
   }, []);
 
+  const prevStateRef = useRef<string>(JSON.stringify(state ?? {}));
```
ibacher, line 24: To meet React's rule of hooks this needs to come in the body of this hook where all the other hooks are, i.e., before the `!moduleName` check.

packages/framework/esm-react-utils/src/extensions.test.tsx
```diff
@@ -301,3 +303,46 @@ function registerSimpleExtension(
...
+  });
+
+  it('should not dispatch when state content is unchanged but reference changes', () => {
+    const { rerender } = renderHook(({ state }) => useExtensionSlot('test-slot-dedup', state), {
+      initialProps: { state: { encounterType: 'visit-note' } },
+      wrapper,
+    });
+
+    updateExtensionSlotStateSpy.mockClear();
+    rerender({ state: { encounterType: 'visit-note' } });
```
ibacher, line 332: Ideally, we'd actually do this several times.

ibacher (changes requested):
While this has correctly identified an issue, the issue is not that we need to do "deep equality" check, but that we're not properly checking if the state has changed in a way that matters. The right tool to use for this is actually to just store the state in a ref (not a string representation of the state) and use `esm-utils` `shallowEqual()` function to test if the state has changed. This should be somewhat similar to what's done in, e.g., `useDefineAppContext()`.

ibacher:
So, actually, after some re-thinking, I've come around on the idea that deep equality here probably makes sense. However, I don't think `JSON.stringify()` is the right way to do that. This is because we don't actually restrict what is supplied in the `state` to things that are correctly serializable via JSON. More specifically, it's acceptable to pass functions via `state` and, in fact, we use that with some frequency where parents and children need to "communicate".

I think we should:

1. Switch `prevStateRef` to just storing the state
2. Use `lodash-es/isEqual` for a "proper" deep equality check
3. Don't initialize `prevStateRef` until the effect runs at least once (i.e., initialize it to `unknown`

Doing this actually will save us two calls to `JSON.stringify(state)` per render cycle, which is a relatively heavy operation and meet the goal of checking for deep equality.

packages/framework/esm-react-utils/src/useExtensionSlot.ts
```diff
@@ -21,9 +21,15 @@ export function useExtensionSlot(slotName: string, state?: ExtensionSlotCustomSt
...
   useEffect(() => {
     if (!isInitialRender.current) {
-      updateExtensionSlotState(slotName, state);
+      const serialized = JSON.stringify(state ?? {});
+      if (prevStateRef.current !== serialized) {
+        prevStateRef.current = serialized;
+        updateExtensionSlotState(slotName, state);
+      }
     }
   }, [slotName, state]);
```
ibacher, line 34: In addition to the implementation concerns I pointed to, I think it would be a good idea for `prevStateRef` to be reset if the `slotName` changes.

sohamdhande:
hey @ibacher . I have made the following changes
Switched from JSON.stringify to lodash-es/isEqual for deep equality
Updated prevStateRef to store the raw state
Initialized prevStateRef lazily so it’s set after the first effect run

This also removes the extra serialization overhead per render. Let me know if this looks good or if you’d like any further tweaks.

sohamdhande:
and also updated the pr title

AshThe25:
Using `JSON.stringify` for deep equality has a few known edge cases that could silently suppress legitimate state updates:

- **`undefined` values are dropped**: `JSON.stringify({ a: undefined })` → `'{}'`, making it equal to `JSON.stringify({})` even though the objects are structurally different. If any state key holds `undefined`, a change to that key won't be detected.
- **Key ordering**: `JSON.stringify({ a: 1, b: 2 })` !== `JSON.stringify({ b: 2, a: 1 })` — two deeply equal objects with different key insertion order would be treated as different, triggering an unnecessary update.

A more reliable approach would be `fast-deep-equal` (very likely already in the dependency tree via other packages) or a simple recursive comparison that handles these cases. Alternatively, since `state` is typed as `ExtensionSlotCustomState`, you could document the constraints (no `undefined` values, stable key order) so consumers know what to expect.

Great idea overall — the performance win from skipping redundant dispatches is real.

AshThe25:
When using `JSON.stringify` for deep equality, there are a couple of edge cases where it could potentially ignore state changes:

- `JSON.stringify({ a: undefined })` → `'{}'` will result in it being the same as `JSON.stringify({})`, but the two objects are different in structure. In case any state property has an `undefined` value, any change to it would go unnoticed.
- Key order: `JSON.stringify({ a: 1, b: 2 })` !== `JSON.stringify({ b: 2, a: 1 })` – two objects with the same properties that were added in a different order would be considered unequal, resulting in an unnecessary state change.

The best solution would be `fast-deep-equal` (which is probably already a dependency of your package), or simply a recursive function to compare the states. Another possible fix is adding some documentation about the limitations of state, such as not allowing `undefined` values or having consistent key order.

Great idea, in general – there's a huge benefit to not triggering unnecessary events.

PR #1738, opened 2026-04-08
(fix) O3-4478: Regression Logged-in users should be redirected to home page

RaniduNethma:
# Requirements

## If applicable

- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below.

## Summary

This PR addresses a regression where users without a session location or default location were being incorrectly redirected after authentication. It updates the login component to explicitly respect the `chooseLocation.enabled` configuration. Additionally, it resolves an incomplete test asserting location selections.

### Approach
- **`login.component.tsx`**: 
  - Destructured `chooseLocation` from the `useConfig()` schema. 
  - Updated hooks (`useEffect`, `handleSubmit`) and added a condition to only redirect the user to `/login/location` if `chooseLocation.enabled` is truthy.
- **`login.test.tsx`**: 
  - Removed outdated `// TODO: Complete the test` comment.
  - Re-structured `mockLogin.mockImplementation` logic to accurately mimic the standard API payload (`sessionStore.session`).
  - Implemented missing `waitFor()` async assertions to correctly test `refetchCurrentUser`.

## Related Issue
Fixes https://openmrs.atlassian.net/issues?filter=10636&selectedIssue=O3-4478

## Other


RaniduNethma:
Hi @ibacher @denniskigen could you please review this when you have a moment?

RaniduNethma:
could you please review this when you have a moment? @ibacher @denniskigen

ELVIS-KATO:
Thanks @RaniduNethma thanks for the contribution to openmrs, But I don't know if you considered handling errors coz it does not appear that errors from setSessionLocation are explicitly handled. Like If setting the session location fails （poor network/server error）, does the user get stuck or see a generic failure page?  or you consider adding user feedback or at least error logging for this path. Otherwise The rest LGTM Thanks!

RaniduNethma:
> Thanks @RaniduNethma thanks for the contribution to openmrs, But I don't know if you considered handling errors coz it does not appear that errors from setSessionLocation are explicitly handled. Like If setting the session location fails （poor network/server error）, does the user get stuck or see a generic failure page? or you consider adding user feedback or at least error logging for this path. Otherwise The rest LGTM Thanks!

Hi @ELVIS-KATO  thanks for the feedback. You're right while the entire block is currently wrapped in a try-catch, a failure in `setSessionLocation` would currently treat the whole login as a failure and reset the form.

To improve the UX, I’ve refactored this to 'fail-soft' if setting the default location fails, it will now log the error and redirect the user to the manual location picker `/login/location` so they can still complete their login without re-entering credentials.

ELVIS-KATO:
Thanks @RaniduNethma again, that looks awesome. This LGTM @denniskigen  @ibacher

RaniduNethma:
> Thanks @RaniduNethma again, that looks awesome. This LGTM @denniskigen @ibacher

Welcome @ELVIS-KATO please excuse the messy commit history @denniskigen @ibacher

RaniduNethma:
@denniskigen @ibacher Kindly requesting workflow approval and a review when convenient. Thank you for your time!

packages/apps/esm-login-app/src/login/login.test.tsx
```diff
@@ -19,12 +19,21 @@ const mockLogin = vi.mocked(refetchCurrentUser);
...
 
 const loginLocations = [
   { uuid: '111', display: 'Earth' },
   { uuid: '222', display: 'Mars' },
 ];
 
+vi.mock('react-router-dom', async () => {
+  const actual = await vi.importActual('react-router-dom');
+  return {
+    ...(actual as any),
```
ibacher: Is there a reason we need the `as any` here?

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -24,6 +25,10 @@ export interface LoginReferrer {
 
 const Login: React.FC = () => {
   const {
+    showPasswordOnSeparateScreen,
```
denniskigen, line 28: Thanks for tackling this regression, @RaniduNethma. Gating the location-picker redirect on `chooseLocation.enabled` is the right fix.

The build is currently failing because this block declares `showPasswordOnSeparateScreen`, `provider: loginProvider`, and `links: loginLinks` twice. Modules run in strict mode, so the duplicate bindings are a SyntaxError, which is what breaks `yarn build`. Only `chooseLocation` was new here; the other three were already destructured in this block. Could you dedupe it like this?

```ts
showPasswordOnSeparateScreen,
provider: loginProvider,
links: loginLinks,
chooseLocation,
announcements = [],
background = { image: '', color: '' },
```

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -53,8 +58,15 @@ const Login: React.FC = () => {
...
+    } else {
+      if (sessionLocation || !chooseLocation.enabled) {
+        openmrsNavigate({ to: loginLinks?.loginSuccess || '${openmrsSpaBase}/home' });
+      } else if (location.pathname !== '/login/location') {
+        // Only redirect to location picker if not already there
+        navigate('/login/location');
+      }
     }
-  }, [username, navigate, location, user, loginProvider]);
+  }, [username, navigate, location, user, loginProvider, chooseLocation, loginLinks]);
```
denniskigen: One smaller thing while you're here: this effect reads `sessionLocation`, but it isn't in the dependency array, so the redirect won't re-run when the session location changes. Could you add it there?

packages/apps/esm-login-app/src/login/login.test.tsx
RaniduNethma: Done, replaced the as any casting by passing a proper generic type parameter to ensure type safety.

packages/apps/esm-login-app/src/login/login.component.tsx
RaniduNethma: Thanks for pointing that out. I have deduplicated the configuration destructuring block to resolve the syntax error and ensure a clean build.

packages/apps/esm-login-app/src/login/login.component.tsx
RaniduNethma: Added `sessionLocation` to the `useEffect` dependency array so the redirect logic triggers correctly whenever the session location state updates.

PR #1750, opened 2026-04-15
(feat) Add support to invoke develop with package name

ibacher:
## Requirements

## Summary

This adds support for `openmrs develop --packages @openmrs/esm-some-app` in addition to existing support for using `--sources`. Packages are resolved either from the nearest package.json or from its workspace definition.

The main advantage here is that you can use this without needing the exact path to the sources, e.g. if switching between repos or the like.

## Other

This feature is mostly inspired by `yarn workspace ...` and `yarn turbo filter=...` as being a more convenient way to refer to packages in monorepos, so I wanted the same thing for the CLI.

### DIFF
--- a/.changeset/early-pumas-mate.md
+++ b/.changeset/early-pumas-mate.md
@@ -1,0 +1,5 @@
+---
+"openmrs": minor
+---
+
+(feat) Add support to invoke develop with package name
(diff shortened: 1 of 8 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher!

.changeset/early-pumas-mate.md
NethmiRodrigo, line 1: Asking just for context, are we keeping these changesets for a specific reason?

NethmiRodrigo (approved):

.changeset/early-pumas-mate.md
ibacher: I meant to move this repo over to using changesets properly but haven't had the time to do so.

PR #1751, opened 2026-04-17
(fix) O3-4020: Return current app context value synchronously on first render

praneeth622:
## Requirements

- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [Contributing Guidelines](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines).

## Summary

Addresses [O3-4020](https://openmrs.atlassian.net/browse/O3-4020): `useAppContext` returned `null`/falsy-empty values in `AdmissionRequestsWorkspace` (and, more generally, in any consumer that mounted before the subscription effect fired or after the owner unmounted).

`useAppContext` had two related issues:

1. **First-render race.** `useState<Readonly<U>>()` initialized the local value as `undefined` and was only populated once the `useEffect` subscription fired. On the very first render, consumers received `undefined` even when the namespace was already registered. This is the timing race that produced the symptom reported by @chibongho (cc @ibacher) — the Ward App's `AdmissionRequestsWorkspace` destructured the context before it had arrived.
2. **Owner-unmount shadowing.** `subscribeToContext` substitutes `Object.freeze({})` when a namespace is unregistered (via `Object.assign({}, null)`). The previous implementation forwarded that `{}` straight to `setValue`, so after the namespace owner (e.g. `DefaultWardView` / `MaternalWardView`) unmounted, every consumer held a truthy empty object instead of `undefined`. Destructuring that object yielded `undefined` properties and required defensive `?? {}` guards — exactly the workaround present in `admission-requests.workspace.tsx`.

### Fix

- Seed the `useState` initializer by calling `getContext(namespace)` synchronously, so the first render already has the current value (or `undefined` when the namespace is absent).
- Inside the subscription callback, prefer the `state` argument forwarded by `subscribeToContext` to avoid re-reading/cloning the context on every update. Fall back to `getContext` only for the ambiguous empty-object case, and set the local value to `undefined` when the namespace is no longer registered.
- Share a small `isBlankNamespace` helper between the state initializer and the validation effect so whitespace-only namespaces are handled uniformly (`namespace.trim().length === 0` rather than `replace(' ', '')`, which only trimmed a single space).

### Tests (new file `useAppContext.test.tsx`)

- Returns `undefined` when the namespace is not registered.
- Returns the currently-registered value on the very first render.
- Applies the selector on the first render.
- Exposes a sibling `useDefineAppContext` value to the consumer after effects flush.
- Returns `undefined` (not `{}`) after the namespace owner unmounts (the O3-4020 scenario).

All 5 new tests pass; full `@openmrs/esm-react-utils` suite remains green (125 passed, 9 skipped).

### Behavior changes

| Scenario | Before | After |
| --- | --- | --- |
| First render, namespace already registered | `undefined` | registered value |
| First render, namespace not registered | `{}` eventually | `undefined` |
| Owner unmounts while consumer mounted | `{}` (frozen) | `undefined` |
| Steady state (owner mounted, value updates) | updated value | updated value |

The default selector `(state) => (state ?? {}) as Readonly<U>` is preserved; it is now only invoked when the state is a real, registered value.

## Screenshots

N/A — this is a behavioral fix in a framework hook; no visible UI changes. Verified via unit tests covering the first-render, selector, sibling-producer, and owner-unmount scenarios.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4020

## Other

The Ward App's `?? {}` defensive workaround in `admission-requests.workspace.tsx` continues to function and remains in place — it is now correctly redundant for the registered case and correctly guards against a genuinely absent owner.

[O3-4020]: https://openmrs.atlassian.net/browse/O3-4020

### DIFF
--- a/.changeset/fix-useappcontext-initial-render.md
+++ b/.changeset/fix-useappcontext-initial-render.md
@@ -1,0 +1,5 @@
+---
+"@openmrs/esm-react-utils": patch
+---
+
+Return current app context value synchronously from useAppContext and undefined when the namespace has not been registered yet (O3-4020).
--- a/.changeset/fix-useappcontext-initial-render.md
+++ b/.changeset/fix-useappcontext-initial-render.md
@@ -1,5 +1,6 @@
 ---
 "@openmrs/esm-react-utils": minor
+"@openmrs/esm-context": patch
 ---
 
-Return current app context value synchronously from useAppContext and undefined when the namespace has not been registered yet (O3-4020).
+Return current app context value synchronously from useAppContext and undefined when the namespace has not been registered yet (O3-4020)
--- a/.changeset/fix-useappcontext-initial-render.md
+++ b/.changeset/fix-useappcontext-initial-render.md
@@ -1,6 +1,6 @@
 ---
 "@openmrs/esm-react-utils": minor
-"@openmrs/esm-context": patch
+"@openmrs/esm-context": minor
 ---
 
 Return current app context value synchronously from useAppContext and undefined when the namespace has not been registered yet (O3-4020)
--- a/.changeset/fix-useappcontext-initial-render.md
+++ b/.changeset/fix-useappcontext-initial-render.md
@@ -1,5 +1,5 @@
 ---
-"@openmrs/esm-react-utils": patch
+"@openmrs/esm-react-utils": minor
 ---
 
 Return current app context value synchronously from useAppContext and undefined when the namespace has not been registered yet (O3-4020).
--- a/packages/framework/esm-context/package.json
+++ b/packages/framework/esm-context/package.json
@@ -19,12 +19,12 @@
   "source": true,
   "sideEffects": false,
   "scripts": {
-    "test": "cross-env TZ=UTC vitest run --passWithNoTests",
-    "test:watch": "cross-env TZ=UTC vitest watch --passWithNoTests",
     "build": "rimraf dist && concurrently \"swc --strip-leading-paths src -d dist\" \"tsc --project tsconfig.build.json\"",
     "build:development": "rimraf dist && concurrently \"swc --strip-leading-paths src -d dist\" \"tsc --project tsconfig.build.json\"",
     "typescript": "tsc --project tsconfig.build.json",
     "lint": "eslint src --ext ts,tsx",
+    "test": "cross-env TZ=UTC vitest run --passWithNoTests",
+    "test:watch": "cross-env TZ=UTC vitest watch --passWithNoTests",
     "coverage": "cross-env TZ=UTC vitest run --coverage --passWithNoTests"
   },
   "keywords": [
--- a/packages/framework/esm-context/src/context.ts
+++ b/packages/framework/esm-context/src/context.ts
@@ -117,15 +117,18 @@ export function subscribeToContext<T extends NonNullable<object> = NonNullable<o
 ) {
   let previous = getContext<T>(namespace);
 
-  // set initial value
-  callback(Object.freeze(Object.assign({}, previous)));
+  callback(snapshot(previous));
 
   return contextStore.subscribe((state) => {
-    let current: Readonly<T> | null | undefined = namespace in state ? (state[namespace] as T) : null;
+    const current: Readonly<T> | null = namespace in state ? (state[namespace] as T) : null;
 
     if (current !== previous) {
       previous = current;
-      callback(Object.freeze(Object.assign({}, current)));
+      callback(snapshot(current));
     }
   });
+}
+
+function snapshot<T extends NonNullable<object> = NonNullable<object>>(value: Readonly<T> | null): Readonly<T> | null {
+  return value === null ? null : Object.freeze(Object.assign({}, value));
 }
--- a/packages/framework/esm-framework/docs/functions/useAppContext.md
+++ b/packages/framework/esm-framework/docs/functions/useAppContext.md
@@ -2,6 +2,8 @@
 
 # Function: useAppContext()
 
+## Call Signature
+
 > **useAppContext**\<`T`\>(`namespace`): `undefined` \| `Readonly`\<`T`\>
 
 Defined in: [packages/framework/esm-react-utils/src/useAppContext.ts:26](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-react-utils/src/useAppContext.ts#L26)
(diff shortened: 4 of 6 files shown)


praneeth622:
Addressed Copilot's three review comments and added a changeset:

1. `useAppContext.ts` subscribe callback — dropped the inner `getContext` call so we no longer double-clone on every update; we now use the state passed into the callback directly and only fall back to `getContext` for the ambiguous empty-object case (where subscribeToContext substitutes a frozen `{}` on unregister).
2. `isBlankNamespace` helper — replaced `namespace.replace(' ', '') === ''` (which only stripped the first space) with `namespace.trim().length === 0`, and reused the helper in both the state initializer and the validation effect.
3. Test rename — renamed the misleading "on the consumer's first render" test to reflect that RTL's `render()` flushes effects before returning, so we're actually asserting post-effect steady state. The pre-effect/first-render contract is covered separately by the "currently-registered value on the very first render" test that pre-registers before `renderHook`.

Added `.changeset/fix-useappcontext-initial-render.md` as a `@openmrs/esm-react-utils` patch.
Requesting a second look from @ibacher  @dkayiwa  @samuelmale

ibacher (approved):

AshThe25:
Snapshot helper is a great extract - no need to repeat `Object.freeze(Object.assign({}, ...))` in two places which makes it more obvious as well as easier to change it when snapshotting mechanism changes.

One thing that needs to be verified: In `subscribeToContext`, `callback(snapshot(previous))` is immediately called at the moment of subscription when `previous` is gotten through the `getContext<T>(namespace)` function, which means that if the namespace hasn't been registered, the result will be `null`, and thus `snapshot(null)` = `null`. In this case, `useAppContext` will return `null` → `undefined` properly during the first render with an unregistered namespace. Worth checking if hook always replaces `null` with `undefined` to avoid getting `null` values from consumers of `T | undefined`.

praneeth622:
@AshThe25 good question, traced it end-to-end. `subscribeToContext` in `context.ts` does invoke the callback with `null` via `snapshot(null)` when the namespace is unregistered. But `useAppContext` in `useAppContext.ts` maps that `null` to `undefined` at both entry points:

- `useState` initializer: when `getContext(namespace)` returns `null`, the initializer returns `undefined` explicitly (`if (current === null) return undefined;`).
- Subscription callback: `if (state == null) { setValue(undefined); return; }` catches both `null` and `undefined`.

So the hook's declared `Readonly<U> | undefined` contract holds for consumers, and a real `null` never escapes. The two tests added in this PR (`returns undefined when the namespace is not registered` and `returns undefined (not {}) after the namespace owner unmounts`) exercise both paths.

AshThe25:
> @AshThe25 good question, traced it end-to-end. `subscribeToContext` in `context.ts` does invoke the callback with `null` via `snapshot(null)` when the namespace is unregistered. But `useAppContext` in `useAppContext.ts` maps that `null` to `undefined` at both entry points:
> 
> * `useState` initializer: when `getContext(namespace)` returns `null`, the initializer returns `undefined` explicitly (`if (current === null) return undefined;`).
> * Subscription callback: `if (state == null) { setValue(undefined); return; }` catches both `null` and `undefined`.
> 
> So the hook's declared `Readonly<U> | undefined` contract holds for consumers, and a real `null` never escapes. The two tests added in this PR (`returns undefined when the namespace is not registered` and `returns undefined (not {}) after the namespace owner unmounts`) exercise both paths.

@praneeth622 Thank you for following up, and I agree that null → undefined (2 layer) is reasonable; however, it’s also worth noting that because useAppContext is a generic type over U, objects will have to narrow to Readonly | undefined before they can be used; additionally, consider adding a note in jsdoc that “undefined” means “the namespace is not yet registered, or was removed, etc.” so that consumers know to handle the undefined case explicitly versus defaulting to an empty object — this may lead to hidden bugs.

praneeth622:
@denniskigen @ibacher @jayasanka-sack

All review feedback addressed and the PR title is now compliant with the conventional commits check. Summary of what changed:

- `useAppContext` now returns the current context value synchronously on first render via a `useState` initializer, so consumers no longer see `undefined` on mount when the namespace is already registered.
- Subscription callback maps `null` to `undefined` so the declared return type `Readonly<U> | undefined` holds end to end (addresses @AshThe25's null-propagation question no extra guard needed; `subscribeToContext` can emit `null`, but both entry points normalize it).
- Title updated to match the repo's pr-title-check regex.

Ready for another look. Happy to make any further changes you'd like.

Thanks!

AshThe25:
> @denniskigen @ibacher @jayasanka-sack
> 
> All review feedback addressed and the PR title is now compliant with the conventional commits check. Summary of what changed:
> 
> * `useAppContext` now returns the current context value synchronously on first render via a `useState` initializer, so consumers no longer see `undefined` on mount when the namespace is already registered.
> * Subscription callback maps `null` to `undefined` so the declared return type `Readonly<U> | undefined` holds end to end (addresses @AshThe25's null-propagation question no extra guard needed; `subscribeToContext` can emit `null`, but both entry points normalize it).
> * Title updated to match the repo's pr-title-check regex.
> 
> Ready for another look. Happy to make any further changes you'd like.
> 
> Thanks!

@praneeth622 the null→undefined normalization at both entry points makes it airtight — good to have it documented in the summary.

praneeth622:
@AshThe25 good call  pushed a JSDoc update in c7ecb9b2 that documents exactly that. Both overloads now carry a `@returns` clause spelling out that `undefined` means the namespace is not yet registered, was unregistered (e.g. owner unmounted), or is blank, and explicitly warning consumers not to default to `{}` since an empty-object default can mask a genuinely missing namespace.

On the generic narrowing: fair point `Readonly<U> | undefined` forces a narrow at the call site, but that's exactly the signal we want. Hiding it behind a default would lose the "registered vs empty" distinction the hook is now deliberately preserving.

Thanks for the thorough review.

AshThe25:
> @AshThe25 good call pushed a JSDoc update in [c7ecb9b](https://github.com/openmrs/openmrs-esm-core/commit/c7ecb9b257fe2d575064557f0a1494cf0ad63804) that documents exactly that. Both overloads now carry a `@returns` clause spelling out that `undefined` means the namespace is not yet registered, was unregistered (e.g. owner unmounted), or is blank, and explicitly warning consumers not to default to `{}` since an empty-object default can mask a genuinely missing namespace.
> 
> On the generic narrowing: fair point `Readonly<U> | undefined` forces a narrow at the call site, but that's exactly the signal we want. Hiding it behind a default would lose the "registered vs empty" distinction the hook is now deliberately preserving.
> 
> Thanks for the thorough review.

Makes sense — the @returns doc makes the three-case semantics (not yet registered, unregistered, blank) explicit right where it matters. And the narrowing rationale tracks: once the JSDoc spells that out, callers have no excuse to default to {} and silently swallow a missing namespace. LGTM

PR #1757, opened 2026-04-22
(fix) Render UserHasAccess children synchronously when session is alr…

moconnell9:
## Requirements

## Summary

`UserHasAccess` previously initialised its user state to `null` and only  populated it inside a `useEffect` — meaning the **first render always returned null** even when the session was already loaded in `sessionStore`.

This caused two observable problems:

- **Flash of missing content during in-app navigation**: when a page transition renders a `UserHasAccess`-gated component after the session is already loaded, the component briefly shows nothing (or a fallback) until the effect fires.
- **Permanently hidden UI on slow/broken networks**: if the `getCurrentUser` observable never emits (e.g. the backend is unreachable), gated content is never shown — including buttons inside already-open modals that the user triggered themselves.

Fix: read `sessionStore` synchronously as the lazy `useState` initialiser so the first render already has the correct user. `useEffect` re-syncs immediately and subscribes for future changes as before. A test is added that proves children render on the first render when the session is already loaded (the Observable never emits in that test).

## Screenshots

Logic change with no visual diff in isolation, but hopes to fix below issue:

Before:
IMG_4913 2

After: (simulated)
Screenshot 2026-04-22 at 5 16 04 PM

## Related Issue

Related patient-chart PR:
https://github.com/openmrs/openmrs-esm-patient-chart/pull/3258

N/A

## Other

None

### DIFF
--- a/packages/framework/esm-react-utils/src/UserHasAccess.tsx
+++ b/packages/framework/esm-react-utils/src/UserHasAccess.tsx
@@ -1,12 +1,16 @@
 /** @module @category API */
-import type { LoggedInUser } from '@openmrs/esm-api';
-import { getCurrentUser, userHasAccess } from '@openmrs/esm-api';
+import type { LoggedInUser, SessionStore } from '@openmrs/esm-api';
+import { getCurrentUser, sessionStore, userHasAccess } from '@openmrs/esm-api';
 import React, { useEffect, useState } from 'react';
 
 export interface UserHasAccessProps {
   privilege: string | string[];
   fallback?: React.ReactNode;
   children?: React.ReactNode;
+}
+
+function getUserFromStore(state: SessionStore): LoggedInUser | null {
+  return state.loaded ? (state.session?.user ?? null) : null;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/framework/esm-react-utils/src/UserHasAccess.tsx
```diff
@@ -38,9 +42,15 @@ export interface UserHasAccessProps {
  *   privileges.
  */
 export const UserHasAccess: React.FC<UserHasAccessProps> = ({ privilege, fallback, children }) => {
-  const [user, setUser] = useState<LoggedInUser | null>(null);
+  // Read the session store synchronously so that components which render after the
+  // session is already loaded (e.g. during in-app navigation) get the correct value
+  // on the very first render rather than flashing blank while useEffect fires.
```
ibacher, line 47:
```suggestion,
```

packages/framework/esm-react-utils/src/UserHasAccess.tsx
```diff
@@ -38,9 +42,15 @@ export interface UserHasAccessProps {
...
+  const [user, setUser] = useState<LoggedInUser | null>(() => getUserFromStore(sessionStore.getState()));
 
   useEffect(() => {
+    // Re-sync in case the store changed between the render and this effect, then
+    // subscribe for future updates.
```
ibacher, line 52:
```suggestion
```

packages/framework/esm-react-utils/src/UserHasAccess.tsx
```diff
@@ -38,9 +42,15 @@ export interface UserHasAccessProps {
...
+    setUser(getUserFromStore(sessionStore.getState()));
     const subscription = getCurrentUser({
```
ibacher, line 54: Strictly speaking, this should subscribe to the store and maybe this whole component can be re-written in terms of `useSyncExternalStore()`

EDSONZ-WASSWA:
Hello @moconnell9  its been a while since the last review, kindly please respond to the review from @ibacher

PR #1758, opened 2026-04-24
(fix) O3-4924: Align expand animation and size prop with Carbon design system

RajPrakash681:
## Requirements

## Summary

The date picker calendar was opening with a slide-from-above animation (translate3d) instead of Carbon's standard scale-expand animation. Also, the size prop was broken only md worked, and the range picker had no size prop at all.

## Screenshots

https://github.com/user-attachments/assets/fc954b10-eb2a-43c9-b4f7-41f709c273c1

## Related Issue

https://issues.openmrs.org/browse/O3-4924

### DIFF
--- a/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
+++ b/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
@@ -163,7 +163,7 @@ describe('OpenmrsDateRangePicker', () => {
       const { container } = render(<OpenmrsDateRangePicker aria-label="datepicker" />);
       const wrapper = getInputsWrapper(container);
       expect(wrapper?.className).toContain('inputsWrapperMd');
-      expect(screen.getByRole('button')).toHaveClass('flatButtonMd');
+      expect(screen.getByRole('button').className).toContain('flatButtonMd');
     });
 
     it('should apply sm size classes when size="sm"', () => {
@@ -171,7 +171,7 @@ describe('OpenmrsDateRangePicker', () => {
       const wrapper = getInputsWrapper(container);
       expect(wrapper?.className).toContain('inputsWrapperSm');
       expect(wrapper?.className).not.toContain('inputsWrapperMd');
-      expect(screen.getByRole('button')).toHaveClass('flatButtonSm');
+      expect(screen.getByRole('button').className).toContain('flatButtonSm');
     });
 
     it('should apply md size classes when size="md"', () => {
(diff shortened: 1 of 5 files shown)


packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
```diff
@@ -152,6 +152,46 @@ describe('OpenmrsDateRangePicker', () => {
...
+      expect(screen.getByRole('button')).toHaveClass('flatButtonMd');
+    });
+
+    it('should apply sm size classes when size="sm"', () => {
+      const { container } = render(<OpenmrsDateRangePicker aria-label="datepicker" size="sm" />);
+      const wrapper = getInputsWrapper(container);
+      expect(wrapper?.className).toContain('inputsWrapperSm');
+      expect(wrapper?.className).not.toContain('inputsWrapperMd');
+      expect(screen.getByRole('button')).toHaveClass('flatButtonSm');
+    });
```
ibacher: I was going to suggest skipping these tests, but since we have them might as well do something like this:

```suggestion
      const wrapper = getInputsWrapper(container)!;
      expect(wrapper.className).toContain('inputsWrapperSm');
      expect(screen.getByRole('button')).toHaveClass('flatButtonSm');
    });
```

I don't think we need the negative assertion. It doesn't really add value. But this also applies to the other tests I think.

packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
```diff
@@ -161,32 +161,29 @@ describe('OpenmrsDateRangePicker', () => {
...
       expect(screen.getByRole('button').className).toContain('flatButtonMd');
     });
 
     it('should apply sm size classes when size="sm"', () => {
       const { container } = render(<OpenmrsDateRangePicker aria-label="datepicker" size="sm" />);
-      const wrapper = getInputsWrapper(container);
-      expect(wrapper?.className).toContain('inputsWrapperSm');
-      expect(wrapper?.className).not.toContain('inputsWrapperMd');
+      const wrapper = getInputsWrapper(container)!;
+      expect(wrapper.className).toContain('inputsWrapperSm');
```
ibacher: Marks here are also suggesting to use `toHaveClass` and that seems like a good idea.

ibacher: classnames are mangled at build, but I don't think we mangle them in Jest.

RajPrakash681: Since the lint rule enforces `.toHaveClass()` over `.className` checks, importing the styles module felt like the cleanest way to satisfy it while keeping the assertions explicit. Happy to simplify also if you prefer just let me know.

ibacher (approved):

RajPrakash681:
thanks @ibacher can you review [this](https://github.com/openmrs/openmrs-esm-stock-management/pull/538) one too

PR #1763, opened 2026-05-01
(BREAKING) O3-5645: Prevent open redirect via unvalidated returnToUrl

omeriinnocent:
## Requirements

## Summary
This pull request adds robust protection against open-redirect vulnerabilities in the login location picker by introducing a new URL validation utility and comprehensive tests. Now, only safe, same-origin SPA paths can be used for post-login navigation, preventing attackers from redirecting users to malicious sites via manipulated URLs.

**Security enhancements:**

* Added a new `isSafeReturnUrl` function in `location-picker-view.component.tsx` to validate that any `returnToUrl` is a same-origin path under the SPA base, blocking open-redirect attacks.
* Updated navigation logic in `LocationPickerView` to use `isSafeReturnUrl` before redirecting, ensuring only validated URLs are used for post-login navigation.

**Testing improvements:**

* Exported `isSafeReturnUrl` and added a dedicated test suite in `location-picker.test.tsx` to verify its behavior against various URL patterns, including external, protocol-relative, and invalid URLs. [[1]](diffhunk://#diff-24199ee0f0001cf082ae8d8c36d6c1784ea64b09ab1b37c84275f7e82dc9497eL24-R25) [[2]](diffhunk://#diff-24199ee0f0001cf082ae8d8c36d6c1784ea64b09ab1b37c84275f7e82dc9497eR349-R466)
* Added integration tests to ensure that the component correctly falls back to a safe default when an unsafe `returnToUrl` is provided, and accepts valid SPA paths.

**Test setup adjustments:**

* Mocked the `navigate` function to verify navigation behavior in tests. [[1]](diffhunk://#diff-24199ee0f0001cf082ae8d8c36d6c1784ea64b09ab1b37c84275f7e82dc9497eR6) [[2]](diffhunk://#diff-24199ee0f0001cf082ae8d8c36d6c1784ea64b09ab1b37c84275f7e82dc9497eR47)

## Screenshots
image

## Related Issue
https://openmrs.atlassian.net/browse/O3-5645


packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
```diff
@@ -16,6 +16,24 @@ import type { ConfigSchema } from '../config-schema';
...
 
+/**
+ * Validates that a returnToUrl is safe to navigate to.
+ * Only same-origin paths under the SPA base are permitted,
+ * preventing open-redirect attacks after login.
+ */
+export function isSafeReturnUrl(url: string): boolean {
+  try {
+    const parsed = new URL(url, window.location.origin);
+    const spaBase = window.getOpenmrsSpaBase().replace(/\/$/, '');
```
ibacher: Don't really need a regex here. `getOpenmrsSpaBase()` always ends in `/`.

```suggestion
    const spaBase = window.getOpenmrsSpaBase().slice(-1);
```

ibacher: Or the framework utility `trimTrailingSlash()`.

omeriinnocent:
Thanks @ibacher  for the suggestion! Since getOpenmrsSpaBase() always returns a trailing /, I replaced the regex with .slice(0, -1) for simplicity. I also checked for a shared trimTrailingSlash utility, but it’s not exported, so this approach fits best.

omeriinnocent:
Hi @ibacher @denniskigen , could you please review this fix when you have a moment?

packages/apps/esm-login-app/src/change-location-link/change-location-link.test.tsx
```diff
@@ -31,7 +31,7 @@ describe('ChangeLocationLink', () => {
     await user.click(changeLocationButton);
 
     expect(mockNavigate).toHaveBeenCalledWith({
-      to: '${openmrsSpaBase}/login/location?returnToUrl=/openmrs/spa/home&update=true',
+      to: `\${openmrsSpaBase}/login/location?returnToUrl=${encodeURIComponent('/openmrs/spa/home')}&update=true`,
```
ibacher: The test should still be using a static string I think without calling `encodeURIComponent()` otherwise this assertion is basically just `expect(a(1)).toBe(a(1))` which isn't terribly useful.

omeriinnocent: Thanks @ibacher , for pointing that out this — I’ve updated the test to use a static, pre-encoded string (${openmrsSpaBase}/login/location?returnToUrl=%2Fopenmrs%2Fspa%2Fhome&update=true) instead of calling encodeURIComponent in the assertion. I think this way now, the test verifies the actual output rather than mirroring the implementation.

packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
```diff
@@ -16,6 +16,24 @@ import type { ConfigSchema } from '../config-schema';
 import type { LoginReferrer } from '../login/login.component';
 import styles from './location-picker.scss';
 
+/**
+ * Validates that a returnToUrl is safe to navigate to.
+ * Only same-origin paths under the SPA base are permitted,
+ * preventing open-redirect attacks after login.
+ */
+export function isSafeReturnUrl(url: string): boolean {
```
ibacher, line 23: So I think there's a small flaw here which is that this function assumes that the redirect must be under `${spaBase}`, but I don't see a good reason to enforce that as long as the URL origin and the `window.location.origin` match...

NethmiRodrigo: @omeriinnocent you’ve marked this conversation as resolved, but you haven’t actually implemented this change...

NethmiRodrigo (approved):

denniskigen:
@omeriinnocent thanks for hardening this, the same-origin check is a clean fix. One convention question, mainly for @ibacher: should this keep the `(BREAKING)` prefix?

Looking at the diff, nothing in the `@openmrs/esm-framework` public API changes. `isSafeReturnUrl` is internal to esm-login-app. Same-origin `returnToUrl` flows still behave as before, and the normal relative referrer flows still work. The only thing that goes away is cross-origin redirects after login, including protocol-relative external redirects, which is exactly the open redirect being closed. That reads to me as a security `fix` rather than a breaking change, and `(BREAKING)` would trigger a major version bump that may not be warranted.

The one case where `(BREAKING)` would be right is if a deployment intentionally relied on redirecting to a different origin via `returnToUrl` after login. Do we know of any? If not, I'd lean toward retitling to `(fix)` and flagging it as a security fix in the release notes.

ibacher:
> One convention question, mainly for @ibacher: should this keep the (BREAKING) prefix?

I don't know of any deployments using this functionality so it's non-breaking for everything I know of. However, it does make a previously allowed configured state to be invalid, so even in the absence of this, it's a breaking change (or, in a more limited sense, it should produce a major version bump for the login app).

omeriinnocent:
Thanks @NethmiRodrigo @denniskigen @ibacher for the guidance and support.

PR #1766, opened 2026-05-05
(chore) Add a semi-automated documentation workflow

ibacher:
## Requirements

## Summary

Currently, we use a Husky hook to run documentation generation. This process is somewhat slow which leads to many contributors skipping it, resulting both in the docs getting out of date and weird documentation changes ending up in unrelated PRs. Instead of pushing this on the local dev, the goal here is to add docs to PRs as part of a semi-automated workflow.

The way this intended to work:

1. Contributor submits PR
2. If changes are made to files that may affect docs, a check is failed.
3. Reviewers review as normal.
4. Once PR is approved by a user with write access to the repo, a job will spin-up which will generate the updated docs and add them to the PR.
5. At this point the failing check will be moved to success
6. Future commits to the PR (that do not mess with any of the docs workflows) from this point on will automatically trigger rebuilding the docs

There are some additional controls in place around PRs that touch the workflow files requiring approval from a /dev/5 because, of necessity, many of these jobs run in the scope of the PR and malicious changes to these files could result in secrets leaking.

I've done my best to try and ensure that secrets are never visible in a context where user-supplied code is run, which is why many workflows have a large number of tasks or pass things through artifacts.

There's also a fallback to manually trigger the docs and a /dev/5 can label any PR with `skip-docs` to opt-out of every stage of this process.

## Other

This removes the Husky hook. The script is still available locally and there are some notes in the README about using it.


denniskigen (approved):
Some of the best in class CI documentation I've read, thanks @ibacher.

PR #1771, opened 2026-05-07
(feat) Add matchLocale framework function

ibacher:
## Requirements

## Summary

matchLocale() implements the Lookup algorithm described in RFC 4647 to select the "best matching" locale against a supplied locale list. This is the algorithm the [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) objects use to support locale lookup. It ensures, for example, that if I'm in the `en` locale and there are translations available for `en-GB`, but not plain `en` then `en-GB` is matched and that if I'm in the `en-GB` and `en` is available, `en` is matched.

This implementation has one non-standard addition to the general algorithm which is that the `ht` locale will fallback to `fr-HT`.

While locales are canonicalized for comparison, this method guarantees that it returns matches as the user specified.

### DIFF
--- a/.changeset/ten-impalas-report.md
+++ b/.changeset/ten-impalas-report.md
@@ -1,0 +1,6 @@
+---
+"@openmrs/esm-framework": minor
+"@openmrs/esm-utils": minor
+---
+
+(feat) Add matchLocale framework function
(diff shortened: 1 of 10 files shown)


denniskigen (approved):
LGTM. Should we add a stub for `match-locale` to `mock-jest.tsx`?

ibacher:
> Should we add a stub for match-locale to mock-jest.tsx?

Yes, yes we should

ibacher:
https://github.com/openmrs/openmrs-esm-core/pull/1771/commits/0eeab2f690ad3d26a09287660a84746d04032eb7 :tada: :tada:

NethmiRodrigo (approved):
Thanks @ibacher!

PR #1778, opened 2026-05-19
(fix) Restore lg height for Search component

denniskigen:
## Requirements

## Summary

`<Search size="lg">` has been rendering at `md` height (40px) across O3, so any layout that places it next to a paired lg `Button` ends up with a visible step between the input and the button (e.g. the patient search bar in the orders workspace).

The root cause is in the bundled Carbon: `.cds--search-input` consumes the layout-token clamp (`height: var(--cds-layout-size-height-local)`), but nothing in the bundle bridges the `cds--search--lg` class to `--cds-layout-size-height`. With no provider for the token in the search subtree, the clamp falls through to its `md` fallback.

This patch adds the missing bridge in the styleguide's overrides so the existing clamp resolves to lg (48px). It can be removed once `@carbon/styles` is bumped to a version that wires this up natively (see [carbon-design-system/carbon#21721](https://github.com/carbon-design-system/carbon/pull/21721), which pairs the SCSS refactor with a JSX update that emits `cds--layout--size-${size}` alongside `cds--search--${size}`).

Design reference: [Carbon Search — Style](https://www.carbondesignsystem.com/components/search/style/) (lg = 48px, matching the lg Button it's paired with).

<details>
<summary>DevTools evidence (before fix)</summary>

The `<Search size="lg">` wrapper carries `cds--search--lg` but the bundled Carbon styles only use that class to position the magnifier icon. No rule sets `--cds-layout-size-height` for the search subtree, so the `clamp()` on `.cds--search-input` resolves to its `md` fallback (40px). DevTools shows `var(--cds-layout-size-height-lg)`, `-xs`, `-min`, `-max` struck through (unresolved); only `-md` is defined.

</details>

### Before

CleanShot 2026-05-19 at 12 17 20@2x

### After

CleanShot 2026-05-19 at 12 15 55@2x


NethmiRodrigo (approved):
Thanks @denniskigen! LGTM

PR #1781, opened 2026-05-28
(chore) Workspace2 is not longer experimental; deprecate currentVisit

chibongho:
## Requirements

## Summary

Workspace v2 is no longer experimental; accordingly, we remove the `@experimental` tag.
"Current visit" is not well defined outside of the patient chart. We mark it as `@deprecated` and will remove it in a future release.

### DIFF
--- a/packages/framework/esm-react-utils/src/useVisit.ts
+++ b/packages/framework/esm-react-utils/src/useVisit.ts
@@ -22,7 +22,8 @@ export interface VisitReturnType {
   currentVisit: Visit | null;
 
   /**
-   * @deprecated
+   * @deprecated "current visit" is not well defined outside of the patient chart.
+   * Use `visitContext` in the patient chart instead.
    */
   currentVisitIsRetrospective: boolean;
   isLoading: boolean;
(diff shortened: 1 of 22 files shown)


ibacher (approved):

denniskigen (approved):

packages/framework/esm-react-utils/src/useVisit.ts
```diff
@@ -14,7 +14,16 @@ export interface VisitReturnType {
...
   activeVisit: Visit | null;
+
+  /**
+   * @deprecated "current visit" is not well defined outside of the patient chart.
+   * Use `visitContext` in the patient chart instead.
+   */
   currentVisit: Visit | null;
+
+  /**
+   * @deprecated
```
denniskigen: Thanks for marking these legacy visit fields as deprecated. Could we also give `currentVisitIsRetrospective` the same replacement guidance as `currentVisit`? The generated API docs currently show an empty “Deprecated” section for this field, so consumers do not get a migration cue for the boolean even though it only makes sense alongside the deprecated visit value.

chibongho: I added it, and I think the "Add docs" job ran fine afterwards. I'm not sure why some of the checks are still pending. Probably safe to merge in?

denniskigen:
@chibongho this is why e2e tests were failing: https://github.com/openmrs/openmrs-esm-core/commit/f36d54777376cc8ce804cc10db6b7ff7da498536

denniskigen (approved):

PR #1791, opened 2026-06-05
(fix) Add install-script to bundle-size workflow

NethmiRodrigo:
## Requirements

## Summary

[The `bundle-size.yml` workflow was failing with](https://github.com/openmrs/openmrs-esm-core/actions/runs/27000979662/job/79681621929):

```
Unable to locate executable file: turbo. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable.
```

**Root cause:** `preactjs/compressed-size-action@v3` is a floating tag. On 2026-06-05, the `v3` tag was advanced to a new build that contained a breaking change originally merged in December 2025 ([`feat!: allow to run any script for clean/install/build`](https://github.com/preactjs/compressed-size-action/pull/126)).

This change had two effects:

1. **Auto-install removed:** The action previously auto-detected the package manager from the lockfile and ran `yarn --frozen-lockfile` when a `yarn.lock` was present. The new version removed `getPackageManagerAndInstallScript()` entirely — `install-script` is now fully opt-in and nothing installs if it isn't provided.

2. **Build script execution changed:** The action previously ran the build as `yarn run <build-script>`, which adds `node_modules/.bin` to PATH automatically. The new version executes `build-script` directly via `exec()` with no PATH augmentation, so binaries in `node_modules/.bin` (including `turbo`) are not found even after installing dependencies.

**Fix:** Add `install-script: "yarn install --immutable"` so dependencies are installed, and prefix the build script with `yarn` (`yarn turbo run build --color`) so that `turbo` is resolved through `node_modules/.bin`.

## Screenshots

N/A — CI workflow fix only.


NethmiRodrigo:
The reason we aren’t seeing this issue say in patient-chart is because they use the `v2` tag of `preactjs/compressed-size-action`

denniskigen (approved):
Thanks, @NethmiRodrigo!

PR #1793, opened 2026-06-08
(feat) Prevent the shared SWR cache being torn down when a decorated component unmounts

denniskigen:
## Requirements

## Summary

`openmrsComponentDecorator` wraps every decorated component in its own `<SWRConfig>` that points at a single module-level cache (`swrCache`, added in #1397 to give one global SWR cache regardless of module-federation load order).

SWR ties a provider cache's global state to the lifecycle of the **first** `<SWRConfig>` boundary that initializes it: that boundary's unmount runs `SWRGlobalState.delete(cache)`. Because every decorated component, extension, and workspace mounts its own `<SWRConfig>` over the one shared cache, whichever mounts first owns that deleter. When it unmounts while others are still mounted (routine on a busy page, e.g. navigating the patient chart or closing a workspace), the shared state is removed — the survivors' next render calls `SWRGlobalState.get(cache)`, which now returns `undefined`, and destructuring that throws `TypeError: undefined is not iterable`. The single-spa error boundary then replaces the page with "An error has occurred".

This is a **single-instance lifecycle bug**: it reproduces with one `swr/_internal` instance and no module federation (see the regression test). It is distinct from the separate multi-instance case — a cache registered in one `swr/_internal` instance being read from another — which `singleton: true` is there to prevent.

This is the cause of the intermittent patient-chart e2e failures (a different chart spec fails each run, e.g. `visit-context-navigation`, `drug-orders`, `attachments`, always a chart page timing out because it was replaced by the error boundary).

### Fix

Keep #1397's single shared cache, but **pre-initialize it once at module load** (`initCache(swrCache)`). That transfers ownership of the cache's `SWRGlobalState` from the first `<SWRConfig>` boundary to this (singleton) framework module: every `<SWRConfig>` then takes SWR's "already initialized" path and none of them registers the deleting cleanup, so the shared state is never torn down. This preserves #1397's guarantee of one global cache regardless of load order (and leaves the global `mutate()` behavior unchanged); it does **not** revert to SWR's default cache.

`ComponentDecoratorOptions['swrConfig']` is also tightened to omit `provider` (alongside the already-omitted `fetcher`), so a caller can't accidentally swap in a different cache provider per component. No current caller passes one.

### Tests

Adds `openmrsComponentDecorator.swr-cache.test.tsx`:
- a regression test that mounts two decorated components over the shared cache, unmounts the first (the owner), re-renders the survivor, and asserts it keeps working (fails without the fix, passes with it);
- a test confirming cached data is still shared across separately decorated components.

## Other

This deliberately keeps the single-global-cache design from #1397 (which was introduced to fix data-consistency problems with SWR's default cache under module federation, see #1395 / #774). The only change is to stop binding that cache's lifecycle to a component boundary. Thanks @ibacher for catching that an earlier revision, which dropped the provider, would have reintroduced the #1397 problem.

Because the crash is intermittent (it depends on component mount order), the patient-chart e2e passing once on this PR isn't conclusive by itself; the failing specs can be run repeatedly (this branch vs. `main` as a control) to show the crash rate going to zero.

### DIFF
--- a/.changeset/fix-swr-shared-cache-teardown.md
+++ b/.changeset/fix-swr-shared-cache-teardown.md
@@ -1,0 +1,5 @@
+---
+"@openmrs/esm-react-utils": patch
+---
+
+Prevent the shared SWR cache from being torn down when a decorated component unmounts. Every component decorated with `openmrsComponentDecorator` mounts its own `<SWRConfig>` over one shared cache; SWR deletes that cache's global state when the first such boundary unmounts, crashing still-mounted components ("undefined is not iterable"). The cache is now pre-initialized so the framework module owns its lifecycle.
--- a/packages/framework/esm-react-utils/src/openmrsComponentDecorator.swr-cache.test.tsx
+++ b/packages/framework/esm-react-utils/src/openmrsComponentDecorator.swr-cache.test.tsx
@@ -1,0 +1,102 @@
+import React, { useReducer, useState } from 'react';
+import { describe, expect, it } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import useSWR from 'swr';
+import { openmrsComponentDecorator } from './openmrsComponentDecorator';
+
+// Regression test for the shared-SWR-cache lifecycle crash.
+//
+// Every component decorated by `openmrsComponentDecorator` mounts its own
+// `<SWRConfig>` over a single shared cache. SWR ties that cache's global state
+// to the first `<SWRConfig>` to initialize it, deleting it when that boundary
+// unmounts. So before the fix, unmounting the first-mounted decorated component
+// while another is still mounted wiped the shared state, and the survivor threw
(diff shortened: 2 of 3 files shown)


NethmiRodrigo (approved):
Thanks @denniskigen!

ibacher:
#1397 was added precisely because we ran into issues with SWR's default cache (my recollection of this is that we landed on #1397 because of problems with consistently loading the shared cache via Module Federation). My concern here is that this fixes E2E tests at the cost of bringing back a setup that had known-bad state in production. (See also #1395, #1346, and #774).

It could be that the shift to RSpack has rendered this work-around redundant though. Let me do some digging.

ibacher (approved):

ibacher:
I like the revised approach. The issue here arises when there are multiple `swr/_internal` instances loaded for some reason (even though we set `strictVersion` to `false` and `singleton` to `true` which should basically prevent that from happening).

PR #1795, opened 2026-06-12
(feat) Fail assemble by default if things are obviously broken

ibacher:
## Requirements

## Summary

Way back in 5.0.0, when we added support for `routes.json`, we implemented a _warning_ if a the `dist/routes.json` file couldn't be found as expected. The idea was to give feedback to implementations that they needed to update any custom ESMs to support the new `routes.json` mechanism. However, we've recently run into some issues with the RefApp that could benefit from sanity-checking that we will be able to install and run all specified modules.

This PR repurposes that warning into an error that fails if a specified module doesn't have a `routes.json`. It also fails if the `routes.json` cannot be parsed or if the module entry point cannot be found. In any of these cases, the framework would not be able to use the specified module.

Users can bypass these new strong enforcements by passing `--no-ensure-entrypoints` to the `assemble` command.


packages/tooling/openmrs/src/commands/assemble.ts
NethmiRodrigo: In cases where `sourceDir` will be `’’`, won’t this mean that the fileName would have two trailing slashes?
Claude suggests:

> The fix
> 
> Build the prefix once so the filter and replace can't diverge:

```
const prefix = sourceDir ? `${packageRoot}/${sourceDir}/` : `${packageRoot}/`;
// …
.filter((m) => m.startsWith(prefix))
.map((m) => {
  const fileName = m.replace(prefix, ‘’);
  …
});
```

packages/tooling/openmrs/src/commands/assemble.test.ts
```diff
@@ -457,6 +476,67 @@ describe('runAssemble', () => {
...
+      mockExistsSync.mockImplementation((p: any) => String(p) === '/path/to/config.json');
+      const { logWarn } = await import('../utils');
+
+      await expect(runAssemble(defaultArgs({ ensureEntrypoints: false }))).resolves.not.toThrow();
+
+      expect(logWarn).toHaveBeenCalledWith(expect.stringContaining('Routes file'));
+      expect(logWarn).toHaveBeenCalledWith(expect.stringContaining('Code entrypoint'));
+    });
+  });
+
```
NethmiRodrigo, line 597: Should we add a test that exercises the `sourceDir === '.’` branch?

packages/tooling/openmrs/src/commands/assemble.ts
```diff
@@ -257,16 +273,20 @@ export async function runAssemble(args: AssembleArgs) {
       const [fileName, version] = await extractFiles(tgzBuffer, resolve(args.target, baseDirName));
       const dirName = `${baseDirName}-${version}`;
 
+      // The routes are how the framework ensures that it loads the correct module before any pages
+      // or extensions; if it's missing the app won't load in the browser.
```
denniskigen, line 278: Small accuracy note on this comment, separate from the grammar fix Copilot flagged. "If it's missing the app won't load in the browser" reads a little stronger than what actually happens: a missing `routes.json` means the module's pages and extensions never get registered, but the module's code still loads via the import map and the rest of the app keeps working. The user-facing error just below already says this well ("no pages or extensions for `${esmName}` will be available"). Since you're touching this comment anyway, could we align it with that, maybe something like "if it's missing, none of the module's pages or extensions will be registered"?

ibacher: I actually think the comment, as written, is correct. The only way the code gets loaded is via a call to the framework's `importDynamic()` (this is because we're using a `systemjs-importmap` not a standard import map). `importDynamic()` is only normally called from two places, 1) the i18next "backend" (which is usually called via `useTranslation()` inside the app) or `loadLifecycles()`, which is the technical implementation of how the routes part loads code.

Concretely, on test3, where (currently) the onboarding app is in the import map but not the routes, you should be able to see that there's no attempt made to load any code associated with the onboarding app.

packages/tooling/openmrs/src/commands/assemble.ts
ibacher: Tbh, I already think adding code to handle `.` was a silly thing to do and this would just compound it. The build [already relies](https://github.com/openmrs/openmrs-esm-core/blob/401e227eaf324ce8d2eb6332a411a6b8e5f86067/packages/tooling/webpack-config/src/index.ts#L162) on `dirname(entryModule)` not resolving to `.`. I'm more inclined to just back this whole part of the change out.

ibacher:
I think all the comments here have been dealt with?

denniskigen (approved):
LGTM! Thanks, @ibacher.

PR #1797, opened 2026-06-21
(feat) O3-5767: Redirect to Location header URL on session endpoint calls

dilankavishka:
## Requirements

## Summary

I raised this PR for achieving the goal that when a request is made to the session endpoint and the response includes a `Location` header, the frontend will automatically redirect the user to that location. 

This ensures that any session-based routing or redirect instructions sent from the server (e.g., during login) are respected by the client immediately.

## Screenshots

N/A

## Related Issue

https://issues.openmrs.org/browse/O3-5767

## Other

N/A


dilankavishka:
@jayasanka-sack @janithRS Could you review this PR please?

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -157,6 +157,23 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
   return window.fetch(url, fetchInit as RequestInit).then(async (r) => {
     const response = r as FetchResponse<T>;
     const { redirectAuthFailure, followRedirects } = await getConfig<EsmApiConfigObject>('@openmrs/esm-api');
+
+    if (url === makeUrl(sessionEndpoint) && response.headers?.has?.('location')) {
```
jayasanka-sack: What are the odds of `response.headers` or `response.headers.has` being undefined here?

I try to avoid adding optional chaining unless there's a real possibility of a value being missing. Otherwise it can hide bugs and make it harder to reason about the code. If there's a specific case where this can happen, could you share it?

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -157,6 +157,23 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
...
+      if (location) {
+        if (!response.ok) {
+          clearHistory();
+        }
+        navigate({ to: location });
+
+        if (!response.ok && redirectAuthFailure.resolvePromise) {
+          return Promise.resolve() as unknown as Promise<FetchResponse<T>>;
+        }
+        return new Promise<FetchResponse<T>>(() => {});
```
jayasanka-sack: I'd move this block inside the if (response.ok) block. The use case we're handling here is when the authentication module sends a Location header on the session endpoint, which always happens with a 200 response. The auth module doesn't interfere with unauthenticated responses from the session endpoint since it's a special whitelisted endpoint.

In any case, if the session endpoint returns a non-200 response, it will be handled by the logic outside the if (response.ok) block anyway, so I don't think we need additional logics here. All we need is to redirect to the location header if it is coming from the session endpoint with a 200 code.

dilankavishka: Updated. Moved the logic to `if (response.ok)` block.

packages/framework/esm-api/src/openmrs-fetch.ts
dilankavishka: I initially added the optional chaining because some of the mock Response objects in `openmrs-fetch.test.ts `were incomplete and missing the headers property, which caused the tests to crash. So I went into the test file and fixed the broken mocks by properly defining the `headers` object so they act like real responses.

Am I wrong in here?

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -157,7 +157,16 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
...
     const response = r as FetchResponse<T>;
     const { redirectAuthFailure, followRedirects } = await getConfig<EsmApiConfigObject>('@openmrs/esm-api');
+
     if (response.ok) {
+      if (url === makeUrl(sessionEndpoint) && response.headers.has('location')) {
+        const location = response.headers.get('location');
+        if (location) {
+          navigate({ to: location });
+        }
+        return new Promise<FetchResponse<T>>(() => {});
```
janithRS: I would vote for adding a comment here about why you are returning a promise which would never resolves

dilankavishka: Sure! Let me update the PR

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -157,7 +157,21 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
   return window.fetch(url, fetchInit as RequestInit).then(async (r) => {
     const response = r as FetchResponse<T>;
     const { redirectAuthFailure, followRedirects } = await getConfig<EsmApiConfigObject>('@openmrs/esm-api');
+
     if (response.ok) {
+      if (url === makeUrl(sessionEndpoint) && response.headers.has('location')) {
```
ibacher: 1. This doesn't really feel integrated into the rest of the flow here. In fact, it feels interruptive, like we're stacking functionality instead of making this flow. For example, do we need special-handling for 204s or can we leverage that for both 204s _and_ requests to the session endpoint to handle location?
2. You might've noticed in the places that we already use the `Location` header here, we check against configuration before changing where the user goes. That seems like a pattern we may want to continue.
3. Functionally, this violates my rule that reads for the session endpoint _must_ return the data from the session endpoint. If it was wrong to do something different at the API level, it's also wrong to do that here for the exact same reasons. Again, the fetch response must always return the data from the session API.

ibacher (changes requested):

packages/framework/esm-api/src/openmrs-fetch.ts
dilankavishka: Thank you for reviewing this PR. I noted your points and updated the PR according to the feedback.

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -157,15 +157,19 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
...
-            navigate({ to: location });
-          }
+      const isSessionRedirect = url === makeUrl(sessionEndpoint) && response.headers.has('location');
+      const is204Redirect = response.status === 204 && response.headers.has('location');
+
+      if (followRedirects && (isSessionRedirect || is204Redirect)) {
+        const location = response.headers.get('location');
+        if (location) {
+          navigate({ to: location });
         }
```
janithRS: Can't we simplify this like this?

```
const location = response.headers.get('location');

const shouldRedirect = followRedirects && location && 
((url === makeUrl(sessionEndpoint)) || response.status === 204);

if (shouldRedirect) {
  navigate({ to: location });
}
```

dilankavishka: Wow. That's great!! Thank you so much and I updated the code.

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -157,15 +157,16 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
...
-        if (followRedirects && response.headers.has('location')) {
-          const location = response.headers.get('location');
-          if (location) {
-            navigate({ to: location });
-          }
-        }
+      const location = response.headers.get('location');
+      const isRedirect = followRedirects && location && (url === makeUrl(sessionEndpoint) || response.status === 204);
+
+      if (isRedirect) {
```
jayasanka-sack: I'd go with `shouldRedirect`.

"if we should redirect, navigate."

dilankavishka: Updated

jayasanka-sack (approved):
Thanks for working on this, @dilankavishka!

PR #1800, opened 2026-06-24
(feat) O3-5763: Add shared VisitSummary component to esm-styleguide

UjjawalPrabhat:
## Requirements

## Summary
Adds a shared, props-based `VisitSummary` to `@openmrs/esm-styleguide` (exported via `@openmrs/esm-framework`) so apps outside the patient chart — e.g. Service Queues, can render the standard visit summary without depending on `@openmrs/esm-patient-common-lib`. This is the base the [patient-chart](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3414) and [service-queues](https://github.com/openmrs/openmrs-esm-patient-management/pull/2598) PRs build on.

## Related Issue
[O3-5763](https://issues.openmrs.org/browse/O3-5763)

## Other
Merge order: **core -> patient-chart <-> queue**.

[O3-5763]: https://openmrs.atlassian.net/browse/O3-5763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/framework/esm-styleguide/src/visit-summary/medications-summary.component.tsx
+++ b/packages/framework/esm-styleguide/src/visit-summary/medications-summary.component.tsx
@@ -1,0 +1,162 @@
+import React from 'react';
+import { capitalize } from 'lodash-es';
+import { Tag, Tooltip } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
+import { formatDate } from '@openmrs/esm-utils';
+import { EmptyCard } from '../empty-card';
+import { type OrderItem } from './types';
+import styles from './visit-detail-overview.module.scss';
+
+interface MedicationSummaryProps {
+  medications: Array<OrderItem>;
+  drugOrderTypeUUID: string;
+}
+
+const MedicationSummary: React.FC<MedicationSummaryProps> = ({ medications, drugOrderTypeUUID }) => {
+  const { t } = useTranslation();
+
+  const isDiscontinued = (order: OrderItem['order']) => {
+    if (!order?.dateStopped) {
+      return false;
+    }
+
+    return new Date(order.dateStopped) < new Date();
+  };
+
+  const drugOrders = medications?.filter((medication) => {
+    return medication?.order?.orderType?.uuid === drugOrderTypeUUID;
+  });
+
+  if (drugOrders.length === 0) {
+    return (
+      <EmptyCard displayText={t('medications__lower', 'medications')} headerTitle={t('medications', 'Medications')} />
+    );
+  }
+
+  return (
+    <div className={styles.medicationRecord}>
+      {drugOrders.map(
+        (medication, index) =>
+          (medication?.order?.dose != null || medication?.order?.dosingInstructions) && (
+            <React.Fragment key={index}>
+              <div className={styles.medicationContainer}>
+                <div>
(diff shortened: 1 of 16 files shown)


packages/framework/esm-styleguide/src/visit-summary/medications-summary.component.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { type OrderItem } from './types';
+import styles from './visit-detail-overview.module.scss';
+
+interface MedicationSummaryProps {
+  medications: Array<OrderItem>;
+  drugOrderTypeUUID: string;
+}
+
+const MedicationSummary: React.FC<MedicationSummaryProps> = ({ medications, drugOrderTypeUUID }) => {
+  const { t } = useTranslation();
```
ibacher, line 16: `useTranslation()` won't work inside the styleguide and alternative setups are needed for this to appropriately work. This is why I was suggesting making this stuff an _extension_ provided by the patient-chart-app.

ibacher: Components tend to need to be heavily re-written to work in the styleguide or we need to decide that it's fine for all the translation keys used here to live in the core translations and that feels like something we're not ready to do.

packages/framework/esm-styleguide/src/visit-summary/medications-summary.component.tsx
```diff
@@ -0,0 +1,162 @@
+import React from 'react';
+import { capitalize } from 'lodash-es';
+import { Tag, Tooltip } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
```
ibacher, line 4: Please see our [docs on import order](https://o3-docs.openmrs.org/en-US/docs/coding-conventions/code-organization/)

packages/framework/esm-styleguide/src/visit-summary/types.ts
ibacher, line 1: Types definitions like this do not belong in the styleguide, but maybe in esm-api directory.

packages/framework/esm-styleguide/src/visit-summary/useEncountersByVisit.ts
ibacher, line 1: Data-fetching hooks are also not allowed in the styleguide.

ibacher (changes requested):
@UjjawalPrabhat Pointing to some things here around the conventions of how things go into the styleguide. I don't think this is the right fit for things in this format. In particular (as noted below):

1. `useTranslation()` does not work in the styleguide the way it works in apps. Consequently, this renders every component migrated effectively untranslatable.
2. We never should have data fetching hooks inside the styleguide. External data that needs to be fed into styleguide components must be passed as props from the appropriate extensions.
3. Type definitions and the like never live in the styleguide either.

Porting code to the styleguide isn't just about moving it. It needs to retain the conventions of the existing styleguide code. If we're just "moving" stuff, then it should generally be packaged up as extensions in some application. Putting things in the styleguide requires some redesign and rethinking of organization, so while ultimately, it would make sense to have a component to render a "medication summary", the types of components in this PR are not how it should be done.

PR #1807, opened 2026-07-03
(fix) Resolve workspace2 lifecycles by workspace uuid instead of stack index

denniskigen:
## Requirements

## Summary

`ActiveWorkspaceWindow` loads workspace lifecycles asynchronously and stored them in an array that the render reads by stack index. When the stack changes shape in quick succession, for example a child workspace closes and a different one opens at the same position, the newly opened workspace can be handed the previous workspace's lifecycle from the stale array. Since the parcel mounts whatever config it first receives and ignores later config changes, the wrong component then renders permanently, with the new workspace's definition props. Every workspaceProp the component destructures is undefined at that point.

We hit this in the ward app's "Add patient to ward" flow: the start visit form closes on save and immediately launches the admission workspace, which then rendered a blank visit form (the closed workspace's component) instead of the admission UI, firing requests like `GET /ws/rest/v1/patient/undefined` from hooks that received undefined props. Whether it reproduced depended on whether the close and the launch landed in the same React commit.

This change keys the loaded lifecycles by the opened workspace's uuid, so a stale entry is unreachable by construction: a newly opened workspace has a fresh uuid and shows the loading placeholder until its own lifecycle resolves. The load effect also ignores resolutions from superseded runs so an out of order resolution cannot clobber the current map.

The included test replaces the workspace at a stack position while the replacement's lifecycle is still loading and asserts the stale component is not mounted. It fails against the previous index-based lookup and passes with this change.


ibacher:
Can we just verify that either this `uuid` property is always available or that this continues to work correctly when it's not.

denniskigen:
`uuid` is a required property on the `OpenedWorkspace` interface (it's typed as `uuid: string` in esm-extensions):

https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-extensions/src/workspaces2.ts#L8-L14

Every opened workspace is created through a single factory, `newOpenedWorkspace()` in `workspace2.ts`, and that factory always assigns a fresh `uuid` via `uuidV4()`. So there is no code path where a workspace ends up in the stack without a uuid.

ibacher (approved):

PR #1814, opened 2026-07-14
(fix) Honor the invalid prop in OpenmrsDatePicker mocks

denniskigen:
## Requirements

## Summary

The `OpenmrsDatePicker` and `OpenmrsDateRangePicker` mocks in `mock.tsx` and `mock-jest.tsx` only render `invalidText` when the `isInvalid` prop is set. The real components accept both `invalid` and `isInvalid` (they compute `invalid || isInvalid`), and consumers across the ecosystem typically pass `invalid`. As a result, the error text never renders in consumer tests, making it impossible to assert on date picker validation error messages. This came up in openmrs/openmrs-esm-patient-chart#3035, where a test could not assert on a date picker validation error.

This PR updates both mocks to render the error text when either prop is truthy, mirroring the real components. The date range picker mock in each file was rewrapped by Prettier because the parameter list exceeded the maximum line width after adding the new prop.

The storybook mocks do not stub the date pickers, so no changes were needed there.

### DIFF
--- a/packages/framework/esm-framework/mock-jest.tsx
+++ b/packages/framework/esm-framework/mock-jest.tsx
@@ -111,7 +111,7 @@ export const navigateAndLaunchWorkspace = jest.fn();
 export const useWorkspaces = jest.fn();
 export const useWorkspace2Context = jest.fn();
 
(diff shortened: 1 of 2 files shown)


ibacher (approved):
Good call on this!

PR #1815, opened 2026-07-14
(feat) Add a distribution-controlled version property

ibacher:
## Requirements

## Summary

This adds an implementation-settable version property that can be used to display the application version. This version is set at the time that the `assemble` command is run by passing in the `--openmrs-version` property. While SemVer are encouraged, you can actually use any string that makes sense.

By default, this version is displayed on the login screen and primary navigation. When not set, the version will read "Prerelease" (this is translatable). For no version display, set `--openmrs-version ''`.

## Screenshots

Screenshot of login screen displaying version

Screenshot of primary navigation displaying version

## Other

I've rejected at least one attempt at doing something similar using the configuration system. Configuration files just aren't the right place for this. Realistically, we should probably have a global property to control this, but that requires some coordination with the backend team.

The version number is generally available as the `window.openmrsVersion` property. This may be slightly confusing with the `spaVersion` property. The `spaVersion` property is the version of the app shell / root container and is mostly intended for technical audiences / debugging. `openmrsVersion` is intended for user display.


packages/framework/esm-globals/src/type-utils.ts
```diff
@@ -50,10 +50,18 @@ export function isOpenmrsAppRoutes(routes: OpenmrsAppRoutes | unknown): routes i
...
 export function isOpenmrsRoutes(routes: OpenmrsRoutes | unknown): routes is OpenmrsRoutes {
-  if (routes && typeof routes === 'object') {
+  if (
+    routes &&
+    typeof routes === 'object' &&
+    'routes' in routes &&
+    routes['routes'] &&
+    typeof routes['routes'] === 'object'
+  ) {
     const maybeRoutes = routes as OpenmrsRoutes;
```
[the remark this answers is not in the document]
ibacher: Reasonable, but the App Shell and the version of the CLI used to build the assemble command need to be in sync already, so this should be a non-issue.

packages/tooling/openmrs/src/utils/importmap.ts
```diff
@@ -299,7 +303,10 @@ export async function mergeImportmapAndRoutes(
 
     routesDecl.value = JSON.stringify({
       ...routes,
-      ...additionalRoutes,
+      routes: {
+        ...routes.routes,
+        ...additionalRoutes,
+      },
     });
```
denniskigen, line 383: If the backend serves a legacy flat registry, there is no nested `routes` property here, so all the remote modules stay at the top level where the app shell now ignores them. Normalizing before the merge keeps `develop` working against servers assembled with older CLIs:

```suggestion
    const normalizedRoutes =
      routes && typeof routes === 'object' && 'routes' in routes ? routes : { routes };

    routesDecl.value = JSON.stringify({
      ...normalizedRoutes,
      routes: {
        ...normalizedRoutes.routes,
        ...additionalRoutes,
      },
    });
```

packages/framework/esm-dynamic-loading/src/route-maps.ts
```diff
@@ -43,10 +43,14 @@ async function readBaseMap(): Promise<OpenmrsRoutes> {
...
-  const merged: OpenmrsRoutes = {};
+  const merged: OpenmrsRoutes = { routes: {} };
   for (const map of maps) {
     if (map && typeof map === 'object') {
-      Object.assign(merged, map);
+      Object.assign(merged.routes, map.routes);
+
+      if (!merged.version && map.version) {
+        merged.version = map.version;
+      }
```
denniskigen: Because `''` is falsy, `mergeRouteMaps` does not copy the empty string written by `assemble --openmrs-version ''`. The merged map therefore has `version === undefined`, and the shell displays "Prerelease" instead of hiding the version. This change preserves the documented opt-out:

```suggestion
      if (typeof merged.version === 'undefined' && typeof map.version !== 'undefined') {
        merged.version = map.version;
      }
```

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -114,16 +118,39 @@ function connectivityChanged() {
...
+  } else if (typeof initialRouteMap.version === 'undefined') {
+    Object.defineProperty(window, 'openmrsVersion', {
+      value:
+        window.spaVersion === 'local' ? getCoreTranslation('localVersion') : getCoreTranslation('prereleaseVersion'),
+      writable: false,
+      configurable: false,
+    });
+  } else {
+    Object.defineProperty(window, 'openmrsVersion', {
+      value: undefined,
```
denniskigen: Could we preserve `initialRouteMap.version` here instead of assigning `undefined`? The public `Window.openmrsVersion` type declares this value as a string, and both framework mocks preserve the empty string. Keeping `''` still renders no text and keeps production, the mocks, and the public type consistent.

packages/framework/esm-globals/src/types.ts
```diff
@@ -45,6 +45,10 @@ declare global {
      * The build number of the app shell. Set when the app shell is built by webpack.
      */
     spaVersion: string;
+    /**
+     * While called "openmrsVersion", this is intended to be an application-level version for the frontend.
+     */
+    openmrsVersion: string;
```
denniskigen: One naming question before this ships: elsewhere in the ecosystem, "OpenMRS version" means the platform version, such as `requireOpenmrsVersion` in the add-on index, `require_version` in `config.xml`, and the `${openMRSVersion}` Maven property. This value represents the frontend application version of the current distribution. Would `applicationVersion` be clearer? In support contexts, "What OpenMRS version are you running?" usually asks for the platform version, while this global would expose the distribution version instead. I see the PR description already weighs this against `spaVersion`, but the platform collision feels like the riskier one.

packages/framework/esm-globals/src/type-utils.ts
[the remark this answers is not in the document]
denniskigen: That is true for assembled distros, but `openmrs develop` works differently. It serves the shell bundled with the local CLI and fetches the routes registry from the remote backend, so they may be produced by different CLI versions. dev3 and o3.openmrs.org both serve the flat format today. Once a frontend repo resolves this CLI version, merging a local app leaves the remote modules at the top level. The new shell reads only the nested `routes` property, so it registers the local app but silently drops the remote distro routes. I suggested normalizing the legacy format in `mergeImportmapAndRoutes`.

packages/framework/esm-globals/src/types.ts
ibacher: That's a reasonable point. I'm open to suggestions on what to call this...

packages/tooling/openmrs/src/utils/importmap.ts
ibacher: Hmmm... This actually points to a deeper flaw in this PR, but I don't want to change the runtime code like this without good reason. In the case you outline, the actual problem here is not "what the backend serves" because the "routes.registry.json" is, in the normal case, proxied via the express server. The issue is that the code to read the remote import map hasn't been updated, and that's the point where it makes sense to bridge between these versions, I think.

packages/framework/esm-globals/src/types.ts
denniskigen: I'd go with `applicationVersion`. It matches the docstring's own description ("an application-level version for the frontend"), so the caveat sentence can go away, and `window.applicationVersion` next to `window.spaVersion` makes the split self-explanatory: the application as assembled vs. the app shell build. The CLI flag would become `--application-version`. If we'd rather emphasize who sets it, `distributionVersion` also works, since it's whatever the distro passed at assemble time.

packages/framework/esm-dynamic-loading/src/route-maps.ts
ibacher: Good catch. Yeah, this code is from before I landed on that.

packages/framework/esm-globals/src/types.ts
ibacher: Done.

ibacher:
@denniskigen I think I've addressed all the points in your review?

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -114,16 +118,39 @@ function connectivityChanged() {
...
+  if (initialRouteMap.version && initialRouteMap.version.length > 0) {
+    Object.defineProperty(window, 'applicationVersion', {
+      value: initialRouteMap.version,
+      writable: false,
+      configurable: false,
+    });
+  } else if (initialRouteMap.version === undefined) {
+    Object.defineProperty(window, 'applicationVersion', {
+      value:
+        window.spaVersion === 'local' ? getCoreTranslation('localVersion') : getCoreTranslation('prereleaseVersion'),
```
denniskigen, line 137: One behavior question before this ships: `undefined` only tells us that no application version was configured, but this branch presents it as “Prerelease.” The Reference Application currently runs `openmrs assemble` without `--application-version`, so adopting this version would put “Prerelease” on its production login screen unless its build is also changed. Would it be safer for the absent case to display nothing, while distributions that want a development label set one explicitly? We could keep “Local Build,” since `spaVersion === 'local'` directly identifies a local shell. This would also remove the empty-string opt-out and special case.

CleanShot 2026-07-15 at 21 38 04@2x

packages/tooling/openmrs/src/utils/importmap.ts
```diff
@@ -79,7 +79,64 @@ async function fetchRemoteRoutes(fetchUrl: string) {
...
+function normalizeRoutesJson(content: string): string | null {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(content);
+  } catch {
+    return null;
+  }
+
+  const normalized = JSON.stringify(normalizeRoutesRegistry(parsed));
+  return checkRoutesJson(normalized) ? normalized : null;
```
denniskigen, line 144: Nit: `normalizeRoutesRegistry` converts JSON primitives, `null`, and arrays to `{ routes: {} }`, so a routes file containing `"a string"` or `5` is accepted as an empty registry. Unparseable JSON takes the warning path, but these invalid registry shapes are served silently. Could we reject non-object values before normalizing?

```suggestion
  if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
    return null;
  }

  const normalized = JSON.stringify(normalizeRoutesRegistry(parsed));
  return checkRoutesJson(normalized) ? normalized : null;
```

packages/framework/esm-dynamic-loading/src/route-maps.test.ts
```diff
@@ -276,7 +277,101 @@ describe('route-maps', () => {
...
+        script.type = 'openmrs-routes';
+        script.textContent = JSON.stringify(registry);
+        document.head.appendChild(script);
+      }
+
+      const { setupRouteMapOverrides, getCurrentRouteMap } = await import('./route-maps');
+      await setupRouteMapOverrides();
+
+      const map = await getCurrentRouteMap();
+      expect(map.version).toBe('2.0.0');
```
denniskigen, line 350: Could we add an empty-string case? This is the documented opt-out and the behavior that previously regressed because `''` is falsy. A focused test would ensure merging preserves it:

```ts
it('preserves an empty top-level version so the shell can suppress the display', async () => {
  (window as any).spaEnv = 'production';
  vi.resetModules();

  document.querySelectorAll("script[type='openmrs-routes']").forEach((el) => el.remove());
  const script = document.createElement('script');
  script.type = 'openmrs-routes';
  script.textContent = JSON.stringify({ version: '', routes: { '@openmrs/esm-foo': { pages: [] } } });
  document.head.appendChild(script);

  const { setupRouteMapOverrides, getCurrentRouteMap } = await import('./route-maps');
  await setupRouteMapOverrides();

  const map = await getCurrentRouteMap();
  expect(map.version).toBe('');
});
```

packages/shell/esm-app-shell/src/run.ts
ibacher: Yeah. I was using “Prerelease” as a stand-in for “unversione.” That said “Prerelease” is actually what I intended to show on dev3, but I’d welcome other suggestions. Do you think we should treat `undefined` as just bypassing this feature?

packages/tooling/openmrs/src/utils/importmap.ts
ibacher: What about just warning for those as well?

packages/shell/esm-app-shell/src/run.ts
denniskigen: Yep

packages/shell/esm-app-shell/src/run.ts
```diff
@@ -114,16 +118,40 @@ function connectivityChanged() {
...
-    })
-    .then(() => start());
+  const applicationVersion = initialRouteMap.version;
+  if (applicationVersion === undefined || applicationVersion === null || applicationVersion.trim().length === 0) {
+    Object.defineProperty(window, 'applicationVersion', {
+      value: undefined,
+      writable: false,
+      configurable: false,
+    });
+  } else if (applicationVersion === 'prelease') {
```
denniskigen: Thanks for making the absent-version case opt in. Should this sentinel be `prerelease`? The new assemble test writes `prerelease`, but this condition checks `prelease`. Consequently, `--application-version prerelease` displays the raw lowercase value and never reaches the translated "Prerelease" or "Local Build" branch. Could we correct the spelling and add a focused shell test?

packages/framework/esm-framework/mock.tsx
```diff
@@ -27,6 +27,7 @@ export function setupPaths(config: any) {
   window.spaBase = config.spaPath;
   window.spaEnv = config.env || 'production';
   window.spaVersion = process.env.BUILD_VERSION ?? 'local';
+  window.applicationVersion = config.applicationVersion ?? 'local';
```
denniskigen: Could we leave `window.applicationVersion` undefined when `config.applicationVersion` is absent? The runtime now treats an omitted version as disabling the feature, while both framework mocks default it to `"local"`. This means consumers see a version in tests when the real shell renders none.

packages/shell/esm-app-shell/src/run.ts
ibacher: Typo. Think I fixed that.

ibacher:
@denniskigen I think I did everything except the test you asked for. `run.ts` needs to be refactored to be testable, but I don't want to do that here.

denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1818, opened 2026-07-21
(fix) O3-5816: Stop LocationPicker crashing when no login locations are configured

suubi-joshua:
## Requirements

## Summary
On installs with zero login locations, the location picker crashes with `TypeError: can't access property "id", n is undefined`
[Screencast from 14-07-26 20:08:20.webm](https://github.com/user-attachments/assets/fc741d08-f76b-4e82-bd35-9c0580b50347)

## Description
I think the root cause was a `useEffect` in `location-picker-view.component.tsx` called `changeLocation()` with no arguments whenever `locationCount === 0`, passing undefined into `setSessionLocation`. This ran automatically on mount, independent of user interaction.

In this PR I have removed the `locationCount === 0` auto-redirect branch from the `useEffect` entirely, there's no valid location to redirect to, so the fix is to not attempt it, not to guard it more carefully
Added a locationCount > 0 guard to the single-location auto-select branch
Empty state (location-picker-view.component.tsx):

Shows "No login locations configured" with a message to contact the system administrator when `locationCount === 0`
Hides the welcome heading, `LocationPicker`, `checkbox`, and Confirm button

Testing
[Screencast from 21-07-26 18:31:45.webm](https://github.com/user-attachments/assets/c9759d20-6013-4087-921e-f6dd8fd8ec18)

## Screenshots
image

## Related Issue

https://issues.openmrs.org/browse/O3-5816

### DIFF
--- a/packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
+++ b/packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
@@ -119,11 +119,12 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
     if (isUpdateFlow) {
       return;
     }
-    if (defaultLocation && !isSubmitting && !hasNoLocations) {
+    if (isLoadingLocationCount) return;
+    if (defaultLocation && !isSubmitting) {
       setActiveLocation(defaultLocation);
       changeLocation(defaultLocation, true);
     }
-  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow, hasNoLocations]);
+  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow, isLoadingLocationCount]);
 
   const handleSubmit = useCallback(
     (evt: React.FormEvent<HTMLFormElement>) => {
--- a/packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
+++ b/packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
@@ -119,12 +119,11 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
     if (isUpdateFlow) {
       return;
     }
-    if (isLoadingLocationCount) return;
     if (defaultLocation && !isSubmitting) {
       setActiveLocation(defaultLocation);
       changeLocation(defaultLocation, true);
     }
-  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow, isLoadingLocationCount]);
+  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow]);
 
   const handleSubmit = useCallback(
     (evt: React.FormEvent<HTMLFormElement>) => {
--- a/packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
+++ b/packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
@@ -10,6 +10,7 @@ import {
   useConfig,
   useConnectivity,
   useSession,
+  WarningIcon,
 } from '@openmrs/esm-framework';
 import { useDefaultLocation, useLocationCount } from './location-picker.resource';
 import type { ConfigSchema } from '../config-schema';
@@ -60,6 +61,8 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
     [user],
   );
 
+  const hasNoLocations = !isLoadingLocationCount && locationCount === 0;
+
   const [activeLocation, setActiveLocation] = useState(() => {
     if (currentLocationUuid && hideWelcomeMessage) {
       return currentLocationUuid;
@@ -98,13 +101,11 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
     [state?.referrer, config.links.loginSuccess, updateDefaultLocation, searchParams],
   );
 
-  // Handle cases where the location picker is disabled, there is only one location, or there are no locations.
+  // Handle cases where the location picker is disabled or there is only one location.
   useEffect(() => {
     if (isLoadingLocationCount) return;
 
-    if (locationCount === 0) {
-      changeLocation();
-    } else if (locationCount === 1 || !chooseLocation.enabled) {
+    if (locationCount === 1 || (!chooseLocation.enabled && locationCount > 0)) {
       if (firstLocation?.resource?.id) {
         changeLocation(firstLocation.resource.id, true);
       } else {
@@ -118,11 +119,11 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
     if (isUpdateFlow) {
       return;
     }
-    if (defaultLocation && !isSubmitting) {
+    if (defaultLocation && !isSubmitting && !hasNoLocations) {
       setActiveLocation(defaultLocation);
       changeLocation(defaultLocation, true);
     }
-  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow]);
+  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow, hasNoLocations]);
 
   const handleSubmit = useCallback(
     (evt: React.FormEvent<HTMLFormElement>) => {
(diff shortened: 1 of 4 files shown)


suubi-joshua:
@claude review

dkayiwa:
@suubi-joshua for claude to review, you need to first copy over this file: https://github.com/openmrs/openmrs-module-authentication/blob/main/.github/workflows/claude-pr-review.yml

suubi-joshua:
> @suubi-joshua for claude to review, you need to first copy over this file: https://github.com/openmrs/openmrs-module-authentication/blob/main/.github/workflows/claude-pr-review.yml

Ohhh okay

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -1,6 +1,9 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import '@testing-library/jest-dom/vitest';
-import { screen, waitFor } from '@testing-library/react';
+import React from 'react';
```
ibacher: This shouldn't, strictly, be necessary here.

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -464,3 +467,137 @@ describe('returnToUrl open-redirect protection', () => {
     });
   });
 });
+
+const mockZeroLocationsResponse: FetchResponse<unknown> = {
```
ibacher: This seems more correct:

```suggestion
const mockZeroLocationsResponse: FetchResponse<fhir.Bundle> = {
```

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -464,3 +467,137 @@ describe('returnToUrl open-redirect protection', () => {
...
+    mockUseConfig.mockReturnValue(mockConfig);
+    mockSetSessionLocation.mockResolvedValue(undefined);
+    mockNavigate.mockClear();
+
+    mockOpenmrsFetch.mockImplementation(async (url) => {
+      if (String(url).includes('_count=1')) {
+        return mockZeroLocationsResponse;
+      }
+      return mockLoginLocations as FetchResponse<unknown>;
+    });
```
ibacher: This doesn't feel like something that should be setup in a `beforeEach()`. This feels like each test should set this mock appropriately and then we don't need to abuse the `?_count=1` to determine which branch to take here.

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -464,3 +467,137 @@ describe('returnToUrl open-redirect protection', () => {
...
+  },
+} as FetchResponse<unknown>;
+
+const regularUser: LoggedInUser = {
+  display: 'Regular User',
+  uuid: 'user-uuid',
+  userProperties: {},
+} as LoggedInUser;
+
+describe('Zero login locations', () => {
```
ibacher: This should still be embedded in the initial describe. We're not working on a different component, just a sub-feature of the component

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -464,3 +467,137 @@ describe('returnToUrl open-redirect protection', () => {
...
+    });
+
+    expect(
+      screen.getByText(
+        'This installation has no login locations configured. Please contact your system administrator.',
+      ),
+    ).toBeInTheDocument();
+  });
+
+  it('does not render the confirm button when no locations exist', async () => {
```
ibacher: It feels a little weird to have multiple tests that effectively run the same machinery and then just differ by `expect()`. I know some guides suggest that kind of thing, but usually that's too fine-grained, especially when the `expect()` is very light-weight compared to the text. This also puts us in the weird position of asserting both "the replacement widget is rendered" AND "the widget it replaces is not rendered."

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -464,3 +467,137 @@ describe('returnToUrl open-redirect protection', () => {
...
+    renderWithFreshCache();
+
+    await waitFor(() => {
+      expect(screen.getByText('No login locations configured')).toBeInTheDocument();
+    });
+
+    expect(mockSetSessionLocation).not.toHaveBeenCalled();
+  });
+
+  it('does not auto-login with saved default when no login locations exist', async () => {
```
ibacher: I'm actually not sure this test validates something we want... I sort of get the impulse, but if the backend doesn't _reject_ the choice, is there a good reason to do so here?

suubi-joshua: You're right, if the backend accepts the UUID, the frontend shouldn't second-guess it. The guard was there as a extra safety check since a saved default can outlive its Login Location tag, but the backend is the authority on validity.

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
suubi-joshua: I have JSX in `renderWithFreshCache` so removing this might fail the test

ibacher: Ah... fair. I keep hoping we've fixed tthis.

packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
```diff
@@ -118,11 +119,12 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
     if (isUpdateFlow) {
       return;
     }
+    if (isLoadingLocationCount) return;
     if (defaultLocation && !isSubmitting) {
       setActiveLocation(defaultLocation);
       changeLocation(defaultLocation, true);
     }
-  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow]);
+  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow, isLoadingLocationCount]);
```
denniskigen: Thanks for preserving the backend-authoritative saved-location behavior here. Could we remove this loading guard? `defaultLocation` has already been independently validated, and this effect does not use `locationCount`. As written, a slow or pending count request now blocks automatic login for users with a valid saved location, which differs from `main`.

```suggestion
    if (defaultLocation && !isSubmitting) {
      setActiveLocation(defaultLocation);
      changeLocation(defaultLocation, true);
    }
  }, [changeLocation, isSubmitting, defaultLocation, isUpdateFlow]);
```

denniskigen (dismissed):

packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
suubi-joshua: You're right. I added the guard to prevent a race condition where `defaultLocation` fires before count resolves to zero, but I think the dependency is unnecessary. I'll remove it thanks.

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
```diff
@@ -345,6 +363,44 @@ describe('LocationPickerView', () => {
...
+    it('shows an empty state and hides the picker, checkbox, and confirm button when no login locations exist', async () => {
+      mockUseSession.mockReturnValue({
+        user: regularUser,
+      } as Session);
+
+      mockOpenmrsFetch.mockImplementation(async (url) =>
+        String(url).includes('_count=1') ? mockZeroLocationsResponse : (mockLoginLocations as FetchResponse<unknown>),
+      );
+
+      renderWithFreshCache();
```
denniskigen: `useLocationCount` is the only thing that fetches in this test, so the mock doesn't need to branch on the URL. The outer `beforeEach()` already provides a session too.

```suggestion
      mockOpenmrsFetch.mockResolvedValue(mockZeroLocationsResponse);

      renderWithFreshCache();
```

`regularUser` can then come out as well. I ran this and the suite still passes, and it still fails against `main`.

packages/apps/esm-login-app/src/location-picker/location-picker.scss
```diff
@@ -92,3 +92,29 @@
   margin-inline-start: -(layout.$spacing-06);
   margin-inline-end: -(layout.$spacing-06);
 }
+
+.emptyStateContainer {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  margin-top: layout.$spacing-03;
+  padding: layout.$spacing-06;
```
denniskigen:
```suggestion
```

`.paddedContainer` already pads this on all four sides, so the inner `padding` doubles it. Together with the margin below, dropping these takes the card from 234px to 162px tall. Text wrapping is unaffected, it's purely the vertical space.

packages/apps/esm-login-app/src/location-picker/location-picker.scss
```diff
@@ -92,3 +92,29 @@
...
+.emptyStateTitle {
+  @extend .bodyLong01;
+  font-weight: 600;
+  margin-bottom: layout.$spacing-03;
+}
+
+.emptyStateMessage {
+  @extend .bodyLong01;
+  color: $color-gray-70;
+  margin-bottom: layout.$spacing-05;
```
denniskigen:
```suggestion
```

Ditto. The message is the last thing in the empty state, so this margin only pads the bottom of the card.

packages/apps/esm-login-app/src/location-picker/location-picker-view.component.tsx
```diff
@@ -142,43 +143,62 @@ const LocationPickerView: React.FC<LocationPickerProps> = ({ hideWelcomeMessage,
...
-              {t('welcome', 'Welcome')} {currentUser}
-            </p>
-            <p className={styles.welcomeMessage}>
-              {t(
-                'selectYourLocation',
-                'Select your location from the list below. Use the search bar to find your location.',
-              )}
-            </p>
+            {hasNoLocations ? (
+              <div className={styles.emptyStateContainer}>
```
denniskigen: The empty state swaps in after the count resolves, so screen readers currently get no announcement. `role="status"` covers it:

```suggestion
              <div className={styles.emptyStateContainer} role="status">
```

Nit, ignorable: the two adjacent `{!hasNoLocations && ...}` blocks below could collapse into one fragment.

denniskigen (commented):
This is very close, @suubi-joshua. I confirmed the crash on `main`: with zero locations the effect calls `setSessionLocation(undefined)`, and your test fails on `main` and passes here, so it pins the regression properly. A few small things below.

packages/apps/esm-login-app/src/location-picker/location-picker.test.tsx
denniskigen: Two notes from trying these out.

`renderWithFreshCache` does need to stay. Swapping in `renderWithRouter` fails the test: `useSwrImmutable` keys on the URL and never revalidates, so the earlier tests leave a non-zero count cached for the same count URL and the empty state never renders. Worth a one-line comment saying so, otherwise someone deletes it later.

And the type annotation doesn't compile:

```
error TS2740: Type '{ data: { ... } }' is missing the following properties from
type 'FetchResponse<Bundle>': headers, ok, redirected, status, and 12 more.
```

So the `as` cast is right here, and it matches how the other fixtures in this file are written.

denniskigen (approved):
LGTM! Thanks for the careful iteration on this, @suubi-joshua. Every round of feedback got addressed exactly, and the regression test still fails on `main` and passes here. The full suite passes locally too. Nice work seeing this one through.

One non-blocking observation for the record, not for this PR: a user parked on the empty state has no logout or back affordance (same as the normal picker), so on a real zero-location install they stay there until an admin fixes the location tags. Something for design to keep in mind if this screen ever gets revisited.

denniskigen (approved):

suubi-joshua:
Thanks @denniskigen for the reviews and feedback as well. Yes i share the same concern with you, it seems the admin is the hero at that point. But hopefully this will suffice for now.

PR #1819, opened 2026-07-22
(feat) O3-5800: Add a two-factor authentication page and TOTP enrollment modal

dilankavishka:
## Requirements

## Summary
In this PR I've included the implementation of the new Two-Factor Authentication page and the TOTP modal for enrollment and verification. When the user clicks the Two-Factor Authentication button in the My Account user panel, they are redirected to the Two-Factor Authentication page.

The feature is gated behind the `twoFactorAuth.enabled` config property in `@openmrs/esm-login-app` and is disabled by default. Enrollment and verification use the `/ws/rest/v1/auth/totp/enrollment` endpoints provided by the [authentication module](https://github.com/openmrs/openmrs-module-authentication).

This PR also adds the `TwoFactorAuthentication` and `MobileCheck` icons and the `DigitalTrust` pictogram to the styleguide.

## Screenshots
https://github.com/user-attachments/assets/b96364c3-6cd0-43df-b8b5-7e2c54d120d3

Note: this recording predates the latest changes to the enrollment modal.

## Related Issue
https://issues.openmrs.org/browse/O3-5800

## Other
**Note**: After the user is enrolled, the `Set up` button changes to a disabled `Remove` button. The removal flow will be implemented in a follow-up PR since the remove endpoint is not implemented in the backend yet.


packages/apps/esm-login-app/src/routes.json
```diff
@@ -52,12 +58,23 @@
       "online": true,
       "offline": true,
       "order": 1
+    },
+    {
+      "name": "two-factor-authentication",
```
denniskigen, line 63: Thanks for wiring this into the account menu. Could we gate this link and the page behind a config option that defaults to disabled? The enrollment endpoints are currently only in authentication `2.4.0-SNAPSHOT` (the reference distro is still on `2.3.0`), and verification also requires `authentication.scheme` to be configured as a `TwoFactorAuthenticationScheme`. As written, supported installations show Setup even though the workflow returns a 404 or rejects verification. We can flip the default on in the distro once the backend release and runtime config are in place.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+          showSnackbar({
+            title: t('totpEnabledSuccessfully', 'Two-Factor Authentication Enabled'),
+            subtitle: t(
+              'totpEnabledSuccessfullyMessage',
+              'Your account is now protected with a TOTP authenticator app.',
+            ),
+            kind: 'success',
+          });
+          close();
+        } else {
```
denniskigen: This `else` never runs. The backend throws on an invalid or expired code (`IllegalRequestException`, HTTP 400), so `openmrsFetch` rejects into the `catch` and a resolved response always has `isValidCode: true`. The upshot is that a wrong code currently shows the raw `error.message` ("Server responded with 400…"), which isn't something we want in front of the user.

Could we drop the dead `else` and read the localized message from the response body? Note it's on `translatedMessage`, not `message` (which is the bracketed key):

```js
} catch (error) {
  setErrorMessage(
    error?.responseBody?.error?.translatedMessage ??
    t('totpVerificationFailed', "We couldn't verify that code. Please try again."),
  );
}
```

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+                  <Tag type="green" size="sm">
+                    {t('enabled', 'Enabled')}
+                  </Tag>
+                )}
+              </div>
+              <p>{t('authenticatorAppDescription', 'Use an authentication app to generate one-time codes')}</p>
+            </div>
+            <div>
+              {isTotpEnabled ? (
+                <Button kind="danger--ghost">{t('remove', 'Remove')}</Button>
```
denniskigen: The enrolled state reads clearly here. Since removal isn't implemented yet, could we render this disabled with a short explanation (or hide it)? An enabled danger action that silently does nothing is confusing, especially for an account-security control.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <ModalHeader closeModal={close} title={t('setupAuthenticatorApp', 'Setup Authenticator App')} />
+      <ModalBody>
+        <div>
+          <p className={styles.banner}>
+            {t(
+              'useAuthenticatorApp',
+              'Use a phone app like Microsoft Authenticator, Google Authenticator or Authy etc to get two factor authentication codes.',
+            )}
+          </p>
+          <h4 className={styles.scanInstruction}>Scan the QR Code using an authenticator app from your phone</h4>
```
denniskigen: Nice work translating most of this flow. A few user-facing strings are still hardcoded English: the scan instruction heading here (line 127), the `Cancel` (163) and `Enable 2FA` (169) buttons, and the `setErrorMessage(...)` strings (48, 70, 101, 107). Could we wrap those in the `t()` function as well so they're translatable?

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+      try {
+        setLoadingEnrollment(true);
+        setErrorMessage('');
+
+        const response = await openmrsFetch('/ws/rest/v1/auth/totp/enrollment', {
+          method: 'POST',
+        });
+
+        if (active && response.data) {
+          setQrCodeUri(response.data.qrCodeUri || '');
```
denniskigen: The QR flow reads clearly. Could we also surface `response.data.secret` behind a "Can't scan the QR code?" option with a copyable manual setup key? QR-only enrollment blocks people setting up on the same device they're viewing, anyone without a working camera, and screen-reader users. The backend already returns the secret, so it's just a matter of exposing it.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+                  <InlineNotification kind="error" subtitle={errorMessage} hideCloseButton />
+                </div>
+              )}
+              <div className={styles.qrContainer}>
+                {qrCodeUri && (
+                  <img className={styles.qrImage} src={qrCodeUri} alt={t('scanQrCode', 'Scan the QR Code')} />
+                )}
+              </div>
+              <div className={styles.inputCode}>
+                <TextInput
```
denniskigen: Could we make this a numeric one-time-code field and attach errors to it directly?
- `inputMode="numeric"` for the right mobile keyboard
- `invalid`/`invalidText` instead of the separate notification, so the error sits on the field and gets announced
- `autoComplete="one-time-code"` in place of `"off"` so password managers can offer to fill it (minor for TOTP, but harmless)

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.scss
```diff
@@ -0,0 +1,88 @@
...
+    margin-bottom: layout.$spacing-04;
+  }
+
+  p {
+    @include type.type-style('body-short-02');
+    margin-bottom: layout.$spacing-02;
+  }
+}
+
+.methodTile {
```
denniskigen, line 50: On narrow screens this stays a single flex row with the icon separator, text, and button, plus heavy padding. Around 320 to 375px, and with longer translations, the text column gets very cramped. Could we stack the content and action at a breakpoint and trim the padding there?

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+      size: 'sm',
+    });
+  }, []);
+
+  return (
+    <div>
+      <PageHeader className={styles.header}>
+        <PageHeaderContent
+          title={title ? title : t(dashboardTitle.key, dashboardTitle.value)}
+          illustration={<ServiceQueuesPictogram />}
```
denniskigen: `ServiceQueuesPictogram` looks out of place here, it makes the page read like another clinical dashboard. Could we swap it for a security or 2FA illustration, or leave the illustration off until one's available?

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const TwoFactorAuth: React.FC<TwoFactorAuthProps> = ({ title }) => {
+  const { t } = useTranslation();
+  const config = useConfig<ConfigSchema>();
+  const { dashboardTitle } = config.twoFactorAuth;
+  const session = useSession();
+
+  const userProperties = session.user?.userProperties;
+  const secondaryAuthTypes = userProperties?.['authentication.secondaryType'] || '';
+  const isTotpEnabled = secondaryAuthTypes.includes('totp');
```
denniskigen: Nit: `authentication.secondaryType` is a comma-separated list of exact scheme IDs, so `includes('totp')` would also match something like `custom-totp`. Might be safer as:

```js
const isTotpEnabled = secondaryAuthTypes.split(',').includes('totp');
```

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth-config-schema.ts
```diff
@@ -0,0 +1,25 @@
+import { Type } from '@openmrs/esm-framework';
```
denniskigen, line 1: Nit: no JSX in this file, so it should be `.ts` per the [file naming conventions](https://o3-docs.openmrs.org/docs/frontend-modules/file-and-folder-conventions).

denniskigen (commented):
Really nice work on this, @dilankavishka. The page and enrollment flow come together well, and the styleguide and icon plumbing follow the existing patterns closely. I've left some comments below.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
+import React, { useState, useRef, useEffect, useCallback } from 'react';
```
denniskigen: The link and page have tests, but the modal has the riskiest logic (enroll, verify, and the error handling above) and none yet. Could we add a couple covering the success path and a failed verification once the error handling is sorted?

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+    },
+    [verificationCode, close],
+  );
+
+  return (
+    <Form onSubmit={handleVerificationCodeSubmit}>
+      <ModalHeader closeModal={close} title={t('setupAuthenticatorApp', 'Setup Authenticator App')} />
+      <ModalBody>
+        <div>
+          <p className={styles.banner}>
```
denniskigen: A few small copy things:
- "Set up" is the verb, "Setup" the noun, so "Set up Authenticator App" and a "Set up" action button.
- "After signing in with your password" assumes password auth, but this app also supports OAuth and custom providers. Something like "After signing in, you'll verify your identity with a second method" is provider-neutral.
- Optional: numbered steps in the modal ("1. Scan the QR code", "2. Enter the six-digit code") and a shorter intro line would tighten it up.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+const TotpEnrollment: React.FC<TotpEnrollmentProps> = ({ close }) => {
+  const { t } = useTranslation();
+  const [loadingEnrollment, setLoadingEnrollment] = useState(true);
+  const [submittingVerificationCode, setSubmittingVerificationCode] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [qrCodeUri, setQrCodeUri] = useState('');
+  const [verificationCode, setVerificationCode] = useState('');
+
+  const codeInputRef = useRef<HTMLInputElement>(null);
```
denniskigen: A few tiny things, all optional:
- `codeInputRef` is attached to the input but never used. Either focus it on mount or drop it.
- In `launchTotpModal` (two-factor-auth.component.tsx:32), the `closeModal` prop isn't read: `showModal` already injects `close`, which is what this component destructures.
- `close(): () => void` should be `close(): void` (same typo as the change-password modal it was copied from).
- Nit: in `two-factor-auth.scss`, `white-space: nowrap` on `.menuItemButton` may clip the account-menu label with longer translations, worth checking.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+          },
+          body: {
+            code: verificationCode,
+          },
+        });
+
+        if (response.data && response.data.isValidCode) {
```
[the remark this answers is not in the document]
denniskigen: Confirming this one against the backend: the secret is saved with an explicit `saveUser` before the endpoint returns `isValidCode: true`, so by the time `refetchCurrentUser()` runs the enrollment is already committed. If that refetch rejects we fall into the `catch` and show an error on what actually succeeded, and reopening then fails with `secretAlreadyConfigured`. Could we show the success snackbar and `close()` as soon as `isValidCode` is true, then refresh the session best-effort so a refresh hiccup can't undo it?

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,176 @@
...
+        });
+
+        if (active && response.data) {
+          setQrCodeUri(response.data.qrCodeUri || '');
+        }
+      } catch (error) {
+        console.error('Failed to initiate TOTP enrollment:', error);
+        if (active) {
+          setErrorMessage('Failed to initiate TOTP enrollment. Please try again.');
+        }
```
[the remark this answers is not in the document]
denniskigen: Agree with this. The QR request only runs on mount, and the secret expires after ~120s server-side, so it's reasonably easy to land in a state where the message says "try again" but there's no way to without closing and reopening. A retry action that re-runs enrollment would cover it.

denniskigen:
Looks like this needs a changeset: it adds two exported styleguide icons (`MobileCheckIcon`, `TwoFactorAuthenticationIcon`) plus new login-app functionality, so it should bump those packages.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth-config-schema.ts
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.scss
dilankavishka: updated for responsiveness

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
dilankavishka: I registered a new pictogram which called `digital trust` and I think that is the most suitable pictogram for this Two-Factor Authentication header. But this icon need some UI improvement and let me create a separate ticket when the pictogram is ready.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Added unit test for the modal and added more unit tests and e2e test as well

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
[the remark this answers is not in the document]
dilankavishka: I've fixed this by implementing `ActionableNotification`

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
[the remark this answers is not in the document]
dilankavishka: Fixed

packages/apps/esm-login-app/src/routes.json
dilankavishka: Added config and set false as default

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: fixed this by implement the enter secret manually

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+  useConfig,
+  showModal,
+  MobileCheckIcon,
+  useSession,
+} from '@openmrs/esm-framework';
+import type { ConfigSchema } from '../config-schema';
+import styles from './two-factor-auth.scss';
+
+interface TwoFactorAuthProps {
+  title?: string | JSX.Element;
```
janithRS, line 17: string | JSX.Element could be simplified to ReactNode.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+                  <Tag type="green" size="sm">
+                    {t('enabled', 'Enabled')}
+                  </Tag>
+                )}
+              </div>
+              <p>{t('authenticatorAppDescription', 'Use an authentication app to generate one-time codes')}</p>
+            </div>
+            <div>
+              {isTotpEnabled ? (
+                <Button kind="danger--ghost">{t('remove', 'Remove')}</Button>
```
janithRS: on click handler is not attached in here. Is this expected?

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const secondaryAuthTypes = userProperties?.['authentication.secondaryType'] || '';
+  const isTotpEnabled = secondaryAuthTypes.split(',').includes('totp');
+
+  const launchTotpModal = React.useCallback(() => {
+    showModal('totp-enrollment-modal', {
+      size: 'sm',
+    });
+  }, []);
+
+  if (!config.twoFactorAuth.enabled) {
```
janithRS, line 36: If twoFactorAuth can ever be absent during loading or due to incomplete configuration, the component will throw instead of returning null. So let's avoid accessing twoFactorAuth before checking its availability.

```
const twoFactorAuthConfig = config.twoFactorAuth;

if (!twoFactorAuthConfig?.enabled) {
  return null;
}

const { dashboardTitle } = twoFactorAuthConfig;
```

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return null;
+  }
+
+  return (
+    <div>
+      <PageHeader className={styles.header}>
+        <PageHeaderContent
+          title={title ? title : t(dashboardTitle.key, dashboardTitle.value)}
```
janithRS: Can be simplified as

`title={title ?? t(dashboardTitle.key, dashboardTitle.value)}`

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const TwoFactorAuth: React.FC<TwoFactorAuthProps> = ({ title }) => {
+  const { t } = useTranslation();
+  const config = useConfig<ConfigSchema>();
+  const { dashboardTitle } = config.twoFactorAuth;
+  const session = useSession();
+
+  const userProperties = session.user?.userProperties;
+  const secondaryAuthTypes = userProperties?.['authentication.secondaryType'] || '';
+  const isTotpEnabled = secondaryAuthTypes.split(',').includes('totp');
```
janithRS, line 28: Hmm since this is false until the session finishes loading, wouldn't this show the Set up button even for users who already have TOTP enabled?

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
denniskigen: Thanks for going through this, Janith. `ReactNode` reads nicer but won't compile here. `title` is forwarded into `PageHeaderContent`, whose `title` is `string | JSX.Element`, so the compiler rejects the wider type: `Type 'ReactNode' is not assignable to type 'string | Element'`. Widening `PageHeaderContentProps` in the styleguide would have to come first, so let's leave it out of this PR.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
denniskigen: `useConfig` suspends until the config is loaded, and the schema gives `twoFactorAuth` an `enabled: false` default, so the key is always present by the time the component body runs. Safe as is.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
denniskigen: `useSession` suspends as well, so there's no partial-session state to guard against. The `Suspense` boundary comes from `openmrsComponentDecorator`, which wraps everything registered through `getSyncLifecycle`, so the page renders nothing until the session resolves. Enrolled users won't see the Set up button. `useSession.test.tsx` covers this in "should suspend and then resolve to the session".

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,213 @@
...
+        close();
+
+        refetchCurrentUser().catch((error) => {
+          console.warn('Failed to refresh current user after TOTP enrollment:', error);
+        });
+      } catch (error: unknown) {
+        const fetchError = error as OpenmrsFetchError;
+
+        setErrorMessage(
+          fetchError.responseBody.error.translatedMessage ??
```
denniskigen: Thanks for switching this to the backend's translated error message. One gap: not every rejection carries the REST error envelope. `openmrsFetch` can reject with a plain network error, and `OpenmrsFetchError.responseBody` can be a string or `null`. So `responseBody.error` can itself throw, leaving the modal with no feedback at all. An `instanceof OpenmrsFetchError` check plus a guarded lookup would keep the translated message where it exists and fall back to `totpVerificationFailed` otherwise.

Worth noting the local `OpenmrsFetchError` interface at the top of this file types `responseBody` as `{ error: { translatedMessage: string } }`, which is why the compiler doesn't catch this. Importing the real class from `@openmrs/esm-framework` would surface the wider type.

packages/apps/esm-login-app/src/index.ts
```diff
@@ -25,3 +27,6 @@ export const logoutButton = getSyncLifecycle(logoutButtonComponent, options);
 export const changeLocationLink = getSyncLifecycle(changeLocationLinkComponent, options);
 export const changePasswordLink = getSyncLifecycle(changePasswordLinkComponent, options);
 export const changePasswordModal = getAsyncLifecycle(() => import('./change-password/change-password.modal'), options);
+export const twoFactorAuthLink = getSyncLifecycle(twoFactorAuthLinkComponent, options);
+export const twoFactorAuth = getSyncLifecycle(TwoFactorAuth, options);
```
denniskigen: Small thing: `twoFactorAuth` is exported here but nothing in `routes.json` references it. The page is reached through the `root` component and the `two-factor-auth` route instead, so this export is unused. Worth dropping unless you're planning to register it as its own page.

denniskigen:
@dilankavishka I took the liberty of fixing up the pictogram in a commit. Our pictograms use two brand teals (`#CEE6E5` and `#7BBCB9`) and flat filled shapes, so the Carbon icon read as out of place next to the rest of the set. Here's what it looks like now:

CleanShot 2026-07-28 at 01 17 39@2x

jayasanka-sack:
I did some experiments and came up with this layout for the modal. The goal was to make it as self-intuitive as possible for users.

The design is heavily inspired by the pairing screens used in Android TV apps like Netflix and YouTube.

The “Which apps can I use?” link could point to a wiki page by default with a recommended list of supported apps and links to download them from the relevant app stores. We can make that link configurable so different implementations can point users to their own documentation if needed.


image

https://talk.openmrs.org/t/gsoc-2026-integrate-o3-with-the-authentication-module/49459/21?u=jayasanka

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
dilankavishka: My plan is to build the actual removal flow (with the API call) in a separate follow-up PR.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
dilankavishka: done

packages/apps/esm-login-app/src/index.ts
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: updated

dilankavishka:
Thank you so much @denniskigen @jayasanka-sack @janithRS for reviewing this pr. I have updated the totp enrollment modal and here how it looks right now.

Screenshot 2026-07-30 at 22 27 32


@jayasanka-sack As we discussed in the yesterday gsoc-goofies call, I will update the pin code input in a separate PR, because my plan is to include that implementation in my next PR. (Right now, i'm building that)

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const launchTotpModal = React.useCallback(() => {
```
denniskigen:
```suggestion
  const launchTotpModal = useCallback(() => {
```

We can derive `useCallback` from the top-level React import.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,235 @@
...
+                labelText={t('enterVerificationCode', 'Enter 6-digit code from your app')}
+                helperText={t(
+                  'codeExpiryWarning',
+                  'Code expires after 30 seconds. If it is rejected, wait for the next one.',
+                )}
+                inputMode="numeric"
+                maxLength={6}
+                value={verificationCode}
+                invalidText={errorMessage}
+                invalid={!!errorMessage}
```
denniskigen: The enrollment request can fail before a verification code exists, but the shared `errorMessage` also marks this input invalid. Testing this locally, the initiation failure appeared both beneath the empty code field and in the retry notification, making it look like the user entered a bad code. Could initiation and verification errors be stored separately so only verification failures set `invalid` here?

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  const handleVerificationCodeSubmit = useCallback(
+    async (event: React.FormEvent) => {
+      event.preventDefault();
+      event.stopPropagation();
+
+      if (verificationCode.length !== 6) {
+        setErrorMessage('Verification code must be 6 digits.');
+        return;
+      }
```
denniskigen: `inputMode="numeric"` only hints which keyboard to display and does not restrict the value. Testing this locally, entering `abcdef` enabled the confirmation button, and the same length-only check would allow it to be submitted. Could we validate against six ASCII digits, or sanitize non-digits on change, and use the same check when deciding whether submission is enabled?

```suggestion
      if (!/^\d{6}$/.test(verificationCode)) {
        setErrorMessage(t('totpCodeMustBeSixDigits', 'The verification code must be 6 digits.'));
        return;
      }
```

(New key, so `yarn turbo run extract-translations` after.) I know the pin-code input is coming in your follow-up PR, so this is just a stopgap; fine to fold it into that PR instead, but the hardcoded string should be translated either way.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.scss
```diff
@@ -0,0 +1,124 @@
...
+  @include type.type-style('helper-text-01');
+  color: $text-02;
+  margin-bottom: layout.$spacing-03;
+}
+
+.secretSnippet {
+  max-width: 23rem;
+}
+
+.twoColumnLayout {
```
denniskigen, line 70: Not blocking this PR, but flagging since we tightened the page tile for narrow screens earlier: in the modal these two columns stay side by side on small screens and the content overlaps. At a 375px viewport the QR image paints over the step 2 text, and the setup key snippet spills across both columns with the end of the key clipped. Stacking the columns below a small breakpoint plus `max-width: 100%` on `.qrImage` would cover it. This could come as a follow-up.

packages/apps/esm-login-app/src/routes.json
```diff
@@ -21,6 +21,12 @@
       "route": "change-password",
       "online": true,
       "offline": true
+    },
+    {
+      "component": "root",
+      "route": "two-factor-auth",
+      "online": true,
+      "offline": true
```
denniskigen: One follow-up from the earlier offline thread: this page and the `two-factor-authentication` extension are still registered with `"offline": true`, but enrollment immediately requires REST calls, so offline users can only reach an error. Could we set `offline: false` on these two entries? The other `root` registrations (login, change-password) should keep it.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,235 @@
...
+                  actionButtonLabel={t('tryAgain', 'Try Again')}
+                  onActionButtonClick={initiateEnrollment}
+                />
+              ) : qrCodeUri ? (
+                <>
+                  <img className={styles.qrImage} src={qrCodeUri} alt={t('scanQrCode', 'Scan the QR Code')} />
+                  <div className={styles.manualSetupContainer}>
+                    <p className={styles.cantScanText}>{t('cantScanQrCode', "Can't scan the QR code?")}</p>
+                    <p className={styles.manualSetupLabel}>{t('manualSetupKey', 'Enter this setup key in your app')}</p>
+                    <CodeSnippet type="single" className={styles.secretSnippet}>
```
denniskigen: Nice improvement making the setup key easier to read. Could we pass `copyText` here? Carbon otherwise copies the child's `innerText`, so the copy button currently puts `JBSW Y3DP EHPK 3PXP ...` on the clipboard, spaces included (verified locally by capturing the copy event):

```suggestion
                    <CodeSnippet type="single" className={styles.secretSnippet} copyText={secret}>
```

That keeps the grouped display while copying the exact Base32 value.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,235 @@
...
+          {t('cancel', 'Cancel')}
+        </Button>
+        <Button
+          type="submit"
+          disabled={loadingEnrollment || submittingVerificationCode || verificationCode.length !== 6}
+        >
+          {submittingVerificationCode ? (
+            <InlineLoading description={t('verifying', 'Verifying...')} />
+          ) : (
+            t('confirmAndEnable', 'Confirm and Enable authenticator app')
```
denniskigen: A few copy nits: sentence case for the button ("Confirm and enable authenticator app"); "confirm the setting up" reads awkwardly, maybe "Enter the code the app shows to finish setting up."; and "after your password" assumes password login, "each time you sign in" alone keeps it provider-neutral.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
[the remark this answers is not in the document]
denniskigen: Thanks for adding the actionable retry for QR-generation failures. I checked this against the backend and tested the flow locally, and one recovery path is still missing. The pending secret expires after `enrollmentWindowSeconds` (default 120s), and on the first expired verify attempt, [`verifyEnrollment`](https://github.com/openmrs/openmrs-module-authentication/blob/main/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java) removes the pending secret from the session before throwing. So the user sees "The enrollment session has expired. Please try again", but the QR still on screen is now dead: every retry fails with "No pending enrollment found". This notification never renders in that state because `qrCodeUri` is still populated, so the only way out is closing and reopening the modal, and nothing tells the user that. The module's own tests confirm both halves (`shouldThrowExceptionIfSessionIsExpired`, `shouldThrowExceptionIfNoStashedSecretInVerifyEnrollment`).

Could we surface a way to restart enrollment after a verification failure too, e.g. a "Generate a new QR code" action alongside the field error, wired to `initiateEnrollment`? Best not to hardcode a 120s timer client-side, since the window is configurable server-side.

denniskigen (commented):
Thanks for the continued iteration on this, @dilankavishka. The new modal layout is a significant improvement and the successful desktop state is clear and easy to follow. I ran the branch locally to exercise the desktop UI states, used a stubbed enrollment response to inspect the QR-code flow, and checked the recovery behavior against the backend implementation. I left a few focused comments below. The expired-enrollment recovery path is the most important.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import TotpEnrollment from './totp-enrollment.modal';
+
+const mockT = (key: string, defaultText: string) => defaultText;
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: mockT,
+  }),
+}));
+
+vi.mock('@openmrs/esm-framework', async () => {
```
denniskigen: This package's Vitest config already aliases `@openmrs/esm-framework` to the shared mock, where `openmrsFetch`, `refetchCurrentUser`, and `showSnackbar` are already `vi.fn()` implementations. The local framework mock is only needed because `@openmrs/esm-api/mock` does not currently export `OpenmrsFetchError`.

Could we export the real class from `packages/framework/esm-api/mock.ts` and `mock-jest.ts`, per the mocks checklist in the PR template, and then remove this `@openmrs/esm-framework` mock?

```ts
export { OpenmrsFetchError } from './src/openmrs-fetch';
```

The stable `react-i18next` mock should remain because `t` participates in the enrollment effect's dependency chain. I verified that exact arrangement passes both modal tests.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const [verificationCode, setVerificationCode] = useState('');
+  const [secret, setSecret] = useState('');
+  const formattedSecret = secret.match(/.{1,4}/g)?.join(' ') || secret;
+
+  const initiateEnrollment = useCallback(async () => {
+    try {
+      setLoadingEnrollment(true);
+      setErrorMessage('');
+
+      const response = await openmrsFetch('/ws/rest/v1/auth/totp/enrollment', {
```
denniskigen: Nit: could we move the enrollment and verification requests into `two-factor-auth.resource.ts`, following the existing `change-password.resource.ts` pattern, and construct the URLs from the shared `restBaseUrl`?

```ts
openmrsFetch(`${restBaseUrl}/auth/totp/enrollment`, ...)
openmrsFetch(`${restBaseUrl}/auth/totp/enrollment/verify`, ...)
```

This keeps REST calls in the package's resource layer, avoids duplicating `/ws/rest/v1`, and lets the component test mock the focused resource module instead of framework networking.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.test.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { vi, describe, beforeEach, it, expect } from 'vitest';
+import { showModal, useSession, useConfig } from '@openmrs/esm-framework';
+import userEvent from '@testing-library/user-event';
+import { render, screen } from '@testing-library/react';
+import TwoFactorAuth from './two-factor-auth.component';
+import type { ConfigSchema } from '../config-schema';
+
+vi.mock('@openmrs/esm-framework', async () => {
```
denniskigen: I removed this entire framework mock locally and both component tests still passed. The shared framework mock already provides `showModal`, `useConfig`, and `useSession` as Vitest mocks. It also receives `MobileCheckIcon` and `DigitalTrustPictogram` through the styleguide mock, while `ServiceQueuesPictogram` is no longer used by this component. Could we remove this local framework mock and use the shared mocks directly?

The same cleanup applies to the framework mock in `two-factor-auth-link.extension.test.tsx`. Its `navigate`, `useConfig`, and icon replacements are already available through the shared mocks. One assertion needs updating there: the shared stub renders the icon name as text, so `screen.getByTestId('mock-icon')` becomes `screen.getByText('TwoFactorAuthenticationIcon')`. With that change, all six tests in the folder pass. While you're in that file, the two consecutive `beforeEach` blocks both call `vi.clearAllMocks()`, so they can be consolidated.

denniskigen (commented):
A few test-convention notes from a second pass, all verified locally against this branch.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,235 @@
...
+          if (translatedMessage) {
+            errorMessage = translatedMessage;
+          }
+        }
+        setErrorMessage(errorMessage);
+      } finally {
+        setSubmittingVerificationCode(false);
+      }
+    },
+    [verificationCode, close],
```
denniskigen: This callback reads `t` when constructing verification errors, but `t` is missing from the dependency list. Could we include it so the callback follows the hook dependency contract and uses the current translator if the language changes while the modal is open?

```suggestion
    [verificationCode, close, t],
```

packages/framework/esm-styleguide/src/icons/icons.tsx
```diff
@@ -771,6 +781,15 @@ export const TreeViewAltIcon = memo(
   }),
 );
 
+/**
+ * This icon is used in the My account and the Two Factor Authentication page header as well.
```
denniskigen: Small documentation mismatch: this icon is currently used in the account-menu link, while the page header uses `DigitalTrustPictogram` and the method tile uses `MobileCheckIcon`. Since this comment feeds the generated framework API documentation, could we update it to describe the actual use? `MobileCheckIcon` above is missing its description entirely, so worth filling that one in at the same time.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.test.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
```
denniskigen, line 1: Nit: per the [file conventions](https://o3-docs.openmrs.org/en-US/docs/coding-conventions/code-organization), test files drop the kind suffix, so these would be `two-factor-auth.test.tsx`, `totp-enrollment.test.tsx`, and `two-factor-auth-link.test.tsx` (the docs pair `patient-banner.component.tsx` with `patient-banner.test.tsx`). Can be ignored, but cheap to do while the mock cleanup touches all three files anyway.

dilankavishka: Updated

packages/framework/esm-styleguide/src/icons/icons.tsx
dilankavishka: Added new comments for newly added icons and pictogram

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.test.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: created new resource file and updated

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.test.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: done

packages/apps/esm-login-app/src/routes.json
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.scss
dilankavishka: fixed this

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: fixed

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
dilankavishka: updated

dilankavishka:
Thank you so much @denniskigen for reviewing this PR and I've addressed all the comments and suggestions you've raised :)

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
janithRS: Yup, make sense. Let's leave it for the future :)

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
janithRS: Then I think a comment would be helpful here

janithRS:
Yeah the overall design looks good :) Nice work Dilan

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,227 @@
...
+            </div>
+          </div>
+
+          <div className={styles.rightColumn}>
+            <div className={styles.qrContainer}>
+              {loadingEnrollment ? (
+                <div className={styles.center}>
+                  <InlineLoading description={t('generatingQrCode', 'Generating QR Code...')} />
+                </div>
+              ) : !qrCodeUri && initiationError ? (
```
denniskigen, line 197: Separating the two error states makes this one more definite: a verification failure only sets `verificationError`, and this notification is gated on `initiationError`, so it can never render for one.

[`verifyEnrollment`](https://github.com/openmrs/openmrs-module-authentication/blob/main/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java#L287-L290) removes the pending secret on the first expired attempt and throws `secretAlreadyExpired`, and every attempt after that throws `noPendingEnrollmentTotpSecretFound`. So once the window passes, the QR on screen is dead, the field error is all the user gets, and closing and reopening the modal is the only way out. Nothing says so.

Could we give verification failures a way back to a fresh enrollment, say a "Generate a new QR code" action next to the field error wired to `initiateEnrollment`? Best not to hardcode 120s client-side since `enrollmentWindowSeconds` is configurable. `initiateEnrollment` should probably reset `verificationCode` and `verificationError` too, since it only clears `initiationError` today.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.scss
```diff
@@ -0,0 +1,128 @@
...
+  display: flex;
+  flex-direction: row;
+  gap: layout.$spacing-05;
+  align-items: stretch;
+
+  @media (max-width: 40px) {
+    flex-direction: column;
```
denniskigen:
```suggestion
  @media (max-width: 672px) {
    flex-direction: column;
```

`40px` never matches a real viewport, so the columns don't stack anywhere. The 375px overlap from last round is unchanged: the 240px `.inputCode` runs past its own 164px column into the QR column, and the setup key snippet ends up about 90px beyond the viewport. `two-factor-auth.scss` already uses `672px` for the same stacking.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.resource.ts
```diff
@@ -0,0 +1,19 @@
+import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+
+export function inititateTotpEnrollment() {
```
denniskigen:
```suggestion
export function initiateTotpEnrollment() {
```

Nit: extra `t` in `initiate`. Also imported in `totp-enrollment.modal.tsx:16` and in the test.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,227 @@
...
+                helperText={t(
+                  'codeExpiryWarning',
+                  'Code expires after 30 seconds. If it is rejected, wait for the next one.',
+                )}
+                inputMode="numeric"
+                maxLength={6}
+                value={verificationCode}
+                invalidText={verificationError}
+                invalid={!!verificationError}
+                onChange={(e) => setVerificationCode(e.target.value)}
```
denniskigen:
```suggestion
                onChange={(e) => setVerificationCode(e.target.value.replace(/\D/g, ''))}
```

Nit: `inputMode="numeric"` is only a keyboard hint, so letters still land in the field. Typing `abcdef` leaves the Confirm button greyed out with nothing saying why. Stripping non-digits here keeps the field to digits and makes the disabled state read as "keep typing".

While you're here, the `totpCodeMustBeSixDigits` check on line 62 can't be reached: the only submit button is disabled until the regex passes, and Enter won't submit a form whose submit button is disabled. That branch and its en.json string can go.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+          illustration={<DigitalTrustPictogram />}
+        />
+      </PageHeader>
+
+      <main className={styles.pageContainer}>
+        <div className={styles.mainCard}>
+          <div className={styles.descriptionText}>
+            <h3>{t('protectAccountTitle', 'Protect your account with an extra verification step.')}</h3>
+            <p>
+              {t('protectAccountDescription', 'After signing in, you will verify your identity with a second method')}
```
denniskigen:
```suggestion
              {t('protectAccountDescription', 'After signing in, you will verify your identity with a second method.')}
```

Nit: en.json has the trailing period and this default doesn't. en.json wins at runtime, but the next `yarn turbo run extract-translations` would overwrite it from here and drop it.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth-link.extension.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const config = useConfig<ConfigSchema>();
+
+  if (!config.twoFactorAuth.enabled) {
+    return null;
+  }
+
+  const handle2faSetupClick = () => {
+    // Navigate to the Two Factor Authentication Page
+    openmrsNavigate({ to: '${openmrsSpaBase}/two-factor-auth' });
+    document.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
```
denniskigen, line 21: Nit, can be ignored. This works, but it took me a minute to see why: `useOnClickOutside` listens for `mousedown` on `window`, and `document` isn't inside the user menu's ref, so the panel closes. The comment above explains the `navigate` call rather than this line. Could we say what this one is for?

The cleaner fix probably belongs in the primary nav: `UserMenuPanel` already has `hidePanel` but doesn't pass it into the `user-panel-slot` state the way `top-nav-actions-slot` gets it. Fine as a follow-up.

denniskigen (commented):
This is very close, @dilankavishka. Almost everything from the last round is addressed. Two things left, and the expired-enrollment one is the item I flagged as most important last time.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
denniskigen: Seconding this. A one-line comment noting the handler lands in the follow-up PR would save the next reader this thread.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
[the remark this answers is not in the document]
denniskigen: Still open on the expired-enrollment side. I've moved the detail to a new thread at line 183, where the gate that excludes verification errors lives.

janithRS:
Forgot one thing. When we were doing UI testing last week I remember we had some issues with tab and mobile view resizing. How did that go? :)

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth-link.extension.tsx
dilankavishka: For now, I will just add a comment for this and let me address this in a separate PR.

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/two-factor-auth.resource.ts
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.scss
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Fixed by adding a Generate new QR code capability

dilankavishka:
> Forgot one thing. When we were doing UI testing last week I remember we had some issues with tab and mobile view resizing. How did that go? :)

@janithRS, this was fixed with a previous feedback from Dennis :D

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const [verificationCode, setVerificationCode] = useState('');
+  const [secret, setSecret] = useState('');
+  const formattedSecret = secret.match(/.{1,4}/g)?.join(' ') || secret;
+
+  const initiateEnrollment = useCallback(async () => {
+    try {
+      setLoadingEnrollment(true);
+      setInitiationError('');
+      setVerificationError('');
+      setVerificationCode('');
```
denniskigen, line 39:
```suggestion
      setLoadingEnrollment(true);
      setInitiationError('');
      setVerificationError('');
      setVerificationCode('');
      setQrCodeUri('');
      setSecret('');
```

`initiateEnrollment` is also the regeneration path now. If that request fails, `qrCodeUri` still holds the old enrollment, so the `!qrCodeUri && initiationError` branch is skipped and the old QR code and setup key come back with no error. Can we clear both values before retrying and cover the rejected retry in a test?

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
```diff
@@ -0,0 +1,244 @@
...
+                      {verificationError}
+                      {''}
+                      <Link
+                        onClick={(event) => {
+                          event.preventDefault();
+                          initiateEnrollment();
+                        }}
+                      >
+                        {t('generateNewQrCode', 'Generate a new QR code')}
+                      </Link>
```
denniskigen, line 178:
```suggestion
                      {verificationError}
                      <Link
                        as="button"
                        type="button"
                        size="sm"
                        className={styles.regenerateAction}
                        onClick={initiateEnrollment}
                      >
                        {t('generateNewQrCode', 'Generate a new QR code')}
                      </Link>
```

Without `href`, Carbon renders this as an `<a>` that is not exposed as an interactive control or reachable by keyboard. `as="button"` gives the recovery action the correct semantics, and `type="button"` prevents it from submitting the enclosing form.

The validation text is 12px while the default Carbon link is 14px, so `size="sm"` gives them matching typography. At the 240px field width, each current English backend message puts the action on a new line with about 1px of separation. Can we make that layout explicit with Carbon's 4px spacing token?

```scss
.regenerateAction {
  display: block;
  margin-top: layout.$spacing-02;
}
```

Worth covering the semantics with `expect(screen.getByRole('button', { name: /generate a new qr code/i })).toBeInTheDocument()`.

denniskigen (commented):
This is very close, @dilankavishka. I found two issues on the new regeneration path, with the failed retry being the important one.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
denniskigen: Thanks, the digit stripping is fixed. The other half is still open: `totpCodeMustBeSixDigits` and its translation key are still present, but the submit button is disabled until the same regex passes, and Enter does not submit while it is disabled. Can we remove both?

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated and add a new test.

packages/apps/esm-login-app/src/two-factor-auth/totp-enrollment.modal.tsx
dilankavishka: Updated

denniskigen (approved):
LGTM! Thanks, @dilankavishka. One small leftover in the thread above: `totpCodeMustBeSixDigits` and its `en.json` key can still be removed.

dilankavishka:
Updated the en.json file @denniskigen.

Also, @denniskigen, thank you so much for reviewing my PR from start to end. I learned so much from your reviews and gained so much through this work.

Thank you @jayasanka-sack and @janithRS for encouraging me throughout this work, reviewing my work, and pushing me to do better.

Yes, I made some mistakes and ignored some minor things that are also very important to consider. I also learned the importance of making an effort to write the code in detail so that anyone reading it can understand it.

I learned through the mistakes I made, and I’m truly grateful for your guidance, support, and patience.

❤️

denniskigen (approved):

denniskigen (approved):

PR #1822, opened 2026-07-31
(fix) O3-5842: Match Carbon's control height in the date pickers

dkayiwa:
## Requirements

- [x] I have updated the esm-framework and storybook mocks ([mock.tsx](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx), [mock-jest.tsx](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock-jest.tsx), and [storybook mocks](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/tooling/storybook/mocks/)) to reflect any API changes. — the mocks do not touch `size`, so there is nothing to update there. The type signatures are unchanged, but see **On the API surface** below: the *runtime default* of `size` changed, and the generated docs do need a regeneration.

## Summary

`OpenmrsDatePicker` rendered its input group at 41px where Carbon's md-size text inputs are 40px, so a date picker next to a `TextInput` or `ComboBox` in a form row put the field bottoms and underlines a pixel apart (reported from the filter row of the audit log app).

**Root cause.** The height and the 1px bottom border lived on different elements. `.inputGroup` carried `border-block-end: 1px solid` but had no height of its own; the height came from the children (`.inputWrapperMd` / `.flatButtonMd`, both `2.5rem`). With `block-size: auto`, the border is added *on top* of the 40px children, so the group measures 41px. Carbon puts the height and the border on the *same* `border-box` element — see `.cds--text-input` (`block-size: layout.size('height')` + `border-block-end: 1px solid $border-strong`) and `.cds--date-picker__input` — so the border sits *inside* the 40px.

**Fix.** The group now owns both the height and the border, and its children fill it (`block-size: 100%`). The height is read from Carbon's layout size tokens via `layout.use('size', $default: 'md', $min: 'sm', $max: 'lg')` / `layout.size('height')`, keyed off the `cds--layout--size-*` class the components set — exactly the mechanism `cds--text-input` uses (`TextInput.js` adds `` [`${prefix}--layout--size-${size}`]: size ``). Nothing is hardcoded against Carbon's sizing, so a Carbon upgrade that changes those tokens carries over automatically.

This also fixes the `sm` and `lg` variants, which were 33px and 49px against Carbon's 32px and 48px, and the same bug in `OpenmrsDateRangePicker`.

**Removed / renamed classes.** `inputWrapperSm/Md/Lg`, `inputsWrapperSm/Md/Lg` and `flatButtonSm/Md/Lg` are all gone — the group sizes its own children now. The one rule that had to survive is the calendar popover's previous/next-month buttons, which are a fixed `2.5rem` square and deliberately *don't* track the field size; that rule was renamed `flatButtonMd` → `.calendarNavButton` so the name stops implying it is the md member of a set that no longer exists. None of the removed classes had consumers outside this directory, and none of them were part of the exported sass surface.

**Size is inherited, not defaulted.** `size` no longer defaults to `'md'` in the components. Emitting `cds--layout--size-md` unconditionally would have opted the picker *out* of any surrounding Carbon layout context: inside a `cds--layout--size-sm` container an unsized `TextInput` is 32px, but an unsized picker would have stayed 40px. With no class, the size comes from the ancestor context, falling back to the `$default: 'md'` in `.inputGroup`'s `layout.use()` — so an unsized picker outside any layout context is still 40px, and inside one it now matches its neighbours. This mirrors `TextInput`, which also only emits the class when a size is passed.

Values outside `sm`/`md`/`lg` are treated as "no size" rather than interpolated into the class name. The prop is typed, but plenty of callers are untyped JavaScript, and interpolating whatever arrives produced tokens that look like Carbon classes but match nothing (`size={40}` → `cds--layout--size-40`), or — for a value containing whitespace — a *second* class that wins the cascade, so `size="md cds--layout--size-lg"` rendered at `lg`.

**Measured heights** (Chromium, `deviceScaleFactor: 1`; `getBoundingClientRect().height` of the picker's `[role="group"]` against a real Carbon `TextInput`/`ComboBox` of the same size rendered in the same document, with the styleguide's `_all.scss` loaded):

| size | Carbon TextInput | picker before | picker after |
|------|-----------------|---------------|--------------|
| sm   | 32 | **33** | 32 |
| md   | 40 | **41** | 40 |
| lg   | 48 | **49** | 48 |
| md (`OpenmrsDateRangePicker`) | 40 | **41** | 40 |

Verified three ways with the same numbers:

1. This repo's Storybook (`yarn storybook`, `Components/DatePicker` and `Components/DateRangePicker` stories, driven with Playwright), in LTR and RTL.
2. A standalone rspack page rendering the components side by side with Carbon's `TextInput` and `ComboBox`.
3. **A running OpenMRS instance.** The app shell was built from this branch (`yarn run:shell` with `OMRS_PROXY_TARGET` pointed at a local OpenMRS 8081) so the real frontend modules from the instance's own importmap render against this styleguide, then Patient Registration and Appointments were driven with Playwright. On Patient Registration the date-of-birth picker measures **41px on `main` and 40px with this branch**, against the page's `cds--text-input` fields at 40px; every element below the picker shifts up by exactly 1px between the two builds. The group's computed `--cds-layout-size-height-local` is `clamp(max(0px, 2rem), 2.5rem, min(999999999px, 3rem))` — i.e. the height really is coming from Carbon's layout token, clamped to sm..lg, and not from a hardcoded value.

**No regressions found.** In the browser, at all three sizes: the calendar opens, renders 35 day cells, a date can be picked, the popover closes and the value lands in the input, and the group height holds after a value is set. The invalid, disabled and default-value stories still render correctly (error outline, warning icon and error text intact); text and the calendar icon remain vertically centred. The popover is portaled, so it cannot inherit the group's layout class.

## Screenshots

Not attached — the change is a 1px height correction; the measurements above are the meaningful evidence. Before, in a mixed row, the date picker's underline sat 1px below the neighbouring inputs'; after, all the underlines share a baseline.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5842

## Other

**On the tests.** The package runs under vitest with happy-dom, which has no layout engine and cannot report a rendered height, so there is no honest way to assert "40px" in the unit tests. Instead the `size prop` suites in `datepicker.test.tsx` and `date-range-picker.test.tsx` assert the mechanism that produces the height: the group carries `cds--layout--size-{sm,md,lg}` when a size is given, and carries **no** `cds--layout--size-*` class at all when one isn't, so that it inherits the surrounding layout context. The same suites cover the full range an untyped caller can actually pass — `null`, `''`, `'xl'`, `'Md'`, a number, an object, and a whitespace-injection string — and assert none of it reaches the DOM as a layout class. `utils.test.ts` covers `getLayoutSizeClass` directly over the same range. The height itself was verified in a real browser as described above. `yarn test` for `@openmrs/esm-styleguide` passes (289 passed, 1 skipped), as do `eslint` and `prettier --check` on the changed files.

**On the API surface.** No exported type changed and no new symbol is exported (`getLayoutSizeClass` is internal to `src/datepicker/` and is not re-exported from `datepicker/index.tsx`). Two things did change that a reviewer should see explicitly:

- The runtime default of `size` (previously always `'md'`, now inherited — see above). Callers that relied on a picker staying 40px inside a non-md layout context will now see it match that context, which is the intended behaviour.
- The TSDoc on `size` in `OpenmrsDatePickerProps` and `OpenmrsDateRangePickerProps`, which means the committed API docs are stale for this branch.

The `drift` check is therefore red on purpose and cannot be cleared from this branch: it requires the `Docs / Regenerated` check run, which only the bot posts, and the bot only runs after an approving review from a write-access reviewer. Committing regenerated docs by hand does not help — the same check hard-fails on any manual edit under `packages/framework/esm-framework/docs/`. **Please just approve; the bot will push the docs commit and `drift` will go green.** `skip-docs` would be the wrong lever here, because the docs genuinely do drift.

### DIFF
--- a/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
+++ b/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
@@ -173,6 +173,23 @@ describe('OpenmrsDateRangePicker', () => {
       expect(inputGroup).toHaveClass(styles.inputGroup);
       expect(inputGroup.className).not.toMatch(/cds--layout--size-/);
     });
+
+    // See the note in datepicker.test.tsx: untyped JavaScript callers can pass anything, and none
+    // of it may reach the DOM as a `cds--layout--size-*` token.
+    it.each([
+      ['null', null],
+      ['an empty string', ''],
+      ['an unsupported step', 'xl'],
+      ['the wrong case', 'Md'],
+      ['a number', 40],
+      ['an object', {}],
+      ['a value that would inject a second class', 'md cds--layout--size-lg'],
+    ])('should render without a layout class when size is %s', (_label, size) => {
+      render(<OpenmrsDateRangePicker aria-label="datepicker" size={size as unknown as 'md'} />);
+      const [inputGroup] = screen.getAllByRole('group');
+      expect(inputGroup).toHaveClass(styles.inputGroup);
+      expect(inputGroup.className).not.toMatch(/cds--layout--size-/);
+    });
   });
 
   describe('calendar popover', () => {
--- a/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
+++ b/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
@@ -157,7 +157,6 @@ describe('OpenmrsDateRangePicker', () => {
     // See the note in datepicker.test.tsx: the group owns both the height and the bottom border,
     // and gets its height from Carbon's layout size tokens via these classes.
     it.each([
-      [undefined, 'cds--layout--size-md'],
       ['sm' as const, 'cds--layout--size-sm'],
       ['md' as const, 'cds--layout--size-md'],
       ['lg' as const, 'cds--layout--size-lg'],
@@ -166,6 +165,13 @@ describe('OpenmrsDateRangePicker', () => {
       const [inputGroup] = screen.getAllByRole('group');
       expect(inputGroup).toHaveClass(styles.inputGroup);
       expect(inputGroup).toHaveClass(expectedClass);
+    });
+
+    it('should not set a layout class when no size is given', () => {
+      render(<OpenmrsDateRangePicker aria-label="datepicker" />);
+      const [inputGroup] = screen.getAllByRole('group');
+      expect(inputGroup).toHaveClass(styles.inputGroup);
+      expect(inputGroup.className).not.toMatch(/cds--layout--size-/);
     });
   });
 
--- a/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
+++ b/packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
@@ -154,38 +154,19 @@ describe('OpenmrsDateRangePicker', () => {
   });
 
   describe('size prop', () => {
-    /* eslint-disable testing-library/no-container, testing-library/no-node-access */
-    const getInputsWrapper = (container: HTMLElement) =>
-      Array.from(container.querySelectorAll('div')).find((el) => el.className.includes('inputsWrapper')) ?? null;
-
-    it('should apply md size classes by default', () => {
-      const { container } = render(<OpenmrsDateRangePicker aria-label="datepicker" />);
-      const wrapper = getInputsWrapper(container)!;
-      expect(wrapper).toHaveClass(styles.inputsWrapperMd);
-      expect(screen.getByRole('button')).toHaveClass(styles.flatButtonMd);
+    // See the note in datepicker.test.tsx: the group owns both the height and the bottom border,
+    // and gets its height from Carbon's layout size tokens via these classes.
+    it.each([
+      [undefined, 'cds--layout--size-md'],
+      ['sm' as const, 'cds--layout--size-sm'],
+      ['md' as const, 'cds--layout--size-md'],
+      ['lg' as const, 'cds--layout--size-lg'],
+    ])('should size the input group with the Carbon layout class for size=%s', (size, expectedClass) => {
+      render(<OpenmrsDateRangePicker aria-label="datepicker" size={size} />);
+      const [inputGroup] = screen.getAllByRole('group');
+      expect(inputGroup).toHaveClass(styles.inputGroup);
+      expect(inputGroup).toHaveClass(expectedClass);
     });
-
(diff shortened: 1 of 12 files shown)


packages/framework/esm-styleguide/src/datepicker/datepicker.test.tsx
```diff
@@ -194,39 +194,51 @@ describe('OpenmrsDatePicker', () => {
...
+      const [inputGroup] = screen.getAllByRole('group');
+      expect(inputGroup).toHaveClass(styles.inputGroup);
+      expect(inputGroup.className).not.toMatch(/cds--layout--size-/);
+    });
+
+    // The prop is typed, but the framework is consumed from untyped JavaScript, so the component
+    // has to survive whatever a caller actually passes. Nothing outside the three supported steps
+    // may reach the DOM as a `cds--layout--size-*` token: the group must stay unsized and inherit
+    // the layout context instead.
+    it.each([
```
denniskigen, line 228: Nit: this invalid-input matrix now runs three times: once against `getLayoutSizeClass` directly in `utils.test.ts`, and once through a full component render in each picker's test file. Since both components just delegate to the helper, `utils.test.ts` can own the full matrix and a single representative case here (say, the whitespace-injection one) would prove the wiring. Would cut about 40 lines across the two component test files with no coverage loss. Can be ignored.

packages/framework/esm-styleguide/src/datepicker/date-range-picker.test.tsx
```diff
@@ -154,38 +154,42 @@ describe('OpenmrsDateRangePicker', () => {
...
+    it('should not set a layout class when no size is given', () => {
+      render(<OpenmrsDateRangePicker aria-label="datepicker" />);
+      const [inputGroup] = screen.getAllByRole('group');
+      expect(inputGroup).toHaveClass(styles.inputGroup);
+      expect(inputGroup.className).not.toMatch(/cds--layout--size-/);
+    });
+
+    // See the note in datepicker.test.tsx: untyped JavaScript callers can pass anything, and none
+    // of it may reach the DOM as a `cds--layout--size-*` token.
+    it.each([
```
denniskigen, line 179: Ditto.

denniskigen (approved):
LGTM! Thanks, @dkayiwa. Approving so the docs bot can regenerate and clear `drift`.

PR #1827, opened 2026-08-04
(feat) Adopt ESLint 9 and the shared @openmrs/eslint-config

denniskigen:
## Requirements

## Summary

This is the pilot migration for [`@openmrs/eslint-config`](https://www.npmjs.com/package/@openmrs/eslint-config), the new shared flat-config package that centralizes the ESLint rules that have been copy-pasted (and slowly drifting) across the `openmrs-esm-*` repositories. Since the config's `base` preset was ported from this repo's `.eslintrc`, esm-core is the natural first adopter.

What changed:

- Bumped `eslint` from `^8.55.0` to `^9.13.0` (resolves to 9.39.5), added `@openmrs/eslint-config`, and removed the seven ESLint plugin devDependencies the shared config now provides.
- Replaced `.eslintrc` and `.eslintignore` with an `eslint.config.mjs` that spreads the shared config. Flat config stops config discovery at the project root, which the old `.eslintrc` never did (it lacked `root: true`).
- Dropped `--ext ts,tsx` from the 28 workspace lint scripts. The flag was removed in ESLint 9 and flat config makes it unnecessary.
- Removed 20 stale `eslint-disable` directives across 15 files. ESLint 9 reports unused directives by default, where v8 silently ignored them. Most were artifacts of the typescript-eslint v8 rule renames that the legacy config never accounted for. These are comment and whitespace-only changes.
- Updated `turbo.json`'s lint cache input, removed dependabot's now-resolved "ESLint v9 requires flat config migration" ignores, pointed its linting group at `eslint` and `@openmrs/eslint-config`, and refreshed the copilot instructions.

## Other

One incidental fix: the old `.eslintignore` pattern `packages/*/dist` never actually matched the built `dist` directories (workspaces are two levels deep, and `*` does not cross `/`). The new config uses `**/dist/**`, which does.

Once this has baked for a bit, the plan is to roll the shared config out to the other active O3 repos in waves, then graduate it to `1.0.0`.


ibacher (approved):
Thanks for catching the `dist` stuff. Before we roll this out, we should release 1.0.0 of the `@openmrs/eslint-config` artifact. It might also be nice if the declared eslint version matched the installed version, but that's not a big deal.

denniskigen:
Thanks for the review, Ian!

Good call on 1.0.0. I'd originally planned to keep it on 0.x until the active repos were migrated, but the caret semantics make your ordering strictly better: `^0.1.0` only accepts `0.1.*`, so every config release during the rollout would need a version-bump PR in each already-migrated repo. `^1.0.0` absorbs minors, and the versioning policy in the README (new rules land as warn in a minor, error only in a major) does the actual stability work regardless of the number. I'll cut 1.0.0 before migrating the next repo and point this rollout at `^1.0.0` ranges.

On the eslint version: agreed, I'll raise the declared floor to the tested version (`^9.39.0`) as part of the 1.0.0 release so the peer range reflects what's actually been validated.

One note on the docs check: automatic regeneration was disabled here because the PR touches `turbo.json`, so I've triggered the on-demand rebuild with the attack-files bypass after auditing the diff (it only changes the lint task's cache inputs). The regenerated docs are already committed on the branch, so the rebuild should verify a zero diff and clear the drift check.

PR #1832, opened 2026-08-04
(fix) O3-5874: Handle regional locales in the change language modal

denniskigen:
## Requirements

## Summary
On installs whose allowed locales include a region (`en_US`, `sw_KE`), the change language modal crashes on open with `RangeError: Invalid language tag: en_US`. The user menu's language item crashes too, as soon as the menu opens.

## Description
The session endpoint reports `locale` and `allowedLocales` in Java's `Locale#toString()` form (`en_US`, `sw_KE`). Those aren't valid BCP 47 tags, and `Intl.DisplayNames` throws instead of degrading when handed one. Both the modal and the user menu language item pass the session values to `Intl.DisplayNames` raw. The framework normalizes underscores when it sets `document.lang`, but it never writes that back to the session store and never touches `allowedLocales`, so these components see the raw form. We ship regional translation bundles (`en_US`, `sw_KE`, `pt_BR`, `zh_CN`), so this hits supported configurations.

In this PR I have:

- Added `toLanguageTag` and `getLocaleDisplayName` helpers that normalize before calling `Intl`, and fall back to the raw identifier for anything `Intl` still rejects
- Used them in both components. The outbound value is unchanged, the backend still gets the underscore form it expects, and a test covers the round trip by selecting `sw_KE` and asserting the POSTed value
- Fixed the modal swallowing failed updates. `handleSubmit` never awaited the request, so a rejection left the spinner running forever with nothing shown. It now awaits and shows the backend's message, the same way the change password modal does
- Made the last locale in a long list visible. The list had a max-height with no overflow, so past roughly nine locales it overflowed into the ModalBody, and Carbon's scroll fade mask kept the last entry invisible even at full scroll (still clickable, though). The list now scrolls itself, with an end margin so the fade lands on empty space on short viewports. There's no unit test for this one since it's CSS layout; the screenshots below are the evidence

One thing this doesn't fix: selecting `uz@Latn` still fails, because the session POST parses the submitted value with `LocaleUtils.toLocale`, which rejects both `uz@Latn` and `uz-Latn`. Supporting it needs a backend change in webservices.rest. The modal now at least shows that error instead of hanging.

## Screenshots

Captured against a local dev server with the session's `allowedLocales` stubbed to include regional locales; the backend was not modified.

#### On main with an `en_US` session: RangeError from ChangeLanguageLink as soon as the app loads

image-1785876169621

#### "Italiano" obscured from view at the end of the locales list

CleanShot 2026-08-04 at 23 51 10@2x

#### User menu rendering the current language
image-1785876173430

#### Modal listing the full allowed-locale set, regional locales included

image-1785876176815

#### List scrolled to its end, last locales visible above Carbon's fade

image-1785876182434

#### Rejected update surfacing the backend's message, submit re-enabled

image-1785876186781

## Related Issue
https://openmrs.atlassian.net/browse/O3-5874


packages/apps/esm-primary-navigation-app/src/utils.ts
```diff
@@ -1,3 +1,30 @@
...
+
+/**
+ * Converts an OpenMRS locale identifier into a BCP 47 language tag.
+ *
+ * The REST API reports locales in Java's `Locale#toString()` form (`en_US`, `sw_KE`), and some
+ * configurations use POSIX-style variants (`uz@Latn`). Neither is a valid language tag, and the
+ * `Intl` constructors throw `RangeError` rather than degrading, so always convert before handing
+ * a locale to `Intl`.
+ */
+export function toLanguageTag(locale: string) {
```
ibacher: Should this be something the framework provides?

ibacher (approved):

packages/apps/esm-primary-navigation-app/src/utils.ts
denniskigen: Yes, and it's in this PR now. `toLanguageTag` and `getLocaleDisplayName` live in `esm-utils` and go out through the framework.

Turned out we had the rule three times already: `canonicalize` in `match-locale.ts`, the `replaceAll` in `setUserLanguage`, and the one in `getLocale` with the "just in case" comment. `getLocale` and `matchLocale` now share the helper, so this removes two of the three copies rather than adding a fourth. It canonicalizes via `Intl.Locale` rather than only swapping separators, so it fixes subtag casing too.

Two things I'd like your read on:

- `toLanguageTag` treats `@` as a separator, which none of the copies did. That's what the modal needs for `uz@Latn`, but it also means `matchLocale` now resolves `uz@Latn` instead of skipping it with a warning, and a malformed modifier that happens to parse as a subtag (`en@garbage` becomes `en-garbage`) will match rather than warn. Nothing calls `matchLocale` yet, so nothing depends on the old strictness, and it's called out in the changeset. I can narrow the conversion to `@` plus four letters, which is the only form we ship, if you'd rather keep it strict.

- `setUserLanguage` keeps its own underscore-only conversion on purpose. i18next reads that value back off `document.lang` and uses it as the translation bundle key, and the bundles are named `uz@Latn.json`, so converting `@` there would send the loader after a bundle that doesn't exist. There's a comment on it now saying so.

One side effect worth knowing about: `getLocale` returns canonical tags now, so legacy codes get rewritten (`iw` to `he`, `tl` to `fil`). Everything that consumes it passes the value to `Intl`, where that's what you want, except the `preferredDateLocale` lookup in the datepicker hook, which keys off it. That schema documents canonical keys, so it should only bite a map keyed `tl` or `en-us`.

packages/apps/esm-primary-navigation-app/src/components/change-language/change-language-link.test.tsx
```diff
@@ -20,9 +20,10 @@ describe('ChangeLanguageLink', () => {
     ['en', /english/i],
     ['fr', /français/i],
```
ibacher: We should probably also make these case-sensitive?

ibacher (approved):

packages/apps/esm-primary-navigation-app/src/components/change-language/change-language-link.test.tsx
denniskigen: Good catch, done!

PR #1840, opened 2026-08-12
(fix) Read past the first page in the docs workflow lookups

denniskigen:
## Requirements

## Summary

Two lookups in the docs workflows only ever saw the first page of results.

`gh api` returns 30 items per page, so an approval past the first 30 reviews was invisible to the docs-build preflight, which aborted with "No write-access approval present". [#1819](https://github.com/openmrs/openmrs-esm-core/pull/1819) hit this: 65 reviews, with both approvals at positions 64 and 65. No retry path could go green because the approval, push, and on-demand rebuild paths all share that lookup.

The check-run lookup had the same blind spot. #1819's head SHA has exactly 30 check runs, which is exactly the page size, and a recent `main` SHA has 118, so `Docs / Regenerated` can easily fall outside page 1.

Both lookups now paginate. `gh` applies `--jq` per page, and `--slurp` cannot be combined with `--jq`, so the review loop dedupes with `sort -u` instead of jq's `unique`, and the check-run lookups count matches with `wc -l` instead of jq's `length`, which would otherwise emit one count per page. `per_page=100` keeps the review lookup to a single request.

The check-run lookups also pass `check_name`, which filters server-side, so they stay at one request whatever else the SHA has accumulated. `--method GET` is load-bearing there: without it, `-f` makes the request a POST and it 404s.

I verified both shapes against live data. The review lookup finds the approval where the old one returned nothing. The check-run lookup returns 1 on #1823's head, which has a real successful `Docs / Regenerated`, and 0 on #1819's, which has none.

## Other

The label lookups are still single-page. A PR with more than 30 labels is not realistic, so I left them alone.

### DIFF
--- a/.github/workflows/docs-build-on-approval.yml
+++ b/.github/workflows/docs-build-on-approval.yml
@@ -58,8 +58,8 @@ jobs:
           # Only the bot's App token can post the Docs / Regenerated check run
           # (checks:write is App-scoped), so finding it on the head SHA is a
           # credible "docs are current" signal — skip the rebuild.
-          regenerated=$(gh api "repos/${REPO}/commits/${HEAD_SHA}/check-runs" \
-            --jq '[.check_runs[] | select(.name == "Docs / Regenerated" and .conclusion == "success")] | length')
+          regenerated=$(gh api --paginate "repos/${REPO}/commits/${HEAD_SHA}/check-runs" \
+            --jq '.check_runs[] | select(.name == "Docs / Regenerated" and .conclusion == "success") | .id' | wc -l)
           if [ "${regenerated}" -gt 0 ]; then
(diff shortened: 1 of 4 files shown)


ibacher (approved):

denniskigen:
Thanks for the review, @ibacher. I pushed another commit that reworks the approach: the check-run lookup now filters server-side with `check_name` instead of paginating and counting, so it stays at one request however many check runs the SHA picks up. `--method GET` is required there. Without it, `-f` makes the request a POST and the call 404s.

I checked the shapes against #1819's and #1823's live data and they seem to work fine.

Can you take another look?

ibacher (approved):
Yeah, updated code appears to work AFAICT (I ran 4 different PRs manually)

PR #1844, opened 2026-08-17
(feat) O3-5838: Implement TOTP verification page

dilankavishka:
## Requirements

## Summary
I raised this PR to introduce a new Time-based One-Time Password (TOTP) verification challenge flow to the login app, enhancing two-factor authentication support. It adds a dedicated page for users to enter TOTP codes and build a custom pin code input component which is used for the verification page.
The verification page is shown the first time a user enables 2FA and logs in. If the user selects “Remember this device”, the device will be remembered for 30 days, and the verification page will not appear again during that period. Once the 30-day period expires, the user will be asked to enter the TOTP code again.
If the user does not select “Remember this device”, they will be redirected to the verification page every time they log in.

## Screenshots
1. Full user flow:

https://github.com/user-attachments/assets/5617b81b-8e31-404f-92fa-8dc7320b959c

3. When user enters an invalid code:
Screenshot 2026-08-17 at 11 38 50

## Related Issue
https://issues.openmrs.org/browse/O3-5838

## Other
There are also two ongoing bug fixes in the authentication module backend. Here are the related tickets.
1. https://openmrs.atlassian.net/browse/AUT-32
2. https://openmrs.atlassian.net/browse/AUT-31


dilankavishka:
@denniskigen @jayasanka-sack @janithrs Could you review this PR please?

packages/apps/esm-login-app/src/root.component.tsx
```diff
@@ -16,6 +17,7 @@ const Root: React.FC = () => {
         <Route path="logout" element={<RedirectLogout />} />
         <Route path="change-password" element={<ChangePassword />} />
         <Route path="two-factor-auth" element={<TwoFactorAuth />} />
+        <Route path="login/totp-challenge" element={<TotpVerificationChallengePage />} />
```
denniskigen, line 20: The session request already follows the authentication module's `Location` header, so we shouldn't need another navigation call in the login page. We do need `authentication.scheme.totp.config.loginPage` to point to this SPA route; its default is still `/loginTotp.page`. Can we link the paired RefApp configuration change from [O3-5838](https://issues.openmrs.org/browse/O3-5838)?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+        },
+      });
+
+      if (response.data && response.data.authenticated) {
+        const sessionStore = await refetchCurrentUser();
+        const session = sessionStore?.session;
+
+        if (session?.sessionLocation) {
+          let to = loginLinks?.loginSuccess || '/home';
+          if (location?.state?.referrer) {
```
denniskigen: Can we carry the original referrer through the backend challenge redirect and cover the full flow? `/login` stores it only in React Router state, but the session `Location` is followed through `navigate()`, and single-spa creates the new history entry with null state. `location.state` is therefore null here, so successful verification falls back to `loginSuccess`.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+    try {
+      setIsVerifying(true);
+      setVerificationError('');
+
+      const sessionUrl = rememberDevice ? '/ws/rest/v1/session?rememberMe=true' : '/ws/rest/v1/session';
+
+      const response = await openmrsFetch(sessionUrl, {
+        method: 'GET',
+        headers: {
+          'X-Totp-Code': code,
```
denniskigen, line 45: Can we avoid hardcoding the challenge contract here? `TotpAuthenticationScheme` supports configurable `codeHeader` and `codeLength`. A deployment that changes either cannot complete this screen.

It's probably better to expose the effective values to the frontend, or make `X-Totp-Code` and six digits the fixed contract for this flow. The enrollment UI already assumes six digits, so that part may need a follow-up.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const [verificationError, setVerificationError] = useState('');
+
+  const handleVerify = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    try {
+      setIsVerifying(true);
+      setVerificationError('');
+
+      const sessionUrl = rememberDevice ? '/ws/rest/v1/session?rememberMe=true' : '/ws/rest/v1/session';
```
denniskigen, line 40: Can we line this up with the backend configuration? `rememberMeEnabled` defaults to false, and both the duration and request parameter name are configurable. As written, this can offer a checkbox that does nothing or promise the wrong duration.

We shouldn't add a second source of truth in frontend config. The effective backend capability should drive the UI.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+          }
+          openmrsNavigate({ to });
+        } else {
+          navigate('/login/location');
+        }
+      } else {
+        setVerificationError(t('totpVerificationFailed', 'We could not verify that code. Please try again.'));
+      }
+    } catch (error: unknown) {
+      let errorMessage = t('invalidVerificationCode', 'Invalid verification code. Please try again.');
```
denniskigen: Can we distinguish request failures from rejected codes here? I took the app offline with a complete code and this showed “Invalid verification code.” A failed session refresh after an accepted code lands here too.

It's probably better to disable submission while offline and reserve this message for an authentication rejection.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+          </div>
+
+          <Button
+            type="submit"
+            className={styles.verifyButton}
+            renderIcon={(props) => <ArrowRightIcon size={24} {...props} />}
+          >
+            {isVerifying ? t('verifying', 'Verifying...') : t('verify', 'Verify')}
+          </Button>
+        </form>
```
denniskigen: Can we include a way to cancel this challenge and start over, clearing the partial server session before returning to login? A plain link to `/login` will bring the user back here because the backend retains the primary-auth candidate, and `/logout` currently skips its session DELETE while the user is only partially authenticated.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+    <div className={styles.wrapper}>
+      <label className={styles.label}>{t('enterCode', 'Enter the code from your authenticator app')}</label>
+      <div className={styles.otpContainer} onPaste={handlePaste}>
+        {otp.map((data, index) => (
+          <TextInput
+            className={styles.otpInput}
+            key={index}
+            id={`otp-input-${index}`}
+            type="tel"
+            inputMode="numeric"
```
denniskigen:
```suggestion
type="text"
inputMode="numeric"
pattern="[0-9]*"
```

`tel` describes a telephone number. `inputMode="numeric"` is what gives us the numeric keyboard here.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+            autoComplete="one-time-code"
+            autoFocus={index === 0}
+            maxLength={1}
+            value={data}
+            labelText={`Digit ${index + 1}`}
```
denniskigen:
```suggestion
labelText={t('digitLabel', 'Digit {{number}}', { number: index + 1 })}
```

This label is visually hidden, but it's still the input's accessible name. It needs to be translated.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+      onComplete(pastedData);
+      inputRefs.current[length - 1]?.focus();
+    } else {
+      inputRefs.current[pastedData.length]?.focus();
+    }
+  };
+
+  return (
+    <div className={styles.wrapper}>
+      <label className={styles.label}>{t('enterCode', 'Enter the code from your authenticator app')}</label>
```
denniskigen: Can we make this a `legend` for a `fieldset` containing the six inputs? This label isn't associated with anything, so assistive tech gets six “Digit N” fields without the authenticator-code context.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+    } else if (event.key === 'ArrowRight' && index < length - 1) {
+      inputRefs.current[index + 1]?.focus();
+    }
+  };
+
+  // Handle copy and paste to allow pasting the entire code
+  const handlePaste = (event: React.ClipboardEvent<HTMLInputElement>) => {
+    event.preventDefault();
+
+    const pastedData = event.clipboardData.getData('text').slice(0, length);
```
denniskigen:
```suggestion
const pastedData = event.clipboardData.getData('text').replace(/\s/g, '').slice(0, length);
```

Authenticator apps often display codes as `123 456`, and the backend deliberately accepts whitespace. This currently prevents the paste and silently drops it.

denniskigen (commented):
Good work, @dilankavishka. The pin code input feels nice to use and its tests are a good start. I've left some suggestions below, mainly around the challenge handoff, keeping the submitted code in sync with the boxes, and matching the authentication module's configuration.

Can we add a changeset and use [O3-5838](https://issues.openmrs.org/browse/O3-5838) to link the AUT-24 module release, the paired RefApp config and version bump, and clarify whether AUT-31 and AUT-32 are release dependencies?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.scss
```diff
@@ -0,0 +1,28 @@
+@use '@carbon/layout';
+@use '@carbon/type';
+
+.otpContainer {
+  display: flex;
```
denniskigen, line 5:
```suggestion
.otpContainer {
  direction: ltr;
  display: flex;
```

Authenticator codes are fixed left-to-right sequences. Without this, Digit 1 appears on the right and focus advances left in Arabic and Hebrew locales.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+          />
+        </div>
+      )}
+      <Tile className={loginStyles.loginCard}>
+        <div className={loginStyles.center}>
+          <Logo t={t} />
+        </div>
+
+        <div className={styles.title}>
+          <h5>{t('twoFactorVerification', 'Two-Factor Verification')}</h5>
```
denniskigen: Can we make this the page-level `h1` and keep the current visual size in CSS? It's the only heading on this route, so starting at `h5` skips the document hierarchy.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+      setVerificationError(errorMessage);
+    } finally {
+      setIsVerifying(false);
+    }
+  };
+
+  return (
+    <div className={containerClassName} style={containerStyle}>
+      {verificationError && (
+        <div className={loginStyles.announcements}>
```
denniskigen: nit: Can we center this against the login card? At narrow widths Carbon caps the notification at `18rem` while this wrapper stays `23rem`, so it sits against the left edge. The notification text should stay left-aligned. Feel free to ignore this one.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+          <InlineNotification
+            kind="error"
+            title={t('error', 'Error')}
```
denniskigen:
```suggestion
            title={getCoreTranslation('error')}
```

nit: Core translations already have this one, and the login page uses it. We can drop the duplicate `error` key from this app.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+      if (response.data && response.data.authenticated) {
+        const sessionStore = await refetchCurrentUser();
+        const session = sessionStore?.session;
+
+        if (session?.sessionLocation) {
+          let to = loginLinks?.loginSuccess || '/home';
+          if (location?.state?.referrer) {
+            if (location.state.referrer.startsWith('/')) {
+              to = `\${window.getOpenmrsSpaBase()}${location.state.referrer}`;
```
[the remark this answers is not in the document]
denniskigen:
```suggestion
to = `\${openmrsSpaBase}${location.state.referrer}`;
```

The finding is right, but it's probably better to match the existing login flow here. `navigate()` only interpolates `${openmrsBase}` and `${openmrsSpaBase}`. `getOpenmrsSpaBase()` includes a trailing slash, so evaluating it directly as suggested gives us a double slash.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+    // Take only the last character
+    const newOtp = [...otp];
+    newOtp[index] = value.substring(value.length - 1);
+    setOtp(newOtp);
+
+    // If the user filled the last box, trigger onComplete prop
+    const combinedOtp = newOtp.join('');
+    if (combinedOtp.length === length) {
+      onComplete(combinedOtp);
+    }
```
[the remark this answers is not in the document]
denniskigen: After entering `123456`, click Digit 3, press Backspace, then Verify. The inputs show `12_456`, but the request still sends `X-Totp-Code: 123456` because the parent is only notified when all six boxes are filled. A partial paste over a completed code does the same.

Can we report every accepted change to the parent and disable Verify until the current value has six digits, and while verification is in progress? It'd be good to cover deletion followed by submission in a page-level test too.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
janithRS: Agree with Dennis here. Remember me should be enabled by default.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+    if (background.image) {
+      return { '--login-bg-image': `url(${interpolateUrl(background.image)})` } as React.CSSProperties;
+    }
+    if (background.color) {
+      return { '--login-bg-color': background.color } as React.CSSProperties;
+    }
+    return undefined;
+  }, [background]);
+  const [code, setCode] = useState('');
+  const [rememberDevice, setRememberDevice] = useState(false);
```
janithRS: Since we have an interaction here with the checkbox, I would suggest to add a test to cover that scenario. Since the request URL is dynamically built based on the input, it's worth having an assertion here. WDYT?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+            if (location.state.referrer.startsWith('/')) {
+              to = `\${window.getOpenmrsSpaBase()}${location.state.referrer}`;
+            }
+          }
+          openmrsNavigate({ to });
+        } else {
+          navigate('/login/location');
+        }
+      } else {
+        setVerificationError(t('totpVerificationFailed', 'We could not verify that code. Please try again.'));
```
janithRS, line 66: nit: To me `We could not verify that code. Please try again` and `Invalid verification code. Please try again.` sounds the same. Wouldn't it be better to merge those 2?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    // Only numbers are allowed to paste
+    if (!/^\d+$/.test(pastedData)) return;
+
+    // Spread the pasted code across the input boxes
+    const newOtp = [...otp];
+    pastedData.split('').forEach((char, index) => {
+      newOtp[index] = char;
+    });
+    setOtp(newOtp);
```
janithRS, line 66: Looks like paste only overwrites indices 0..pastedData.length-1. If paste shorter than length and boxes after already held digits, stale digits stay (e.g. "123456" then paste "99" at start gives "993456" not "990000"). Is this intentional? If not we need to clear rest of array first.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+    if (!/^[0-9]*$/.test(value)) return;
+
+    // Take only the last character
+    const newOtp = [...otp];
+    newOtp[index] = value.substring(value.length - 1);
+    setOtp(newOtp);
+
+    // If the user filled the last box, trigger onComplete prop
+    const combinedOtp = newOtp.join('');
+    if (combinedOtp.length === length) {
```
janithRS: Here the onComplete fires every time combined length hits length, even repeat edits after already complete, fine if parent is idempotent, worth confirming.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>, index: number) => {
+    if (event.key === 'Backspace') {
+      if (!otp[index] && index > 0 && inputRefs.current[index - 1]) {
+        inputRefs.current[index - 1]?.focus();
+      }
+    } else if (event.key === 'ArrowLeft' && index > 0) {
+      event.preventDefault();
+      inputRefs.current[index - 1]?.focus();
+    } else if (event.key === 'ArrowRight' && index < length - 1) {
+      inputRefs.current[index + 1]?.focus();
```
janithRS, line 48: Missing event.preventDefault() unlike ArrowLeft, probably harmless but asymmetric.

janithRS:
Overall nice job Dilan, The UI feels really nice and stable. I've left a few comments and suggestions, and something which needed confirmation 👍🏽

packages/apps/esm-login-app/src/root.component.tsx
janithRS: A bit confused about this comment. @denniskigen the redirect URL always comes from the runtime properties config. So I think there is no manual redirect path set. Also looks like it's the totp login page getting registered here. Am I missing something?

denniskigen: You're right about the mechanism, @janithRS. The gap is the default: `authentication.scheme.totp.config.loginPage` defaults to `/loginTotp.page` ([TotpAuthenticationScheme.java#L90](https://github.com/openmrs/openmrs-module-authentication/blob/main/omod/src/main/java/org/openmrs/module/authentication/web/TotpAuthenticationScheme.java#L90)), not this route. So nothing redirects here until a deployment sets that property to `/spa/login/totp-challenge`. Can we track that config change on [O3-5838](https://issues.openmrs.org/browse/O3-5838) so it doesn't get lost? Without it, this page ships unreachable.

janithRS: I see, yeah that's correct.

jayasanka-sack: The default config is pointing to the O2 TOTP page. When configuring TOTP, we’re expected to set the redirect URL explicitly instead of relying on the defaults.

And yes, we should update those defaults since we’re moving away from the O2 module, but that shouldn’t block this PR.

We still need to register the routes, and then configure the module to use those routes. Not the other way around where the UI has to match whatever happens to be configured in the module.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: I actually just refactored this . `onComplete` now fires on every single change (not just when it hits the target length) so the parent always has the most up to date state.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: Thank you for this. this is fixed now :)

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Added a new test case with this

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.scss
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: Updated.

Note: https://github.com/openmrs/openmrs-module-authentication/pull/55 was the fix that i did to remove the whitespaces :)

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Add a new `Back to login` button for this DELETE action

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Fixed

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Now, the Remember me was enabled by default. 

Note: At the platform call discussion, we came up with a decision that not to show the `Remember This device for 30 days`, instead we're only showing the `Remember This device`. @jayasanka-sack would you like to help me in here to explain this more. :D

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: @jayasanka-sack need your help in here as well to describe what we discussed in the Platform call. :)

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Fixed

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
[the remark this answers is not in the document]
dilankavishka: Fixed

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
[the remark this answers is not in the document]
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: this is now refactored.

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -125,14 +133,14 @@ const Login: React.FC = () => {
...
           if (session.sessionLocation) {
             let to = loginLinks?.loginSuccess || '/home';
-            if (location?.state?.referrer) {
-              // Only accept relative paths; absolute or protocol-relative referrers
-              // are silently ignored to prevent open-redirect attacks after login.
-              if (location.state.referrer.startsWith('/')) {
-                to = `\${openmrsSpaBase}${location.state.referrer}`;
-              }
-            }
+            const referrer = location?.state?.referrer || sessionStorage.getItem('loginReferrer');
```
ibacher:
```suggestion
            const referrer = location?.state?.referrer ?? sessionStorage.getItem('loginReferrer');
```

packages/apps/esm-login-app/src/login/login.test.tsx
```diff
@@ -347,4 +348,14 @@ describe('Login', () => {
...
+
+  it('saves the referrer to session storage if present in router state', () => {
+    render(
+      <MemoryRouter initialEntries={[{ pathname: '/login', state: { referrer: '/patient-chart' } }]}>
+        <Login />
+      </MemoryRouter>,
+    );
+
+    expect(sessionStorage.getItem('loginReferrer')).toBe('/patient-chart');
+  });
```
ibacher: Is there a way we can write this test to test the thing we care about (the user is sent to the appropriate referrer) instead of an implementation detail (how we store that while processing).

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,189 @@
...
+  });
+  const containerStyle = useMemo<React.CSSProperties | undefined>(() => {
+    if (background.image) {
+      return { '--login-bg-image': `url(${interpolateUrl(background.image)})` } as React.CSSProperties;
+    }
+    if (background.color) {
+      return { '--login-bg-color': background.color } as React.CSSProperties;
+    }
+    return undefined;
+  }, [background]);
```
ibacher: I think stuff like this actually suggests we should do a bigger refactor and come up with a "login page wrapper component" into which the login component and this slot themselves. That wrapper component can handle the shared commonalities of the two pages, etc.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
ibacher: Yeah, this is what Jaye and I were arguing about. My solution is that we modify the backend to send the expected header (or whatever else the backend will accept) name with the redirect request. Jaye was of the opinion that we hard-code this and just require implementations configure things this way.

Broadly, either of those two options will work. I really don't have a particularly good argument for my stance other than I think it's preferable to make the UI work for as many configurations as possible.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,189 @@
...
+    } catch (error: unknown) {
+      let errorMessage = t('verificationFailedError', 'A network or server error occurred. Please try again.');
+
+      if (error instanceof OpenmrsFetchError) {
+        if (error.response?.status === 401 || error.response?.status === 403) {
+          errorMessage = t('invalidCode', 'Invalid verification code. Please try again.');
+        }
+        if (typeof error.responseBody === 'object' && error.responseBody !== null) {
+          const body = error.responseBody as { error?: { translatedMessage?: string } };
+          const translatedMessage = body.error?.translatedMessage;
```
ibacher: Maybe?

```suggestion
          const translatedMessage = body.error?.translatedMessage ?? error.message;
```

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
ibacher: The PR doesn't match that assertion.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
ibacher: Also doesn't seem to be done.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+          <div className={styles.actionButtons}>
+            <Button
+              type="button"
+              className={styles.backToLoginButton}
+              kind="ghost"
+              onClick={backToLogin}
+              disabled={isVerifying || isBackToLogin}
+            >
+              <ArrowLeftIcon size={10} />
```
ibacher: I didn't see this in the video, but 10 feels kind of small. Is it possible to share a screenshot of this rendered?

packages/apps/esm-login-app/src/root.component.tsx
ibacher: I just want to challenge the part where you're saying "we're moving away from the O2 module". We're not. The defaults should remain primed to O2 for now. Mike Seaton wrote the authentication module basically himself; his implementation needs the O2 defaults, so we'll leave them as-is, but write this up as part of the docs on enable this on O3.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const backToLogin = useCallback(
+    async (event: React.MouseEvent) => {
+      event.preventDefault();
+      setIsBackToLogin(true);
+      try {
+        await performLogout();
+        navigate('/login');
+      } catch (error) {
+        setVerificationError(
+          t('backToLoginFailed', 'Failed to cancel the verification process. Please refresh the page and try again.'),
```
ibacher: I'm not sure I understand what this is trying to achieve? If the session delete failed, the user will still have a cookie for the session, which is likely still valid, so on refresh, wouldn't they just hit this page again?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    // If the user filled the last box, trigger onComplete prop
+    const combinedOtp = newOtp.join('');
+    onComplete(combinedOtp);
+
+    // Move focus to the next input box if the current one is filled
+    if (value && index < length - 1 && inputRefs.current[index + 1]) {
+      inputRefs.current[index + 1]?.focus();
+    }
+  };
```
ibacher, line 35: Have we tested entering fields out of order?

ibacher (commented):
I think this mostly looks code, but weighing in on a few things. @dilankavishka I kind of think there might've been a commit responsive to some of Dennis's comments that didn't get pushed?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: This is how it looks like when the size is **10**:
Screenshot 2026-08-21 at 07 31 37

And here how it's looks like when the size is **15**:
Screenshot 2026-08-21 at 07 33 24

shall I make it to 15?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: To keep the scope of this PR focused on the totp feature, would you prefer I create a ticket and handle that refactor in a follow-up PR? Or would you rather I include the refactoring as part of this PR before we merge?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
ibacher: How about 16? That's one of Carbon's "standard" sizes.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
ibacher: Well, I'm suggesting it precisely because you need to duplicate code here, so I think this will actually have the effect of introducing _fewer_ new lines.

ibacher: But we can move it to a future PR.

packages/apps/esm-login-app/src/login/login.component.tsx
dilankavishka: updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Now we're using `LoginPageWrapper` shared component

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Updated

packages/apps/esm-login-app/src/login/login.test.tsx
dilankavishka: Updated! I rewrote this test to simulate a successful login and assert that openmrsNavigate is actually called with the correct referrer URL

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: I added a new test case for this

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Updated the error message

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: Perfect! I updated that to 16 :)

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: I created a new `LoginPageWrapper` component. Both the `Login` component and the `TotpVerificationChallengePage` now act as simple children slotted into this wrapper. I also moved all the relevant layout test cases into a new `login-page-wrapper.test.tsx` file.

One quick question regarding the `Footer`, I currently included it in the shared wrapper so it displays on both pages. If we prefer to only show the footer on the main login page, I can either move it back out of the wrapper, or keep it in the wrapper but hide it specifically on the totp verification page. I'm personally fine with keeping it visible on both for consistency, but what do you think is best?

jayasanka-sack:
@dilankavishka please fix the items we discussed during today’s code review and let us know once they’re done.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
denniskigen: We haven't fixed the original issue here. Checking the box by default only changes the frontend. The backend still defaults `rememberMeEnabled` to `false`, and when it is disabled it ignores `rememberMe=true`, so the user can be told the device is remembered and still be challenged on the next login.

It's probably best to hide this until the backend exposes the effective capability and parameter name. If we're fixing these values as the O3 contract instead, can we land and link the paired RefApp configuration first?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
denniskigen: The handoff itself looks fixed, but I can't see the regression test from this thread. The new login test succeeds directly while router state is still present, so it never mounts the TOTP page or exercises the `sessionStorage` fallback. Can we cover the challenge handoff and assert that the original referrer is used after TOTP verification?

dilankavishka: Added the test for this

dilankavishka:
@jayasanka-sack I have addressed the issues we discussed during the code review. could you please take another look when you have a moment?

- Refactored back to login flow. It now correctly goes to the /logout route.
- Removed the 403 status check.
- Added a showFooter prop to the LoginPageWrapper (the footer correctly only appears on the main login page and is hidden on the TOTP verification page)
- Refactored the test suite and added new test cases.

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
dilankavishka: @denniskigen, I understand your concern about the default state causing a false promise. The backend does support this via the `authentication.scheme.2fa.config.rememberMeEnabled=true` runtime property. since we wants this enabled by default in O3, should we establish this property as part of the standard O3 contract?

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/totp-verification-challenge-page.component.tsx
```diff
@@ -0,0 +1,138 @@
...
+      } else {
+        setVerificationError(t('totpVerificationFailed', 'We could not verify that code. Please try again.'));
+      }
+    } catch (error: unknown) {
+      let errorMessage = t('verificationFailedError', 'A network or server error occurred. Please try again.');
+
+      if (error instanceof OpenmrsFetchError) {
+        if (error.response?.status === 401) {
+          errorMessage = t('invalidCode', 'Invalid verification code. Please try again.');
+        }
```
jayasanka-sack: This block is basically dead code as the session endpoint never returns a 401. If the code is invalid it will send `{ authenticated: false, ... }` in the body. Which is what the else block on above catches. So we don't need this check.

dilankavishka: Updated!

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+import styles from './verification-code-input.scss';
+
+interface VerificationCodeInputProps {
+  length: number;
+  onChange: (code: string) => void;
+}
+
+const VerificationCodeInput: React.FC<VerificationCodeInputProps> = ({ length = 6, onChange }) => {
+  const { t } = useTranslation();
+  const [otp, setOtp] = useState<string[]>(new Array(length).fill(''));
```
jayasanka-sack: Currently the otp lives in two places (this one and in the parent component). I'd make this a controlled component.

Happy to hop on a call if you want to discuss this further.

dilankavishka: Please let me know a convenient time for you. Let's take a call :)

jayasanka-sack (approved):
Merging this after our call and leaving the controlled component thingy as a follow-up enhancement. Thanks @dilankavishka!

packages/apps/esm-login-app/src/two-factor-auth/verification-challenge/verification-code-input.component.tsx
dilankavishka: @jayasanka-sack!

As we discussed in our call, I'll update this to be a controlled component. Since I also need to address the autofill issue, and reuse this 6-box component inside the TOTP enrollment modal as well with a follow-up PR.

I've created a ticket to tackle all three of these at once: [HERE](https://openmrs.atlassian.net/browse/O3-6006)

PR #1853, opened 2026-08-25
(feat) Federate most of the Module Federation runtime

ibacher:
## Requirements

## Summary

Effectively, this moves most of the Module Federation runtime into the app-shell, which means that apps built with a version of the rspack or webpack config after this PR is merged can no longer be run on _older_ versions of the app shell.

RSpack bundles the Module Federation runtime into every app. This is desirable for some configurations, as it guarantees the app can always be loaded. However, it also adds some overhead to the initial chunk of every app (~85kB of which this PR reclaims ~50kB).

As an unintended consequence, this now puts us fully on Module Federation 2.0.

I've verified with testing that we, at least, don't duplicate the framework, even with older apps. Other federated libraries may accidentally be duplicated, but apart from SWR (where our framework already overwrites its internal cache), I'm not aware of any place where we are dependent on library-global state. Basically, this means this should be safe.

### DIFF
--- a/.changeset/wacky-pens-refuse.md
+++ b/.changeset/wacky-pens-refuse.md
@@ -5,3 +5,5 @@
 ---
 
 (feat) Federate most of the Module Federation runtime
+
+Apps built with these versions of `@openmrs/rspack-config` or `@openmrs/webpack-config` require an app shell that provides the Module Federation runtime. Deploy the app shell first. Rolling the shell back to a pre-change version while those apps remain deployed will prevent them from starting. The new shell remains compatible with apps built using older tooling.
--- a/.changeset/wacky-pens-refuse.md
+++ b/.changeset/wacky-pens-refuse.md
@@ -1,0 +1,7 @@
+---
+"@openmrs/esm-app-shell": major
+"@openmrs/rspack-config": major
+"@openmrs/webpack-config": major
+---
+
+(feat) Federate most of the Module Federation runtime
--- a/packages/shell/esm-app-shell/src/federation-runtime.ts
+++ b/packages/shell/esm-app-shell/src/federation-runtime.ts
@@ -15,11 +15,11 @@ import * as sdk from '@module-federation/sdk';
 
 declare global {
   // `var` is what declares a property on `globalThis`; `let`/`const` do not.
-  // eslint-disable-next-line no-var
-  var _FEDERATION_ERROR_CODES: typeof errorCodes;
-  // eslint-disable-next-line no-var
-  var _FEDERATION_SDK: typeof sdk;
-  // eslint-disable-next-line no-var
+   
+  var _OPENMRS_FEDERATION_ERROR_CODES: typeof errorCodes;
+   
+  var _OPENMRS_FEDERATION_SDK: typeof sdk;
+   
   var _FEDERATION_RUNTIME_CORE: unknown;
 }
 
@@ -28,8 +28,8 @@ declare global {
  * ever declares `sideEffects: false`.
  */
 export function publishFederationRuntime() {
-  globalThis._FEDERATION_ERROR_CODES = errorCodes;
-  globalThis._FEDERATION_SDK = sdk;
+  globalThis._OPENMRS_FEDERATION_ERROR_CODES = errorCodes;
+  globalThis._OPENMRS_FEDERATION_SDK = sdk;
 
   // Reporting this here means a mis-built app shell says so itself, rather than every app in the
   // distribution claiming the app shell is too old.
--- a/packages/shell/esm-app-shell/src/federation-runtime.ts
+++ b/packages/shell/esm-app-shell/src/federation-runtime.ts
@@ -1,0 +1,42 @@
+/**
+ * Remote entries built by `@openmrs/rspack-config` or `@openmrs/webpack-config` read the Module
+ * Federation runtime from globals instead of embedding a copy each, so the app shell has to publish it.
+ *
+ * `@module-federation/runtime-core` is published by Module Federation itself, via the runtime plugin
+ * that `experiments.provideExternalRuntime` injects. Because it is this build's copy, its compiled-in
+ * build identifier and `experiments.optimization` settings are the ones every app's federation instance
+ * ends up with. The two helpers below are published here because that doesn't cover them; they hold no
+ * cross-build state. `@module-federation/runtime` and `webpack-bundler-runtime` are deliberately not
+ * published — they cache a federation instance at module scope, so a remote using the app shell's copy
+ * dies with RUNTIME-010.
+ */
+import * as errorCodes from '@module-federation/error-codes';
+import * as sdk from '@module-federation/sdk';
+
+declare global {
+  // `var` is what declares a property on `globalThis`; `let`/`const` do not.
+  // eslint-disable-next-line no-var
+  var _FEDERATION_ERROR_CODES: typeof errorCodes;
+  // eslint-disable-next-line no-var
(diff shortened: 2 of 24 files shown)


packages/shell/esm-app-shell/src/federation-runtime.ts
ibacher, line 1: This is now the "federation runtime". Note, though, that we're basically just creating global variables to hold it's state.

packages/tooling/integration-tests/package.json
ibacher, line 1: Integration tests added (but not published)

packages/tooling/rspack-config/src/index.ts
```diff
@@ -59,7 +59,44 @@ type OpenmrsRspackConfig = Omit<Partial<RspackConfiguration>, 'module'> & {
...
+const moduleFederationPin: string = require('../package.json').dependencies['@module-federation/enhanced'];
+const moduleFederationVersion = parse(moduleFederationPin);
+
+/**
+ * Prepended to this app's entry chunks. Without it, an app running under an app shell too old to
+ * publish the runtime globals fails with a `TypeError` from inside minified runtime code. All three
+ * globals are checked because `_FEDERATION_ERROR_CODES` is what describes the failure, and a runtime
+ * minor differing from the app shell's warns. `@openmrs/webpack-config` has a copy of this; keep them in step.
+ */
+function buildFederationRuntimeGuard(appName: string, expectedMinor: string | undefined) {
```
ibacher, line 76: This is the bit of this PR I don't love, but it's the best way to get us a useful error is something goes wrong.

packages/tooling/rspack-config/src/index.ts
```diff
@@ -59,7 +59,44 @@ type OpenmrsRspackConfig = Omit<Partial<RspackConfiguration>, 'module'> & {
...
+  const missingRuntime =
+    "(function(){var g=typeof globalThis!=='undefined'?globalThis:self;" +
+    "if(typeof g._FEDERATION_RUNTIME_CORE==='undefined'||typeof g._FEDERATION_SDK==='undefined'||typeof g._FEDERATION_ERROR_CODES==='undefined'){" +
+    'throw new Error(' +
+    JSON.stringify(appName) +
+    " + ' cannot start: the OpenMRS app shell serving this page does not provide the Module Federation runtime. " +
+    'This app was built with newer OpenMRS tooling than the app shell, so either upgrade @openmrs/esm-app-shell, ' +
+    "or rebuild this app with tooling matching the app shell.');}";
```
ibacher, line 84: This is the weird failure condition: deploying a _newer_ app on an _older_ app shell will not work since the older app shell will not provide the runtime the newer app expects.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -60,7 +62,44 @@ type OpenmrsWebpackConfig = Omit<Partial<WebpackConfiguration>, 'module' | 'outp
...
+const moduleFederationPin: string = require('../package.json').dependencies['@module-federation/enhanced'];
+const moduleFederationVersion = parse(moduleFederationPin);
+
+/**
+ * Prepended to this app's entry chunks. Without it, an app running under an app shell too old to
+ * publish the runtime globals fails with a `TypeError` from inside minified runtime code. All three
+ * globals are checked because `_FEDERATION_ERROR_CODES` is what describes the failure, and a runtime
+ * minor differing from the app shell's warns. `@openmrs/rspack-config` has a copy of this; keep them in step.
+ */
+function buildFederationRuntimeGuard(appName: string, expectedMinor: string | undefined) {
```
ibacher, line 79: Same ugly code, but Webpack version

packages/tooling/integration-tests/turbo.json
```diff
@@ -0,0 +1,27 @@
...
+  "tasks": {
+    "test": {
+      "inputs": [
+        "src/**",
+        "__fixtures__/**",
+        "vitest.config.ts",
+        "$TURBO_ROOT$/packages/tooling/rspack-config/src/**",
+        "$TURBO_ROOT$/packages/tooling/webpack-config/src/**",
+        "$TURBO_ROOT$/packages/tooling/rspack-config/package.json",
+        "$TURBO_ROOT$/packages/tooling/webpack-config/package.json"
```
denniskigen:
```suggestion
        "$TURBO_ROOT$/packages/tooling/webpack-config/package.json",
        "$TURBO_ROOT$/packages/shell/esm-app-shell/package.json"
```
Can we add this to both `test` and `coverage`? The pin test reads the app shell’s `package.json`, but it isn’t part of the cache key. I tested changing the pin, and Turbo replayed the cached pass even though the test failed when run directly.

.changeset/wacky-pens-refuse.md
```diff
@@ -0,0 +1,7 @@
+---
+"@openmrs/esm-app-shell": major
+"@openmrs/rspack-config": major
+"@openmrs/webpack-config": major
+---
+
+(feat) Federate most of the Module Federation runtime
```
denniskigen, line 7:
```suggestion
(feat) Federate most of the Module Federation runtime

Apps built with these versions of `@openmrs/rspack-config` or `@openmrs/webpack-config` require an app shell that provides the Module Federation runtime. Deploy the app shell first. Rolling the shell back to a pre-change version while those apps remain deployed will prevent them from starting. The new shell remains compatible with apps built using older tooling.
```
Can we document the rollout and rollback contract here? This changeset drives three major releases, but the current release note doesn’t explain the compatibility requirement.

packages/shell/esm-app-shell/src/federation-runtime.ts
```diff
@@ -0,0 +1,42 @@
...
+  var _FEDERATION_RUNTIME_CORE: unknown;
+}
+
+/**
+ * A function rather than a module side effect so the call can't be tree-shaken away if the app shell
+ * ever declares `sideEffects: false`.
+ */
+export function publishFederationRuntime() {
+  globalThis._FEDERATION_ERROR_CODES = errorCodes;
+  globalThis._FEDERATION_SDK = sdk;
```
denniskigen: nit: Can we use names we own for these two globals, perhaps `_OPENMRS_FEDERATION_SDK` and `_OPENMRS_FEDERATION_ERROR_CODES`? `_FEDERATION_RUNTIME_CORE` and `_FEDERATION_RUNTIME_CORE_FROM` belong to Module Federation, but these two are ours. Keeping them in the same namespace risks a collision if upstream adds them later, and both sides of the contract are still cheap to rename here.

.changeset/wacky-pens-refuse.md
```diff
@@ -0,0 +1,7 @@
+---
+"@openmrs/esm-app-shell": major
+"@openmrs/rspack-config": major
+"@openmrs/webpack-config": major
```
denniskigen, line 4: Shouldn’t `openmrs` be a major here too? `openmrs/default-rspack-config` and `openmrs/default-webpack-config` directly re-export these configs, and that’s what most apps import. Otherwise a patch release of `openmrs` could start producing apps that require the new app shell.

ibacher: One thing: I sorted resisted marking this as BREAKING because it's only breaking in the reverse sense, i.e., the new version of the app shell can correctly run modules built from with the older version of the tool, but modules built with the newer tooling need a newer version of the shell, which is more or less the rule we've always required anyways.

denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1855, opened 2026-08-27
(fix) Import map and routes should be cached

ibacher:
## Requirements

## Summary

I realized today while looking at network logs that there are circumstances where we are fetching the importmap multiple times in a single page. It turns out that that was due to the fact that `getCurrentPageMap()` was constantly re-fetching and re-parsing the import map _and_ that `getCurrentPageMap()` is called by default in `importDynamic()` when an app is loaded for the first time, so effectively, we were fetching one import map per-app, even though we don't expect the contents to change.

I had previously realized that even though we pre-fetch the import map and routes registry, these pre-fetches didn't actually take, but had never looked into why.

This PR fixes both of these things, so that under ideal circumstances (single import map or routes registry loaded successfully in pre-fetch), there is only ever _one_ request to either of these, the browser doing the pre-fetch. However, there are some allowance in the code so that if fetching an import map or routes _fails_ that result is not cached, so that the app has a chance to recover; no recovery steps are added here, but are something we may consider in the future.

## Other

Let me know if this description is unclear; I know this is kind of deeper in the weeds than most people are.

### DIFF
--- a/packages/framework/esm-dynamic-loading/src/import-maps.test.ts
+++ b/packages/framework/esm-dynamic-loading/src/import-maps.test.ts
@@ -347,7 +347,7 @@ describe('import-maps', () => {
       vi.resetModules();
     });
 
-    it('fetches the import map without credentials so the preloaded response is reused', async () => {
+    it('fetches the import map with the default credentials mode so the preloaded response is reused', async () => {
       setRemoteImportMap();
       fetchMock.mockResponse(JSON.stringify({ imports: { '@openmrs/esm-remote': '/remote.js' } }));
 
@@ -356,7 +356,9 @@ describe('import-maps', () => {
 
       await getCurrentPageMap();
 
-      expect(fetchMock).toHaveBeenCalledWith('http://localhost/importmap.json', { credentials: 'omit' });
+      // `crossorigin="anonymous"` on the preload link gives it a `same-origin` credentials mode,
+      // which is also `fetch()`'s default; passing anything else here misses the preload cache.
+      expect(fetchMock).toHaveBeenCalledWith('http://localhost/importmap.json');
     });
 
     it('fetches the import map only once across repeated reads', async () => {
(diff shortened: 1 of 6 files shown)


packages/framework/esm-dynamic-loading/src/import-maps.ts
```diff
@@ -12,30 +12,53 @@ let devMode = false;
...
   const maps: ImportMap[] = [];
+  let complete = true;
 
   for (let i = 0; i < scripts.length; i++) {
     const script = scripts[i];
     try {
       if (script.src) {
-        const response = await fetch(script.src);
+        // `credentials: 'omit'` matches the `crossorigin="anonymous"` preload
+        const response = await fetch(script.src, { credentials: 'omit' });
```
denniskigen: [`crossorigin="anonymous"` gives the preload a `same-origin` credentials mode](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attributes), which is also the default for `fetch()`. Credentials mode is part of the preload key. I verified this in Chrome with the same markup as `index.ejs`: the default fetch reused the preload, producing one request with the same-origin cookie, while `credentials: "omit"` produced a second request without the cookie.

Can we leave this as `fetch(script.src)` here and in `route-maps.ts`? The explicit `omit` defeats the preload reuse this PR is meant to restore, and it can break a same-origin map endpoint that relies on the session cookie.

packages/framework/esm-dynamic-loading/src/import-maps.ts
```diff
@@ -12,30 +12,53 @@ let devMode = false;
...
         maps.push(await response.json());
```
denniskigen, line 51: `fetch()` does not reject merely because a response has a 4xx or 5xx status. If the server returns a transient 503 with a JSON body, `response.json()` succeeds and `complete` remains true. The import-map loader merges that error body into an empty map, the route-map loader drops it during validation, and both cache the incomplete result for the rest of the page. I reproduced this against this head with a 503 response followed by a valid map; the second call stayed empty instead of consuming the valid response.

Can we throw when `!response.ok` before parsing, here and in `route-maps.ts`, and cover this case in the retry tests? That sends HTTP failures through the existing catch, marks the read incomplete, and clears `baseMapPromise` as intended.

packages/framework/esm-dynamic-loading/src/import-maps.test.ts
```diff
@@ -334,6 +334,73 @@
     });
   });
 
+  describe('remote import map caching', () => {
+    function setRemoteImportMap(src = 'http://localhost/importmap.json') {
+      const script = document.createElement('script');
+      script.type = 'systemjs-importmap';
+      Object.defineProperty(script, 'src', { value: src, writable: false });
```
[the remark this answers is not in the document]
ibacher: Pointing to localhost and done in a test.

packages/framework/esm-dynamic-loading/src/route-maps.test.ts
```diff
@@ -375,6 +375,75 @@
     });
   });
 
+  describe('remote route map caching', () => {
+    function setRemoteRouteMap(src = 'http://localhost/routes.registry.json') {
+      const script = document.createElement('script');
+      script.type = 'openmrs-routes';
+      Object.defineProperty(script, 'src', { value: src, writable: false });
```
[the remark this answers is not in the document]
ibacher: Pointing to localhost and done in a test.

packages/framework/esm-dynamic-loading/src/import-maps.ts
ibacher: Thanks for catching this!

packages/framework/esm-dynamic-loading/src/import-maps.ts
ibacher: Well-spotted. I'm too used to `openmrsFetch()`.

denniskigen:
LGTM, @ibacher. Thanks. Can we add a changeset?

denniskigen (approved):

PR #1856, opened 2026-08-27
(BREAKING) Extension system should not re-derive in full on every write

ibacher:
## Requirements

## Summary

This is an attempt to reduce much of the churn described here: https://claude.ai/code/artifact/14b92358-1ef6-4152-a4ec-d220bae9e03e.

### PR Overview

This PR overhauls the extension system's internal recomputation model to eliminate O(n²) behavior where any single write (a config change, an extension registering, a slot mounting) triggered a full re-derivation of state for every module, extension, and slot in the app. Derived state is now cached per-entity keyed on its actual inputs, and stores only publish a new value when it actually changed, so consumers no longer re-render on unrelated updates elsewhere in the app. The config system's implementer-tools output is now computed lazily on read instead of eagerly on every write. This is a breaking change to the shape and timing of internal store updates.

Alongside that, the branch clarifies the extension system's data model by splitting what was previously a single "extension" concept into a registration (what a module registers, by name) and a rendering (one live, on-screen copy of a configured instance in a slot — since one instance can render many times, e.g. once per row in a list). Renderings are tracked in their own store, mutated in place for performance rather than copied on every mount/unmount. This also adds support for a per-rendering displayConditionExpression, evaluated against that specific rendering's local state rather than globally. Supporting hooks (useAssignedExtensions, useExtensionSlot, useStore, and a new useShallowStableValue) and Extension.tsx were updated to work with this model, along with substantial new test coverage validating the recomputation and config-propagation behavior.

### Relation to performance

The extension system recomputation this ameliorates is a cause of many of the performance issues we're aiming to fix, but not all of them. Specifically, this should mostly reduce re-render churn, reduce the CPU drag from constant re-computation, and save a little bit of memory due to reshaping of the store's internals. It does not, by itself, resolve all the performance short-comings.

### Notable breakages

* Store no longer tracks `assignedExtensions`. Instead, the store tracks `candidateExtensions`, which is the list of extensions assigned to a slot but which have not had their `Display conditions` `expression` property evaluated yet. This is because the evaluation has the ability to read the extension slot state, which can be different per instance of an extension slot. Use `getAssignedExtensions()` or `useAssignedExtensions()` to get the correct list.
* Relatedly `useAssignedExtensionIds()` is now implemented in terms of `useAssignedExtensions()` and has been deprecated as it is no longer a "fast-path" to get extension metadata.
* `useStore()` no longer returns a copy of the store's state, but the state directly. Use with caution.
* All parcels will hard fail if they do not render within 15 seconds of loading. This is necessary to ensure that we can clean them up. Note that this 15 seconds is purely for the `bootstrap()`, `mount()`, and `unmount()` functions, which are not network dependent.
* Extensions added through configuration now retain the order they are defined in configuration (although, this is the intended behavior).
* Extension rendering state is no longer tracked in the extension store.

## Other

To make this explicit: the state of the extension system, other than as exposed through `getAssignedExtensions()` should be treated as framework internals. While there are valid use-cases for accessing this data, the shape of the data in the extension store is not part of the public API.

### DIFF
--- a/packages/framework/esm-extensions/src/extensions.ts
+++ b/packages/framework/esm-extensions/src/extensions.ts
@@ -20,10 +20,10 @@ import {
   getExtensionSlotsConfigStore,
   getExtensionsConfigStore,
 } from '@openmrs/esm-config';
-import { evaluateAsBoolean } from '@openmrs/esm-expression-evaluator';
+import { evaluateAsBoolean, type VariablesMap } from '@openmrs/esm-expression-evaluator';
 import { type FeatureFlagsStore, featureFlagsStore } from '@openmrs/esm-feature-flags';
 import { subscribeConnectivityChanged } from '@openmrs/esm-globals';
-import { isOnline as isOnlineFn } from '@openmrs/esm-utils';
+import { isOnline as isOnlineFn, shallowEqual } from '@openmrs/esm-utils';
 import { isEqual, merge } from 'lodash-es';
 import { checkStatusFor } from './helpers';
 import {
@@ -33,8 +33,10 @@ import {
   type ExtensionSlotCustomState,
   type ExtensionSlotInfo,
   type ExtensionSlotState,
+  getExtensionInstancesStore,
   getExtensionInternalStore,
   getExtensionStore,
+  scheduleRecomputation,
   updateInternalExtensionStore,
 } from './store';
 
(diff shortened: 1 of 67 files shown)


ibacher:
@pirupius This targets the bigger source of churn I identified earlier today. It doesn't completely solve things, but I think it solves the parts of the problem that are related to the extension system. I'd appreciate if you could test this out on your instance and report any issues.

@denniskigen FYI.

Note that I've tried to keep this confined to the extension system. I am aware that scrolling rapidly back-and-forth on the list of results will eventually cause the app to stall out. This appears to be related to `useVisit()` (which I'll address in a separate PR) and the patient chart store (which will also be it's own thing, but I need to hung that down).

pirupius:
@ibacher this is my test with the patient search and running the shell and this is how it behaves. There's definitely an improvement. 


https://github.com/user-attachments/assets/80063cfb-a5f1-4cfd-a24a-238d70a24e4d

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -447,6 +559,8 @@ export function getAssignedExtensions(slotName: string): Array<AssignedExtension
     isOnline,
     sessionState.session,
   );
+
+  return state === undefined ? assigned : filterExtensionsByDisplayConditions(assigned, state, sessionState.session);
```
denniskigen: Shouldn’t the no-state call still evaluate display conditions against `{ session }`, as it did before this PR and as `useAssignedExtensions` does now?

I reproduced a downstream regression in [dispensing-app](https://github.com/openmrs/openmrs-esm-dispensing-app/blob/main/src/prescriptions/prescription-expanded.component.tsx#L19): when both condition extensions have `expression: "false"`, `getAssignedExtensions("dispensing-condition-and-diagnoses")` still returns them, so the app creates the tab, while the state-aware `ExtensionSlot` filters them out and renders an empty panel.

First render:

Screenshot 2026-08-28 at 16 48 16

Next render:

Screenshot 2026-08-28 at 16 49 08

Could we preserve the previous no-state behavior by always calling `filterExtensionsByDisplayConditions`? It already constructs `{ session }` when state is absent. If returning the unfiltered list is intentional, I think that needs an explicit API distinction and a changeset so existing callers know their behavior has changed.

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -43,98 +45,170 @@ const extensionStore = getExtensionStore();
...
+
+  scheduleRecomputation(recomputeExtensionOutputStore);
+}
+
+function recomputeExtensionOutputStore() {
+  // Writing the output store notifies its subscribers synchronously, and a subscriber that dirties
+  // a slot would otherwise re-enter here and recurse without bound. What it dirties is picked up
+  // by the next recomputation.
+  if (isRecomputingOutputStore) {
+    return;
```
denniskigen, line 78: Could we flush invalidations that arrive while the output store is being published?

I reproduced this with an `extensionStore` subscriber that calls `attach()` once. `markSlotsDirty` records the slot, but the nested recomputation returns at `isRecomputingOutputStore`, and nothing schedules another pass. The public output store remains stale until an unrelated write occurs.

Could we loop or reschedule after the current pass when `dirtySlots` is still non-empty? A focused regression test where an output-store subscriber performs one input write would lock this down.

packages/framework/esm-react-utils/src/useExtensionSlot.ts
```diff
@@ -1,33 +1,30 @@
...
     throw Error('ComponentContext has not been provided. This should come from @openmrs/esm-react-utils.');
   }
 
-  useEffect(() => {
-    registerExtensionSlot(moduleName, slotName, state);
-    isInitialRender.current = false;
-  }, []);
+  // Callers almost always pass `state` as an object literal, so its identity changes every render
+  // while its contents don't. Reading it from a ref keeps it out of the effect dependencies below.
+  const stateRef = useRef(state);
```
denniskigen: Small cleanup: `stateRef` is no longer read now that state stays local to `useAssignedExtensions`. Could we remove it and the `useRef` import?

denniskigen:
> @ibacher this is my test with the patient search and running the shell and this is how it behaves. There's definitely an improvement.
> 
>  Screen.Recording.2026-08-28.at.17.00.22.mov

Am I seeing things, or did the search results UI snap back to the first page when you clicked on page 2 after searching for "men"?

ibacher:
> Am I seeing things, or did the search results UI snap back to the first page when you clicked on page 2 after searching for "men"?

Yeah, that's a weird one...

pirupius:
That's right @denniskigen it happened

packages/framework/esm-extensions/src/extensions.ts
ibacher: Yes, it should! This is one of the reasons I've left this as a draft; it's still full of rough edges like this. That said, this should be fixed in the recent commit.

packages/framework/esm-react-utils/src/useExtensionSlot.ts
ibacher: Gone!

ibacher:
Alright. Reproduced the snap-back bug. It's unrelated to this change set and is actually a memoization flaw in `usePagination`. We have this effect:

```tsx
useEffect(() => {
  goTo(1);
}, [query, goTo]);
```

Which is meant to fire when the _query_ changes, but it also fires when the `goTo()` reference changes. Goto is declared like this:

```tsx
const goTo = useCallback(
  (page: number) => { setPage(Math.max(1, Math.min(totalPages, page))); },
  [setPage, totalPages],
);
```

So it changes if either the `setPage` handler or `totalPages` change. Here what triggers it is that `totalPages` is calculated from `data.length`, but with `useSWRInfinite()` the size of the data grows with each page load and the advanced patient search [loads every page](https://github.com/openmrs/openmrs-esm-patient-management/blob/f4163794c31570904d1b77259cfadd135136fb91/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx#L47). 

So basically if you try to page through results before they've fully settled you get this behaviour. It's probably just not _noticeable_ because performance has been terrible, so the results have settled before the click registered. At least that's my read.

packages/framework/esm-extensions/src/extensions.ts
ibacher: This was fixed by adding a loop that does up to 20 sweeps. After 20 sweeps, we wait for the next write to avoid infinite loops.

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -330,7 +423,9 @@ function getOrder(
       // after all others
       return 2000 + assignedIndex;
     } else {
-      return -1;
+      // Added by configuration with no order of any kind — after everything else, and among
+      // themselves in the order the configuration lists them.
+      return 3000;
```
denniskigen, line 426: This keeps multiple extensions introduced only through `config.add`, with no other ordering information, in their configuration order. On `main`, both received the `-1` sentinel, so the comparator returned `-1` instead of `0` for a tie and could reorder them. Since that is observable behavior, could we mention the ordering stabilization in the changeset?

packages/framework/esm-extensions/src/extensions.ts
```diff
@@ -422,12 +497,67 @@ function getAssignedExtensionsFromSlotData(
...
+ * @param extensions The extensions assigned to the slot, from `getAssignedExtensions` or the store
+ * @param state The state of this rendering of the slot
+ * @param session The current session, which conditions may refer to as `session`
+ * @param slotName The slot these were assigned to, used only to identify them if a condition throws
+ * @returns Those of `extensions` that should be displayed, in the same order. Always a new array.
+ */
+export function filterExtensionsByDisplayConditions(
+  extensions: Array<AssignedExtension>,
+  state?: ExtensionSlotCustomState,
+  session: Session | null = null,
```
denniskigen, line 513: Since this helper is public, should the optional `session` parameter default to `sessionStore.getState().session` rather than `null`? A caller filtering extensions read directly from `getExtensionStore()` could reasonably omit it, causing session-dependent conditions to evaluate against `null` and potentially hide an extension. If keeping the helper pure is intentional, could the JSDoc explicitly say callers must provide the current session when expressions may reference it?

packages/framework/esm-extensions/src/public.ts
```diff
@@ -5,17 +5,20 @@ export {
...
 export { type CancelLoading, renderExtension } from './render';
 export {
   type ExtensionMeta,
+  type ExtensionRendering,
   type ExtensionRegistration,
   type ExtensionStore,
   type AssignedExtension,
   type ConnectedExtension,
   type ExtensionSlotState,
+  batchExtensionUpdates,
```
denniskigen: Could we mirror these two new public exports in the Vitest, Jest, and Storybook extension mocks? `@openmrs/esm-framework/mock` re-exports the individual mocks, so downstream tests using the standard alias will otherwise have missing exports for `filterExtensionsByDisplayConditions` and `batchExtensionUpdates`.

ibacher: I'm similarly planning on removing this from the public API. This should remain an implementation detail, I think.

packages/framework/esm-extensions/src/extensions.ts
ibacher: Honestly, this isn't intended to remain part of the public API. The public API is still `getAssignedExtensions()` which does do that lookup.

pirupius:
Hi @ibacher @denniskigen i've made a follow up PR to patient management based on the work on this branch to fix the pagination issue https://github.com/openmrs/openmrs-esm-patient-management/pull/2708 

Is it possible to get this once it's ready so that other people can stress test this work?

denniskigen:
Hey @ibacher, I've merged the latest `main` into this branch to resolve the conflict with #1858. Heads up that the merge commit carries a bit more than conflict resolution. More specifically:

- `renderExtension()` now mounts through `renderParcel()` from #1858 while keeping this branch's rendering registration and cleanup semantics.
- I extended the cleanup guards to the `mountPromise`/`unmountPromise` handlers via a small `forgetSafely()` helper. Previously a throwing `unregisterExtensionRendering()` in those paths would surface as an unhandled rejection. There are four new tests in `render.test.ts` covering the registration record and the contained cleanup failures.
- I also added the release changeset for this PR, including a note about the `config.add` ordering stabilization.

Tests and typecheck are green across the affected packages. Happy to adjust any of this if it doesn't match what you had in mind.

ibacher:
Marking this as "ready-to-review" as I'm pretty sure this has captured the "shape" of the changes.

packages/framework/esm-react-utils/src/Extension.tsx
```diff
@@ -21,73 +21,114 @@ export const Extension: React.FC<ExtensionProps> = ({ state, children, ...divPro
...
+        if (isUnmounted.current) {
+          unmountParcel(newParcel);
+        }
+      },
+      // Cleared so a later reattach can try again; `renderExtension` reports its own failures.
+      () => {
+        isRendering.current = false;
+      },
+    );
   }, []);
```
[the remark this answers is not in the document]
ibacher: This is a real issue, especially because an `<ExtensionSlot />` that hasn't been registered will initially load with no state.

packages/framework/esm-extensions/src/store.ts
```diff
@@ -70,8 +102,11 @@ export interface ExtensionStore {
 
 export interface ExtensionSlotState {
   moduleName?: string;
-  assignedExtensions: Array<AssignedExtension>;
-  state?: ExtensionSlotCustomState;
+  /**
+   * Candidates only. Call `getAssignedExtensions()` for the extensions a given rendering of
+   * this slot should actually display.
+   */
+  candidateExtensions: Array<AssignedExtension>;
```
denniskigen, line 109: This one is blocking. Can we migrate and release the existing consumers before this lands? [Patient chart](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx#L85) and [stock management](https://github.com/openmrs/openmrs-esm-stock-management/blob/main/src/dashboard/home-dashboard.component.tsx#L13) both read `assignedExtensions` directly. Apps resolve the framework at runtime, so both receive `undefined` for this field as soon as a shell ships the rename.

The [patient-management E2E](https://github.com/openmrs/openmrs-esm-core/actions/runs/33548684773) is catching exactly this: all nine failed traces contain `undefined.flatMap` from the current patient-chart bundle.

A `candidateExtensions` alias would not preserve the old behavior. `assignedExtensions` evaluated display expressions, while `candidateExtensions` intentionally does not. It’s probably better to fix and release both consumers first.

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -100,12 +110,29 @@ function handleModalStateUpdate({ modalStack, modalContainer }: ModalState) {
...
+          renderModalIntoDOM(modalFrame, instance.modalName, instance.props).then(
+            (parcel) => {
+              // Release the parcel if it's been closed while mounting
+              if (isClosing(instance)) {
+                unmountModalParcel(instance.modalName, parcel);
+                return;
+              }
+
+              instance.parcel = parcel;
+              instance.state = 'MOUNTED';
```
denniskigen: Can we wait for `parcel.mountPromise` before setting this to `MOUNTED`, and call `reportError()` and `closeInstance()` if it rejects? `renderParcel()` returns the parcel handle before mounting finishes, so a failed mount currently leaves the modal on the stack with the overlay and body scroll lock in place until another close path runs. The new timeout makes a hung mount take this path after 15 seconds.

Can we also observe `parcel.unmountPromise`? It is separate from the promise returned by `unmount()`, so catching only this call still leaves a rejected promise unhandled. The tests cover a rejected `unmount()` call, but not either lifecycle promise.

packages/framework/esm-react-utils/src/useCandidateExtensions.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param slotName The name of the slot to get the candidate extensions for.
+ * @internal
+ */
+export function useCandidateExtensions(slotName: string) {
+  // Subscribes so this re-runs when the slot's extensions change; `getCandidateExtensions` reads the
+  // same store, and returns the array below.
+  useExtensionSlotStore(slotName);
+
+  return getCandidateExtensions(slotName);
```
denniskigen: nit: Can we return `useExtensionSlotStore(slotName).candidateExtensions` directly here? We already subscribe to that stable snapshot, then throw it away and call a getter that returns a fresh `[]` for an unregistered slot.

denniskigen (dismissed):
This is very close, @ibacher. The changeset and cleanup coverage look good; I’ve left two things I think we need to resolve before merging, plus a nit.

packages/framework/esm-extensions/src/store.ts
ibacher: Forget what I said. I was thinking that we wanted to fix those based on the `candidateExtensions`, but actually the `assignedExtensions` are ultimately what we want...

packages/framework/esm-react-utils/src/useCandidateExtensions.ts
ibacher: Yeah, this is mostly because I want people to prefer `getCandidateExtensions()` over directly loading from the store (because that's easier to keep stable) and I worry that this becomes the reference implementation for how people will build this.

But you're right that it's sort of silly to throw away the same value we're returning here.

packages/framework/esm-extensions/src/store.ts
ibacher: Stock Management: https://github.com/openmrs/openmrs-esm-stock-management/pull/622
Patient Chart: https://github.com/openmrs/openmrs-esm-patient-chart/pull/3614

packages/framework/esm-styleguide/src/modals/index.tsx
```diff
@@ -100,12 +118,50 @@ function handleModalStateUpdate({ modalStack, modalContainer }: ModalState) {
...
+              if (!parcel) {
+                // `renderModalIntoDOM` only returns without one when it had nowhere to render it,
+                // which it has already reported.
+                closeInstance(instance);
+                return;
+              }
+
+              instance.parcel = parcel;
+              modalContainer.prepend(modalFrame);
+              modalFrame.style.visibility = modalStore.getState().modalStack.indexOf(instance) ? 'hidden' : 'unset';
```
denniskigen:
```ts
        case 'MOUNTING':
        case 'MOUNTED':
```

Can we handle `MOUNTING` in the visibility switch below too? `renderParcel()` can return while `mountPromise` is still pending. If another modal opens in that window, the first frame has already been inserted but is skipped there, so both frames remain visible. Resolving the first mount does not trigger another store update.

packages/framework/esm-styleguide/src/modals/index.tsx
ibacher: I think this is fixed.

denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1857, opened 2026-08-28
(fix) Match workspace scopes relative to SPA base

denniskigen:
## Requirements

## Summary

Workspace `scopePattern` values are route patterns, but core currently matches them against the browser's full pathname. On the default deployment, an anchored patient-chart pattern such as `^/patient/([^/]+)/chart` is therefore tested against `/openmrs/spa/patient/...` and never matches. The current unanchored pattern only works because the regex finds `/patient/...` in the middle of that full path, which is also why core repeatedly warns that the pattern should start with `^`. Adding the recommended anchor without changing core would make the workspace close on every same-app navigation.

Downstream code already treats this as a SPA-relative contract. [Patient chart](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/routes.json) uses a dynamic patient scope. Bed management, patient lists, wards, service queues, and appointments in [patient management](https://github.com/openmrs/openmrs-esm-patient-management/tree/main/packages) use static SPA-relative scopes, as does [laboratory](https://github.com/openmrs/openmrs-esm-laboratory-app/blob/main/src/routes.json). In total, those seven public apps define nine workspace groups this way. The [workspace documentation](https://github.com/openmrs/openmrs-contrib-o3-docs/blob/main/content/en-US/docs/workspaces/creating-workspaces.mdx) also shows an anchored SPA-relative pattern, and [`create-o3-app`](https://github.com/openmrs/create-o3-app/blob/main/src/templates/engine.ts) generates one from the module's first route.

This changes `shouldCloseOnUrlChange` to remove the configured SPA base before matching. The matching contract now works at `/openmrs/spa`, a custom base, or the domain root. Navigation outside the configured SPA base closes the workspace even when the destination has the same route shape. If the pattern does not match both SPA-relative pathnames, core retries the full path so existing private or downstream patterns that include the SPA base keep working. The public type documentation now states the same contract explicitly.

The new tests cover:

- navigation within the same patient scope and between different patients;
- the domain root, the default SPA base, and a custom SPA base;
- navigation from a URL exactly equal to the configured SPA base;
- navigation outside the configured SPA base;
- SPA-base prefix collisions such as `/openmrs/spa2`;
- route-boundary collisions such as `/chart-extra`; and
- legacy patterns that include `/openmrs/spa`.

## Screenshots

Grabbed from the console when running the app shell locally:

Screenshot 2026-08-28 at 13 04 00

### DIFF
--- a/.changeset/fix-workspace-scope-pattern.md
+++ b/.changeset/fix-workspace-scope-pattern.md
@@ -1,0 +1,6 @@
+---
+'@openmrs/esm-globals': patch
+'@openmrs/esm-styleguide': patch
+---
+
+Match workspace scope patterns against pathnames relative to the configured SPA base while preserving legacy full-path patterns.
--- a/packages/framework/esm-framework/docs/interfaces/FeatureFlagDefinition.md
+++ b/packages/framework/esm-framework/docs/interfaces/FeatureFlagDefinition.md
@@ -2,7 +2,7 @@
 
 # Interface: FeatureFlagDefinition
 
-Defined in: [packages/framework/esm-globals/src/types.ts:350](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-globals/src/types.ts#L350)
+Defined in: [packages/framework/esm-globals/src/types.ts:354](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-globals/src/types.ts#L354)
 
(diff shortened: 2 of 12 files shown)


ibacher (approved):

ibacher:
Thanks @denniskigen!

PR #1858, opened 2026-08-28
(fix) Mount all single-spa parcels in a single long-lived parcel

ibacher:
## Requirements

## Summary

single-spa@6 has a bug with root parcels where it does not appropriately clean-up its internal tracking when a _root_ parcel unmounts though it does clean-up after non-root parcels. This is fixed in the as-yet unreleased single-spa@7. However, here we move to using a single long-running root parcel and mount all other parcels as children of that parcel.

Because single-spa, among other things, tracks the DOM node that belongs to the parcel, without this, we end up with a large number of detached DOM nodes cached in the single-spa registry.

### DIFF
--- a/.changeset/bright-teeth-design.md
+++ b/.changeset/bright-teeth-design.md
@@ -1,5 +1,5 @@
 ---
-"@openmrs/esm-extensions": minor
+"@openmrs/esm-extensions": patch
 "@openmrs/esm-framework": patch
 ---
 
--- a/.changeset/bright-teeth-design.md
+++ b/.changeset/bright-teeth-design.md
@@ -1,0 +1,6 @@
+---
+"@openmrs/esm-extensions": minor
+"@openmrs/esm-framework": patch
+---
+
+(fix) Mount all single-spa parcels in a single long-lived parcel
--- a/packages/framework/esm-extensions/src/render.ts
+++ b/packages/framework/esm-extensions/src/render.ts
@@ -1,5 +1,5 @@
 /** @module @category Extension */
-import { mountRootParcel, type Parcel, type ParcelConfig } from 'single-spa';
+import { mountRootParcel, type AppProps, type Parcel, type ParcelConfig } from 'single-spa';
 import { getExtensionNameFromId, getExtensionRegistration } from './extensions';
 import { checkStatus } from './helpers';
 import { updateInternalExtensionStore } from './store';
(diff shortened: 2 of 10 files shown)


packages/framework/esm-extensions/src/render.ts
```diff
@@ -8,7 +8,53 @@ export interface CancelLoading {
...
+ * single-spa only removes a parcel from its owner's registry when it unmounts if that owner has a
+ * name, and the internal object backing `mountRootParcel()` has none (fixed in single-spa 7).
+ * Anything mounted with it is therefore retained for the lifetime of the page, along with the
+ * `domElement` it was given and the whole subtree rendered into it. Extensions mounted through a
+ * named owner are pruned as they should be.
+ *
+ * The host parcel is mounted at most once, lazily, and never unmounted; if mounting it fails we
+ * fall back to `mountRootParcel()`, since leaking is better than not rendering.
+ */
+function getParcelMounter(): Promise<MountParcel> {
```
denniskigen, line 36: This fixes the extension parcels, but three production call sites still bypass the named host: modals and the v1 and v2 workspace renderers continue to use `mountRootParcel` and are unmounted on close or removal. In single-spa 6.0.3, those parcels remain in the anonymous root owner’s registry for the same reason. Since the title and summary refer to all parcels, could we either route these callers through the shared mounter here, or narrow the stated scope and track them in a follow-up?

.changeset/bright-teeth-design.md
```diff
@@ -0,0 +1,6 @@
+---
+"@openmrs/esm-extensions": minor
```
denniskigen: Could we make `@openmrs/esm-extensions` a patch bump here? The repository’s release guidance classifies bug fixes as patch releases, and this change does not add or change public API. Since the changeset text is release-facing, could we also describe the user-visible impact, for example: “Release detached DOM nodes after extensions unmount by mounting them under a named single-spa host parcel.”

denniskigen (commented):
Nice fix, @ibacher. I verified this against the single-spa source. In 6.0.3, `rootParcels` has no name, so `unmountThisParcel()` skips removing root parcels from the owner’s registry. The 7.0 branch fixes this by naming that owner. Extensions mounted through this named host are pruned as expected, and the react-utils extension suite passes locally through the real single-spa path.

.changeset/bright-teeth-design.md
ibacher: Fair. I considered this minor because it does reorganize things internally, but that's not actually reflected anywhere you can see.

ibacher: Fixed in latest commit.

packages/framework/esm-extensions/src/render.ts
ibacher: Good call. Implemented this most recent commit.

ibacher:
> The 7.0 branch fixes this by naming that owner.

Yeah, Claude actually kept trying to steer this into just being a code patch for 6.0.3 since that's a one-line fix. Unfortunately, it's also just not workable with how everything is structured.

denniskigen (approved):

denniskigen:
Thanks, @ibacher!

PR #1859, opened 2026-08-28
(feat) Remove `useVisitContextStore` re-render amplification

ibacher:
## Requirements

## Summary

This fixes a major request amplification issue with `useVisit` and `useVisitContextStore`. First, `useVisit` called the `useVisitContextStore` hook twice on every render, first to get some data and then to register a `mutateVisit` callback. Instead, we register a ref as the callback and fill in the refs contents with the actual function, so that `useVisit` only writes to the store once.

More critically, registering the callback required updating the store's state, which triggered re-renders in unrelated components just because we were registering a new `useVisit` hook. Measured with a test, 400 components using `useVisit` would cause 80,600 renders. Likewise, every call to `mutateVisit()`, because it was bound as an action, actually _writes_ to the store, causing re-renders even before invalidating the SWR cache.

Moving the callback tracking out of the store and into a global variable removes both of these re-render amplifications. Re-renders now should only occur when the active visit (or retrospective visit) is updated.

`mutateVisitCallbacks` has been removed from the VisitStoreState; I _think_ it's ok to treat the contents of the store as an internal detail. The implementation surface (including `useVisitContextStore()`) is identical. 

## Other

Ultimately, instead of storing functions globally, we should probably move to some kind of system for signally when mutations are need as that would involve much less book-keeping here and much less chance of triggering excessive renders.

### DIFF
--- a/packages/framework/esm-emr-api/src/visit-utils.ts
+++ b/packages/framework/esm-emr-api/src/visit-utils.ts
@@ -25,15 +25,6 @@ export enum VisitStatus {
 export interface VisitStoreState {
   patientUuid: string | null;
   manuallySetVisitUuid: string | null;
-
-  /**
-   * Stores a record of SWR mutate callbacks that should be called when
-   * the Visit with the specified uuid is modified. The callbacks are keyed
-   * by unique component IDs.
-   */
-  mutateVisitCallbacks: {
-    [componentId: string]: () => void;
-  };
 }
 
 /**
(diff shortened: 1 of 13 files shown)


chibongho (approved):
Thanks. This seems like a better pattern.

> Ultimately, instead of storing functions globally, we should probably move to some kind of system for signally when mutations are need as that would involve much less book-keeping here and much less chance of triggering excessive renders.

Maybe use `fireOpenmrsEvent`?

PR #1861, opened 2026-09-01
(feat) Make pagination functions stable references

ibacher:
## Requirements

## Summary

`usePagination()` and friends return three functions callers can use to navigate: `goTo()`, `goToNext()` and `goToPrevious()`. However, these functions were not stable references; more exactly, they depended on `totalPages` which depended on the amount of data to be paged. But in cases where the page data don't arrive all at once (e.g., `useSWRInfinite()`), these functions would become new instances which can trigger a weird bug where things will be inappropriately clamped to page 1 of the results.

This fixes this and a few other edge cases by moving most of the clamping into the `usePagination()` render rather than trying to resolve it completely inside those functions, allowing us to keep those functions as stable references in more situations.

A similar fix is applied to `useServerPagination()`.

### DIFF
--- a/packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
+++ b/packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
@@ -223,7 +223,7 @@ export function useServerPagination<T, R>(
     totalPages,
     totalCount: totalCount.current,
     currentPage,
-    currentPageSize,
+    currentPageSize: currentPageSize.current,
     paginated: totalPages > 1,
     showNextButton: currentPage < totalPages,
     showPreviousButton: currentPage > 1,
--- a/packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
+++ b/packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
@@ -223,7 +223,7 @@ export function useServerPagination<T, R>(
     totalPages,
     totalCount: totalCount.current,
     currentPage,
-    currentPageSize: currentPageSize.current,
+    currentPageSize,
     paginated: totalPages > 1,
     showNextButton: currentPage < totalPages,
     showPreviousButton: currentPage > 1,
--- a/packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
+++ b/packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
@@ -170,7 +170,7 @@ export function useServerPagination<T, R>(
         return;
       }
 
-      // page coung is not known prior to running any query when `bound` will be `NaN`, so in this
+      // page count is not known prior to running any query when `bound` will be `NaN`, so in this
       // case we assume the requested page is fine; after results return, things will settle properly
       // clamped
       const bound = Math.ceil(totalCount.current / pageSize);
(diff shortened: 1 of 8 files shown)


packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
```diff
@@ -134,34 +156,74 @@ export function useServerPagination<T, R>(
...
+    goTo(currentPageRef.current - 1);
+  }, [goTo]);
 
   return {
     data: getData(data?.data),
     totalPages,
     totalCount: totalCount.current,
     currentPage,
-    currentPageSize,
+    currentPageSize: currentPageSize.current,
```
denniskigen, line 226:
```suggestion
    currentPageSize,
```

Can we keep returning the ref here, or ship this under a major? Unwrapping it is probably the better shape since `totalCount` is already a number, but it changes the documented return type from `MutableRefObject<number>` to `number`. [Patient-chart has a typed fixture built against the current shape](https://github.com/openmrs/openmrs-esm-patient-chart/blob/e6570f685c67bb6251547c7b686d127510b2655b/packages/esm-patient-chart-app/src/visit/visits-widget/visit-detail-overview.test.tsx#L26-L35) that will fail typechecking when it picks up this change. This isn't needed for the callback-stability fix either way.

ibacher: Actually, I might revert this. Patient-chart does have that fixture, but that's only in test, not in production code. And the test code doesn't even actually _read_ that value, i.e., it's an overly-specific mock rather than something we need to preserve.

denniskigen: Thanks. This restores the ref, but the `currentPageSize` test and the final changeset paragraph still expect and document a number. The build is now failing on that test. Can we update those too?

ibacher: Like I said above, I think the right thing to do here is to restore `currentPageSize` as a number and just fix the fixtures. That shouldn't be breaking as we're only using a Ref in tests to correspond to the type and I can find no evidence of actually using the Ref to get the current value.

packages/framework/esm-react-utils/src/useOpenmrsPagination.ts
```diff
@@ -134,34 +156,74 @@ export function useServerPagination<T, R>(
...
+  // which is what keeps this callback referentially stable, apart from a change in `pageSize`.
   const goTo = useCallback(
     (page: number) => {
-      if (0 < page && page <= totalPages) {
+      if (!Number.isInteger(page) || page < 1) {
+        console.warn(`useServerPagination: ignoring goTo(${page}); page must be a positive integer.`);
+        return;
+      }
+
+      // page coung is not known prior to running any query when `bound` will be `NaN`, so in this
```
denniskigen:
```suggestion
      // page count is not known prior to running any query when `bound` will be `NaN`, so in this
```

nit: typo

packages/framework/esm-react-utils/src/usePagination.ts
```diff
@@ -22,29 +27,36 @@ export function usePagination<T>(data: Array<T> = [], resultsPerPage = defaultRe
...
+  const goTo = useCallback((page: number) => {
+    if (!Number.isInteger(page) || page < 1) {
+      console.warn(`usePagination: ignoring goTo(${page}); page must be a positive integer.`);
+      return;
     }
-  }, [page, totalPages, setPage]);
+    setRequestedPage(page);
+  }, []);
+
+  // overflow clamping happens in render; see the setRequestPage() call above
```
denniskigen:
```suggestion
  // overflow clamping happens in render; see the setRequestedPage() call above
```

nit: typo in the setter name

denniskigen (approved):
LGTM! Thanks, @ibacher. Two tiny nits below.

PR #1863, opened 2026-09-02
(fix) Pin framework to app shell version

ibacher:
## Requirements

## Summary

With Module Federation v1, the default resolution scheme with singleton shares was that the first-loaded wins. This served as our guarantee that `@openmrs/esm-framework` would always be loaded from the App Shell. With Module Federation v2, this changes to a policy that the highest-loaded version wins by default. Unfortunately, this means that the "loaded" version can change mid-application, so we add a plugin to pin things to the "correct" framework version.

## Other

@denniskigen This is the other release blocker I'm aware of.

This is written in maybe a slightly more convoluted way than would be expected as I wanted to explore using the same machinery to pin other versions, e.g., `@openmrs/esm-patient-common-lib` to the copy from `@openmrs/esm-patient-chart-app`. That, however, would be a behavioural change, so it's not supported here. This is only restoring the implicit v1 behaviour.

### DIFF
--- a/packages/framework/esm-state/src/state.ts
+++ b/packages/framework/esm-state/src/state.ts
@@ -13,6 +13,30 @@ interface StoreEntity {
 
 const availableStores: Record<string, StoreEntity> = {};
 
+/**
+ * Each copy of this module owns a separate `availableStores`, so a page running more than one copy
+ * has stores, config and extension state that cannot see each other. Nothing fails outright, which
+ * is what makes it expensive: the symptoms are missing state and "store already exists" errors far
+ * from the cause. The count lives on a global registry key rather than in module scope, since a
+ * second copy would not share module scope — that is the whole problem being detected.
+ *
+ * A hot reload re-evaluates this module and so also trips this, which is the honest answer: after a
+ * hot reload the page really is running two copies.
+ */
+const instanceCountKey = Symbol.for('openmrs.esm-state.instanceCount');
+const globalCounters = globalThis as unknown as Record<symbol, number | undefined>;
+const instanceCount = (globalCounters[instanceCountKey] ?? 0) + 1;
+globalCounters[instanceCountKey] = instanceCount;
+
+if (instanceCount > 1 && !isTestEnvironment()) {
+  console.error(
+    `Detected ${instanceCount} copies of @openmrs/esm-framework running on this page. Each copy keeps ` +
+      'its own store registry, so stores, configuration and extension state are not shared between ' +
+      'them. This usually means a frontend module loaded its own bundled framework instead of the one ' +
+      'the app shell provides.',
+  );
+}
+
 // spaEnv isn't available immediately. Wait a bit before making stores available
 // on window in development mode.
 globalThis.setTimeout?.(() => {
--- a/packages/shell/esm-app-shell/src/framework-share.test.ts
+++ b/packages/shell/esm-app-shell/src/framework-share.test.ts
@@ -1,0 +1,220 @@
+// Reproduces the duplicate-framework bug against Module Federation's own resolver rather than a model
+// of it, so the tests still mean something if Module Federation changes how it picks a provider. Each
+// "without the pin" case is the bug: the resolver hands the consuming module a provider other than the
+// app shell's, and taking that provider loads a second copy of the framework — a second store
+// registry, a second config store, a second set of extensions, and roughly a megabyte of duplicated
+// download. Nothing throws when that happens, which is why it needs a test rather than a smoke check.
+import { getRegisteredShare } from '@module-federation/runtime-core';
+import type { Shared, ShareScopeMap } from '@module-federation/runtime-core/types';
+import type { ResolvedShare } from './federation-plugins';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { pinFrameworkToAppShell } from './framework-share';
+
+const appShell = '@openmrs/esm-app-shell';
+const frontendModule = '@openmrs/esm-ward-app';
+const publicKey = '@openmrs/esm-framework';
+const internalKey = '@openmrs/esm-framework/src/internal';
+const pluginName = 'openmrs-pin-framework-to-app-shell';
+
+type VersionMap = Record<string, Shared>;
+
+function provider(version: string, from: string, { loaded = false } = {}) {
+  return {
+    version,
+    from,
+    loaded: loaded || undefined,
+    loading: null,
+    lib: null,
+    get: () => Promise.resolve(() => ({})),
+    scope: ['default'],
+    strategy: 'version-first',
+    useIn: [],
+    deps: [],
+    shareConfig: { requiredVersion: `^${version}`, singleton: true, eager: false, strictVersion: false },
+  } as unknown as Shared;
+}
+
+/** What a published frontend module asks for, as its remote entry declares it. */
+function consumer(pkgName = publicKey) {
+  return {
+    from: frontendModule,
+    version: '10.0.1-pre.5263',
+    scope: ['default'],
+    strategy: 'version-first',
+    shareConfig: { requiredVersion: '10.x', singleton: true, eager: false, strictVersion: false },
+  } as unknown as Shared & { pkgName?: string };
+}
+
+function federationGlobal() {
+  return globalThis.__FEDERATION__ as unknown as {
+    __GLOBAL_PLUGIN__: Array<{ name: string; resolveShare?: (args: unknown) => unknown }>;
+    __INSTANCES__: Array<{ registerPlugins: (plugins: Array<unknown>) => void }>;
+  };
+}
+
+/**
+ * Sets the page up the way `initializeSpa` does: the app shell has registered its framework and
+ * nothing else has loaded yet, so this is the only moment the app shell's own entry is identifiable.
+ * The two keys share one version map because `initializeSpa` aliases them onto each other.
+ */
+function pinWithAppShellProvider() {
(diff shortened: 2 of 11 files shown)


packages/shell/esm-app-shell/src/framework-share.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const plugin: ModuleFederationRuntimePlugin = {
+    name: pluginName,
+    resolveShare(args) {
+      if (!frameworkShareKeys.has(args.pkgName)) {
+        return args;
+      }
+
+      // The framework is not shared with tree shaking enabled, so there is no shaken variant to pick.
+      args.resolver = () => ({ shared: preferLoadedProvider(appShellProviders), useTreesShaking: false });
```
denniskigen:
```suggestion
      const defaultResolver = args.resolver;
      args.resolver = () => {
        const shared = preferLoadedProvider(appShellProviders);
        // Apps built with older tooling embed their own `runtime-core` (0.x), whose resolver contract is
        // the `Shared` entry itself; 2.x wraps it. Both runtimes pick this plugin up from
        // `__GLOBAL_PLUGIN__`, so answer in whichever shape the calling runtime's own resolver uses.
        const probe = defaultResolver() as unknown;
        const wrapped = typeof probe === 'object' && probe !== null && 'shared' in probe && !('version' in probe);
        // The framework is not shared with tree shaking enabled, so there is no shaken variant to pick.
        return (wrapped ? { shared, useTreesShaking: false } : shared) as ReturnType<typeof defaultResolver>;
      };
```

This breaks apps built with the old tooling. `__GLOBAL_PLUGIN__` is read by every MF runtime on the page, so the `runtime-core` 0.22.0 embedded in anything built with the pre-#1853 rspack config picks this up too. That runtime expects the resolver to hand back the `Shared` entry itself, not `{ shared, useTreesShaking }`, so it gets nothing with `lib`/`get` on it and dies with `i.get is not a function` as soon as it consumes the framework.

Reproduced in the browser: this branch's shell plus `esm-patient-flags-app` built from patient-chart's current tooling (runtime 0.22.0). Same app loads fine with the pin disabled, and with the suggestion above.

That's the "old apps keep working in the new shell" promise from the O3-5873 changeset. CI won't catch it since the e2e frontend is assembled from `next`, where everything's already an external-runtime build.

packages/shell/esm-app-shell/src/framework-share.ts
```diff
@@ -0,0 +1,77 @@
...
+  type ShareVersionMap,
+} from './federation-plugins';
+
+/**
+ * openmrs-pin-framework-to-app-shell
+ *
+ * This is a runtime Module Federation plugin to induce Module Federation to work as we expect.
+ * Specifically, e expect the version of @openmrs/esm-framework used to match the version loaded
+ * by the shell regardless of the versions available elsewhere. With Module Fedderation v2's
+ * default "version-first" loading strategy, the exact version of a resolution can be changed as
```
denniskigen, line 17: Do you have a repro for the flip? Looking at 2.8.1's `findSingletonVersionOrderByVersion`, once the shell's copy is loaded the reducer never leaves it, and the shell consumes the framework when the `run` chunk loads, before any remote registers. In the repro above the old app offered 10.0.1-pre.5158 against the shell's 10.0.0 and the shell still won without the pin. The only order where a remote won for me was one registering before the shell's own consume. Not against keeping the pin as a safety net, just wondering if the trigger is something else.

packages/shell/esm-app-shell/src/framework-share.ts
```diff
@@ -0,0 +1,77 @@
...
+  type Shared,
+  type ShareVersionMap,
+} from './federation-plugins';
+
+/**
+ * openmrs-pin-framework-to-app-shell
+ *
+ * This is a runtime Module Federation plugin to induce Module Federation to work as we expect.
+ * Specifically, e expect the version of @openmrs/esm-framework used to match the version loaded
+ * by the shell regardless of the versions available elsewhere. With Module Fedderation v2's
```
denniskigen:
```suggestion
 * Specifically, we expect the version of @openmrs/esm-framework used to match the version loaded
 * by the shell regardless of the versions available elsewhere. With Module Federation v2's
```

packages/shell/esm-app-shell/src/framework-share.ts
```diff
@@ -0,0 +1,77 @@
...
+      }
+
+      // The framework is not shared with tree shaking enabled, so there is no shaken variant to pick.
+      args.resolver = () => ({ shared: preferLoadedProvider(appShellProviders), useTreesShaking: false });
+      return args;
+    },
+  };
+
+  registerFederationPlugin(plugin);
+}
```
denniskigen, line 79: Do we want to push a test for this next to `federation-runtime.test.ts`? Two `ModuleFederation` instances sharing one scope map, plus a third on the 0.22.0 runtime, covers both the pin and the mixed-runtime case.

denniskigen (dismissed):
Thanks for digging into this, @ibacher. One blocker below, plus a question and a nit.

packages/shell/esm-app-shell/src/framework-share.ts
ibacher: Good catch! Fixing!

packages/shell/esm-app-shell/src/framework-share.ts
ibacher: So, admittedly, the way I caught this was with a debug build (`yarn run:shell`) which exposes `window.stores` which was being re-written 3 times. This was initially what lead to [this](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3605). That's also why there's a guard in esm-state specifically in this PR.

ibacher: Here's a [Gist](https://gist.github.com/ibacher/edb1e0fdefd91789bf398267829283dc) of a detector script, modified from the code originally used to find this issue. Needs to be run with `MODE=real-duplicate` and `NODE_PATH` pointing to something containing Playwright.

As a practical matter, it's somewhat hard to induce this: you need to be running an _older_ version of the app shell and a version of some app that includes a _newer_ version of the framework. The easiest way to be in that state is `yarn run:shell` (which will almost always be behind because the version is reported as whatever the base version is), but it should reproduce if you build an instance with an older version of the shell and run with the latest version of an app.

packages/framework/esm-state/src/state.ts
```diff
@@ -13,6 +13,30 @@ interface StoreEntity {
...
+ *
+ * A hot reload re-evaluates this module and so also trips this, which is the honest answer: after a
+ * hot reload the page really is running two copies.
+ */
+const instanceCountKey = Symbol.for('openmrs.esm-state.instanceCount');
+const globalCounters = globalThis as unknown as Record<symbol, number | undefined>;
+const instanceCount = (globalCounters[instanceCountKey] ?? 0) + 1;
+globalCounters[instanceCountKey] = instanceCount;
+
+if (instanceCount > 1 && !isTestEnvironment()) {
```
NethmiRodrigo: Might be a stupid question, but why are we logging this error only if it is not a test environment? The reason being, I'd imagine we'd want to know if this happens in dev3 or test3

packages/shell/esm-app-shell/src/framework-share.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  });
+
+  it('registers once, however many times it is called', () => {
+    pinWithAppShellProvider();
+    pinWithAppShellProvider();
+
+    expect(federationGlobal().__GLOBAL_PLUGIN__.filter((p) => p.name === pluginName)).toHaveLength(1);
+  });
+
+  it('leaves resolution alone when the app shell provides no framework at all', () => {
```
denniskigen, line 215: Nit: this prints the paranoia error into the test output. Spying on `console.error` and asserting it was called would quiet it and prove the branch fires.

denniskigen (commented):
This is very close, @ibacher. The 0.22 path checks out now: I ran a 0.22.0 `ModuleFederation` remote and a 2.8.1 one against the pinned scope and both load the shell's copy through `loadShare` (and `loadShareSync` on 2.8.1). Just the changeset left (if we're enforcing that), plus one optional nit.

```md
---
"@openmrs/esm-app-shell": patch
"@openmrs/esm-state": patch
---

(fix) Pin @openmrs/esm-framework to the app shell's copy under Module Federation 2.0, and report when more than one copy of the framework is running
```

ibacher:
@denniskigen Both fixed.

denniskigen (approved):
LGTM. Thanks, @ibacher!

packages/framework/esm-state/src/state.ts
NethmiRodrigo, line 31: @ibacher Might be a stupid question, but why are we logging this error only if it is not a test environment? The reason being, I'd imagine we'd want to know if this happens in dev3 or test3

ibacher: No, that's a reasonable question! The sort of answer is that dev3 and test3 are not "test environments" in the scope that `isTestEnvironment()` checks for. Rather this checks for literally running with NODE_ENVIRONMENT set to `test` (dev3 and test3 are "production" builds) or in the context of Jest or Vitest or something like that. What this is defending against is mentioned a little further up in the comment:

> A hot reload re-evaluates this module and so also trips this

Basically, if you're running the shell under `run:shell` and you update the framework and trigger the rebuild, this will trigger HMR which would trip this error, even though that's the _expected_ behaviour.

PR #1867, opened 2026-09-04
(fix) Move hard lifecycle stop to framework instead of single-spa

ibacher:
## Requirements

## Summary

In #1856, I added "hard" timeouts of 15s where if a parcel doesn't complete its bootstrap, mount, or unmount call within that time frame, we simply fail it since it would appear to be hung (the boostrap, mount, and unmount calls generally don't load scripts, so the time out here was intended to be a measure of pure JS speed). single-spa provides a mechanism for configuring timeouts like this which #1856 used, with the idea that we could use this to coordinate both single-spa clean-up and the framework's own clean-up. This turned out to be a mistake.

single-spa's "timeouts" actually become `setTimeout()` calls that are not cleared, even when the event they are waiting for has already happened. E.g., the 15s timeout for unmount means that single-spa retains a timeout closure for 15s, even if the unmount operation succeeds faster that that. These timeouts closed over the parcel definitions, including DOM nodes and event listeners, so effectively this change meant that any unmounted extension, modal or workspace was retained in JS memory for 15s, even though the unmount operation itself is normally milliseconds.

By moving the hard timeouts into our framework and reverting to single-spa's defaults, we greatly reduce the window that single-spa will hold the references to its default of 3s. Short of patching single-spa, this seems like the best way we can work around this issue.

NB You'll notice this also uses `setTimeout()`; the difference is that we track this timeout and call `clearTimeout()` if the promise settles.


packages/framework/esm-extensions/src/render.ts
```diff
@@ -21,31 +21,81 @@ type MountParcel = AppProps['mountParcel'];
...
+  return (props) => {
+    let timer: ReturnType<typeof setTimeout> | undefined;
+    const deadline = new Promise<never>((_, reject) => {
+      timer = setTimeout(
+        () => reject(new Error(`Lifecycle function ${which} for parcel ${name} did not settle within ${millis}ms`)),
+        millis,
+      );
+    });
+
+    return Promise.race([Promise.resolve().then(() => lifecycle(props)), deadline]).finally(() => clearTimeout(timer));
```
denniskigen: Can we preserve single-spa’s check that each lifecycle function returns a Promise here? `Promise.resolve().then(...)` converts a non-Promise result into a successful Promise, and `toSingleFn()` does the same for each array item. A mount that forgets to return its asynchronous work is then marked `MOUNTED` immediately and escapes this new deadline instead of failing as it does today. It’s probably best to reject non-Promise results in both paths and cover them with a real single-spa test.

packages/framework/esm-extensions/src/render.ts
```diff
@@ -21,31 +21,81 @@ type MountParcel = AppProps['mountParcel'];
...
+ * closure retains the parcel, the `domElement` it was given and the whole subtree rendered into it.
  */
-const lifecycleTimeouts = {
-  bootstrap: { millis: 15_000, dieOnTimeout: true },
-  mount: { millis: 15_000, dieOnTimeout: true },
-  unmount: { millis: 15_000, dieOnTimeout: true },
+const lifecycleDeadlines = {
+  bootstrap: 15_000,
+  mount: 15_000,
+  unmount: 3_000,
```
NethmiRodrigo: Is there a danger if we increase this timeout? Because, if an unmount for some reason takes longer than 3s (which would probably be extremely rare), wouldn't the parcel be marked as broken? Doesn't single-spa keep parcels whose unmounts are rejected?

ibacher: Good catch! I was initially thinking "no", mostly because this is the single-spa default and I've never seen it log, but it makes sense to give things more time...

packages/framework/esm-extensions/src/render.ts
ibacher: Yes, that's a great idea!

ibacher:
Anything else that needs to be addressed here?

denniskigen (approved):
LGTM! Thanks, @ibacher. The real single-spa tests are a nice touch.

PR #1868, opened 2026-09-04
(BREAKING) O3-5853: Use browserslist to compile JS in modern format

ibacher:
## Requirements

## Summary

When we switched to swc, we accidentally lost support for the browserlist, which means our code has been compiled for ES5 for a while. This fixes that by resolving the browserlist and injecting it at appropriate points into the rspack / webpack builds.

One breakage: webpack no longer honors .swcrc in order to match rspack's behavior.

The idea here is that this will help us output somewhat less convoluted and easier-to-read JS, though as the bundle-size check shows, this is a pretty small effect.

## Related Issue

https://issues.openmrs.org/browse/O3-5853

### DIFF
--- a/packages/shell/esm-app-shell/rspack.config.js
+++ b/packages/shell/esm-app-shell/rspack.config.js
@@ -21,6 +21,16 @@ const frameworkVersion = require('@openmrs/esm-framework/package.json').version;
 
 const timestamp = getTimestamp();
 const production = 'production';
+
+/**
+ * The browserslist queries swc compiles the app shell for. Given no target, swc down-levels to ES5 and
+ * every supported browser pays for transform helpers it doesn't need.
+ *
+ * Read from `browserslist-config-openmrs` rather than through this package's own `browserslist` field,
+ * as swc can't follow the extends we were using.
+ */
+const browserTargets = require('browserslist-config-openmrs');
+
 const allowedSuffixes = ['-app', '-widgets'];
 
 const openmrsAddCookie = process.env.OMRS_ADD_COOKIE;
@@ -237,7 +247,9 @@ module.exports = (env, argv = []) => {
       publicPath: '',
       hashFunction: 'xxhash64',
     },
-    target: 'web',
+    // set targets for the rspack runtime code which is not run through swc
+    // targets are expanded here because rspack can't properly handle browserlist extends
+    target: ['web', `browserslist:${browserTargets.join(', ')}`],
     // Module Federation v1.5 is incompatible with lazy compilation
     lazyCompilation: false,
     devServer: {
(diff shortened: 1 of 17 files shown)


packages/tooling/rspack-config/src/index.ts
```diff
@@ -108,6 +113,56 @@ function getFrameworkVersion() {
...
+    }
+
+    seen.add(extended[1]);
+
+    try {
+      // Resolved from the module being built, so that it picks up that module's own shared config.
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      const resolved = require(require.resolve(extended[1], { paths: [root] }));
+
+      return expandBrowserslistExtends(Array.isArray(resolved) ? resolved : [resolved], root, seen);
```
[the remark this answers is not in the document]
ibacher: I see your point and will adjust the PR, but [by convention](https://raw.githubusercontent.com/openmrs/openmrs-rfc-frontend/refs/heads/master/text/0003-browser-support.md) we don't really support that syntax.

ibacher:
I've seen the E2E test failures. Will fix.

packages/shell/esm-app-shell/rspack.config.js
```diff
@@ -369,6 +381,12 @@ module.exports = (env, argv = []) => {
           use: [
             {
               loader: 'builtin:swc-loader',
+              options: {
+                // No `jsc.parser`, so that swc keeps inferring syntax from each file's extension.
+                env: {
+                  targets: browserTargets,
```
denniskigen, line 391: Traced the e2e failure to `src/events.ts`: `getCurrentUser()` can emit synchronously, and the callback reads `subscription` before initialization. The old ES5 output masked this by using `var`. With the new targets it throws, skips `setupOptionalDependencies()`, and leaves an error toast blocking the test clicks. `filter` plus `take(1)` should avoid the self-reference.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -239,7 +293,16 @@ export default (env: Record<string, string>, argv: Record<string, string> = {})
...
             exclude: /node_modules/,
-            use: require.resolve('swc-loader'),
+            use: {
+              loader: require.resolve('swc-loader'),
+              options: {
+                env: {
+                  targets: browserTargets,
+                },
+                // ignore a project .swcrc to match rspack behavior
+                swcrc: false,
```
denniskigen, line 404: Can we document this as a breaking change and show how to move existing `.swcrc` options into `scriptRuleConfig`? I checked an app using the automatic JSX runtime in `.swcrc`: the previous config honored it, while this emits `React.createElement`. Apps relying on that setting can still build successfully but fail when rendering without a `React` binding.

packages/shell/esm-app-shell/rspack.config.js
ibacher: Fixed.

denniskigen (approved):
LGTM. Thanks, @ibacher, for fixing the startup issue and adding regression coverage for both bundlers!

PR #1869, opened 2026-09-04
(BREAKING) Switch from the style-loader to CSS loader in production mode

ibacher:
## Requirements

## Summary

This is the implementation of [RFC-0033](https://github.com/openmrs/openmrs-rfc-frontend/pull/34), removing our use of CSS-in-JS from apps and the framework. In order to keep the output of both rspack and webpack in sync, it implements a slightly larger surface than expected, basically ensuring we're using a consistent browserslist.

This also removes the "main" chunk we were generating, but which was unreachable, since our entry point was always the module federation loader and then the index chunk.

This adds a rspack / webpack plugin to enforce the rule that an app must not change global styles of Carbon components, but may make any adjustments provided those are appropriately scoped.

Finally, this cleans-up some SCSS imports in this repo that imported _far_ more stuff than we really needed (i.e., we were importing all of the button rules to import a couple of tokens).

## Other

This has some overlap with both #1868 in using browserslists. The code should be identical, but this should be merged after that. It also has some minor conflicts with #1866. I'm not sure which should land first, but the conflicts will be easy to address.

I expect that this PR will overstate the change in the bundle-size. Dropping the main chunk drops some duplicated but never actually loaded code. There's a real loss from the size of the styleguide and implementer-tools bundles. We should probably audit our other repos for unnecessarily broad imports, though these _likely_ will actually trigger the plugin which is how I noticed them. Note, though, that these gains are also somewhat overstated since we bundle the styleguide CSS everywhere (currently working on that).


packages/tooling/carbon-css-guard/src/index.ts
```diff
@@ -0,0 +1,301 @@
...
+ */
+const carbonAnchorPattern = /^\.cds--/;
+
+const carbonPrefix = '.cds--';
+
+/**
+ * Exempt from the anchorless rule. `:root` and `:host` carry custom properties, which add to the
+ * cascade rather than restyling anything, and are how a module contributes its own design tokens.
+ */
+const anchorlessExemptPattern = /^:(?:root|host)\b/;
```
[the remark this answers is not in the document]
ibacher: These exclusions, as noted in the comment, are intentional.

packages/apps/esm-implementer-tools-app/src/configuration/interactive-editor/value-editor.styles.scss
```diff
@@ -1,5 +1,6 @@
 @use '@carbon/layout';
-@use '@carbon/styles/scss/components/button';
+// `.../components/button` would emit Carbon's entire button stylesheet; the tokens module emits none.
+@use '@carbon/styles/scss/components/button/tokens' as button;
```
denniskigen, line 3: nit: and this is probably something for me to handle, but can we follow up with a lint rule for Carbon imports that emit component styles? It'd catch these earlier than the production build. We'd still need the guard for the compiled output, and imports of tokens, functions, and mixins should remain allowed.

packages/tooling/carbon-css-guard/src/index.ts
```diff
@@ -0,0 +1,301 @@
...
+
+/** Whether this single (comma-free) selector restyles the page rather than the module's own markup. */
+function isGlobal(selector: string): boolean {
+  const trimmed = selector.trim();
+
+  if (trimmed.length === 0) {
+    return false;
+  }
+
+  const anchors = stripNegations(trimmed).match(anchorPattern) ?? [];
```
denniskigen, line 229: Should `[data-extension-id=…]` and `[data-extension-slot-name=…]` count as anchors? `:global([data-extension-id='sticky-notes-button']):empty` targets wrappers for that specific extension, but currently gets treated like `[dir=rtl]`. It's probably fine either way, but patient-chart has three files that fail on these selectors alone.

denniskigen (approved):
LGTM. Thanks, @ibacher! Production builds passed with both bundlers, and I didn't find regressions in the UI flows I tested.

denniskigen:
Can we mark this as a breaking tooling change and add migration guidance? Patient-chart has at least six stylesheets across four packages that fail the guard, including print, RTL, and extension-wrapper rules. Seven other stylesheets didn't compile standalone, so that's a lower bound. The other-repos e2e job doesn't rebuild those apps with the new tooling, so CI won't catch this before the bump.

None of that blocks this PR. Can we line up a patient-chart sweep before the bump and decide where the print and RTL rules should live? The error points at the app shell, but a module author can't put them there.

packages/tooling/webpack-config/src/index.ts
```diff
@@ -214,20 +277,30 @@ export default (env: Record<string, string>, argv: Record<string, string> = {})
...
       },
     },
-  };
+  });
+
+  // Production emits real `.css` assets; development keeps `style-loader` for its better HMR story.
+  // See RFC 0033. The chunk-loading runtime resolves a chunk only once its stylesheet has loaded,
+  // so nothing mounts unstyled.
+  const styleLoader = () =>
+    isProd ? { loader: require.resolve(MiniCssExtractPlugin.loader) } : { loader: require.resolve('style-loader') };
```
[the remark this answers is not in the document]
denniskigen: I'd second this. The fixture exercises the shared webpack config in production, but imports no CSS. Can we add a `.scss` import, assert that a `.css` asset is emitted, and check that a bare `:global(.cds--btn)` fails with both bundlers? That would cover extraction and the guard. We verified stylesheet loading locally too, though that needs a separate runtime assertion.

packages/tooling/carbon-css-guard/src/index.ts
ibacher: Yes, this is definitely something we should address.

ibacher: Alright, `[data-extension-*]` attribute selectors should now be treated like any other valid selectors.

packages/tooling/carbon-css-guard/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "keywords": [
+    "openmrs",
+    "microfrontends",
+    "webpack",
+    "rspack",
+    "config"
+  ],
+  "homepage": "https://github.com/openmrs/openmrs-esm-core#readme",
+  "devDependencies": {
+    "@rspack/core": "1.7.9",
```
NethmiRodrigo: Are these used anywhere? Doesn't the plugin declare its own compiler interface?

packages/tooling/carbon-css-guard/src/index.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  return out;
+}
+
+function classify(prelude: string): BlockKind {
+  if (!prelude.startsWith('@')) {
+    return 'style';
+  }
+
+  return /^@(?:-\w+-)?keyframes\b/.test(prelude) ? 'keyframes' : 'at-rule';
```
NethmiRodrigo: Would it make sense to treat `@scope` as a style block? `@scope (.myThing) { ... }` does limit the inner rules to `.myThing`, so it seems like a safe case to let through.
```suggestion
  if (/^@scope\b/.test(prelude)) {
    return 'style';
  }

  return /^@(?:-\w+-)?keyframes\b/.test(prelude) ? 'keyframes' : 'at-rule';
```

packages/tooling/rspack-config/src/index.ts
```diff
@@ -108,6 +113,56 @@ function getFrameworkVersion() {
...
+function browserslistQueries(root: string): Array<string> {
+  const loaded = browserslist.loadConfig({ path: root });
+  const configured = loaded === undefined ? [] : Array.isArray(loaded) ? loaded : [loaded];
+
+  return expandBrowserslistExtends(configured.length > 0 ? configured : defaultBrowserslistQueries, root);
+}
+
+function expandBrowserslistExtends(queries: Array<string>, root: string, seen = new Set<string>()): Array<string> {
+  return queries.flatMap((query) => {
+    const extended = /^extends\s+(\S+)$/.exec(query.trim());
```
NethmiRodrigo: Small one: browserslist accepts `Extends foo`, but this regex does not, so the query would reach Lightning CSS unexpanded.
```suggestion
    const extended = /^extends\s+(\S+)$/i.exec(query.trim());
```

packages/tooling/rspack-config/src/index.ts
```diff
@@ -108,6 +113,56 @@ function getFrameworkVersion() {
...
+      return expandBrowserslistExtends(Array.isArray(resolved) ? resolved : [resolved], root, seen);
+    } catch {
+      // A config that can't be loaded shouldn't take the whole build — or the dev server — down over
+      // which browsers it targets. Fall back to the default queries and say so.
+      console.warn(
+        `Could not load the browserslist config "${extended[1]}". Targeting ${defaultBrowserslistQueries.join(
+          ', ',
+        )} instead.`,
+      );
+
```
NethmiRodrigo: Could we include the error in the warning? It would help tell a missing package apart from a malformed config. Also, since this runs inside `flatMap`, the defaults get added to the other queries rather than used instead of them, so maybe the wording could say that.
```suggestion
    } catch (error) {
      // A config that can't be loaded shouldn't take the whole build — or the dev server — down over
      // which browsers it targets. Fall back to the default queries and say so.
      console.warn(
        `Could not load the browserslist config "${extended[1]}": ${error}. Adding ${defaultBrowserslistQueries.join(
          ', ',
        )} to the targets.`,
      );
```

packages/tooling/rspack-config/src/index.ts
```diff
@@ -440,5 +523,13 @@ export default (env: Record<string, string>, argv: Record<string, string> = {})
     }),
     ...overrides,
   };
-  return mergeWith(baseConfig, additionalConfig, mergeFunction);
+  const config = mergeWith(baseConfig, additionalConfig, mergeFunction);
+
+  // Appended after the merge rather than listed above, because `overrides.plugins` replaces the
+  // plugin array wholesale and an app doing that would drop the guard without noticing. Same
+  // reasoning as the `ExternalsPlugin` block above, which avoids `externals` for the same reason.
```
NethmiRodrigo: I might be reading this wrong, but doesn't `mergeFunction` concatenate arrays? So `overrides.plugins` would add to the list rather than replace it.
```suggestion
  // Appended after the merge so that no `overrides` or `additionalConfig` entry can filter it out.
```

packages/tooling/rspack-config/src/index.ts
```diff
@@ -185,6 +240,10 @@ export const watchConfig: Partial<OpenmrsRspackConfig['watchOptions']> = {};
  * This object will be merged with the webpack optimization
  * object.
  * Make sure to modify this object and not reassign it.
+ *
+ * Arrays here merge by index rather than replacing, so a `minimizer` set on this object lands on top of
+ * the defaults entry by entry instead of taking their place. Use `overrides.optimization` to replace
+ * them outright.
```
NethmiRodrigo, line 300: Same thing here: `overrides.optimization` goes through the same concat, so does it replace the array?
```suggestion
 * Arrays here merge by index rather than replacing, so a `minimizer` set on this object lands on top of
 * the defaults entry by entry instead of taking their place.
```

NethmiRodrigo (commented):
Thanks @ibacher!

packages/tooling/webpack-config/src/index.ts
```diff
@@ -111,6 +118,56 @@ function getFrameworkVersion() {
...
+function browserslistQueries(root: string): Array<string> {
+  const loaded = browserslist.loadConfig({ path: root });
+  const configured = loaded === undefined ? [] : Array.isArray(loaded) ? loaded : [loaded];
+
+  return expandBrowserslistExtends(configured.length > 0 ? configured : defaultBrowserslistQueries, root);
+}
+
+function expandBrowserslistExtends(queries: Array<string>, root: string, seen = new Set<string>()): Array<string> {
+  return queries.flatMap((query) => {
+    const extended = /^extends\s+(\S+)$/.exec(query.trim());
```
NethmiRodrigo: Same as the rspack-config comments above :)

packages/tooling/rspack-config/src/index.ts
ibacher: Yeah, the comment is somewhat misleading. The actual problem is line 524 (of this iteration) which does `...overrides`. That mechanism replaces the array (and we should probably deprecate it).

ibacher:
>  decide where the print and RTL rules should live?

For the most part, the answer to this is that they need to be re-written to be narrower or use a module-specific selector. Some of these rules (e.g., [this one](https://github.com/openmrs/openmrs-esm-patient-chart/blob/b34d459fa67899997a5c7cb4183da47132acb22a/packages/esm-patient-orders-app/src/components/order-details-table.scss#L136)) probably should just be part of the global stylesheet.

NethmiRodrigo (approved):
Thanks @ibacher!

ibacher:
Alright, so status here: We've got 3 PRs to fix 3 repos (+ 1 [commit](https://github.com/openmrs/openmrs-esm-stock-management/commit/babd91e98530d3507d73c4bdbe9b60b735425c6b) I pushed to stock-management because it just removes a completely inert rule.

There are two other apps that need clean-up:

* openmrs-esm-laboratory-app: src/lab-tabs/modals/reject-lab-request-modal.scss has a rule for `section`. It appears this was meant to be `.section`, but it seems like it should be removed and the ineffective reference in reject-lab-request-modal.component.tsx deleted.
*  openmrs-esm-form-builder:
    * src/components/form-editor/form-editor.scss: uses `button`, not `.container button`
    * src/.../rendering-types/inputs/markdown/markdown.scss: has `:global(sup)` and `:global(sub)`. These need to be either removed (they replicate rules that don't exist anymore) or properly anchored

Once those PRs are fixed, this should be mergeable without breaking anything.

PR #1872, opened 2026-09-11
(fix) When using yarn run:shell ensure we serve CSS files

ibacher:
## Requirements

## Summary

#1869 accidentally broke `yarn run:shell` because we now serve CSS files from apps.

Previously, we had a weird `pathRewrite` rule to try and figure out if something was matching a file or not because URLs that aren't for a filename should generally be served the single page. That URL only allowed `.html` or `.js`. This switches to a more generic rule where the proxy doesn't touch paths that are requests for files and more closely mirrors how things work in production.

### DIFF
--- a/.changeset/public-snakes-poke.md
+++ b/.changeset/public-snakes-poke.md
@@ -1,0 +1,5 @@
+---
+"@openmrs/esm-app-shell": patch
+---
+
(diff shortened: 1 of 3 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1874, opened 2026-09-11
(fix) Release broken parcel DOM references

ibacher:
## Requirements

## Summary

Another single-spa fix building on the framework timeouts work.

When single-spa bootstraps or unmounts a parcel and the promise rejects due to w/e error, single-spa retains the parcel in its internal registry, including the props. The most expensive part of this is that single-spa can wind up retaining the `domElement` the parcel belongs to, which might actually be large depending on the content of the parcel. We can't clean-up single-spa's internal registry (and there's some value in being able to see broken parcels in its dev tools), but we own the props that we pass to single-spa which single-spa uses to store the `domElement`, so here, if unmounting fails, we release the props, including the `domElement`.

Basically, this reduces the memory pressure from having a bunch of "ghost" single-spa parcels hanging around at the cost of needing to copy the props passed.

This tries to preserve the semantics of rejections still winding up in the global reject handler.


packages/framework/esm-extensions/src/render.ts
```diff
@@ -182,7 +250,11 @@ export async function renderParcel<T = CustomProps>(
   customProps: ParcelProps & T,
 ): Promise<ReturnType<MountParcel>> {
   const mountParcel = await getParcelMounter();
-  return mountParcel(withLifecycleDeadlines(parcelConfig), customProps);
+  // Copied because single-spa holds onto this object for as long as it holds the parcel, and
+  // {@link releaseProps} empties it; the caller's own object is unchanged.
+  const props = { ...customProps };
+
+  return mountParcel(withLifecycleDeadlines(parcelConfig, trackForRelease(props)), props);
```
denniskigen: Good stuff, @ibacher. Can we track the props passed through `update()` too? single-spa replaces `parcel.customProps` on update, so a subsequent failed unmount clears the initial copy while retaining the updated `domElement`. Our workspace `<Parcel>` includes `domElement` in those updates. I reproduced this against real single-spa.

It’s probably best to copy and retarget the tracked props on update, and release them if the update lifecycle leaves the parcel `SKIP_BECAUSE_BROKEN`. We shouldn’t clear them on every update rejection: calling `update()` on an unmounted parcel rejects too, but that parcel can still remount.

Can we cover update → failed unmount, failed update lifecycle, and rejected update → successful remount in the real single-spa tests?

ibacher: Solid point. Will do.

ibacher: The fix here feels a little dirty, but hey!

denniskigen (approved):
LGTM. Thanks, @ibacher, for covering the update-props cases too!

denniskigen (approved):
LGTM. Thanks, @ibacher! I tried this locally with an unmount that fails on purpose. On main the detached element and its 1,000 children are still there after the timers expire. With this PR they're gone, and the clean-unmount control is collected too.

PR #1876, opened 2026-09-11
(fix) Prevent spurious notifications in subscribeToContext and add unit tests

bhuvan-somisetty:
## Requirements

## Summary
- **Bug Fix in `@openmrs/esm-context`**: In `subscribeToContext`, `previous` was initialized with the clone returned by `getContext(namespace)` instead of the reference from the raw store state. Because `getContext` returns a new frozen object, any subsequent state change across the store (even to an unrelated namespace) caused `current !== previous` to evaluate to `true`, triggering spurious subscriber callbacks. Initializing `previous` with `initialState[namespace]` ensures listeners only fire when their subscribed namespace changes.
- **Unit Tests**: Added comprehensive test suite for `@openmrs/esm-context` covering `registerContext`, `unregisterContext`, `getContext` (with selector and immutability checks), `updateContext`, and `subscribeToContext` (with isolation and unsubscribe checks).

## Screenshots
*N/A (no UI changes)*

## Related Issue
*N/A*

## Other
Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>

### DIFF
--- a/packages/framework/esm-context/src/context.test.ts
+++ b/packages/framework/esm-context/src/context.test.ts
@@ -1,0 +1,181 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import {
+  contextStore,
+  getContext,
+  registerContext,
+  subscribeToContext,
+  unregisterContext,
+  updateContext,
+} from './context';
+
+describe('esm-context', () => {
+  beforeEach(() => {
+    // Reset the context store state between tests
+    contextStore.setState({}, true);
+  });
+
+  describe('registerContext', () => {
+    it('registers a namespace with an empty object as default initial value', () => {
+      registerContext('test-namespace');
+      expect(getContext('test-namespace')).toEqual({});
+    });
+
+    it('registers a namespace with a provided initial value', () => {
+      const initialValue = { user: 'admin', role: 'clinician' };
+      registerContext('user-context', initialValue);
+      expect(getContext('user-context')).toEqual(initialValue);
+    });
+
+    it('throws an error when attempting to re-register an already-registered namespace', () => {
+      registerContext('duplicate-namespace', { count: 1 });
+      expect(() => {
+        registerContext('duplicate-namespace', { count: 2 });
+      }).toThrow(/Attempted to re-register namespace duplicate-namespace/);
+    });
+  });
+
+  describe('unregisterContext', () => {
+    it('removes an existing namespace and all associated data', () => {
+      registerContext('temp-namespace', { active: true });
+      expect(getContext('temp-namespace')).toEqual({ active: true });
+
+      unregisterContext('temp-namespace');
+      expect(getContext('temp-namespace')).toBeNull();
+    });
+
+    it('does not throw when attempting to unregister a non-existent namespace', () => {
+      expect(() => {
+        unregisterContext('non-existent-namespace');
+      }).not.toThrow();
+    });
+  });
+
+  describe('getContext', () => {
+    it('returns null if the namespace is not registered', () => {
+      expect(getContext('unknown-namespace')).toBeNull();
+    });
+
+    it('returns a frozen (immutable) object of the current namespace state', () => {
+      registerContext('immutable-ns', { key: 'value' });
+      const context = getContext<{ key: string }>('immutable-ns');
+
+      expect(context).toEqual({ key: 'value' });
+      expect(Object.isFrozen(context)).toBe(true);
+    });
+
+    it('applies a custom selector function when provided', () => {
+      interface PatientContext {
+        patient: { id: string; name: string };
+        visits: Array<string>;
+      }
+
+      registerContext<PatientContext>('patient-ns', {
+        patient: { id: '123', name: 'John Doe' },
+        visits: ['v1', 'v2'],
+      });
+
+      const selectedName = getContext<PatientContext, { name: string }>('patient-ns', (state) => ({
+        name: state.patient.name,
+      }));
+
+      expect(selectedName).toEqual({ name: 'John Doe' });
+      expect(Object.isFrozen(selectedName)).toBe(true);
+    });
+  });
+
+  describe('updateContext', () => {
+    it('updates an existing namespace state via updater function', () => {
+      registerContext('counter-ns', { count: 0 });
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Added unit tests for `@openmrs/esm-context` and fixed an issue in `subscribeToContext` where subscribers were erroneously triggered when unrelated namespaces in the store were updated.

packages/framework/esm-context/src/context.test.ts
```diff
@@ -0,0 +1,181 @@
...
+      expect(getContext('ns-2')).toEqual({ value: 2 });
+    });
+  });
+
+  describe('subscribeToContext', () => {
+    it('immediately calls callback with initial state', () => {
+      registerContext('sub-ns', { initialized: true });
+      const callback = vi.fn();
+
+      subscribeToContext('sub-ns', callback);
```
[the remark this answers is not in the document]
bhuvan-somisetty: Added an `afterEach` hook and tracked all subscriptions so they are explicitly cleaned up after each test run.

bhuvan-somisetty:
Hey @denniskigen, whenever you have a moment, could you please take a look at this PR and approve the workflows? Thank you!

packages/framework/esm-context/src/context.ts
```diff
@@ -115,7 +115,8 @@ export function subscribeToContext<T extends NonNullable<object> = NonNullable<o
   namespace: string,
   callback: ContextCallback<T>,
 ) {
-  let previous = getContext<T>(namespace);
+  const initialState = contextStore.getState();
+  let previous: Readonly<T> | null = namespace in initialState ? (initialState[namespace] as T) : null;
```
denniskigen: Greptile's point about same-object updaters holds, though it needs a qualification. On the base branch, with repeated same-object updates to an already-registered namespace, only the first store change notified the subscriber. An unrelated store change could use up that first notification too. Subsequent same-object updates were already silent; this change suppresses the first one as well.

Subscribe to an existing namespace, then update it twice:

| Updater | Base branch | This PR |
| --- | --- | --- |
| Mutate and return the same object | Notifies once, then goes silent | Never notifies |
| Return a new object | Notifies twice | Notifies twice |

Can we preserve the [documented mutate-and-return behavior](https://github.com/openmrs/openmrs-esm-core/blob/b7e21cdb8029bea9f80bd35d0fa68f42ca9f6246/packages/framework/esm-react-utils/src/useDefineAppContext.ts#L21-L31) while keeping unrelated namespace updates silent? Copying the updater's result fixes the plain-object case, but `Object.assign({}, …)` changes arrays and class instances into plain objects, so we'd need to account for those too.

packages/framework/esm-context/src/context.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      registerContext('ns-2', { value: 2 });
+
+      updateContext<{ value: number }>('ns-1', (state) => ({ value: state.value + 10 }));
+
+      expect(getContext('ns-1')).toEqual({ value: 11 });
+      expect(getContext('ns-2')).toEqual({ value: 2 });
+    });
+  });
+
+  describe('subscribeToContext', () => {
```
denniskigen, line 127: Can we add a test here for the same-object case, so it can't regress again? Something like:

```ts
it('notifies when an updater mutates and returns the same object', () => {
  registerContext('mutate-ns', { count: 0 });
  const callback = vi.fn();
  unsubscribers.push(subscribeToContext('mutate-ns', callback));
  callback.mockClear();

  const mutate = (state: { count: number }) => {
    state.count += 1;
    return state;
  };
  updateContext('mutate-ns', mutate);
  updateContext('mutate-ns', mutate);

  expect(callback).toHaveBeenCalledTimes(2);
  expect(callback).toHaveBeenLastCalledWith({ count: 2 });
});
```

denniskigen (changes requested):
Thanks, @bhuvan-somisetty. The namespace-isolation fix makes sense, and the tests help pin down the intended behavior. I've left some feedback on the same-object case Greptile flagged.

bhuvan-somisetty:
Good catch, thanks! Pushed a fix, updateContext now shallow-clones the updater's result (type-aware, so arrays/class instances stay intact) so the namespace's reference always changes on an update, whether the updater mutates in place or returns a new object. That keeps unrelated namespace updates silent while making sure our own namespace updates always notify. Added the test you suggested too.

packages/framework/esm-context/src/context.ts
```diff
@@ -95,11 +95,26 @@ export function updateContext<T extends NonNullable<object> = NonNullable<object
...
+ * object, so that `subscribeToContext` can reliably detect the change via identity
+ * comparison without mistaking arrays or class instances for plain objects.
+ */
+function cloneNamespaceValue<T extends NonNullable<object>>(value: T): T {
+  if (Array.isArray(value)) {
+    return [...value] as unknown as T;
+  }
+
+  return Object.assign(Object.create(Object.getPrototypeOf(value)), value) as T;
+}
```
ibacher: I'm not sure about this... There are classes of things this can't properly clone that could be assigned as a namespace value (Maps, Sets, objects with private members). I think we can do away with it and fix things in the subscription.

packages/framework/esm-context/src/context.ts
ibacher, line 133:
```suggestion
let previousValue: Readonly<T> | null = namespace in initialState ? (initialState[namespace] as T) : null;
let previousSnapshot = snapshot(previousValue);

callback(previousSnapshot);

return contextStore.subscribe((state) => {
	const current: Readonly<T> | null = namespace in state ? (state[namespace] as T) : null;
	
	if (current !== previous || !shallowEqual(currentSnapshot, previousSnapshot)) {
		previous = current;
		previousSnapshot = currentSnapshot;
		callback(currentSnapshot);
	}
```

I think this might be a more sensible implementation.

bhuvan-somisetty:
Fixed! Dropped the clone-based workaround and switched subscribeToContext to compare a shallow snapshot instead, so mutate-and-return updaters notify correctly without the Maps/Sets edge case. Thanks for catching that.

ibacher (approved):
Thanks @bhuvan-somisetty!

PR #1877, opened 2026-09-14
(fix) O3-5971: Guard module-load code against invalid temporary config

JordanCalzen:
## Requirements
- [x] I have read the [OpenMRS contributing guidelines](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25469)
- [x] My code follows the OpenMRS code style
- [ ] I have added tests that prove my fix is effective

## Summary
Fixes O3-5971 — an invalid config saved in `localStorage['openmrs:temporaryConfig']`
crashes the app shell at startup with an unhandled ReferenceError. The only
recovery was manually clearing localStorage, which is not discoverable.

**Root cause — two problems compound each other:**

1. `let displayedValidationMessages` was declared at line 1151 but read by
   `logError()` which is called at module load via `setupConfigSubscriptions()`
   at line 95. Under modern SWC output (since PR #1868 added explicit browserslist
   targets) the `let` survives compilation, so the read falls in the temporal dead
   zone and throws a `ReferenceError`. Under the old ES5 output `let` became `var`,
   the read returned `undefined`, and the guard added in PR #1356 accidentally
   recovered — masking the bug for years.

2. The initial `recomputeAllConfigs()` call inside `setupConfigSubscriptions()`
   was not wrapped in the safe wrapper that every subscription-driven call already
   uses, so the error escaped as an unhandled rejection and the SPA never loaded.

**Changes:**
- Moved `let displayedValidationMessages` to the top of the module, immediately
  after imports, so `logError()` is safe to call at any point during module evaluation
- Removed the now-dead guard inside `logError()` that PR #1356 added as a workaround
- Changed the bare `recomputeAllConfigs()` in `setupConfigSubscriptions()` to
  `recomputeAllConfigsSafely()`

## Screenshots
No UI changes — this is a runtime crash fix. The visible difference is the app
shell loading normally instead of showing a blank page when an invalid config
is present in localStorage.

## Related Issue
O3-5971

## Other
To reproduce before the fix:
1. In DevTools console paste:
```js
localStorage.setItem(
  'openmrs:temporaryConfig',
  JSON.stringify({ "fake-module": { "nonexistent-slot": { "badKey": true } } })
);
```
2. Hard reload — app shell never loads, ReferenceError in console
3. After fix — app loads, `console.error` names the offending key


JordanCalzen:
Hi @denniskigen @ibacher ,  I've opened a PR for O3-5971 (invalid implementer tools config 
crashes the app shell at startup) and would appreciate a review when you 
get a chance.

PR #1878, opened 2026-09-14
(fix) O3-5975: Resolve useConfig suspense from stale config promises

abheeee03:
## Requirements

## Summary
Two bugs in `useConfig` 's Suspense promise mechanism could leave a component suspended forever:

- Promise factories never checked current store state — if config was already loaded at creation time, no subscription event would fire and the promise would hang.
- Settled promises were never evicted from cache — a re-suspend for the same key would re-throw an old resolved promise that no longer tracked the store.

Fix: Both factories now check `store.getState()` before subscribing (resolving immediately if already loaded), and each promise evicts its own cache entry on settle so re-suspends always create a fresh, live promise.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5975


packages/framework/esm-react-utils/src/useConfig.ts
```diff
@@ -42,19 +42,44 @@ function readInitialExtensionConfig(store: StoreApi<ExtensionsConfigStore>, exte
   return null;
 }
 
-function createConfigPromise(store: StoreApi<ConfigStore>) {
-  return new Promise<ConfigObject>((resolve) => {
+function createConfigPromise(store: StoreApi<ConfigStore>, cacheId: string) {
+  const p = new Promise<ConfigObject>((resolve) => {
```
ibacher, line 46:
```suggestion
  return new Promise<ConfigObject>((resolve) => {
    // ...
   ).then(() => delete promises[cacheId]);
```

ibacher: That said, I think `.then()` should actually be `.finally()`.

packages/framework/esm-react-utils/src/useConfig.ts
```diff
@@ -42,19 +42,44 @@ function readInitialExtensionConfig(store: StoreApi<ExtensionsConfigStore>, exte
...
+    if (current.loaded && current.config) {
+      resolve(current.config);
+      return;
+    }
     const unsubscribe = store.subscribe((state) => {
       if (state.loaded && state.config) {
         resolve(state.config);
         unsubscribe();
       }
     });
```
ibacher, line 58: This is probably how this should've been written everywhere so the promise resolves as long as the config is loaded.

```suggestion
const unsubscribe = store.subscribe((state) => {
  if (state.loaded) {
    resolve(state.config ?? {});
    unsubscribe();
   }
});
```

packages/framework/esm-react-utils/src/useConfig.test.tsx
```diff
@@ -343,3 +345,214 @@ describe(`useConfig in an extension`, () => {
...
+          },
+        }}
+      >
+        <RenderConfig configKey={configKey} />
+      </ComponentContext.Provider>
+    </React.Suspense>,
+  );
+}
+
+describe('useConfig Suspense — module-config path', () => {
```
ibacher, line 385: I don't see why we need a new top-level test block. I get that it's probably easier to write the tests this way, but it would be better if these test could be worked in alongside the existing tests and especially if the existing tests could be refactored to use the new helpers.

packages/framework/esm-react-utils/src/useConfig.test.tsx
```diff
@@ -343,3 +345,214 @@ describe(`useConfig in an extension`, () => {
...
+
+    // simulate the config becoming available.
+    act(() => {
+      store.setState({ loaded: true, config: { greeting: 'hello' }, translationOverridesLoaded: false });
+    });
+
+    await screen.findByText('hello');
+  });
+
+  it('does not throw a stale settled promise on re-suspend after a store reset', async () => {
```
ibacher, line 419: So I think for this and the test under extension-config pass, it's necessary to run things in this order:

1. Render the component — Component suspends
2. Unmount the component — Now the promise is stale
3. Push the config to the store
4. Re-render the component

With the store populated for the first render, the render won't leave the dangling stale promise but that's exactly what we need to fix.

ibacher (changes requested):
Thanks @abheeee03! This is quite close to being there. I think there's just a few things to address before this is ready-to-go.

PR #1881, opened 2026-09-16
(feat) O3-5823: Add certainty display to shared diagnosis tags

denniskigen:
## Requirements

## Summary
The visit-note changes in openmrs/openmrs-esm-patient-chart#3621 capture diagnosis certainty. We need to display that through the shared `DiagnosisTags` component so apps use the same presentation.

This adds an optional `showCertainty` prop, defaulting to `false`. When enabled, recorded CONFIRMED and PROVISIONAL values appear alongside the name using core translations. Missing or unknown certainty has no label, and primary/secondary colours still come from configuration.

Long names truncate while the certainty stays visible. Only truncated names become interactive: hover, focus or click reveals the full name in a tooltip, Escape closes it, and the pointer can move onto it. The tooltip is rendered at the document body so scrolling table cells don't clip it. If a focused tag stops truncating because its container widened, it keeps focus until you tab away.

One change for existing callers even with the prop off: the pills are now rendered with Carbon's tag classes rather than the `Tag` component, so a truncated name gets this tooltip instead of Carbon's click-only one.

## Screenshots

Storybook, WithCertainty story.

https://github.com/user-attachments/assets/7a08914b-efc4-4213-b858-52abd66e5d36

## Related Issue
https://openmrs.atlassian.net/browse/O3-5823

Related: openmrs/openmrs-esm-patient-chart#3621

## Other
Callers opt in with `showCertainty` once this ships. A patient-chart follow-up switches the visit summary and visit table to it. This doesn't change visit-level diagnosis aggregation.

The text-suffix treatment is the first option in the ticket. Design hasn't weighed in yet, hence the unchecked box above.


VeronicaMuthee:
Thanks @denniskigen, quick one
1. On this [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3621), @ibacher asked for null certainty to display as Provisional; should we handle that here?
2. Ian also asked for a note on the Diagnosis type documenting how missing certainty is interpreted - do we handle this here also
3. nit — double spacing in the suffix: the suffix span has both a literal leading space (<span …> ({certaintyLabel})</span>) and margin-inline-start: $spacing-01; one of the two should go. (The mocks bake in the space too, so they'd follow.)
4. Also, there is a check failing

denniskigen:
Thanks, @VeronicaMuthee.

1 and 2. @ibacher, I checked the backend: certainty is non-nullable in the database, the validator rejects null, and REST requires it on create. A frontend representation can omit it, though, so defaulting a missing value to Provisional could mislabel a confirmed diagnosis. Would it be better to show nothing and document on `Diagnosis.certainty` that callers need to fetch it to display it?

3. The leading space collapses at the start of the flex item, so the margin supplies the visual gap. The space keeps the combined text content readable as “Asthma (Confirmed)”.

4. That’s the docs drift check. Reviewer approval triggers regeneration.

NethmiRodrigo (approved):
Thanks @denniskigen! LGTM

PR #1885, opened 2026-09-17
(fix) Pass the browserslist targets to the app shell's CSS minimizer

denniskigen:
## Requirements

## Summary
The app shell's production build minifies CSS with rspack's default Lightning CSS minimizer, which we never give browser targets. On its own defaults Lightning CSS downlevels every logical property in the shell's CSS, and that CSS includes all of Carbon, so each `border-inline-start`, `padding-inline-end` and so on becomes physical properties guarded by `:not(:is(:lang(ae), :lang(ar), ...)))`, repeated for `:-webkit-any` and `:-moz-any`, plus mirrored rules for RTL languages. On `main` that is 6,594 fallback rules, about two thirds of 3 MB of CSS, for browsers that predate our browserslist by years.

The shared rspack and webpack configs for modules already pass `browserslist-config-openmrs` to their CSS minimizers (#1869). The shell config was missed. This adds the same `minimizer` array to the shell's `optimization` block, reusing the `browserTargets` the config already hands to swc. `minimizer` replaces rspack's default pair rather than extending it, so the JS minimizer is listed alongside.

The styleguide stylesheet is marked as already minified. The config compiles it with sass, names it `openmrs.<hash>.css` after a hash of that output, and copies it in as a static file. Minifying it afterwards changed its bytes under a name that no longer matched them, which matters because it's served with a one-year `max-age`; this change alone would have shipped a very different file under the same URL. Sass already emits it compressed, so the copy pattern sets `info.minimized` and the minimizer leaves it alone. That costs about 3 KB gzipped and the filename now matches the bytes it serves.

Built from the same commit before and after:

| | Before | After |
| --- | --- | --- |
| Fallback rules, all shell CSS | 6,594 | 0 |
| All shell CSS, raw | 3.03 MB | 1.11 MB |
| Main stylesheet, gzipped | 163 KB | 104 KB |
| JS output | | byte-identical; only the entry file's hash moves, because it embeds the renamed CSS chunk names |

Logical properties now come through as written, the `:-moz-any` and `:-webkit-any` prefixes are gone, and direction resolves from `dir` the way our `html[dir='rtl']` overrides already assume, instead of from `lang`.

## Other
The minimizer honours the targets: the same one-rule build emits fallbacks for `chrome 60`, none for `chrome 120`, output identical to `chrome 120` for `browserslist-config-openmrs`, and fails on an invalid query. Distros pick this up with the next `@openmrs/esm-app-shell` release, since the `openmrs build` command requires the shell's `rspack.config.js`.

Moving the styleguide compilation out of the config factory altogether is left to O3-5966.


packages/shell/esm-app-shell/rspack.config.js
```diff
@@ -392,6 +394,18 @@ module.exports = (env, argv = []) => {
...
+      // given our browserslist; on its own defaults it downlevels every logical property for old browsers.
+      // The styleguide stylesheet is left out: it is named above after a hash of the sass output and copied
+      // in as a static file, so minifying it here would change its bytes under a name that no longer
+      // matches them. Sass already emits it compressed.
+      minimizer: [
+        new SwcJsMinimizerRspackPlugin(),
+        new LightningCssMinimizerRspackPlugin({
+          exclude: /^openmrs\.[a-f0-9]{16}\.css$/,
+          minimizerOptions: { targets: browserTargets },
+        }),
```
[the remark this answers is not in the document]
ibacher: This is the kind of LLM-suggestion nonsense I don't like.

packages/shell/esm-app-shell/rspack.config.js
```diff
@@ -470,7 +480,9 @@ module.exports = (env, argv = []) => {
       new CopyRspackPlugin({
         patterns: [
           { from: resolve(__dirname, 'src/assets') },
-          { from: resolve(cssTmpDir, openmrsCssFilename), to: openmrsCssFilename },
+          // Named above after a hash of the sass output, which already emits it compressed. Marked minified so
+          // the CSS minimizer leaves its bytes matching its name.
```
ibacher: I'm not sure this comment is helpful?

ibacher (approved):

packages/shell/esm-app-shell/rspack.config.js
denniskigen: Removed in latest commit.

PR #1887, opened 2026-09-18
(feat) O3-5986: Let attachments be recorded on and read by encounter

denniskigen:
## Requirements

## Summary

Images uploaded from the visit note form save fine and show up on the Attachments page, but they're gone when we reopen the note for editing. `createAttachment` only ever sends the patient, so the obs ends up encounterless and nothing ties it back to the note.

This PR effectuates the framework half of the fix. `createAttachment` takes an optional encounter UUID and sends it as the `encounter` form field. `getAttachments` and `useAttachments` take the same optional UUID and hit `/attachment?patient=...&encounter=...` when it's set. We leave `includeEncounterless` out of that request on purpose: the attachments module ignores the encounter filter whenever that flag is present ([AttachmentResource.search](https://github.com/openmrs/openmrs-module-attachments/blob/master/omod/src/main/java/org/openmrs/module/attachments/rest/AttachmentResource.java)). Both go through a new `getAttachmentsUrl` helper so the fetch and the SWR cache key are always the same string, and consumers can build that key when they need to invalidate it.

The `includeEncounterless` boolean is ignored when an encounter is given. That mirrors the backend, where the two filters are mutually exclusive, and keeps the existing two-argument calls working. An options object would read better but it's more churn for the same trade-off.

The patient-chart follow-up PR passes the note's encounter UUID on upload and loads the saved images when editing.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5986


ibacher (approved):
Thanks @denniskigen!

PR #1892, opened 2026-09-21
(chore) Adopt the shared Prettier configuration

denniskigen:
## Requirements

## Summary
We're replacing the copied Prettier settings with `@openmrs/prettier-config@^1.0.0` (the latest release of https://github.com/openmrs/openmrs-contrib-prettier-config) and pinning Prettier to `3.9.8`.

The shared configuration preserves the existing formatting settings. The formatter upgrade changes 15 source files; those changes are mechanical. Existing formatting issues elsewhere are left untouched.


packages/framework/esm-styleguide/src/page-header/page-header.component.tsx
```diff
@@ -24,8 +24,7 @@ export type ExcludeOptionalKeys<T, U> = {
 
 /** @internal */
 export type XOR<T, U> =
-  | (T & { [K in ExcludeOptionalKeys<U, T>]?: never })
-  | (U & { [K in ExcludeOptionalKeys<T, U>]?: never });
```
ibacher, line 28: I sort of like this part of the formatting we had previously. Especially for expressions like this, it reads much clearer than the compressed version.

ibacher: Although, maybe this is a result of prettier never being properly applied?

denniskigen: These files were already correctly formatted under Prettier 3.1.0. The shared config preserves the same options; the formatting changes come from the version bump.

It’s probably better to pin 3.8.5 here. That keeps the other formatting improvements without collapsing these unions or adding `prettier-ignore`. We’d align patient management and the shared config README too. Changing `printWidth` would cause much wider churn, so I’d leave that alone.

PR #1893, opened 2026-09-21
(BREAKING) Remove workspace v1

chibongho:
## Requirements

- [X] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.

## Summary

Removes the old Workspace V1 code.


packages/framework/esm-globals/src/type-utils.ts
```diff
@@ -23,12 +23,6 @@ export function isOpenmrsAppRoutes(routes: OpenmrsAppRoutes | unknown): routes i
       }
     }
 
-    if (Object.hasOwn(routes, 'workspaces')) {
```
chibongho, line 26: Not sure what this function does. We never put similar checks here for `workspace2`, `workspaceWindows2`, `workspaceGroups2`, but I can add those if needed.

chibongho:
Shoot... we still have some stuff in patient management and patient chart using the old API. Marking as draft for now.

packages/framework/esm-globals/src/type-utils.ts
ibacher: It's a type-guard. It's used in the place we parse a raw JSON string into an `OpenmrsAppRoutes` to make sure the structure looks like an `OpenmrsAppRoutes`. It also stops us from processing anything that doesn't have the right "shape". Probably a good idea to cover wrokspace2 at least here?

PR #1894, opened 2026-09-21
(BREAKING) Remove offline & rxjs support from framework

ibacher:
## Requirements

## Summary

Removes esm-offline, esm-offline-app, and rxjs from underlying APIs. Offline support has been out of date for a while, i.e., our version of `dexie` is 6-years old and the version of Workbox no longer really works with the current setup at all. Similarly, while providing rxjs via the framework made sense several years ago, we're now on a very old version that is likely holding back any consumers that would actually use rxjs. Since `Promises` can be trivially converted to `Observables` losslessly, especially given how we used them, this should allow downstream apps to independently update their rxjs version.

Retains a small compatibility shim with the most commonly-used public functions now stubbed. These deprecated shims will be removed in a future version.

There is a small implementation of subscribable notifications added to esm-styleguide to support the notification system.


packages/framework/esm-api/src/current-user.ts
```diff
@@ -26,77 +25,61 @@ export const sessionStore = createGlobalStore<SessionStore>('session', {
...
  *   without the surrounding session information.
- * @returns An Observable that produces {@link LoggedInUser} objects.
+ * @returns A Promise resolving to a {@link LoggedInUser} object.
  */
-function getCurrentUser(opts: { includeAuthStatus: false }): Observable<LoggedInUser>;
-function getCurrentUser(opts = { includeAuthStatus: true }): Observable<Session | LoggedInUser> {
+function getCurrentUser(opts: { includeAuthStatus: false }): Promise<LoggedInUser>;
+function getCurrentUser(opts = { includeAuthStatus: true }): Promise<Session | LoggedInUser> {
   if (lastFetchTimeMillis < Date.now() - 1000 * 60 || !sessionStore.getState().loaded) {
     refetchCurrentUser();
```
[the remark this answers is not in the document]
ibacher: This is a reasonable point, but I don't think the behaviour is meaningfully changed from what it was before this PR, right?

packages/apps/esm-login-app/src/login/login.component.tsx
```diff
@@ -30,7 +29,7 @@ const Login: React.FC = () => {
     provider: loginProvider,
     links: loginLinks,
   } = useConfig<ConfigSchema>();
-  const isLoginEnabled = useConnectivity();
+  const isLoginEnabled = navigator.onLine;
```
NethmiRodrigo: Do we need this anymore? Can't we remove this like it was removed in the location picker?

packages/apps/esm-login-app/src/redirect-logout/redirect-logout.component.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { navigate, setUserLanguage, useConfig, useConnectivity, useSession } from '@openmrs/esm-framework';
+import { navigate, setUserLanguage, useConfig, useSession } from '@openmrs/esm-framework';
 import { clearHistory } from '@openmrs/esm-framework/src/internal';
 import { type ConfigSchema } from '../config-schema';
 import { performLogout } from './logout.resource';
 
 const RedirectLogout: React.FC = () => {
   const config = useConfig<ConfigSchema>();
-  const isLoginEnabled = useConnectivity();
+  const isLoginEnabled = navigator.onLine;
```
NethmiRodrigo: Ditto here

packages/framework/esm-api/src/current-user.ts
```diff
@@ -23,80 +22,95 @@ export const sessionStore = createGlobalStore<SessionStore>('session', {
...
+        unsubscribe();
+        resolve(select(state.session));
       }
-    };
-    handler(sessionStore.getState());
-    // The observable subscribe function should return an unsubscribe function,
-    // which happens to be exactly what `subscribe` returns.
-    return sessionStore.subscribe(handler);
+    });
   });
```
NethmiRodrigo, line 118: Won't this cause the promise to hang if the `refetch` fails? Because it would reset the store to `loaded: false`?

packages/framework/esm-api/src/current-user.ts
```diff
@@ -23,80 +22,95 @@ export const sessionStore = createGlobalStore<SessionStore>('session', {
...
+let inFlightRefresh: Promise<SessionStore> | null = null;
+
+/**
+ * Fetches the session if the store is unloaded or its session is older than `sessionMaxAgeMillis`,
+ * joining a fetch already in flight rather than starting a second one. Returns `null` if the store
+ * already holds a fresh session. The returned promise rejects if the fetch fails, but the rejection
+ * is already handled and reported, so a caller that only needs the fetch started can ignore it.
+ */
+function refreshSessionIfStale(): Promise<SessionStore> | null {
+  if (sessionStore.getState().loaded && lastFetchTimeMillis >= Date.now() - sessionMaxAgeMillis) {
```
NethmiRodrigo: Since the timestamp is set when the request starts, won't a component that calls when a request is still running count the session as fresh but get the old data?

packages/framework/esm-framework/mock.tsx
```diff
@@ -66,10 +65,6 @@ export const getHistory = vi.fn(() => ['https://o3.openmrs.org/home']);
 export const clearHistory = vi.fn();
 export const goBackInHistory = vi.fn();
 
-/* esm-offline */
-export const useConnectivity = vi.fn().mockReturnValue(true);
```
NethmiRodrigo, line 70: Bit of a far fetch I guess, but should we remove this, because its still exported in `deprecated.ts`, so if there's any tests that do `vi.mocked(useConnectivity)` might need it?

packages/framework/esm-framework/mock-jest.tsx
```diff
@@ -65,10 +64,6 @@ export const getHistory = jest.fn(() => ['https://o3.openmrs.org/home']);
 export const clearHistory = jest.fn();
 export const goBackInHistory = jest.fn();
 
-/* esm-offline */
-export const useConnectivity = jest.fn().mockReturnValue(true);
```
NethmiRodrigo, line 69: Ditto, and the storybook esm-react-utils.ts mock

packages/framework/esm-framework/src/deprecated.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * @deprecated Offline support has been removed from the framework. This is a
+ * no-op kept only so that frontend modules that still call
+ * `useConnectivity` don't throw.
+ */
+export function useConnectivity() {
+  return true;
+}
```
NethmiRodrigo, line 87: The patient list calls getDynamicOfflineDataEntries every time the "Add to list" modal opens and registration calls `getSynchronizationItems` when the server does not return the patient. Should we add empty getters for those as well?

packages/framework/esm-react-utils/src/UserHasAccess.tsx
```diff
@@ -41,10 +41,15 @@ export const UserHasAccess: React.FC<UserHasAccessProps> = ({ privilege, fallbac
...
 
   useEffect(() => {
-    const subscription = getCurrentUser({
-      includeAuthStatus: false,
-    }).subscribe(setUser);
-    return () => subscription.unsubscribe();
+    const store = getSessionStore();
+    const unsubscribe = store.subscribe(({ loaded, session }) => {
+      setUser(loaded ? session?.user ?? null : null);
+    });
```
NethmiRodrigo, line 47: Should we do the same thing here that the `useSession` hook does and keep the last session if the refetch fails?

packages/framework/esm-react-utils/src/useVisitTypes.ts
```diff
@@ -25,18 +27,6 @@ import { useEffect, useState } from 'react';
...
-        setVisitTypes(visitTypes);
-      },
-      (error) => console.error(error),
-    );
-
-    return () => visitTypesSub.unsubscribe();
-  }, []);
-
-  return visitTypes;
+  const { data } = useSWRImmutable('visit-types', getVisitTypes);
```
NethmiRodrigo: Like the other hooks in here, should the key here be the request URL?

```suggestion
  const { data } = useSWRImmutable(`${restBaseUrl}/visittype`, getVisitTypes);
```

NethmiRodrigo (commented):
Thanks @ibacher! A couple of updates we should probably make as well

1. `.tx/config:20-26` - esm-offline-tools-app resource
2. `.github/dependabot.yml:34-39, 43-44` - rxjs, dexie, workbox-* ignore rules
3. `packages/shell/esm-app-shell/package.json:9-12` - OMRS_OFFLINE in the scripts
4. `packages/framework/esm-styleguide/src/components/_main-content.scss:15-24` - `.omrs-offline`
5. `packages/framework/esm-config/src/types.ts:40-43`- online / offline on `DisplayConditionsConfigObject`
6. The main README.md links to esm-offline-tools-app, [here](https://github.com/openmrs/openmrs-esm-core/blob/8173bbe9f9938ff3202e6a8a0ea3dba4c622757c/README.md?plain=1#L37) and [here](https://github.com/openmrs/openmrs-esm-core/blob/8173bbe9f9938ff3202e6a8a0ea3dba4c622757c/README.md?plain=1#L58)

packages/apps/esm-login-app/src/login/login.component.tsx
ibacher: I removed it in [1b5429e](https://github.com/openmrs/openmrs-esm-core/pull/1894/commits/1b5429ee0cfa4cdf21735d6635cdff95960419b4). How'd that get back in there?

packages/framework/esm-api/src/current-user.ts
ibacher: Out of date, but `< sessionMaxAgeMillis` out of date, which is the window this starts-off guaranteeing. This the right at the request start, two requests could be started at the same time because there are at least two pathways here.

packages/framework/esm-framework/mock.tsx
ibacher: Hmm... I suppose that's a good point.

packages/framework/esm-framework/mock-jest.tsx
ibacher: Storybook is not a concern because there's nothing outside of storybook itself that requires this be a mock (and it's mock is equivalent to this). We need to do this here because `vi.mocked()` is just Typescript sugar.

packages/framework/esm-react-utils/src/UserHasAccess.tsx
ibacher: I think for this, the better default behaviour is to deny until we've definitely got the latest session.

packages/framework/esm-react-utils/src/useVisitTypes.ts
ibacher: No, I don't think so. Normally, we use the URL as the key because:

1. It's usually the right thing (i.e., the request is only the same if the parameters are the same)
2. `useSWR()` calls the `fetcher` with the key as a parameter so we can just write: `useSWR('some-url', openmrsFetch);` and under the hood it calls `openmrsFetch('some-url')`.

In this case the request is immutable, so we don't want it to change based on the URL.

packages/framework/esm-api/src/current-user.ts
ibacher: Yes, but if the `refetch` fails, we don't know who the current user is. This gives the caller an opportunity to wait until we have an up-to-date session if they genuinely need it.

packages/apps/esm-login-app/src/login/login.component.tsx
ibacher: Ah! I missed that this was a different spot than the one I noticed. Removed.

packages/framework/esm-api/src/openmrs-fetch.ts
```diff
@@ -256,67 +255,6 @@ export function openmrsFetch<T = any>(path: string, fetchInit: FetchConfig = {})
...
- * subscription.unsubscribe()
- * ```
- *
- * #### Cancellation
- *
- * To cancel the network request, simply call `subscription.unsubscribe();`
- *
- * @category API
- */
-export function openmrsObservableFetch<T>(url: string, fetchInit: FetchConfig = {}) {
```
denniskigen, line 288: Can we migrate the reports resource and its callers alongside this removal? Running or rescheduling a report still calls `openmrsObservableFetch`. It throws before `.subscribe()`, so the error handler never runs and the form stays submitting.

packages/framework/esm-framework/src/deprecated.ts
```diff
@@ -0,0 +1,87 @@
...
+export function getDynamicOfflineDataEntries<T = unknown>(_type?: string): Promise<Array<T>> {
+  return Promise.resolve([]);
+}
+
+/**
+ * @deprecated Offline support has been removed from the framework. This is a stub kept only so that
+ * frontend modules that still read the sync queue keep working. It always resolves to an empty array,
+ * since nothing is queued for synchronization any more.
+ */
+export function getSynchronizationItems<T = unknown>(_type?: string): Promise<Array<T>> {
```
denniskigen, line 48: Can we also cover `getFullSynchronizationItems`, or migrate its AFE caller before this ships? AFE checks the queue before every submission, including online submissions. I reproduced it failing before reaching the online save path.

packages/framework/esm-framework/src/deprecated.ts
```diff
@@ -0,0 +1,87 @@
...
+  // no-op: dynamic offline data was removed from the framework
+}
+
+/**
+ * @deprecated Offline support has been removed from the framework. This is a stub kept only so that
+ * frontend modules that still read the dynamic offline data registry keep working. It always resolves
+ * to an empty array, since nothing is registered for offline use any more.
+ */
+export function getDynamicOfflineDataEntries<T = unknown>(_type?: string): Promise<Array<T>> {
+  return Promise.resolve([]);
```
denniskigen, line 40: Can we remove the fake offline list from patient-management alongside this change? Returning `[]` makes “Offline patients” appear in the Add to list modal, but selecting it calls the removed `putDynamicOfflineData` and fails.

packages/framework/esm-api/src/openmrs-fetch.ts
ibacher: Yeah. It's also [over here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/281126263e3ede446bd40b816983f258a39f39d1/packages/esm-patient-allergies-app/src/allergies/allergy-intolerance.resource.ts#L61), although that looks to be dead code.

PR #1898, opened 2026-09-23
(fix) Navigate should refuse non-HTTP URLs

ibacher:
## Requirements

## Summary

`navigate()` now rejects any non-HTTP(S) URLs, avoiding navigating to unintended locations. Also patches a small related issue with `isSafeUrl()` that missed that `navigate()` interpolates the URL so the URL it checks and the URL navigated to may not match.


packages/framework/esm-navigation/src/navigation/navigate.ts
```diff
@@ -54,6 +55,12 @@ export function navigate({ to, templateParams }: NavigateOptions): void {
   if (isSpaPath) {
     navigateToUrl(target);
   } else {
+    const targetURL = new URL(target, window.location.origin);
```
denniskigen: Can we check the protocol before branching on `isSpaPath`? When `spaPath` is `/`, `openmrsSpaBase` becomes an empty string, so every target goes through `navigateToUrl()` and skips this check. I reproduced a `javascript:` URL executing through single-spa in that configuration. Can we cover that case in the tests too?

packages/framework/esm-navigation/src/navigation/navigate.test.ts
```diff
@@ -66,4 +66,10 @@ describe('navigate', () => {
     expect(navigateToUrl).toHaveBeenCalledWith('/openmrs/spa/qux/page');
     expect(window.location.assign).not.toHaveBeenCalled();
   });
+
+  it('rejects not http URLs', () => {
+    expect(() => navigate({ to: 'ftp://${openmrsSpaBase}/qux/page' })).toThrow(Error);
```
denniskigen:
```suggestion
  it('rejects non-http URLs', () => {
    expect(() => navigate({ to: 'ftp://${openmrsSpaBase}/qux/page' })).toThrow(Error);
    expect(() => navigate({ to: 'javascript:alert(1)' })).toThrow(Error);
```
It'd be nice to cover `javascript:` explicitly since that's the case we really care about.

packages/framework/esm-navigation/src/navigation/interpolate-string.ts
```diff
@@ -36,11 +36,15 @@ function trimTrailingSlash(str: string) {
...
+  const result = interpolateString(template, {
     openmrsBase: window.openmrsBase,
     openmrsSpaBase: openmrsSpaBase,
     ...additionalParams,
-  }).replace(/^\/\//, '/'); // remove extra initial slash if present
+  });
+
+  // remove a additional / at the beginning if added by interpolation
+  // e.g. /${openmrsSpaBase} -> //openmrs/spa
+  return template.startsWith('//') ? result : result.replace(/^\/\//, '/');
```
denniskigen: Can we add a test for this? Protocol-relative templates are now kept as-is instead of collapsing to a same-origin path. `isSafeReturnUrl` depends on that now, so it's probably worth pinning down.

.changeset/navigate-non-http.md
```diff
@@ -0,0 +1,8 @@
+---
+"@openmrs/esm-login-app": patch
+"@openmrs/esm-navigation": patch
+"@openmrs/esm-translation": patch
```
denniskigen:
```suggestion
"@openmrs/esm-translations": patch
```

denniskigen (changes requested):
Thanks, @ibacher. Checking the interpolated URL in `isSafeReturnUrl` is a nice catch. There's one case where the protocol check gets skipped, plus a couple of test suggestions.

NethmiRodrigo (approved):

packages/framework/esm-navigation/src/navigation/interpolate-string.ts
ibacher: Tbf, this is sort of pinned in that the tests for `isSafeReturnUrl` with protocol-relative templates broke before I added this, but yeah, we should have an explicit test for it.

PR #1899, opened 2026-09-23
(feat) O3-6000: Allow extensions and workspaces to access their metadata

chibongho:
## Requirements

- [X] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.

## Summary

This PR:
- for workspaces, add the optional `meta` field when they are declared in `routes.json`.
  - the `meta` value is exposed through the existing `useWorkspace2Context()`.
- for extensions, with `meta` already an optional field, we store that value in the extensions's `ComponentContext`
  - the `meta` is exposed through a new `useExtensionContext()` hook., symmetric to the `useWorkspace2Context()`. (I did not write a `useComponentContext()` hook to expose the entire `CompontentContext`, only its `extension` field).


packages/framework/esm-extensions/src/types.ts
```diff
@@ -1,7 +1,14 @@
+import { type ExtensionMeta } from './store';
+
 export interface ExtensionData {
   extensionSlotName: string;
   extensionSlotModuleName: string;
   extensionId: string;
+  /**
+   * The extension's `meta`. Only set for components rendered as extensions; a component rendered
+   * some other way (a modal, say) has no extension context.
+   */
```
ibacher: nit: this comment doesn't really add much value.

```suggestion
```

packages/framework/esm-react-utils/src/public.ts
```diff
@@ -16,6 +16,7 @@ export * from './useConnectivity';
 export * from './useDebounce';
 export * from './useDefineAppContext';
 export * from './useEmrConfiguration';
+export * from './useExtensionContext';
```
ibacher: I think when we add things to `public.ts`, they also need to be added to `index.ts`. `public.ts` controls what's publicly visible in the types, but `index.ts` is what's available at runtime.

packages/framework/esm-styleguide/src/workspaces2/workspace2.component.tsx
```diff
@@ -41,6 +41,13 @@ export interface Workspace2DefinitionProps<
   workspaceProps: WorkspaceProps | null;
   windowProps: WindowProps | null;
   groupProps: GroupProps | null;
+  /**
+   * The workspace's `meta` from its registration (the `meta` field of its `workspaces2` entry in
+   * `routes.json`). The framework always supplies an object, empty when the workspace declares no
+   * meta; it is optional only so existing prop literals need not be updated. Values are typed as
+   * `any` to match `useExtensionContext().extensionMeta`, so fields can be read without a cast.
```
ibacher:
```suggestion
   * `routes.json`).
```

ibacher (commented):
Basically LGTM. A couple of very minor nits.

packages/framework/esm-styleguide/src/workspaces2/workspace2.component.tsx
```diff
@@ -41,6 +41,13 @@ export interface Workspace2DefinitionProps<
...
+   */
+  workspaceMeta?: Readonly<Record<string, any>>;
```
denniskigen:
```suggestion
  workspaceMeta?: Readonly<Record<string, unknown>>;
```

One additional nit: Can we use `unknown` here, consistent with `WorkspaceDefinition2.meta`? The metadata shape varies by workspace, so consumers should narrow the values before using them.

chibongho: Changed this, and `useExtensionContext().extensionMeta` to `Readonly<Record<string, unknown>>`. Note that the current `useExtensionSlotMeta()` hook uses type `ExtensionMeta`, of type `Record<string, any>`. I left it alone as changing that will be a breaking change.

denniskigen:
LGTM! Thanks, @chibongho!

denniskigen (approved):

